PR #25, opened 2022-10-19
Move to FIPS-compliant pre-compiled boringssl library

briansonnenberg:


boringssl-fips/build/crypto/chacha/chacha-x86_64.S
```diff
@@ -0,0 +1,1632 @@
+# This file is generated from a similarly-named Perl script in the BoringSSL
```
howardjohn, line 1: I think we just need the `*.a` files in build/ (but everything under include/). Can we clean these up a bit? Especially the 800k line one :-)

howardjohn (commented):
This is great, works perfectly. Just a few nits

PR #28, opened 2022-10-20
Add support for reading from local file instead of XDS

howardjohn:
This is useful for testing, etc without an XDS server


briansonnenberg (approved):

examples/localhost.yaml
```diff
@@ -0,0 +1,6 @@
+# This shows an example local config for ztunnel that adds a workload for localhost.
+# This allows local testing by sending requests through the local ztunnel other servers running on localhost.
```
howardjohn, line 2:
```suggestion
# This allows local testing by sending requests through the local ztunnel to other servers running on localhost.
```

Issue #11, opened 2022-10-19
Develop certificate provider module that can reach out to Istiod CA over gRPC and fetch and rotate certificates


howardjohn:

howardjohn:
The "fetching certs" part is done. Now we need something to cache them and rotate them.

What we want:

* In SecretManager, store a cache of certificates we own
* For each cert, at 50% of its lifetime, refresh it. See `secretcache.go` in agent for prior art here, although it can likely be a bit simpler here (I think we can just spawn a task with `sleep(TTL); refresh` or something)
* When a new workload on the node is discovered, pre-fetch its cert (if we don't have it yet)
* `SecretManager.fetch_certificate` uses the cache when possible.
* If we race on explicit fetch_certificate call and pre-fetch, we should only have 1 request to the CA and both await it; a CA request has high cost

stevenctl:
zTunnel can request cert on behalf of workload

Certs are pre-fetched on workload discovery

Certs are cached and re-used 

Certs in cache are automatically rotated

SkyfireFrancisZ:
first PR merged o Fri 11/19

SkyfireFrancisZ:
Split cache on discovery into separate issue, then mark this one done.

PR #30, opened 2022-10-21
tls: use real certificates

howardjohn:
This drops the test-certs for real certs.

WARNING: this tanks performance since every connection generates a cert
on the fly. https://github.com/istio/ztunnel/issues/11 will fix this and
Brian is working on this right now.

Note: certs are still not verified since we don't properly get the root
cert, etc.

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -9,6 +9,7 @@ use tracing::{debug, error, info, warn};
 
 use xds::istio::workload::Workload as XdsWorkload;
 
+use crate::identity::Identity;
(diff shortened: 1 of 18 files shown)


src/workload.rs
```diff
@@ -60,6 +63,17 @@ pub struct Workload {
     pub native_hbone: bool,
 }
 
+impl Workload {
```
briansonnenberg, line 66: Combine with the impl Workload below?

briansonnenberg (approved):

PR #32, opened 2022-10-21
Create `presubmit` test for CI and setup build container

howardjohn:


howardjohn:
Goes with https://github.com/istio/test-infra/pull/4343

Issue #10, opened 2022-10-19
Audit .clone() usages (especially for String); replace with references where possible


howardjohn:

ilrudie:
I did a pretty simple audit of clones in 418.  It looks like we have some places where we have clones of owned values stored in different places nested within a parent data structure. This may be for legitimate reasons because in some instances the nested types are being used independently and it's not feasible to have borrows being used in this way. I didn't want to spend a lot of time potentially refactoring our types and end up with a big merge conflict so I picked some low hanging fruit and tried to keep 418 pretty focused on that. I suspect more work can be done to optimize our types though which may include changing some owned values to references.

howardjohn:
This is pretty much done, especially with Strng

PR #36, opened 2022-10-22
Audit clone usages

digitalsanctum:
* Address some of the `clone()` usages #10
* Introduce `WorkloadBuilder` in an effort to clean up how `Workload` is constructed
* Simplify `byte_to_ip` function and added tests

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -1,6 +1,7 @@
 use std::collections::{HashMap, HashSet};
+use std::convert::Into;
 use std::net::{IpAddr, SocketAddr};
-use std::str::FromStr;
+use std::ops::Deref;
 use std::sync::{Arc, Mutex};
 use std::{fmt, net};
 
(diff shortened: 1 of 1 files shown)


src/workload.rs
```diff
@@ -118,51 +181,40 @@ impl fmt::Display for Upstream {
...
-            Some(IpAddr::V6(Ipv6Addr::from(v)))
-        }
-        n => return Err(WorkloadError::ByteAddressParse(n)),
-    })
+fn byte_to_ip(b: &bytes::Bytes) -> Result<Option<IpAddr>, WorkloadError> {
+    let s = std::str::from_utf8(b).expect("error converting bytes to string");
+    if s.is_empty() {
+        return Ok(None);
+    }
+    match IpAddr::from_str(s) {
```
howardjohn, line 189: Is this doing allocations to convert bytes -> string -> ip instead of just bytes -> IP?

I expect IP addresses to be one of the largest memory usages since we want to scale to >100k IPs, so optimizing this is pretty important

howardjohn: I am hoping to add benchmarks so we don't have to guess at this :-)

src/workload.rs
```diff
@@ -118,51 +181,40 @@ impl fmt::Display for Upstream {
...
+        let workload = WorkloadBuilder::new(
+            resource.name,
+            resource.namespace,
+            resource.native_hbone,
+            resource.node,
+            protocol,
+            service_account,
+            workload_ip,
+        )
+        .waypoint_address(waypoint_ip)
```
howardjohn, line 215: Can you help my understand why we want to some fields in the `new` but this one as a function? Is it because its an Option? It seems like we are always setting it though.

Do we lose some type safety by switching from named struct to unnamed param list? For example, name and namespace could accidentally be swapped and not noticed?

digitalsanctum: Generally, you want required fields in the `new` and optional fields as a function.

The unnamed param list is just a shortcut if you keep the param names the same as the field names.

howardjohn: But we seem to have lost type safety (I can trivially cause a bug by swapping `name` and `namespace` - impossible with the struct shortcut) and readability, while making it more verbose .

Its not clear to me what value we are getting out of just using the struct directly?

howardjohn: TBH this is my first Rust project... so while I have an OK handle on how to get things to compile at this point, my knowledge of best practices is pretty slim - very open to expert opinions here. I like the builder abstracting logic like that, but don't love that its easy to miss required fields. It does at least seem like we only actually make a Workload in 1 single place, so any minor deficiencies aren't a huge issue and I am ok with whatever is preferred. 

At the very least abstracting out some of the conversion logic could make sense

digitalsanctum: Okay, I'll revert the builder stuff then.

src/workload.rs
```diff
@@ -118,28 +120,32 @@ impl fmt::Display for Upstream {
...
+            Some(IpAddr::from(v))
         }
         16 => {
-            let v: [u8; 16] = (&*b).try_into().expect("size already proven");
-            Some(IpAddr::V6(Ipv6Addr::from(v)))
+            let v: [u8; 16] = b.deref().try_into().expect("size already proven");
+            Some(IpAddr::from(v))
+        }
+        n => {
+            println!("invalid ip address length: {n}");
```
howardjohn, line 135: can we remove print? At the very least we should using `tracing::warning!` probably, but not sure we need it since we already have a ByteAddressParse which should get bubbled up somewhere

howardjohn (approved):
thanks!

PR #46, opened 2022-11-03
Cert cache + refresh logic

briansonnenberg:
Initially used tokio::DelayQueue for refreshing, but after discussing with John changed to independent tasks spawned when we fetch the cert. Eventually will want some protection against sending many CA requests at once.

### DIFF
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -14,9 +14,10 @@
 
 use std::fmt;
 use std::collections::HashMap;
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, RwLock};
 use tokio::time::{sleep, Duration};
 use tracing::instrument;
+use tokio::sync::watch;
 
 use super::CaClient;
 use super::Error;
@@ -50,37 +51,46 @@ impl fmt::Display for Identity {
 #[derive(Clone)]
 pub struct SecretManager {
     client: CaClient,
-    cache: Arc<Mutex<HashMap<Identity, tls::Certs>>>,
+    cache: Arc<RwLock<HashMap<Identity, watch::Receiver<Option<tls::Certs>>>>>,
 }
 
 impl SecretManager {
     pub fn new(cfg: crate::config::Config) -> SecretManager {
-        let client = CaClient::new(cfg.auth);
-        let cache: HashMap<Identity, tls::Certs> = Default::default();
-        return SecretManager { client: client, cache: Arc::new(Mutex::new(cache)) };
+        let caclient = CaClient::new(cfg.auth);
+        let cache: HashMap<Identity, watch::Receiver<Option<tls::Certs>>> = Default::default();
+        SecretManager {
+            client: caclient,
+            cache: Arc::new(RwLock::new(cache))
+        }
     }
 
-    pub async fn refresh_handler(id: Identity, ctx: SecretManager, initial_sleep_time: u64) {
-        info!("refreshing certs for id {} in {:?} seconds", id, initial_sleep_time);
-        sleep(Duration::from_secs(initial_sleep_time)).await;
+    pub async fn refresh_handler(id: Identity, mut ctx: SecretManager,
+                                 initial_sleep_time: Duration,
+                                 tx: watch::Sender<Option<tls::Certs>>) {
+        sleep(initial_sleep_time).await;
         loop {
-            match ctx.client.clone().fetch_certificate(&id.clone()).await {
+            match ctx.client.fetch_certificate(&id).await {
                 Err(e) => {
                     // Cert refresh has failed. Drop cert from the cache.
                     warn!("Failed cert refresh for id {:?}: {:?}", id, e);
-                    { // lock cache
-                        let mut locked_cache = ctx.cache.lock().unwrap();
+                    {
+                        let mut locked_cache = ctx.cache.write().unwrap();
                         locked_cache.remove(&id.clone());
-                    } // unlock cache
+                    }
                     return;
                 }
                 Ok(fetched_certs) => {
-                    info!("refreshed certs {:?}", fetched_certs);
-                    { // lock cache
-                        let mut locked_cache = ctx.cache.lock().unwrap();
-                        locked_cache.insert(id.clone(), fetched_certs.clone());
-                    } // unlock cache
-                    let sleep_dur = Duration::from_secs(fetched_certs.get_seconds_until_refresh());
+
+                    let sleep_dur = fetched_certs.get_duration_until_refresh();
+                    match tx.send(Some(fetched_certs.clone())) {
+                        Err(_) => {
+                            let mut locked_cache = ctx.cache.write().unwrap();
+                            locked_cache.remove(&id.clone());
+                        },
+                        Ok(_) => {
+                            info!("refreshed certs for id: {:?}", id);
+                        }
+                    }
                     info!("refreshing certs for id {} in {:?} seconds", id, sleep_dur);
                     sleep(sleep_dur).await;
                 }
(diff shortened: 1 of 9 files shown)


src/tls/boring.rs
```diff
@@ -69,13 +70,39 @@ impl CsrOptions {
...
+        self.cert.not_after()
+    }
+
+    pub fn get_seconds_until_refresh(&self) -> u64 {
+        let current = Asn1Time::days_from_now(0).unwrap();
+        let start: &Asn1TimeRef = self.not_before();
+        let end: &Asn1TimeRef = self.not_after();
+
+        let total_lifetime = start.diff(end).unwrap();
+        let total_lifetime_secs = total_lifetime.days * 86400 + total_lifetime.secs;
```
howardjohn, line 94: A lot of this may be more ergonomic if we can just convert the Asn1Time stuff to standard rust Duration/Time?

howardjohn (commented):
Overall looks good!

src/identity/manager.rs
```diff
@@ -32,16 +36,68 @@ impl fmt::Display for Identity {
...
+            let locked_cache = self.cache.lock().unwrap();
+            let cache_certs: std::option::Option<&tls::Certs> = locked_cache.get(id);
+            info!("cache certs for req: {:?}", cache_certs);
+            if cache_certs.is_some() {
+                return Ok(cache_certs.unwrap().clone())
+            }
+        } // unlock cache
+
+        // No cache entry, fetch it and spawn refresh handler
+        let fetched_certs = self.client.clone().fetch_certificate(id.clone()).await?;
```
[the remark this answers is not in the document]
briansonnenberg: How about a state flag in the cache indicating a request is in progress or something along those lines?

src/identity/manager.rs
```diff
@@ -32,16 +36,68 @@ impl fmt::Display for Identity {
...
+        info!("fetched certs {:?}", fetched_certs);
+        { // lock cache
+            let mut locked_cache = self.cache.lock().unwrap();
+            locked_cache.insert(id.clone(), fetched_certs.clone());
+        } // unlock cache
+
+        tokio::spawn(SecretManager::refresh_handler(id.clone(),
```
stevenctl, line 97: Would it make more sense to just immediately spawn a refresh handler, and return a handle to some channel that notifies when it's got a cert/cert changed etc. Then wait on that handler.

Having a set of id -> handler would be the "request in progress" indicator

src/identity/manager.rs
```diff
@@ -32,16 +36,68 @@ impl fmt::Display for Identity {
...
+                                                        self.clone(),
```
stevenctl, line 98: the `self.clone` feels odd to me. if we add some Clone field to SecretManager an unsuspecting dev may assume it's a shared reference when it's not. Should we pass a client and reference to the cache here explicitly? Should the cache get wrapped up in some accessors?

I'm not sure of the idiomatic thing to do.

src/identity/manager.rs
```diff
@@ -32,16 +36,68 @@ impl fmt::Display for Identity {
...
+                    // Cert refresh has failed. Drop cert from the cache.
+                    warn!("Failed cert refresh for id {:?}: {:?}", id, e);
+                    { // lock cache
+                        let mut locked_cache = ctx.cache.lock().unwrap();
+                        locked_cache.remove(&id.clone());
+                    } // unlock cache
+                    return;
+                }
+                Ok(fetched_certs) => {
+                    info!("refreshed certs {:?}", fetched_certs);
```
hzxuzhonghu, line 64: is this secure?

src/identity/manager.rs
```diff
@@ -32,16 +36,68 @@ impl fmt::Display for Identity {
...
+
+    pub async fn refresh_handler(id: Identity, ctx: SecretManager, initial_sleep_time: u64) {
+        info!("refreshing certs for id {} in {:?} seconds", id, initial_sleep_time);
+        sleep(Duration::from_secs(initial_sleep_time)).await;
+        loop {
+            match ctx.client.clone().fetch_certificate(id.clone()).await {
+                Err(e) => {
+                    // Cert refresh has failed. Drop cert from the cache.
+                    warn!("Failed cert refresh for id {:?}: {:?}", id, e);
+                    { // lock cache
```
hzxuzhonghu, line 57: do we need {} ?

briansonnenberg: It is used to scope the lock.  Can replace with an explicit drop, but I personally like the visual indicator of where the lock is held.  Either way.

src/identity/manager.rs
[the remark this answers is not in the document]
briansonnenberg: There is currently a race condition that will result in multiple refresh handlers being spawned.  Working on it.

src/identity/manager.rs
```diff
@@ -32,16 +37,141 @@ impl fmt::Display for Identity {
...
+                            warn!("Cert sender has been dropped.");
+                            break;
+                        }
+                        Ok(_) => {
+                            continue;
+                        }
+                    }
+                }
+            }
+        }
```
howardjohn, line 175: Can we add some unit tests to this? We should be able to mock out the actual CA call with a fake CaClient (might require some refactoring to allow CaClient to be generic)

src/proxy/outbound.rs
```diff
@@ -46,25 +47,32 @@ impl Outbound {
...
                             cfg,
                         };
                         tokio::spawn(async move {
                             let res = oc.proxy(stream).await;
                             match res {
-                                Ok(_) => info!("outbound proxy complete"),
-                                Err(ref e) => warn!("outbound proxy failed: {}", e),
+                                Ok(_) => info!(
+                                    "outbound proxy complete ({}ms)",
+                                    start_outbound_instant.elapsed().as_millis()
```
howardjohn, line 69: can we add a format that is dynamic like Go's duration format? we don't want to use ms to log a connection that took 10hrs

src/tls/boring.rs
```diff
@@ -69,13 +71,39 @@ impl CsrOptions {
...
     key: pkey::PKey<pkey::Private>,
 }
 
+impl Certs {
+    pub fn not_before(&self) -> &Asn1TimeRef {
+        self.cert.not_before()
+    }
+    pub fn not_after(&self) -> &Asn1TimeRef {
+        self.cert.not_after()
+    }
```
howardjohn, line 87: these functions do not seem needed, or at least not `pub`, since we only call them internally

howardjohn: If anything we could keep them but convert them to stdlib Time so its easier to deal with

src/identity/manager.rs
```diff
@@ -32,16 +37,141 @@ impl fmt::Display for Identity {
...
+        let cache: HashMap<Identity, watch::Receiver<Option<tls::Certs>>> = Default::default();
+        SecretManager {
+            client: caclient,
+            cache: Arc::new(RwLock::new(cache)),
+        }
+    }
+
+    pub async fn refresh_handler(
+        id: Identity,
+        mut ctx: SecretManager,
```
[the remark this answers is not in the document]
briansonnenberg: Need to clone self so that the compiler is not scared of self's lifetime being less than task.  Steven's suggestion of passing the necessary fields individually is probably what I'll do, but it's a bit messy.

src/identity/manager.rs
```diff
@@ -32,16 +37,141 @@ impl fmt::Display for Identity {
...
+    ) {
+        sleep(initial_sleep_time).await;
+        loop {
+            match ctx.client.fetch_certificate(id.clone()).await {
+                Err(e) => {
+                    // Cert refresh has failed. Drop cert from the cache.
+                    warn!("Failed cert refresh for id {:?}: {:?}", id, e);
+                    {
+                        let mut locked_cache = ctx.cache.write().unwrap();
+                        locked_cache.remove(&id.clone());
```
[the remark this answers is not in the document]
briansonnenberg: This is true.  I guess we can just hold and retry until it expires.

src/identity/manager.rs
```diff
@@ -32,16 +37,141 @@ impl fmt::Display for Identity {
...
+                    }
+                    return;
+                }
+                Ok(fetched_certs) => {
+                    let sleep_dur = fetched_certs.get_duration_until_refresh();
+                    match tx.send(Some(fetched_certs.clone())) {
+                        Err(_) => {
```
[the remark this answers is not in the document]
briansonnenberg: "This method fails if the channel has been closed, which happens when every receiver has been dropped."  Should pretty much never happen.  Should just return after though.

src/identity/manager.rs
```diff
@@ -32,16 +37,141 @@ impl fmt::Display for Identity {
...
     }
 
     #[instrument(skip_all, fields(%id))]
-    pub async fn fetch_certificate(&self, id: Identity) -> Result<tls::Certs, Error> {
-        self.client.clone().fetch_certificate(id).await
+    pub async fn fetch_certificate(&mut self, id: &Identity) -> Result<tls::Certs, Error> {
+        {
+            loop {
+                let mut tx_option: Option<watch::Sender<Option<tls::Certs>>> = None;
+                let mut cache_rx = None;
```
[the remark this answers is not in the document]
briansonnenberg: Wanted to do something like that, but there are some issues with .awaiting in the context of a locked cache, or even in a scope with a lock that isn't being used.

src/identity/manager.rs
```diff
@@ -32,16 +37,141 @@ impl fmt::Display for Identity {
...
+                {
+                    let read_locked_cache = self.cache.read().unwrap();
+                    match read_locked_cache.get(id) {
+                        None => {
+                            drop(read_locked_cache);
```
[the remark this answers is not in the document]
briansonnenberg: The overwhelmingly more common case is get cert from the cache, so I wanted to avoid exclusive locking the cache on it.

src/identity/manager.rs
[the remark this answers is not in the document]
howardjohn: There must be a better way to do this but I cannot find it. Its annoying since `drop(mutex)` isn't respected..

src/tls/boring.rs
```diff
@@ -242,7 +284,42 @@ where
...
+    const PKEY: &[u8] = include_bytes!("cert.key");
+
+    pub fn test_certs() -> Certs {
+        let cert = x509::X509::from_pem(CERT).unwrap();
+        let key = pkey::PKey::private_key_from_pem(PKEY).unwrap();
+        Certs { cert, key }
+    }
+
+    // Creates an invalid dummy cert with overridden expire time
+    pub fn generate_test_certs(seconds_until_expiry: u64) -> Certs {
```
howardjohn, line 301: u64 -> Duration will make this more obvious

src/identity/manager.rs
```diff
@@ -30,18 +38,214 @@ impl fmt::Display for Identity {
...
+                trust_domain: "cluster.local".to_string(),
+                namespace: "istio-system".to_string(),
+                service_account: format!("ztunnel{}", i),
+            };
+            sm.fetch_certificate(&id).await.unwrap();
+        }
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
+    async fn test_stress_caching() {
```
howardjohn, line 242: can we have a test covering the refresh?

src/identity/manager.rs
```diff
@@ -44,18 +52,306 @@ impl fmt::Display for Identity {
...
+                    Err(e) => {
+                        let mut locked_cache = self.cache.write().unwrap();
+                        locked_cache.remove(id);
+                        return Err(e);
+                    }
+                }
+            }
+
+            // We found a channel receiver in the cache, so use it to get the cert.
+            loop {
```
stevenctl, line 188: Why must the cache receiver send `Option`? Seems like we could just block here until the receiver gets something instead of looping. I don't see any `.send(None)`.

briansonnenberg: L142:
`let (tx, rx) = watch::channel(None);`

That is sending None.  Without an Option we can't initialize the receiver until we have the cert.

src/identity/manager.rs
```diff
@@ -44,18 +52,306 @@ impl fmt::Display for Identity {
...
+            dur.as_secs() / expected_update_interval.as_secs() - 1
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
+    async fn test_stress_caching() {
+        let mut tasks: Vec<tokio::task::JoinHandle<()>> = Vec::new();
+        let secret_manager = SecretManager::new_mock(Duration::from_secs(10));
+        for _n in 0..8 {
+            tasks.push(tokio::spawn(stress_many_ids(secret_manager.clone(), 5000)));
```
howardjohn, line 307: this test is pretty slow. Do we need 5000? Would 100 be sufficient

src/tls/boring.rs
howardjohn: this still could be done

howardjohn: I think this would also be needed for  the test to work properly on short time scales, since right now we are only handling second precision

src/identity/manager.rs
```diff
@@ -44,18 +52,306 @@ impl fmt::Display for Identity {
...
+                    Some(cert_rcvr) => {
+                        cache_rx = Some(cert_rcvr.clone());
+                    }
+                }
+            }
+
+            // We made a transmitter, so fetch the cert and send the result.
+            if let Some(tx) = tx_option {
+                info!("Fetching cert.");
+                let certs = self.client.fetch_certificate(id).await;
```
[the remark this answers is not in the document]
briansonnenberg: I thought about that, and it makes sense and is neater, but it means an extra rx.borrow().clone() from the fetch_certificate context in the first-request case.  It's probably best to do it anyway.

src/identity/manager.rs
```diff
@@ -44,18 +52,306 @@ impl fmt::Display for Identity {
...
+                Err(e) => {
+                    warn!("Failed cert refresh for id {:?}: {:?}", id, e);
+                    let mut write_locked_cache = cache.write().unwrap();
+                    let Some(certs_rx) = write_locked_cache.get(&id) else {
+                        // Should not be possible, but if there is no receiver
+                        // in the cache, then no one is using these certs that
+                        // are being refreshed, so let's stop refreshing them.
+                        return;
+                    };
+                    if certs_rx.borrow().clone().unwrap().is_expired() {
```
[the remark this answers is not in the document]
briansonnenberg: Because we do not spawn the refresh handler with None for the cert entry.  We only have None until the initial request has been completed and the cert populated.  After that we spawn the refresh thread.

howardjohn:
https://docs.rs/tokio/latest/tokio/time/fn.advance.html may be useful for the test

PR #47, opened 2022-11-03
Removed unused golang related scripts and variables

litong01:
These golang related variables and scripts do not seem to be used any more, removing them to make the build process less confusing.
Signed-off-by: Tong Li <litong01@us.ibm.com>


litong01:
> These come from common-files, this will be reverted automatically even if we merge.
> 
> These are common to all repos, regardless of language

@howardjohn  John, are they copied from other repo?

howardjohn:
istio/common-files

ericvn:
@litong01 Files in the common directory are: https://github.com/istio/common/blob/master/common/Makefile.common.mk#L105

litong01:
Not needed since common files pulled from istio/common

PR #48, opened 2022-11-03
Removed unused GOBIN variable from Makefile

litong01:
Signed-off-by: Tong Li <litong01@us.ibm.com>


howardjohn:
this is also from common files

litong01:
> this is also from common files

@howardjohn haha, wow. ok.

PR #52, opened 2022-11-04
Initial support for on demand xds

howardjohn:
This is tied to https://github.com/istio/istio/pull/41796, since the resource name changes.

This does NOT enable on-demand XDS yet; its off-by-default. It sort of works though. However, it does some nice cleanup and allows for further development, and allows us to run with https://github.com/istio/istio/pull/41796 which makes the control plane much more efficient (along with supporting on-demand)

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -13,7 +13,7 @@ use xds::istio::workload::Workload as XdsWorkload;
 
 use crate::identity::Identity;
 use crate::workload::WorkloadError::ProtocolParse;
-use crate::xds::{HandlerContext, XdsUpdate};
+use crate::xds::{Demander, HandlerContext, XdsUpdate};
 use crate::{config, xds};
 
 #[derive(Debug, Hash, Eq, PartialEq, Clone, Copy, serde::Serialize, serde::Deserialize)]
@@ -73,12 +73,6 @@ impl Workload {
             namespace: self.namespace.clone(),
             service_account: self.service_account.clone(),
         }
-    }
-}
-
-impl Workload {
-    fn resource_name(&self) -> String {
-        self.name.to_owned() + "/" + self.namespace.as_str()
     }
 }
 
@@ -168,7 +162,7 @@ impl TryFrom<&XdsWorkload> for Workload {
 }
 
 pub struct WorkloadManager {
-    workloads: Arc<Mutex<WorkloadInformation>>,
+    workloads: WorkloadInformation,
     xds_client: xds::AdsClient,
     local_client: LocalClient,
 }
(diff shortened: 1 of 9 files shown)


src/workload.rs
```diff
@@ -275,40 +278,79 @@ impl LocalClient {
...
-            if let Some(_workload) = self.workloads.remove(&ip) {
-                // TODO: store VIPs in workload so we can remove them
-            } else {
-                panic!(
-                    "removed name {} with ip {}, but was not found in workload map",
-                    name, ip
-                )
+    fn remove(&mut self, ip: String) {
+        use std::str::FromStr;
+        let ip: IpAddr = match IpAddr::from_str(&ip) {
```
stevenctl, line 339: `let else` syntax just came out.. we're on 1.65, right? 

```suggestion
        let Ok(ip) = IpAddr::from_str(&ip) else {
          error!("...")
          return;
       }
       // use `ip` as needed. 
```

src/workload.rs
```diff
@@ -219,9 +213,9 @@ impl xds::Handler<XdsWorkload> for Arc<Mutex<WorkloadInformation>> {
 
 impl WorkloadManager {
     pub fn new(config: config::Config) -> WorkloadManager {
```
stevenctl, line 215: seems more appropriate to take a reference here

src/xds/client.rs
```diff
@@ -62,13 +63,15 @@ impl<T: prost::Message> Handler<T> for NopHandler {
 pub struct Config {
     workload_handler: Box<dyn Handler<Workload>>,
     initial_watches: Vec<String>,
+    on_demand: bool,
 }
 
 impl Config {
-    pub fn new() -> Config {
+    pub fn new(config: crate::config::Config) -> Config {
```
stevenctl, line 70: again, seems more appropriate to take a reference

src/xds/client.rs
```diff
@@ -181,86 +243,142 @@ impl AdsClient {
...
-                    Resource::Workload(inner) => {
-                        let name = inner.name.clone();
-                        let up = XdsUpdate::Update(inner);
-                        updates.push(up);
-                        workloads.insert(name);
-                    }
+        let demand = &mut self.demand;
+        loop {
+            tokio::select! {
+                demand_event = demand.recv() => {
```
stevenctl, line 257: Can we break up `run_internal`? It's kind of hard to read. 

* outbound_stream(watches vec<&str>, discovery_req_rx, Receiver<DiscoveryRequest>) -> Stream
* One for each of these select blocks.

stevenctl:
LGTM

stevenctl (approved):

src/workload.rs
howardjohn: I changed some other places but in this case `let else` is not amazing since it won't destruct the `Err(e)` for us

PR #59, opened 2022-11-07
Fix calls to pods on the same node

howardjohn:


src/proxy/outbound.rs
```diff
@@ -232,11 +232,14 @@ impl OutboundConnection {
...
-            // so we need to explicitly send it to ourselves.
+            // Sending to a node on the same node (ourselves).
             // In the future this could be optimized to avoid a full network traversal.
             req.request_type = RequestType::DirectLocal;
-            req.gateway = "127.0.0.1:15088".parse().unwrap();
+            // We would want to send to 127.0.0.1:15008 in theory. However, the inbound listener
+            // expects to lookup the desired certificate based on the destination IP. If we send directly,
+            // we would try to lookup an IP for 127.0.0.1.
+            // Instead, we send to the actual IP, but iptables in the pod ensures traffic is redirected to 15008.
+            req.gateway = SocketAddr::from((req.gateway.ip(), 15088));
```
hzxuzhonghu, line 242: 15008？

howardjohn: No, 15088 is setup to redirect back to 15008.

This is not a great hack but expediant way to get things working and it matches what we do in Envoy today (hence why the iptables rule exists today).

hzxuzhonghu: got it

howardjohn: long term we can probably just do it in memory since we are calling ourselves

hzxuzhonghu: how about `req.gateway.set_port`

howardjohn: req.gateway is not `mut` so would need more refactoring, but could be done.

I think there is an issue to refactor this code a bit

hzxuzhonghu (approved):
So lgtm

stevenctl (approved):

PR #62, opened 2022-11-08
Implement basic graceful drain/exit of proxy

howardjohn:
Currently the proxy just runs forever on SIGTERM. This makes it exit, and puts some *very basic* graceful draining in place.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,11 +9,15 @@ default = []
 gperftools = ["dep:gperftools"]
 console = ["dep:console-subscriber"]
 
+[lib]
+path = "src/lib.rs"
+
 [[bin]]
 name = "ztunnel"
 path = "src/main.rs"
 
 [dependencies]
+#tikv-jemallocator = { version = "0.5", features = ["profiling", "stats"]}
 anyhow = "1.0.65"
 async-stream = "0.3.3"
 async-trait = "0.1.58"
@@ -30,6 +34,7 @@ lazy_static = "1.4.0"
 libc = "0.2.126"
 log = "0.4"
 num_cpus = "1.13.1"
+once_cell = "1.16.0"
 pprof = { version = "0.10.1", features = ["protobuf", "protobuf-codec"] }
 prost = "0.11"
 prost-types = "0.11.1"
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -46,6 +46,7 @@ hyper-boring = { version= "2.1.2", features = ['fips'] }
 anyhow = "1.0.65"
 serde_yaml = "0.9.13"
 async-trait = "0.1.58"
+drain = "0.1.1"
```
stevenctl, line 49: I see this is from linkerd and is probably solid. 

As we add more reviewers (not sure how many will be project-root approvers), we gotta have a well-defined dependency evaluation process.

src/proxy/mod.rs
```diff
@@ -27,11 +28,12 @@ impl Proxy {
         cfg: config::Config,
         workloads: WorkloadInformation,
         secret_manager: identity::SecretManager,
+        drain: Watch,
     ) -> Result<Proxy, Error> {
         // We setup all the listeners first so we can capture any errors that should block startup
         let inbound_passthrough = InboundPassthrough::new(cfg.clone());
         let inbound = Inbound::new(cfg.clone(), workloads.clone(), secret_manager.clone()).await?;
-        let outbound = Outbound::new(cfg.clone(), secret_manager, workloads).await?;
+        let outbound = Outbound::new(cfg.clone(), secret_manager, workloads, drain).await?;
```
stevenctl, line 36: at some point as we keep passing the `drain` down, does it make sense to clone it (even though there aren't other uses)

makes it clear that each consumer should take a clone. i think a codebase for a language we're not that familiar with it's nice when you're surrounded by examples of how you can get the same functionality. 

random thought, not sure if it actually makes much sense. the clone should be nearly free.

stevenctl (commented):
LGTM. Can we add any tests here. We should probably stop merging untested PRs soon.

src/proxy/mod.rs
howardjohn: Yeah, probably so

howardjohn:
> LGTM. Can we add any tests here. We should probably stop merging untested PRs soon.

I agree. This one will be a good test of our test setup, since it involves the whole main() structure, so will be a ~full e2e test.

stevenctl (approved):

PR #63, opened 2022-11-09
use UNSPECIFIED for socketAddr and reduce a parameter for the Builder::new

zhlsunshine:
1. Use UNSPECIFIED for socketAddr of Default config 
2. Reduce a parameter for the Builder::new


src/admin/mod.rs
```diff
@@ -44,9 +44,9 @@ pub struct Server {
 pub struct Readiness(Arc<AtomicBool>);
 
 impl Builder {
-    pub fn new(addr: SocketAddr, f: WorkloadInformation) -> Self {
+    pub fn new(f: WorkloadInformation) -> Self {
```
hzxuzhonghu, line 47: Maybe need to pass address later

hzxuzhonghu (commented):
LGTM othewise

src/admin/mod.rs
howardjohn: Yeah we should put this in Config. Even though it's not configurable right now, in the future we will make everything in Config customizable

zhlsunshine: Okay, got it. I handled the conflict and re-submitted.

howardjohn (approved):

PR #100, opened 2022-11-09
Add some starter unit tests for `build_request`

howardjohn:


stevenctl (commented):
LGTM. I'll try to investigate a very minimal macro we can re-use to get some kind of basic table-testing structure. 

I think the macro should basically take a function, an identifier and some args and generate

```
#[test]
fn <fn_name>_<identifier>() {
  fn_name(<args...>)
}
````


Usage:

```
mod tests {
  fn the_actual_test(a: uint) {
    assert_eq!(a, 1)
  }
  
  subtest!(the_actual_test, "1", 1)
  
}
```

stevenctl (approved):

PR #103, opened 2022-11-10
Implement `baggage` header for outbound calls

howardjohn:
This allows the waypoint proxy to emit telemetry. Currently, its not used by the server-side ztunnel, but it could be in the future if/when we add L4 telemetry

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -1,4 +1,3 @@
-use rand::prelude::IteratorRandom;
 use std::collections::{HashMap, HashSet};
 use std::convert::Into;
 use std::net::{IpAddr, SocketAddr};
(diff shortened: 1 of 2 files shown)


src/workload.rs
```diff
@@ -398,6 +410,10 @@ impl WorkloadStore {
                     namespace: "".to_string(),
                     node: "".to_string(),
                     service_account: "".to_string(),
+                    workload_name: "".to_string(),
```
stevenctl, line 413: Can you do default  to save some boilerplate? 

```
..default()
```

src/workload.rs
```diff
@@ -537,7 +553,12 @@ mod tests {
             name: "".to_string(),
             namespace: "".to_string(),
             service_account: "".to_string(),
+            workload_name: "".to_string(),
```
stevenctl, line 556: We should either rely on `default::default()` more or have a test module with some example workloads to re-use in tests or both.

stevenctl (approved):

PR #104, opened 2022-11-10
improve set_gateway_ip

hzxuzhonghu:
Set it only when gateway_ip is None，

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -407,18 +407,15 @@ impl WorkloadStore {
     }
 
     fn set_gateway_ip(us: &mut Upstream) {
-        match us.workload.gateway_ip {
-            Some(_) => (),
-            None => {
-                let mut ip = us.workload.workload_ip;
-                if let Some(addr) = us.workload.waypoint_address {
-                    ip = addr;
-                }
-                us.workload.gateway_ip = Some(match us.workload.protocol {
-                    Protocol::Hbone => SocketAddr::from((ip, 15008)),
-                    Protocol::Tcp => SocketAddr::from((ip, us.port)),
-                });
+        if let None = us.workload.gateway_ip {
+            let mut ip = us.workload.workload_ip;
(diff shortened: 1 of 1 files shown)


src/workload.rs
```diff
@@ -407,15 +407,18 @@ impl WorkloadStore {
...
-            ip = us.workload.waypoint_address.unwrap();
-        }
-        if us.workload.gateway_ip.is_none() {
-            us.workload.gateway_ip = Some(match us.workload.protocol {
-                Protocol::Hbone => SocketAddr::from((ip, 15008)),
-                Protocol::Tcp => SocketAddr::from((us.workload.workload_ip, us.port)),
-            });
+        match us.workload.gateway_ip {
+            Some(_) => (),
+            None => {
```
howardjohn, line 412: nit: this can be `if let None = us.workload.gateway_ip { ... }`. Of just `if us.workload.gateway_ip.is_none()`?

src/workload.rs
```diff
@@ -407,15 +407,18 @@ impl WorkloadStore {
...
+                let mut ip = us.workload.workload_ip;
+                if let Some(addr) = us.workload.waypoint_address {
+                    ip = addr;
+                }
+                us.workload.gateway_ip = Some(match us.workload.protocol {
+                    Protocol::Hbone => SocketAddr::from((ip, 15008)),
+                    Protocol::Tcp => SocketAddr::from((ip, us.port)),
```
howardjohn, line 419:
```suggestion
                    Protocol::Tcp => SocketAddr::from((us.workload.workload_ip, us.port)),
```

We cannot access the waypoint over TCP

src/workload.rs
hzxuzhonghu: ok

src/workload.rs
hzxuzhonghu: IC

howardjohn (approved):

PR #107, opened 2022-11-11
add shutdown trigger from the admin server

zhlsunshine:
add shutdown trigger from the admin server based on `signal::Shutdown`


src/admin/mod.rs
```diff
@@ -121,9 +123,13 @@ impl Server {
                 }
             }));
 
+        let shutdown_trigger = shutdown.trigger();
         tokio::spawn(async move {
             info!("Serving admin server at {}", self.addr);
-            server.await
+            if let Err(err) = server.await {
+                eprintln!("Serving admin start failed: {err}");
```
howardjohn, line 130: we should never use println, use `error!`

src/admin/mod.rs
```diff
@@ -121,9 +123,13 @@ impl Server {
                 }
             }));
 
+        let shutdown_trigger = shutdown.trigger();
         tokio::spawn(async move {
             info!("Serving admin server at {}", self.addr);
-            server.await
+            if let Err(err) = server.await {
```
howardjohn, line 129: So this seems to wait for the server to terminate then trigger shutdown. But the server should never terminate.

I think the intent here was to offer a /quitquitquit endpoint like Envoy that explicitly calls `shutdown_trigger.shutdown_now()`

src/admin/mod.rs
zhlsunshine: okay, keep the same by using `error!`

src/admin/mod.rs
zhlsunshine: Hi @howardjohn, okay, got it. will add an new endpoint for admin server's shutdown. And I think that the error catch for admin server is necessary.  Does it make sense?

zhlsunshine:
Hi @howardjohn, already added a new /quitquitquit endpoint for admin server to graceful shutdown.

src/admin/mod.rs
```diff
@@ -19,12 +20,20 @@ use gperftools::profiler::PROFILER;
...
 
 use crate::workload::WorkloadInformation;
-use tracing::info;
+use tracing::{error, info};
+
+lazy_static! {
+    /// Channel used to send shutdown signal - wrapped in an Option to allow
+    /// it to be taken by value (since oneshot channels consume themselves on
+    /// send) and an Arc<Mutex> to allow it to be safely shared between threads
+    static ref SHUTDOWN_TX: Arc<Mutex<Option<Sender<()>>>> = <_>::default();
```
hzxuzhonghu, line 35: Why donot you instead pass a  mpsc channel to the quitquit handler

zhlsunshine: Hi @hzxuzhonghu, I think the static `SHUTDOWN_TX` may be used by other threads in admin server where need to handle shutdown cases, then it can be re-used directly.

zhlsunshine: Hi @hzxuzhonghu & @howardjohn, there are too long life time for admin server from `make_service_fn` and its `async move` to `service_fn` and its `async move`, so I suggest that a global static var can be used here for admin server's shutdown, so what do you think about it?

zhlsunshine: Hi @hzxuzhonghu,  I figured out a way to remove such global static var, and both ztunnel and admin server can use the same shutdown signal. Re-subbmitted.

src/admin/mod.rs
zhlsunshine: Hi @howardjohn, an new endpoint `quitquitquit` has been added.

src/admin/mod.rs
hzxuzhonghu: LGTM

src/admin/mod.rs
howardjohn: Thanks! Do you know what case would actually trigger server.await to return an error btw? It seemed non-obvious to me.

src/admin/mod.rs
```diff
@@ -185,6 +197,18 @@ async fn handle_pprof(_req: Request<Body>) -> Response<Body> {
     }
 }
 
+async fn handle_server_shutdown(
```
howardjohn, line 200: I think for safety we will want to make this POST only

howardjohn (commented):
LGTM with change to POST and test added. Thanks for the effort, this is great

src/admin/mod.rs
zhlsunshine: Hi @howardjohn, u r welcome, I think this kind of server will always occur some accidents especially when the IP and port are known by all. So I think it's necessary to catch the server internal error to analyze.

src/admin/mod.rs
zhlsunshine: sure, that makes sense!

zhlsunshine:
Hi @howardjohn & @hzxuzhonghu, limit the `quit` endpoint in a post method and add unit test for this new endpoint.

tests/proxy.rs
```diff
@@ -20,3 +23,33 @@ async fn test_lifecycle() {
         .expect("app shuts down")
         .expect("app exits without error")
 }
+
+#[tokio::test]
+async fn test_quit_lifecycle() {
+    // need to wait for the previous test release the port resource
+    thread::sleep(Duration::from_secs(1));
```
howardjohn, line 30: This is a sign we are not actually properly shutting down..we should fix this. Ok for now though

howardjohn (approved):

PR #109, opened 2022-11-11
performance: set TCP_NODELAY and increase copy buffer

howardjohn:
io::copy default buffer is 8k, we see increased performance with a more precise 16k buffer.

Results:

Throughput Gb/s
buffer nodelay: 7
nodelay: 4.5
neither: 3.5

Latency p50,p90,p99
buffer nodelay: 0.17ms  0.21ms  0.32ms
nodelay:  0.17ms  0.22ms  0.46ms
neither: 0.17ms  0.22ms  0.54ms

("Neither" is before, "buffer nodelay" is this PR)

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -55,12 +55,12 @@ impl Inbound {
                     cert_manager: self.cert_manager.clone(),
                 },
             };
+            let mut listener = hyper::server::conn::AddrIncoming::from_listener(self.listener)
+                .expect("hbone bind");
+            listener.set_nodelay(true);
             let incoming = hyper::server::accept::from_stream(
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (commented):
Donot we  need same config for upstream connection?

howardjohn:
oddly setting it on outbound drops us ~1Gb/s worse. Not sure why. Needs more investigation. Might make sense to merge this as-is for now and continue to tune it moving forward; hopefully soon we will have proper benchmarks as well.

hzxuzhonghu:
great test, can you share the test script. So anyone want to optimize some thing, it would be possible to reuse your effort

howardjohn:
LOCAL.md with simple iperf3

Issue #67, opened 2022-11-09
Minimal support for SOCKS5 


costinm:
SOCKS is a very simple protocol, easy to implement (probably already has rust impl). This can be optionally compiled in.

It will make it very easy to do local development without eBPF/interception/root - and using common tools (virtually all
apps support an option to use a socks proxy). 

Happy to work on it.

howardjohn:
This is merged

PR #111, opened 2022-11-11
Add Socks5 support for easier local debugging 

costinm:
This is the most basic protocol implementation, supporting only the IPv4 and IPv6 ( no hostname ).

You can use ztunnel without iptables - just run it locally, run istiod locally (see Local.md), and make requests
using a client that supports socks (virtually all):
Example: 

```
curl -x socks5://127.0.0.1:15080  -4 www.google.com
curl -x socks5://127.0.0.1:15080 -6   www.google.com
```


costinm:
Bug: https://github.com/istio/ztunnel/issues/67

This is also usable with any of the Tun2Socks implementations that exist, for experimenting with TUN-based capture. 
It will also allow ztunnel to be used on VMs or environments without root or NET_ADMIN permissions ( hopefully - still need to test )

src/proxy/socks5.rs
```diff
@@ -0,0 +1,176 @@
...
+impl Socks5 {
+    pub async fn new(
+        cfg: Config,
+        cert_manager: identity::SecretManager,
+        workloads: WorkloadInformation,
+    ) -> Result<Socks5, Error> {
+        let listener: TcpListener = TcpListener::bind("127.0.0.1:15080")
+            .await
+            .map_err(Error::Bind)?;
+        match socket::set_transparent(&listener) {
```
howardjohn, line 32: don't think we need this, its for TPROXY

src/proxy/socks5.rs
```diff
@@ -0,0 +1,176 @@
...
+                }
+                Err(e) => error!("Failed TCP handshake {}", e),
+            }
+        }
+    }
+}
+
+async fn handle(oc: OutboundConnection, mut stream: TcpStream) -> Result<(), anyhow::Error> {
+
+    let mut version = [0u8];
```
howardjohn, line 77: this is very nit, but fwiw you can do 

```
let mut buf = [0u8; 2];
let [version, nmethods] = stream.read_exact(&mut buf).await?;
```

to avoid some of the `[0]` indexing and avoid 1 syscall

src/proxy/socks5.rs
```diff
@@ -0,0 +1,176 @@
...
+    let nmethods = nmethods[0];
+
+    if nmethods == 0 {
+        return Err(anyhow::anyhow!("Invalid auth methods"));
+    }
+
+    let mut methods = vec![0u8; nmethods as usize];
+    stream.read_exact(&mut methods).await?;
+
+    let a =  methods.into_iter().find(|m| *m == 0);
```
howardjohn, line 97: nit: `any` instead of `find` just returns a `bool`

src/proxy/socks5.rs
```diff
@@ -0,0 +1,176 @@
...
+    if a == None {
+        return Err(anyhow::anyhow!("unsupported method"));
+    }
+
+    stream.write_all(&[0x05, 0x00]).await?;
```
howardjohn, line 102: might be nice to comment what these magic values are?

howardjohn (commented):
Nice :+1:

src/proxy/socks5.rs
costinm: The joy of cut&pasting... 

BTW - the code seems to use 'orig_dst_addr' - AFAIK this is not working for TPROXY, only for REDIRECT ( with TPROXY - it's the local address ). I'll try to add the shell script that sets TPROXY with a single namespace and test.

howardjohn (approved):

PR #112, opened 2022-11-12
Implement to configure the Tokio worker threads number

haiyuewa:
Builds Tokio Runtime with custom configuration values, instead of fixed setting by macro '#[tokio::main]'.

Since the ztunnel works on node level, the workloads will be heavy, add one environment variable for tuning the Tokio worker threads number.

Use 'istioctl --set values.ztunnel.env.ZTUNNEL_WOKRER_THREADS=5' to set worker threads if needed, and default value is 2.

Signed-off-by: Haiyue Wang <haiyue.wang@intel.com>

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -27,7 +27,12 @@ pub struct Config {
     pub auth: identity::AuthSource,
 
     pub termination_grace_period: time::Duration,
+
+    /// Specify the number of worker threads the Tokio Runtime will use.
+    pub num_worker_threads: usize,
 }
+
+const DEFAULT_WOKRER_THREADS: usize = 2;
 
 impl Default for Config {
     fn default() -> Config {
@@ -53,6 +58,16 @@ impl Default for Config {
             auth: identity::AuthSource::Token(PathBuf::from(
                 r"./var/run/secrets/tokens/istio-token",
             )),
+
+            num_worker_threads: std::env::var("ZTUNNEL_WOKRER_THREADS")
+                .ok()
+                .map(|v| {
+                    v.parse::<usize>()
+                        .ok()
+                        .filter(|n| *n > 0)
+                        .unwrap_or(DEFAULT_WOKRER_THREADS)
+                })
+                .unwrap_or(DEFAULT_WOKRER_THREADS),
         }
     }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,7 +10,7 @@ use ztunnel::*;
 // #[global_allocator]
 // static GLOBAL: tcmalloc::TCMalloc = tcmalloc::TCMalloc;
 
-fn main() {
+fn main() -> anyhow::Result<()> {
     telemetry::setup_logging();
     let config = ztunnel::config::Config {
         ..Default::default()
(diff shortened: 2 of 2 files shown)


src/main.rs
```diff
@@ -10,11 +10,30 @@ use ztunnel::*;
 // #[global_allocator]
 // static GLOBAL: tcmalloc::TCMalloc = tcmalloc::TCMalloc;
 
-#[tokio::main(worker_threads = 2)]
-async fn main() -> anyhow::Result<()> {
+const DEFAULT_WOKRER_THREADS: usize = 2;
+
+fn main() {
     telemetry::setup_logging();
     let config = ztunnel::config::Config {
```
howardjohn, line 17: I'd like to have all config in Config. can we move it there?

The idea is we will have some more unified way to configure things rather than just env vars in the future, so having everything in one place makes that future proofed.

src/main.rs
```diff
@@ -10,11 +10,30 @@ use ztunnel::*;
...
         ..Default::default()
     };
-    app::spawn(signal::Shutdown::new(), config).await
+
+    let worker_threads = std::env::var("ZTUNNEL_WOKRER_THREADS")
```
howardjohn, line 21:
```suggestion
    let worker_threads = std::env::var("ZTUNNEL_WORKER_THREADS")
```

howardjohn (commented):
Looks good. Longer term we will probably want to put the admin parts (XDS and admin server) on a dedicated thread as well.

src/main.rs
haiyuewa: Got the desgin idea, will change it.

src/config.rs
```diff
@@ -27,8 +27,13 @@ pub struct Config {
     pub auth: identity::AuthSource,
 
     pub termination_grace_period: time::Duration,
+
+    /// Specify the number of worker threads the Tokio Runtime will use.
+    pub num_worker_threads: usize,
```
hzxuzhonghu, line 32: I am thinking we should not make every config in this Struct, maybe abstract some scopes

haiyuewa: The current 'Config' is so flat ? Like put the same usages members into new struct:

pub struct xds_config {
 /// Filepath to a local xds file for workloads, as YAML.
    pub local_xds_path: Option<String>,
    /// If true, on-demand XDS will be used
    pub xds_on_demand: bool,"
}

src/main.rs
```diff
@@ -10,11 +10,18 @@ use ztunnel::*;
...
     };
-    app::spawn(signal::Shutdown::new(), config).await
+
+    tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(config.num_worker_threads)
+        .enable_all()
+        .build()
+        .unwrap()
+        .block_on(async move {
+            let _ = app::spawn(signal::Shutdown::new(), config).await;
```
howardjohn, line 25: This seems like we ignore errors now; previously tokio handled them. can we add it back?

howardjohn (commented):
LGTM once we get error handling back. Thanks!

howardjohn (approved):
thanks!

PR #118, opened 2022-11-14
verify certs/sans on outbound

stevenctl:
This is the result of a short bit of hacking around and getting familiar with boringssl, needs to be reorganized and cleaned up a lot. 

A couple of mysteries:

The first request will always log this failure (but the request succeeds)

```
2022-11-14T18:56:32.873206Z  INFO ztunnel::proxy::outbound: Proxying to 127.0.0.1:8080 using HBONE via 127.0.0.1:15008 type Direct
2022-11-14T18:56:32.875826Z  INFO fetch_certificate{id=spiffe://cluster.local/ns/default/sa/default}:fetch_certificate: ztunnel::tls::boring: ssl: X509VerifyResult { code: 20, error: "unable to get local issuer certificate" }
2022-11-14T18:56:32.875869Z  INFO fetch_certificate{id=spiffe://cluster.local/ns/default/sa/default}:fetch_certificate: ztunnel::tls::boring: ssl: X509VerifyResult { code: 21, error: "unable to verify the first certificate" }
2022-11-14T18:56:32.930733Z  INFO ztunnel::proxy::inbound: tls: accepting connection to 127.0.0.1:15008 (spiffe://cluster.local/ns/default/sa/default)
```

This log from the tls verify callback is always printed twice per request:

```
2022-11-14T18:56:50.053081Z  INFO ztunnel::proxy::inbound: tls: accepting connection to 127.0.0.1:15008 (spiffe://cluster.local/ns/default/sa/default)
2022-11-14T18:56:50.106757Z  INFO ztunnel::tls::boring: verified SAN spiffe://cluster.local/ns/default/sa/default
2022-11-14T18:56:50.106858Z  INFO ztunnel::tls::boring: verified SAN spiffe://cluster.local/ns/default/sa/default
```

### DIFF
--- a/src/identity/caclient.rs
+++ b/src/identity/caclient.rs
@@ -71,7 +71,10 @@ impl CaClient {
         Ok(tls::cert_from(
             &pkey,
             resp.cert_chain.first().unwrap().as_bytes(),
-            resp.cert_chain[1..].into_iter().map(|s| s.as_bytes()).collect(),
+            resp.cert_chain[1..]
+                .into_iter()
+                .map(|s| s.as_bytes())
+                .collect(),
         ))
     }
 }
--- a/src/identity/caclient.rs
+++ b/src/identity/caclient.rs
@@ -66,15 +66,17 @@ impl CaClient {
                 )]),
             }),
         };
-        let resp = self.client.create_certificate(req).await?;
-        let resp = resp.into_inner();
-        Ok(tls::cert_from(
-            &pkey,
-            resp.cert_chain.first().unwrap().as_bytes(),
+        let resp = self.client.create_certificate(req).await?.into_inner();
+
+        let leaf = resp.cert_chain.first().unwrap().as_bytes();
+        let chain = if resp.cert_chain.len() > 1 {
             resp.cert_chain[1..]
                 .into_iter()
                 .map(|s| s.as_bytes())
-                .collect(),
-        ))
+                .collect()
+        } else {
+            vec![]
+        };
+        Ok(tls::cert_from(&pkey, leaf, chain))
     }
 }
--- a/src/identity/caclient.rs
+++ b/src/identity/caclient.rs
@@ -45,7 +45,7 @@ impl CaClient {
     }
 
     #[instrument(skip_all)]
-    pub async fn fetch_certificate(&mut self, id: Identity) -> Result<tls::Certs, Error> {
+    pub async fn fetch_certificate(&mut self, id: &Identity) -> Result<tls::Certs, Error> {
         let cs = tls::CsrOptions {
             san: id.to_string(),
         }
--- a/src/identity/caclient.rs
+++ b/src/identity/caclient.rs
@@ -71,6 +71,7 @@ impl CaClient {
         Ok(tls::cert_from(
             &pkey,
             resp.cert_chain.first().unwrap().as_bytes(),
+            resp.cert_chain[1..].into_iter().map(|s| s.as_bytes()).collect(),
         ))
     }
 }
--- a/src/identity/caclient.rs
+++ b/src/identity/caclient.rs
@@ -70,10 +70,7 @@ impl CaClient {
 
         let leaf = resp.cert_chain.first().unwrap().as_bytes();
         let chain = if resp.cert_chain.len() > 1 {
-            resp.cert_chain[1..]
-                .into_iter()
-                .map(|s| s.as_bytes())
-                .collect()
+            resp.cert_chain[1..].iter().map(|s| s.as_bytes()).collect()
         } else {
             vec![]
         };
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -150,7 +150,7 @@ impl OutboundConnection {
                 let mut request_sender = if self.cfg.tls {
                     let id = req.source.identity();
                     let cert = self.cert_manager.fetch_certificate(&id).await?;
-                    let connector = cert.connector(&id)?.configure()?;
+                    let connector = cert.connector(&req.destination_identity)?.configure()?;
                     let tcp_stream = TcpStream::connect(req.gateway).await?;
                     let tls_stream = connect_tls(connector, tcp_stream).await?;
                     let (request_sender, connection) = builder
@@ -338,9 +338,9 @@ struct Request {
     direction: Direction,
     source: Workload,
     destination: SocketAddr,
+    destination_identity: Option<identity::Identity>,
     gateway: SocketAddr,
     request_type: RequestType,
-    destination_identity: Option<identity::Identity>,
 }
 
 #[derive(Debug)]
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -231,6 +231,7 @@ impl OutboundConnection {
                 protocol: Protocol::Tcp,
                 source: source_workload,
                 destination: target,
+                destination_identity: None,
                 gateway: target,
                 direction: Direction::Outbound,
                 request_type: RequestType::Passthrough,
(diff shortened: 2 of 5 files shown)


src/tls/boring.rs
```diff
@@ -162,6 +164,50 @@ impl Certs {
...
+    // TODO other logic from 
+    fn build(self) -> impl Fn(bool, &mut X509StoreContextRef) -> bool { 
+        move |verified,ctx| {
+            if !verified {
+                // log errs from ctx? definitely increment counters
+                return false;
+            }
+            let v = self.clone(); // TODO appease borrow checking in a cleaner way 
+            match v {
+                Verifier::None => true,
```
stevenctl, line 187: TODO move the match outside the closure

howardjohn:
> The first request will always log this failure (but the request succeeds)

I think that part is the control plane connection to the CA probably fwiw

src/proxy/outbound.rs
```diff
@@ -126,8 +126,8 @@ impl OutboundConnection {
 
                 let mut request_sender = if self.cfg.tls {
                     let id = req.source.identity();
-                    let cert = self.cert_manager.fetch_certificate(id).await?;
-                    let connector = cert.connector()?.configure()?;
+                    let cert = self.cert_manager.fetch_certificate(id.clone()).await?;
```
howardjohn, line 129: nit: we can probably user references everywhere we take an Identity

howardjohn: Maybe not (easily) if its stored in a struct though

src/proxy/outbound.rs
```diff
@@ -308,11 +310,15 @@ async fn connect_tls(
 ) -> Result<tokio_boring::SslStream<TcpStream>, tokio_boring::HandshakeError<TcpStream>> {
     connector.set_verify_hostname(false);
     connector.set_use_server_name_indication(false);
-    let addr = stream.local_addr();
-    connector.set_verify_callback(boring::ssl::SslVerifyMode::PEER, move |_, x509| {
-        info!("TLS callback for {:?}: {:?}", addr, x509.error());
-        true
-    });
+    // let addr = stream.local_addr();
```
howardjohn, line 313: This seems like we are not verifying if its commented out; I think we need custom verification callback for SPIFFE possibly

src/tls/boring.rs
```diff
@@ -132,27 +138,23 @@ impl Certs {
...
 
         Ok(conn.build())
     }
-    pub fn connector(&self) -> Result<ssl::SslConnector, Error> {
+    pub fn connector(&self, id: Identity) -> Result<ssl::SslConnector, Error> {
         let mut conn = ssl::SslConnector::builder(ssl::SslMethod::tls_client())?;
 
         conn.set_private_key(&self.key)?;
         conn.set_certificate(&self.cert)?;
+        conn.set_ca_file("./var/run/secrets/istio/root-cert.pem")?;
```
howardjohn, line 154: This is the CA of XDS, not workloads; we can do this for short term but we need to fetch it from the returned CSR like we do in Envoy mode long term

src/tls/boring.rs
```diff
@@ -162,6 +164,50 @@ impl Certs {
...
 }
 
+#[derive(Clone)]
+enum Verifier {
+    // Does not verify the identity.
+    None,
+
+    // Allows exactly one identity
+    Exact(Identity),
+    // TODO other types could be allowlist/denylist or some combo for RBAC
```
howardjohn, line 174: I suspect RBAC will probably be denied at the HBONE request level. But just my initial impression

src/proxy/outbound.rs
stevenctl: AFAICT calling set_verify_callback or set_verify multiple times is overriding and just using the value from the last call. The passed in `connector` has the verifier.

src/tls/boring.rs
stevenctl: I've switched it to add the certs from the chain. 

Envoy does this differently.. seems like in their verification callback they check the intermediates.

https://github.com/envoyproxy/envoy/blob/8618789f14fc8314db3063d7ccde265c5198b2bc/source/extensions/transport_sockets/tls/context_impl.cc#L1235

I tried something similar and it failed with `unable to get local issuer certificate`.

At one point I was trying to register the root/intermediates (we ofc just have a root) using `add_extra_chain_cert` but I don't think that does what I think it does. 

https://github.com/envoyproxy/envoy/blob/8618789f14fc8314db3063d7ccde265c5198b2bc/source/extensions/transport_sockets/tls/context_impl.cc#L1285

src/identity/caclient.rs
```diff
@@ -57,6 +57,7 @@ impl CaClient {
         Ok(tls::cert_from(
             &pkey,
             resp.cert_chain.first().unwrap().as_bytes(),
+            resp.cert_chain[1..].into_iter().map(|s| s.as_bytes()).collect(),
```
howardjohn, line 60: we should validate `len > 1`

src/tls/boring.rs
```diff
@@ -132,28 +142,27 @@ impl Certs {
         conn.check_private_key()?;
 
         // Ensure that client certificates are validated when present.
-        conn.set_verify(ssl::SslVerifyMode::FAIL_IF_NO_PEER_CERT);
-        conn.set_verify_callback(ssl::SslVerifyMode::PEER, |_, _| {
+        conn.set_verify_callback(Self::verify_mode(), Verifier::None.callback());
+        conn.set_verify_callback(Self::verify_mode(), |_, _| {
```
howardjohn, line 146: is having 2 set_verify_callback expected

src/tls/boring.rs
```diff
@@ -132,28 +142,27 @@ impl Certs {
...
     }
-    pub fn connector(&self) -> Result<ssl::SslConnector, Error> {
+    pub fn connector(&self, id: &Identity) -> Result<ssl::SslConnector, Error> {
         let mut conn = ssl::SslConnector::builder(ssl::SslMethod::tls_client())?;
 
         conn.set_private_key(&self.key)?;
         conn.set_certificate(&self.cert)?;
+        for chain_cert in self.chain.iter() {
+            // TODO avoid cloning even though this method requires owned X509
+            // conn.add_extra_chain_cert(chain_cert.clone())?;
```
howardjohn, line 161: linkerd is doing this (including clone). Not sure how it compares to add_cert

src/tls/boring.rs
```diff
@@ -132,28 +142,27 @@ impl Certs {
         conn.check_private_key()?;
 
         // Ensure that client certificates are validated when present.
-        conn.set_verify(ssl::SslVerifyMode::FAIL_IF_NO_PEER_CERT);
-        conn.set_verify_callback(ssl::SslVerifyMode::PEER, |_, _| {
+        conn.set_verify_callback(Self::verify_mode(), Verifier::None.callback());
```
howardjohn, line 145: is None TODO or expected long term?

src/proxy/outbound.rs
```diff
@@ -319,6 +320,7 @@ struct Request {
     destination: SocketAddr,
     gateway: SocketAddr,
     request_type: RequestType,
+    destination_identity: identity::Identity,
```
howardjohn, line 323: prob needs to be an Option<> for non-HBONE cases

src/tls/boring.rs
stevenctl: Long term for inbound I think. Unless we decide to add some extra validation. 

It also seems like it may be better to capture some verification errors outside of these callbacks; if we can pipe the san failures through boringssl we wouldn't need any callback here.

howardjohn (approved):
Looks like a great start. I am hoping to add e2e tests for the full proxy flow at which point we can probably test this reasonably

PR #123, opened 2022-11-15
Fix workload removal from VIP

howardjohn:
Previously VIP always grew

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -454,10 +454,12 @@ pub enum WorkloadError {
 
 #[cfg(test)]
 mod tests {
+    use std::net::{Ipv4Addr, Ipv6Addr};
+
+    use bytes::Bytes;
+
     use crate::xds::istio::workload::Port as XdsPort;
     use crate::xds::istio::workload::PortList as XdsPortList;
-    use bytes::Bytes;
-    use std::net::{Ipv4Addr, Ipv6Addr};
 
     use super::*;
 
@@ -653,6 +655,7 @@ mod tests {
         wi.remove("127.0.0.1".to_string());
         assert_vips(&wi, vec![]);
     }
+
     #[track_caller]
     fn assert_vips(wi: &WorkloadStore, want: Vec<&str>) {
         let mut wants: HashSet<String> = HashSet::from_iter(want.iter().map(|x| x.to_string()));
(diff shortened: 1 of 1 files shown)


src/workload.rs
```diff
@@ -329,6 +329,8 @@ impl WorkloadInformation {
 #[derive(serde::Serialize, Default, Debug)]
 pub struct WorkloadStore {
     workloads: HashMap<IpAddr, Workload>,
+    // workload_to_vip maintains a mapping of workload IP to VIP. This is used only to handle removals.
+    workload_to_vip: HashMap<IpAddr, HashSet<(SocketAddr, u16)>>,
```
zhlsunshine, line 333: Shall we consider the dual stack in advance? If yes, we may need to do reasonable design for it.

hzxuzhonghu, line 333: I think we can record the vips in `Workload`, then can get ride of this map

howardjohn: We need a way to efficiently lookup a VIP though

howardjohn: What is the issue? IpAddr and SocketAddr can both be IPv4 or IPv6, right?

stevenctl (approved):

src/workload.rs
zhlsunshine: I mean that a workload with IPv4 and IPv6 would have 2 entries in `workloads`, is that what we want?

zhlsunshine:
```
pub struct AdsClient {
    config: Config,
    /// Stores all known workload resources (by name)
    workloads: HashSet<String>,
...
```
As you can see, the workloads are identified by name in `AdsClient`, I think there may be an issue.

howardjohn: Yes, we want 2 workloads in that case IMO. `name` is IP

howardjohn: Well its a bit confusing. ResourceName == IP, but `name` is the pod name

zhlsunshine: yes, shall I change it as?
```
workloads: HashSet<IpAddr>,
```

zhlsunshine: Sorry, @howardjohn, I should do this change:
```
pub struct ResourceKey {
    pub resource_name: String,
    pub type_url: String,
}
```

howardjohn: No it's good as is

zhlsunshine: and 
```
pub struct XdsResource<T: prost::Message> {
    pub resource_name: String,
    pub resource: T,
}
```

howardjohn: I'm not sure I understand what you mean

zhlsunshine: > Well its a bit confusing. ResourceName == IP, but `name` is the pod name

I understand that everything is okay for now except there are some field name confusing in `ResourceKey` definition. 
If okay, no anything need to be changed here.

PR #124, opened 2022-11-15
Use env to generate node id for xds client

PlatformLC:
* Make node_id generated from env rather than hardcoded ones.


src/xds/client.rs
```diff
@@ -94,10 +94,21 @@ impl Config {
...
             demand_tx: tx,
+            node_id: generate_node_id(), 
         }
     }
 }
 
+fn generate_node_id() -> String {
+    // TODO: may add new type, but this requires changing together with pilot-discovery 
+    let node_type = "sidecar";
+    let ip = std::env::var("INSTANCE_IP").unwrap_or_else(|_| "1.1.1.1".to_string());
```
zirain, line 105: add fn like envOrDefault?

hzxuzhonghu (commented):
I donot think in the long run, it is viable to set node id via env. Because ztunnel should be deployed via daemonset. So not possible to differentiate it

src/xds/client.rs
```diff
@@ -94,10 +94,21 @@ impl Config {
...
+    let pod_name = std::env::var("POD_NAME").unwrap_or_else(|_| "test".to_string());
```
hzxuzhonghu, line 106: POD_NAME makes sense

PlatformLC:
> I donot think in the long run, it is viable to set node id via env. Because ztunnel should be deployed via daemonset. So not possible to differentiate it

@hzxuzhonghu , thanks for your comments.
Retrieving info from env just refers current pilot-agent implementation.
[pilot-agent](https://github.com/istio/istio/blob/experimental-ambient/pilot/cmd/pilot-agent/options/options.go#L31)
```
	InstanceIPVar        = env.Register("INSTANCE_IP", "", "")
	PodNameVar           = env.Register("POD_NAME", "", "")
	PodNamespaceVar      = env.Register("POD_NAMESPACE", "", "")
```

Regarding deployment via ds.  IMHO, INSTANCE_IP, POD_NAME should be different among ztunnels, while namespace is a bit meaningless. However I think we may follow this format style to make the compatibility with istiod(pilot-discovery) side.

src/xds/client.rs
hzxuzhonghu, line 105: nit : unwrap_or

hzxuzhonghu:
make sense @PlatformLC

src/xds/client.rs
PlatformLC: clippy lint warns about this.

[Clippy Lints](https://rust-lang.github.io/rust-clippy/master/index.html#or_fun_call)

unwrap_or_else is more suitable for  default is a function call.

howardjohn (approved):

Issue #120, opened 2022-11-14
If source not found, ztunnel panics


costinm:
stack backtrace:
   0: rust_begin_unwind
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/panicking.rs:142:14
   2: core::panicking::panic_display
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/panicking.rs:72:5
   3: core::panicking::panic_str
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/panicking.rs:56:5
   4: core::option::expect_failed
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/option.rs:1880:5
   5: core::option::Option<T>::expect
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/option.rs:738:21
   6: ztunnel::proxy::outbound::OutboundConnection::build_request::{{closure}}
             at ./src/proxy/outbound.rs:213:35
   7: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/future/mod.rs:91:19

PR #125, opened 2022-11-15
Fix build request panic

hzxuzhonghu:
Fix https://github.com/istio/ztunnel/issues/120

### DIFF
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -93,11 +93,11 @@ struct OutboundConnection {
 }
 
 impl OutboundConnection {
-    async fn proxy(&self, mut stream: TcpStream) -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


src/proxy/outbound.rs
```diff
@@ -93,11 +93,11 @@ struct OutboundConnection {
 }
 
 impl OutboundConnection {
-    async fn proxy(&self, mut stream: TcpStream) -> Result<(), Error> {
+    async fn proxy(&self, mut stream: TcpStream) -> anyhow::Result<()> {
```
howardjohn, line 96: IMO we should be using proper Error enums where feasible, not just untyped anyhow results. This can become a new enum `UnknownSource`

howardjohn (approved):

src/proxy/outbound.rs
hzxuzhonghu: Sure， can do it in following pr

PR #127, opened 2022-11-15
add testcase for insert_xds_workload

KfreeZ:
add testcase for "insert_xds_workload", it also invokes "find_upstream" to get the upstream from workload and check.

A question: when remove a workload, it is necessary to remove the upstream also? I am afraid it's not removed yet

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -675,19 +675,25 @@ mod tests {
             canonical_revision: "test-cVer".to_string(),
 
             node: "test-Node".to_string(),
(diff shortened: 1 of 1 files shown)


howardjohn:
I think this may be fairly overlapping with https://github.com/istio/ztunnel/pull/123 which handles the removal

KfreeZ:
> I think this may be fairly overlapping with #123 which handles the removal

yes, exactly, haha

PR #128, opened 2022-11-15
fix trust_domain hardcode

zirain:
Signed-off-by: hejianpeng <hejianpeng2@huawei.com>


howardjohn (commented):
fyi: https://github.com/istio/ztunnel/pull/115/files

PR #133, opened 2022-11-16
chore: fix copyright banner

daixiang0:
Signed-off-by: Loong <loong.dai@intel.com>

### DIFF
--- a/proto/xds.proto
+++ b/proto/xds.proto
@@ -1,3 +1,17 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 syntax = "proto3";
 
 // GRPC package - part of the URL. Service is added.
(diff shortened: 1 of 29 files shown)


daixiang0:
I will run `make update-common` after https://github.com/istio/common-files/pull/717 merged.

proto/xds.proto
```diff
@@ -7,166 +21,172 @@ package envoy.service.discovery.v3;
...
-    // A developer-facing error message, which should be in English. Any
-    // user-facing error message should be localized and sent in the
-    // [google.rpc.Status.details][google.rpc.Status.details] field, or localized by the client.
-    string message = 2;
-
-    // A list of messages that carry the error details.  There is a common set of
-    // message types for APIs to use.
-    // google.protobuf.Any
-    repeated google.protobuf.Any details = 3;
+  // The status code, which should be an enum value of
```
howardjohn, line 27: This shouldn't change...?

proto/xds.proto
```diff
@@ -1,3 +1,17 @@
+// Copyright Istio Authors
```
howardjohn, line 1: This is not copyright Istio Authors, its copy+pasted from Envoy

proto/xds.proto
daixiang0: Yeah, it is about proto format, let me revert.

proto/xds.proto
daixiang0: Sure, let me fix it.

daixiang0: I have reverted those changes, not sure what is the right repo and path they from, so keep it as before.

Maybe we can introduce a new target like `update-xds` in another PR.

howardjohn: We don't want to update, it's pruned by hand to remove unused fields to remove bloat

howardjohn (approved):

PR #141, opened 2022-11-16
outbound: implement in memory fast path, bypassing full HBONE connection

howardjohn:
On the surface, this increases performance of same node by a lot. However, this actually makes testing way easier. Before, we had to have iptable rules to rewrite 15088 to 15008 to handle the node-local case. By having the fastpath, its actually easier to test since we can avoid any iptables rules.

In the future, we need to enforce RBAC obviously. The `DirectPath` enum will need to encode whatever info about the client that is needed to perform these checks; today there is none, so no action is needed.

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -3,12 +3,13 @@ use std::os::unix::io::RawFd;
 
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Method, Request, Response, Server, StatusCode};
-use tokio::net::TcpListener;
+use tokio::net::{TcpListener, TcpStream};
 use tokio_stream::StreamExt;
 use tracing::{error, info, warn};
 
 use crate::config::Config;
 use crate::identity;
+use crate::proxy::inbound::InboundConnect::Hbone;
 use crate::tls::TlsError;
 use crate::workload::WorkloadInformation;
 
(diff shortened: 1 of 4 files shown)


src/proxy/inbound.rs
```diff
@@ -107,41 +108,61 @@ impl Inbound {
...
+                tokio::task::spawn(async move {
+                    match request_type {
+                        InboundConnect::DirectPath(mut incoming) => {
+                            tokio::io::copy_bidirectional(&mut incoming, &mut stream)
+                                .await
+                                .expect("internal server copy");
+                        }
+                        Hbone(req) => match hyper::upgrade::on(req).await {
+                            Ok(mut upgraded) => {
+                                super::copy_hbone("hbone server", &mut upgraded, &mut stream)
```
hzxuzhonghu, line 134: How is copy_hbone different from copy_bidirectional

howardjohn: It has its own buffer. We could probably clean it up. However, pretty soon we are going to need more control than `copy_bidirectional` gives us is my guess.

howardjohn: Things like reporting metrics, etc

howardjohn: Also note I didn't add this in the PR, it was always used - just moving code around

src/proxy/inbound.rs
```diff
@@ -107,41 +108,61 @@ impl Inbound {
...
+        match stream {
+            Err(err) => {
+                warn!("connect to {} failed: {}", addr, err);
+                Err(err)
+            }
+            Ok(stream) => {
+                let mut stream = stream;
+                tokio::task::spawn(async move {
+                    match request_type {
+                        InboundConnect::DirectPath(mut incoming) => {
```
stevenctl, line 127: nit: can we move the `match request_type` into an impl on the `InboundConnect` enum to avoid the nesting?

stevenctl (approved):

PR #142, opened 2022-11-17
chore: fix lint issues

daixiang0:
`lint-copyright-banner` would check `.h` and `.proto`, which is not expected for `boringssl-fips` and `proto/xds.proro`.

As it is from `common-files`, one way is to overload it at local, or just remove it.

Update `common-files` as #133 said.


daixiang0:
@howardjohn what do you think?

howardjohn:
If we are removing the lint check we should remove the headers too. They either need to be checked or they shouldn't be there at all, we can't have unchecked requirements

daixiang0:
After override:
```
root:[ztunnel]$ make lint
useradd: UID 0 is not unique
Makefile.core.mk:11: warning: overriding recipe for target 'lint-copyright-banner'
common/Makefile.common.mk:38: warning: ignoring old recipe for target 'lint-copyright-banner'
cargo clippy
    Finished dev [unoptimized + debuginfo] target(s) in 0.09s
root:[ztunnel]$
```

Makefile.overrides.mk
```diff
@@ -1,2 +1,11 @@
 # Use the build container by default
 BUILD_WITH_CONTAINER ?= 1
+
+# sync from common/Makefile.common.mk
+FINDFILES=find . \( -path ./common-protos -o -path ./.git -o -path ./out -o -path ./.github -o -path ./licenses -o -path ./vendor \) -prune -o -type f
+XARGS = xargs -0 -r
+
+# override target in common/Makefile.common.mk, only check golang and rust codes
+lint-copyright-banner:
```
howardjohn, line 9: do we need it in both places?

howardjohn (approved):

Makefile.overrides.mk
daixiang0: It is useless, let me remove it.

PR #145, opened 2022-11-17
Add a script to set tproxy capture

costinm:
For testing on local dev machine - also for running ztunnel as sidecar without the REDIRECT bug (but with NET_ADMIN).
TPROXY is also required for UDP.

### DIFF
--- a/scripts/tproxy.sh
+++ b/scripts/tproxy.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+
+# Init the base set of tables and routes
+init() {
+  # Anything with the mark 15001 will be sent to loopback
+  ip -4 rule add fwmark 15001 lookup 15001
+  ip -4 route add local default dev lo table 15001
+
+  iptables -t mangle -N ZT_CAPTURE_EGRESS
+  iptables -t mangle -A ZT_CAPTURE_EGRESS -j MARK --set-mark 15001
+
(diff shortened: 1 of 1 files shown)


stevenctl (approved):

shashankram:
> For testing on local dev machine - also for running ztunnel as sidecar without the REDIRECT bug (but with NET_ADMIN). TPROXY is also required for UDP.

@costinm what's the bug with REDIRECT and Ztunnel? With ztunnel as a sidecar, where is TPROXY being used?

howardjohn:
https://github.com/istio/istio/issues/38982

costinm:
Also REDIRECT is not usable for UDP - no way to recover the original dst
addr, since REDIRECT modified the packet.

shashankram:
> Also REDIRECT is not usable for UDP - no way to recover the original dst addr, since REDIRECT modified the packet.

@costinm I am trying to understand why TPROXY UDP capture is needed for ztunnels. Is it to support UDP apps or something else? With Ztunnel and HBONE, dport is always 15008|6, so is the REDIRECT bug still an issue here?

costinm:
Since we're talking about UDP - I want to once again raise awareness of the
WebTransport standard for bi-directional stream and
datagram communication over HTTP/2 CONNECT.

While Istio is commonly used only between workloads, having a protocol with
broad support in browsers and devices can greatly expand
our adoption and use ( and bring more security everywhere ).

PR #146, opened 2022-11-17
Update ssl connector setting

hzxuzhonghu:


howardjohn (commented):
I think this is basically a subset of https://github.com/istio/ztunnel/pull/118, does it make sense to just go with that?

hzxuzhonghu:
@howardjohn Tls 1.3 does  not need to set cipher, and also these ciphers are not for tls 1.3

howardjohn (approved):

PR #151, opened 2022-11-17
Initial full request end to end tests

howardjohn:
This just adds a fairly simple "hello world" hitting the full system end to end.

Currently tested: socks5 outbound, TCP proxying, local XDS file config

HBONE is close to being able to be added, just need the fake CA from Brian's PR.

Major changes here:
* Bind to random ports everywhere; this involves passing hbone port around some places
* Move test helpers to src/ not tests/ since you cannot import between test modules.
* Allow turning off xds server
* Create a test Echo server

### DIFF
--- a/src/proxy/socks5.rs
+++ b/src/proxy/socks5.rs
@@ -15,6 +15,7 @@ pub struct Socks5 {
     cfg: Config,
     cert_manager: identity::SecretManager,
     workloads: WorkloadInformation,
+    hbone_port: u16,
     listener: TcpListener,
 }
 
@@ -22,6 +23,7 @@ impl Socks5 {
     pub async fn new(
         cfg: Config,
         cert_manager: identity::SecretManager,
+        hbone_port: u16,
         workloads: WorkloadInformation,
     ) -> Result<Socks5, Error> {
         let listener: TcpListener = TcpListener::bind(cfg.socks5_addr)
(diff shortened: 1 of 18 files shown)


src/proxy/socks5.rs
```diff
@@ -15,13 +15,15 @@ pub struct Socks5 {
     cfg: Config,
     cert_manager: identity::SecretManager,
     workloads: WorkloadInformation,
+    hbone_port: u16,
```
stevenctl, line 18: should this go on `Config` to avoid changing so many signatures?

howardjohn: The issue is we don't know the port until after `inbound` starts

howardjohn: One way we could do it is make it an Option and then swap it in later, but its a bit wonky...

stevenctl: yeah misunderstood at first

src/test_helpers/app.rs
```diff
@@ -0,0 +1,100 @@
...
+            self.proxy_addresses.socks5,
+            IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)),
+        );
+
+        let mut stream = TcpStream::connect(socks_addr).await.expect("must connect");
+
+        let addr_type = if addr.ip().is_ipv4() { 0x01u8 } else { 0x04u8 };
+        stream
+            .write_all(&[
+                0x05u8, // socks5
```
stevenctl, line 72: Where do these come from?

howardjohn: I followed https://medium.com/@nimit95/socks-5-a-proxy-protocol-b741d3bec66c since it was a nice explanation, but largely from the RFC and mirroring the socks5 server we have

stevenctl (approved):
lgtm

PR #154, opened 2022-11-18
Remove the duplication in inbound server

zhlsunshine:
Remove the duplication in inbound server for both TCP and TLS stream by generic mechanism of Rust.

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -12,11 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::clone::Clone;
+use std::marker::Send;
 use std::net::SocketAddr;
 use std::os::unix::io::RawFd;
 
+use hyper::server::{accept::Accept, Builder};
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Method, Request, Response, Server, StatusCode};
+use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::net::{TcpListener, TcpStream};
 use tokio_stream::StreamExt;
 use tracing::{error, info, warn};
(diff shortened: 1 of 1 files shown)


src/proxy/inbound.rs
```diff
@@ -57,25 +61,45 @@ impl Inbound {
         })
     }
 
+    pub(super) async fn start<I>(&self, builder: Builder<I>)
+    where
+        I: Accept,
+        <I as Accept>::Conn: AsyncRead + AsyncWrite + Unpin + Send + 'static,
+        <I as Accept>::Error: std::error::Error + Send + Sync + 'static,
```
hzxuzhonghu, line 68: This is so complex

hzxuzhonghu (commented):
It is very possible that tcp and tls service builder could be different later. I am not sure if this is right

src/proxy/inbound.rs
zhlsunshine: Hi @hzxuzhonghu, yes, there is custom TLS acceptor, so have to declare these traits.

zhlsunshine:
> It is very possible that tcp and tls service builder could be different later. I am not sure if this is right way

Hi @hzxuzhonghu, use generic pattern to remove the duplicated code is the way I am thinking, there are too many Syntactic sugar in Rust

src/proxy/inbound.rs
howardjohn: One of the goals of this project is to *not* use complex Rust feeatures where possible because ~everyone developing on it is new to rust. IMO this is a regression in the readability of the system. There is nothing so wrong with copying a bit of code.

zhlsunshine: Hi @howardjohn & @hzxuzhonghu, I am also a new to use the Rust, I understand that it's better to make the code readability. However, this is not the most complicated if checking the custom TLS acceptor of `BoringTlsAcceptor` in `src/tls/boring.rs`.  The definition and implementation of `impl<C, F> tls_listener::AsyncTls<C> for BoringTlsAcceptor<F>` is the root cause why I have to add all these traits in this generic implementation.

howardjohn: BoringTlsAcceptor is as simple as it can be as far as I can tell. The complexity is required since we are integrating with third party libraries. We don't need such complexity for first party use cases.

zhlsunshine: Yeah, @howardjohn, I understand ztunnel want to avoid the complicated use case of Rust, so we are avoiding the reference usage, however, in frankly, generic pattern with traits bound is widely used and is the most important feature in Rust, including the libs and applications. I hope we can accept it in ztunnel as well.

howardjohn:
obsolete since https://github.com/istio/ztunnel/pull/186. thanks!

PR #155, opened 2022-11-18
tests: add e2e hbone test

howardjohn:
Previously we had a TCP one, this adds HBONE. The challenge here is TLS; we needed a way to use a mock CA.

* Use `dyn CertificateProvider` in proxy to avoid generic proliferation
* Make the test certificate generation generate real certifications -- the old ones were invalid
* Regen the test certs (and add script) so that we have the ca key and cert as well


src/test_helpers/app.rs
```diff
@@ -35,7 +36,8 @@ where
 {
     initialize_telemetry();
 
-    let app = app::build(cfg).await.unwrap();
```
stevenctl, line 38: Not needed for this PR but I think we should try to split up `app::build` (maybe needs an `AppBuilder` struct) so we can easily override things that can't be set via config. (like manually creating a MockCaClient with bad certs, or custom logic to give bad certs for only one ID). 

`app::new` would be where we set our defaults.

tests/proxy.rs
```diff
@@ -101,6 +96,16 @@ async fn test_request() {
     .await;
 }
 
+#[tokio::test]
+async fn test_hbone_request() {
```
stevenctl, line 100: Should be ~trivial to add some negative tests for expired cert

stevenctl (approved):
Nice! Added DNM so you can merge when ready.

src/test_helpers/app.rs
howardjohn: I was making an AppBuilder at first but got annoyed so went with this for now :-)
 but I agree

Issue #156, opened 2022-11-18
Properly report "ready" only when we are ready


howardjohn:
Ready is defined as:
* XDS is synced
* Listeners are serving
* ?

This is needed for reliability, but also testing.

GregHanson:
is `ready` something that should be displayed via an `istioctl` command, such as `istioctl proxy-status`?

howardjohn:
No, I don't think we need that. We don't for Envoy; its displayed in `kubectl get pods` via the k8s readiness probe

howardjohn:
We could make `proxy-status` show `WDS: SYNCED` or something, but its hard since its a table...

howardjohn:
@keithmattix  is working on this but I cannot assign yet as they aren;'t in the org

keithmattix:
Do we feel like it makes sense to gate readiness on the various listeners being [`run()`](https://github.com/istio/ztunnel/blob/master/src/proxy/mod.rs#L77)? @howardjohn mentioned that XDS is the only outstanding component left, but I think a `BlockReady` makes sense at this point in the code

howardjohn:
we do wait until they are bound so I think that's good enough, the connections can start queueing up at that point

PiotrSikora:
Right now, the readiness is available via `/healthz/ready`, but it's not used by ztunnel itself.

I'd argue that `proxy.run()` should wait _until_ ztunnel is ready (i.e. after local and remote XDS configs are processed), otherwise we're accepting and incorrectly dropping connections.

hzxuzhonghu:
Both make sense, @keithmattix's point is make `run` as another condition to be ready

PiotrSikora:
> Both make sense, @keithmattix's point is make `run` as another condition to be ready

Right, but with current implementation, this means that `proxy.run()` would be executed in parallel with other conditions, leading to a race condition between configuration and proxy listeners (see: #195).

What I'm saying is that we shouldn't start accepting connections until all configuration is processed.

Something like this (ordered):

1. Process configs, fetch XDS, bind to ports, etc.
2. Start accepting connections.
3. Mark server as ready.

keithmattix:
The current block ready implementation is just an unordered hash set of strings signifying a task that must be completed before /admin/ready returns true. If we want to create a dependency graph of actions on startup (e.g process local/remote XDS config before starting listeners before marking ready), we'll likely need some tweaks in the design.

howardjohn:
Could we do:

1. Bind to ports
[...]

PR #157, opened 2022-11-18
Add real readiness logic

howardjohn:
We can put this in more places as well (real XDS server), but wanted a minimal one that gets our test running properly

Part of https://github.com/istio/ztunnel/issues/156

### DIFF
--- a/src/admin/mod.rs
+++ b/src/admin/mod.rs
@@ -14,14 +14,9 @@
 
 // Forked from https://github.com/olix0r/kubert/blob/main/kubert/src/admin.rs
 
-use std::{
-    net::SocketAddr,
-    sync::{
-        atomic::{AtomicBool, Ordering},
-        Arc,
-    },
-    time::Duration,
-};
+use std::collections::HashSet;
+use std::sync::Mutex;
+use std::{net::SocketAddr, sync::Arc, time::Duration};
 
 use drain::Watch;
 #[cfg(feature = "gperftools")]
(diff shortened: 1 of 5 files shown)


src/workload.rs
```diff
@@ -308,11 +311,14 @@ impl LocalClient {
         // Currently, we just load the file once. In the future, we could dynamically reload.
         let data = tokio::fs::read_to_string(self.path).await?;
         let r: Vec<Workload> = serde_yaml::from_str(&data)?;
-        let mut wli = self.workloads.lock().unwrap();
-        for wl in r {
-            info!("inserting local workloads {wl}");
-            wli.insert(wl);
+        {
```
zhlsunshine, line 314: want to know what's the purpose to new a block for workload insert for LocalClient?

howardjohn: To drop the mutex

howardjohn: I guess it might not be needed anymore, in initial iteration it was

src/app.rs
```diff
@@ -35,10 +34,13 @@ pub async fn build_with_cert(
 
     let mut tasks: Vec<JoinHandle<()>> = Vec::new();
 
-    let workload_manager = workload::WorkloadManager::new(config.clone());
+    let ready = admin::Ready::new();
+    let proxy_task = ready.register_task("proxy listeners");
+
+    let workload_manager =
+        workload::WorkloadManager::new(config.clone(), ready.register_task("workload manage"));
```
stevenctl, line 41:
```suggestion
        workload::WorkloadManager::new(config.clone(), ready.register_task("workload manager"));
```

src/admin/mod.rs
```diff
@@ -44,33 +39,73 @@ use crate::{config, signal};
...
+    name: String,
+}
+
+impl BlockReady {
+    pub fn subtask(&self, name: &str) -> BlockReady {
+        self.parent.register_task(name)
+    }
+}
+
+impl Drop for BlockReady {
```
stevenctl, line 87: If pretty much every `drop` is explicit, why do we need this vs some explicit "done" method or wrapping it in something like

```
fn subtask(&self, name: &str, f: SomeFnType) {
  // add to set
  f()
  // remove
}

```

howardjohn: The closure won't work great some other cases (in XDS its going to be deeply nested when we process the first request fully). We can remove the drop but then it makes it possible to forget about it and block readiness indefinitely

stevenctl (approved):

PR #160, opened 2022-11-21
remove unnecessary tcp listener on inbound

zhlsunshine:
remove tcp listener on inbound, there may be tcp or tls listener on inbound based on cfg flag, so unnecessary to store the tcp listener on inbound in advance.

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -32,6 +32,7 @@ use super::Error;
 
 pub struct Inbound {
     cfg: Config,
+    sock_addr: SocketAddr,
     cert_manager: Box<dyn CertificateProvider>,
     workloads: WorkloadInformation,
 }
@@ -49,15 +50,17 @@ impl Inbound {
             Err(_e) => info!("running without transparent mode"),
             _ => info!("running with transparent mode"),
         };
+        let sock_addr = listener.local_addr().unwrap();
         Ok(Inbound {
             cfg,
+            sock_addr,
             workloads,
             cert_manager,
         })
     }
 
     pub(super) fn address(&self) -> SocketAddr {
-        self.cfg.inbound_addr
+        self.sock_addr
     }
 
     pub(super) async fn run(self) {
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -73,8 +73,8 @@ impl Inbound {
                     cert_manager: self.cert_manager.clone(),
                 },
             };
-            let mut incoming_listener = hyper::server::conn::AddrIncoming::bind(&addr)
-                .expect("hbone bind");
+            let mut incoming_listener =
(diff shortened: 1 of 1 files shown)


src/proxy/inbound.rs
```diff
@@ -53,17 +52,16 @@ impl Inbound {
...
             cfg,
             workloads,
-            listener,
             cert_manager,
         })
     }
 
     pub(super) fn address(&self) -> SocketAddr {
-        self.listener.local_addr().unwrap()
+        self.cfg.inbound_addr
```
howardjohn, line 60: This is not correct, this is the configured address -- if we bind to port `0` it will not return the real port

howardjohn (changes requested):
The `proxy`s are explicitly designed to bind before `run()` so we can get the ports + report bind errors

src/proxy/inbound.rs
hzxuzhonghu: If it is 0， then how does the redirect configurred

howardjohn: Bind to 0 is for tests. We already do this but there was a small issue in the test so I had to skip it. It would have detected this and failed if I hadn't skipped though.

hzxuzhonghu: IC

zhlsunshine: Hi @howardjohn, okay, got it. So the OS need to assigns a port to Inbound when binding to port `0`. I will check it here.

howardjohn: We also need to bind outside of run() not just check something here

zhlsunshine: Ok, so that's why we need to keep a tcp listener in Inbound, am I correct?

howardjohn: Yes

src/proxy/inbound.rs
```diff
@@ -75,12 +76,12 @@ impl Inbound {
                     cert_manager: self.cert_manager.clone(),
                 },
             };
-            let mut listener = hyper::server::conn::AddrIncoming::from_listener(self.listener)
-                .expect("hbone bind");
-            listener.set_nodelay(true);
+            let mut incoming_listener =
+                hyper::server::conn::AddrIncoming::bind(&addr).expect("hbone bind");
```
howardjohn, line 80: This is changing the semantic. Its now actually binding

howardjohn (changes requested):
Can you clarify what you are trying to solve? There is no issue with storing a TcpListener in the Inbound struct as far as I know.

src/proxy/inbound.rs
zhlsunshine: Yeah, it's really bind here.

zhlsunshine:
> Can you clarify what you are trying to solve? There is no issue with storing a TcpListener in the Inbound struct as far as I know.

Okay, if so I will close this PR if unnecessary.

PR #163, opened 2022-11-21
optimized the workload store struct

zhlsunshine:
optimized the workload store struct:
1. No `Upstream` contains in `WorkloadStore`, eliminating multiple workload clone for `vip` of `WorkloadStore`.
2. Construct `Upstream` when necessary.

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -368,7 +368,7 @@ pub struct WorkloadStore {
     workload_to_vip: HashMap<IpAddr, HashSet<(SocketAddr, u16)>>,
     // vips maintains a mapping of socket address with service port to workload ip and socket address
     // with target ports in hashset.
-    vips: HashMap<SocketAddr, HashSet<(u16, IpAddr)>>,
+    vips: HashMap<SocketAddr, HashSet<(IpAddr, u16)>>,
 }
 
 impl WorkloadStore {
@@ -392,7 +392,7 @@ impl WorkloadStore {
                 self.vips
                     .entry(service_sock_addr)
                     .or_default()
-                    .insert((port.target_port as u16, wip));
+                    .insert((wip, port.target_port as u16));
                 self.workload_to_vip
                     .entry(wip)
                     .or_default()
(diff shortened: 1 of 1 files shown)


src/workload.rs
```diff
@@ -438,15 +438,20 @@ impl WorkloadStore {
...
-            let mut us: Upstream = us.clone();
-            Self::set_gateway_address(&mut us, hbone_port);
-            debug!("found upstream from VIP: {}", us);
-            return Some(us);
+            let (workload_ip, target_sock_addr) =
+                wl_vips.iter().choose(&mut rand::thread_rng()).unwrap();
+            if let Some(wl) = self.workloads.get(workload_ip) {
+                let mut us = Upstream {
+                    workload: wl.clone(),
+                    port: target_sock_addr.port(),
```
howardjohn, line 449: why do we need to store the whole target_sock_addr when we only use the port? Can the `vips` map just be SocketAddr -> SocketAddr (service:serviceport -> workload:workloadport)?

zhlsunshine: Hi @howardjohn, sure, how about define the `vips` like below?
```
vips: HashMap<SocketAddr, HashSet<(serviceport, IpAddr)>>
```
As you know, we just want to use service port and workload ip to construct the Upstream. Is that okay?

howardjohn: I think you could just turn the tuple into a SocketAddr but yeah basically the same

KfreeZ: can we make sure we are not going to lose any information if use SocketAddr to take place of Workload?

zhlsunshine: @KfreeZ according to the workload definition, it's enough for us to identify a workload.

KfreeZ: > @KfreeZ according to the workload definition, it's enough for us to identify a workload.

suppose if we want to use other information of a workload, let's say waypoint address, we have to look it up again in the workloads Hashmap, correct?

zhlsunshine: Yes, it is

KfreeZ: > 

cool, thanks for the response, no more question.

src/workload.rs
```diff
@@ -366,7 +366,9 @@ pub struct WorkloadStore {
     workloads: HashMap<IpAddr, Workload>,
     // workload_to_vip maintains a mapping of workload IP to VIP. This is used only to handle removals.
     workload_to_vip: HashMap<IpAddr, HashSet<(SocketAddr, u16)>>,
-    vips: HashMap<SocketAddr, HashSet<Upstream>>,
+    // vips maintains a mapping of socket address with service port to workload ip and socket address
+    // with target ports in hashset.
+    vips: HashMap<SocketAddr, HashSet<(u16, IpAddr)>>,
```
hzxuzhonghu, line 371: to be consistent with workload_to_vip, maybe HashSet<(IpAddr, u16)

zhlsunshine: okay, it makes sense

zhlsunshine: @hzxuzhonghu done!

howardjohn (approved):

Issue #158, opened 2022-11-19
Should we consider socket splice for passthrough case


hzxuzhonghu:
Using splice can eliminate copying data from kernel space to user space, and then from user space to kernel space again. 

But splice seems not suitable for telemetry, I am not super sure @howardjohn WDYT?

howardjohn:
Yes SGTM. I think you still get the bytes written so it's not too bad for telemetry.

In Go this happens automagically with io.Copy, not sure what libraries there are to do it in Rust but should be feasible

hzxuzhonghu:
This is fixed by #170 and #179

Issue #99, opened 2022-11-09
Isolated (non-k8s) performance tests


stevenctl:
Clear instructions on how to run

There is a well-known location to publish and compare results.

Targeting iperf, nighthawk, netperf tcp crr, fairness (measure tbd)

Measure resource usage (CPU, RAM)

These tests may be the same as the E2E ones, but can easily be run either 1 or 2 machines with only the testing infra and the zTunnel.

howardjohn:
I ran it adhoc, I think we want to make it trivial to run (e.g `cargo bench --test throughput`) or whatever

linsun:
Has a open PR for it, should be done soon.

PR #168, opened 2022-11-21
Add initial throughput and latency benchmark

howardjohn:
For https://github.com/istio/ztunnel/issues/99

This introduces the first benchmarks. We have 3 throughput tests: direct (no proxy, baseline), TCP, and HBONE.

Example results:
```
$ cargo bench --profile release
   Compiling ztunnel v0.0.0 (/usr/local/google/home/howardjohn/go/src/istio.io/ztunnel)
    Finished release [optimized] target(s) in 1m 45s
     Running benches/throughput.rs (out/rust/release/deps/throughput-1d814c184fd554d8)
tcp/direct              time:   [27.686 ms 28.092 ms 28.529 ms]
                        thrpt:  [3.4230 GiB/s 3.4763 GiB/s 3.5273 GiB/s]
                 change:
                        time:   [+16.629% +19.708% +22.844%] (p = 0.01 < 0.05)
                        thrpt:  [-18.596% -16.463% -14.258%]
                        Performance has regressed.
tcp/hbone               time:   [337.63 ms 345.95 ms 354.01 ms]
                        thrpt:  [282.48 MiB/s 289.06 MiB/s 296.18 MiB/s]
                 change:
                        time:   [-36.935% -35.396% -33.782%] (p = 0.01 < 0.05)
                        thrpt:  [+51.016% +54.788% +58.567%]
                        Performance has improved.
tcp/tcp                 time:   [134.12 ms 138.60 ms 143.24 ms]
                        thrpt:  [698.13 MiB/s 721.49 MiB/s 745.58 MiB/s]
                 change:
                        time:   [-23.810% -21.124% -18.050%] (p = 0.01 < 0.05)
                        thrpt:  [+22.026% +26.781% +31.251%]
                        Performance has improved.
```

The echo server implementation is a bit slow; we can use a better impl once https://github.com/istio/ztunnel/issues/158 is done.

### DIFF
--- a/benches/throughput.rs
+++ b/benches/throughput.rs
@@ -1,0 +1,82 @@
+// #[tokio::test]
+// async fn test_direct() {
+//     helpers::initialize_telemetry();
+//     let echo = echo::TestServer::new().await;
+//     let echo_addr = echo.address();
+//     tokio::spawn(echo.run());
+//     let s = TcpStream::connect(echo_addr).await.unwrap();
+//     s.set_nodelay(true).unwrap();
+//     tcp::run(s, 10 * GB).await.unwrap();
+// }
+
+use std::net::SocketAddr;
+use std::thread;
+use std::time::Duration;
+
+use criterion::{BatchSize, BenchmarkId, Criterion, criterion_group, criterion_main, Throughput};
+use futures_util::SinkExt;
+use pprof::criterion::{Output, PProfProfiler};
+use tokio::runtime::Runtime;
+
+use ztunnel::test_helpers::echo;
+
+const KB: u64 = 1024;
+const MB: u64 = 1024 * KB;
+const GB: u64 = 1024 * MB;
+
+async fn do_test(addr: SocketAddr) {
+    tokio::time::sleep(Duration::from_millis(10)).await;
+}
+
+pub fn tcp_direct(c: &mut Criterion) {
+    let mut c = c.benchmark_group("tcp");
+    c.throughput(Throughput::Bytes(10 * MB));
+    let (mut tx, rx) = std::sync::mpsc::channel();
+    thread::spawn(|| {
+        println!("in thread");
+        let server_rt = Runtime::new().unwrap().block_on(async move {
+            println!("in rt");
+            let echo = echo::TestServer::new().await;
+            println!("in echo");
+            let echo_addr = echo.address();
+            let t=tokio::spawn(echo.run());
+            println!("in running");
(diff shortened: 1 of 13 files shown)


benches/throughput.rs
```diff
@@ -0,0 +1,144 @@
...
+            .await;
+        let mut tcp = ta
+            .socks5_connect(helpers::with_ip(echo_addr, "127.0.0.2".parse().unwrap()))
+            .await;
+        let mut direct = TcpStream::connect(echo_addr).await.unwrap();
+        direct.set_nodelay(true).unwrap();
+        info!("setup complete");
+
+        // warmup: send 1 byte so we ensure we have the full connection setup.
+        tcp::run_latency(&mut hbone, 1).await.unwrap();
```
stevenctl, line 83: Should we consider lazily doing this? Probably not needed, but it feels weird to have the unused ones idling.

src/test_helpers/tcp.rs
```diff
@@ -0,0 +1,64 @@
...
+    let throughput = wrote as f64 / elapsed / 0.125e9;
+    info!(
+        "throughput: {:.3} Gb/s, wrote {wrote} in {:?}",
+        throughput,
+        start.elapsed()
+    );
+    Ok(())
+}
+
+pub async fn run_latency(stream: &mut TcpStream, amt: usize) -> Result<(), io::Error> {
```
stevenctl, line 56: Are these under test_helpers instead of benches because we can use them for something else? 

It also feels weird that they log themselves instead of returning their result.

stevenctl (approved):
you're right, the criterion stuff is awesome

src/test_helpers/tcp.rs
howardjohn: We don't actually use the result, the benchmark system measures the time it takes anyways; this is mostly for debugging (info logging is off on benchmark, as well, by default).

in terms of organization - they could probably be under benches/. I refactored this PR literally 10 times before getting to this point, so I was just happy when everything worked without warnings :-)

benches/throughput.rs
howardjohn: Getting setup properly for our needs with criterion seems to be really hard. I guess making it Lazy would actually probably do the trick.. we could do that.

Or have 1 full setup per type (direct, tcp, hbone) and only establish one connection.

howardjohn:
this is broken for the same reason hbone e2e test is, I have a fix ill merge in another PR

PR #171, opened 2022-11-22
inbound graceful shutdown

hzxuzhonghu:


hzxuzhonghu:
I am copying this from admin server shutdown, Thanks for the input. Will try to improve

howardjohn:
my understanding:

* graceful shutdown runs immediately, blocks on signaled().await
* SIGTERM comes, signaled ().await completes
* server starts to drain
* signaled () is dropped, marking the drain as complete
* ztunnel exits immediately

At no point are we waiting for the server to actually exit?

When I tested this, open connections did not block shutdown. With the tweak I mentioned it did

hzxuzhonghu:
The above order is right.

Just we donot need oneshot channel here, because my test shows only when `self.drain.signaled().await;` returns, the `server.await ` could be unblocked

hzxuzhonghu:
So it seems redundant, that's why i did not add it

howardjohn:
> The above order is right.
> 
> Just we donot need oneshot channel here, because my test shows only when `self.drain.signaled().await;` returns, the `server.await ` could be unblocked

But once the drain is dropped the entire app will shutdown. Nothing waits for server.await to finish.

You can try it by simply opening `nc` connection on ztunnel and then send SIGTERM, it closes immediately before inbound terminates

hzxuzhonghu:
Understood your point, but with oneshot channel, curious how it can prevent the process exit.

Doesn't  drain.signaled() will trigger the process exit?

howardjohn:
> Understood your point, but with oneshot channel, curious how it can prevent the process exit.
> 
> Doesn't drain.signaled() will trigger the process exit?

No, that will return from await when the process is exited.

```
let x = signaled ().await // the process has started exiting..
// exit is blocked by `x`
drop(x) // process can now exit
```

is how it works. Note that there can be multiple copies of it, so it only fully exits once all are dropped

src/proxy/inbound.rs
```diff
@@ -118,13 +130,21 @@ impl Inbound {
                 .http2_initial_stream_window_size(self.cfg.window_size)
                 .http2_initial_connection_window_size(self.cfg.connection_window_size)
                 .http2_max_frame_size(self.cfg.frame_size)
-                .serve(service);
+                .serve(service)
+                .with_graceful_shutdown(async {
+                    let shutdown = self.drain.signaled().await;
+                    // TODO: add some more graceful shutdown way
+                    tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
```
hzxuzhonghu, line 137: added 3s here because termination_grace_period is 5s. 

Definitely it needs otherway here

src/proxy/inbound.rs
```diff
@@ -98,13 +103,21 @@ impl Inbound {
                 .http2_initial_stream_window_size(self.cfg.window_size)
                 .http2_initial_connection_window_size(self.cfg.connection_window_size)
                 .http2_max_frame_size(self.cfg.frame_size)
-                .serve(service);
+                .serve(service)
+                .with_graceful_shutdown(async {
+                    let shutdown = self.drain.signaled().await;
+                    // TODO: add some more graceful shutdown way
+                    time::sleep(time::Duration::from_secs(3)).await;
```
howardjohn, line 110: This needs to be configurable or our tests will take forever

src/proxy/inbound.rs
```diff
@@ -98,13 +103,21 @@ impl Inbound {
                 .http2_initial_stream_window_size(self.cfg.window_size)
                 .http2_initial_connection_window_size(self.cfg.connection_window_size)
                 .http2_max_frame_size(self.cfg.frame_size)
-                .serve(service);
+                .serve(service)
+                .with_graceful_shutdown(async {
+                    let shutdown = self.drain.signaled().await;
+                    // TODO: add some more graceful shutdown way
```
howardjohn, line 109: can we comment why we do this? I agree it makes sense but its not intuitive. I think just saying something like

"We need 2 tiers of draining. First, we still accept new connections and just wait. This is to let Kubernetes propagate that we are shutting down and start directing new connections to another instance. Then we stop accepting connections and wait for existing ones to shut down"

This also probably belongs at a higher level; we don't want each component to try to hardcode the delay. Instead in `app.rs` Bound I would put the sleep between `shutdown.wait()` and `drain_tx.drain()` (and again, make it configurable)

src/proxy/inbound.rs
```diff
@@ -98,13 +103,21 @@ impl Inbound {
...
+                    info!("drain end");
+                    drop(shutdown);
+                });
 
             info!("HBONE listener established {}", addr);
 
             if let Err(e) = server.await {
                 error!("server error: {}", e);
             }
+            info!("server exit success drained");
```
howardjohn, line 120: This still doesn't ever get reached. We need to move `drop(shutdown)` until after the `server.await` or we aren't actually waiting for the shutdown (aside from the arbitrary 3s sleep)

src/proxy/inbound.rs
hzxuzhonghu: There should be some way to notify connection close, like close with a timeout.

src/proxy/inbound.rs
hzxuzhonghu: >We need to move drop(shutdown) until after the server.await
I think it make sense

src/proxy/inbound.rs
hzxuzhonghu: yes, maybe infer from the existing one. i will remove it ATM

hzxuzhonghu:
@howardjohn another look

hzxuzhonghu:
ping @howardjohn

howardjohn (approved):

PR #173, opened 2022-11-22
Explicitly fallback on orig_dst lookup failure

howardjohn:
Typically, if you do orig_dst lookup and there was no iptables redirect, you get back the unredirected address: this is what we want.

However, it turns out that if there are no iptables rules in the network namespace, regardless of whether they apply to the request, the orig_dst lookup fails.

To workaround this, we explicitly fallback to the unredirected address instead of relying on the kernel to do it for us.

This fixes the tests in CI, which have no iptables rules setup.

As part of this, we move `BoringTlsAcceptor` from generic to explicit struct. Its possible to make it generic but I don't see a need for now; I expect moving forward we will likely want to have some concrete structs passed between stages (tcp connection, tls connection, http connection) storing metadata along the way at each hop, so having concrete types kind of aligns with that.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -39,20 +39,8 @@ pub fn set_transparent(l: &TcpListener) -> io::Result<()> {
     Ok(())
 }
 
-#[cfg(target_os = "linux")]
-#[allow(unsafe_code)]
-pub fn orig_dst_addr_fd<T: AsRawFd>(sock: T) -> io::Result<SocketAddr> {
-    let fd = sock.as_raw_fd();
-
-    unsafe { linux::so_original_dst(fd) }
-}
-
-#[cfg(not(target_os = "linux"))]
-pub fn orig_dst_addr_fd<T: AsRawFd>(_: T) -> io::Result<SocketAddr> {
-    Err(io::Error::new(
-        io::ErrorKind::Other,
-        "SO_ORIGINAL_DST not supported on this operating system",
-    ))
+pub fn orig_dst_addr_or_default(sock: &tokio::net::TcpStream) -> SocketAddr {
+    orig_dst_addr(sock).unwrap_or_else(|_| sock.peer_addr().expect("must get peer address"))
 }
 
 #[cfg(target_os = "linux")]
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 use std::net::SocketAddr;
-use std::os::unix::io::RawFd;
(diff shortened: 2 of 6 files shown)


stevenctl (approved):

src/socket.rs
```diff
@@ -39,20 +39,8 @@ pub fn set_transparent(l: &TcpListener) -> io::Result<()> {
...
-}
-
-#[cfg(not(target_os = "linux"))]
-pub fn orig_dst_addr_fd<T: AsRawFd>(_: T) -> io::Result<SocketAddr> {
-    Err(io::Error::new(
-        io::ErrorKind::Other,
-        "SO_ORIGINAL_DST not supported on this operating system",
-    ))
+pub fn orig_dst_addr_or_default(sock: &tokio::net::TcpStream) -> SocketAddr {
+    orig_dst_addr(sock).unwrap_or_else(|_| sock.peer_addr().expect("must get peer address"))
```
PiotrSikora, line 43: I've been looking at this code, and I cannot wrap my head around the logic here, but perhaps I'm missing something.

In case of the lookup failure, this will default to a `peer_addr`, which is the IP of the connecting client.

This means that later we proxy traffic between `remote_addr` source (which is `peer_addr`) and `orig` destination (which is also `peer_addr`), which cannot be right.

@howardjohn could you explain?

howardjohn: I think this is probably inverted and it isn't breaking things (yet) since our tests either run to/from localhost OR run with actual iptables (in which case fallback never occurs)

PiotrSikora:
Thanks!

Yes, I think this should default to `local_peer` instead (which is also what's the proper destination address when using `TPROXY`), I'll send a PR in a sec.

However, I'd argue that if we cannot retrieve the original destination, then we should probably abort tunneling.

src/socket.rs
howardjohn: One thing to note is that even if there is no redirect, the orig_dst lookup succeeds - it will just return the same address (local_peer); the only case it fails is if there are 0 iptables rules on the node entirely

PiotrSikora: Right, but we could compare `orig_dst` and `local_addr` and bail out in case they're the same.

But we can do that only for `REDIRECT`s and/or when running `IP_TRANSPARENT` fails, since `TPROXY` overrides `local_addr` with `orig_dst`, and I don't think we can detect when that have happened.

PiotrSikora: PR with the `peer_addr` -> `local_addr` change: #234.

PR #175, opened 2022-11-23
Refactor benchmark a bit

howardjohn:
This changes our "echo server" to be a server that can do one of 3 modes: Read, Write, or Read+Write (ie echo). The client (which now lives in the same package) can also do the same set of three modes.

The throughput benchmark is changed to use Write only; this mirrors `iperf3` (note: even in `iperf3 --bidir` it is in Read or Write mode, never "ReadWrite" mode; bidir opens 2 connections).

The logic in ReadWrite mode is now `loop { write; read }` rather than two joined tasks reading and writing; this improves reliability because otherwise we randomly pick to read or write and get inefficient behavior (many writes then many reads, etc); this also mirrors iperf3.


benches/throughput.rs
```diff
@@ -79,10 +80,15 @@ fn initialize_environment() -> (Arc<Mutex<TestEnv>>, Runtime) {
         direct.set_nodelay(true).unwrap();
         info!("setup complete");
 
+        let client_mode = match mode {
```
stevenctl, line 83: Shouldn't this be just `mode`, possibly with a clone?

howardjohn: Its the inverse; if server is a reader then client must be a writer

PR #176, opened 2022-11-23
bi copy support cross platform

hzxuzhonghu:


src/socket.rs
```diff
@@ -166,3 +168,31 @@ mod linux {
...
+const EINVAL: i32 = 22;
+pub async fn relay(
+    downstream: &mut tokio::net::TcpStream,
+    upstream: &mut tokio::net::TcpStream,
+) -> Result<(), Error> {
+    #[cfg(all(target_os = "linux"))]
+    {
+        match realm_io::bidi_zero_copy(downstream, upstream).await {
+            Ok(()) => Ok(()),
+            Err(ref e) if e.raw_os_error().map_or(false, |ec| ec == EINVAL) => {
```
hzxuzhonghu, line 181: This is referred from https://github.com/zhboner/realm/blob/master/realm_core/src/tcp/plain.rs

src/socket.rs
```diff
@@ -166,3 +168,31 @@ mod linux {
...
+                    Err(e) => Err(e),
+                }
+            }
+            Err(e) => Err(e),
+        }
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    {
+        match tokio::io::copy_bidirectional(downstream, upstream).await {
```
howardjohn, line 193: I think you can replace this with `map(|_| ())`

howardjohn (commented):
lgtm just 1 question

howardjohn (approved):
oops forgot to ask the questions but you answered it already...

src/socket.rs
howardjohn: any idea when it can occur on Linux?

hzxuzhonghu:
```
       EINVAL The target filesystem doesn't support splicing.

       EINVAL The target file is opened in append mode.

       EINVAL Neither of the file descriptors refers to a pipe.

       EINVAL An offset was given for nonseekable device (e.g., a pipe).

       EINVAL fd_in and fd_out refer to the same pipe.

```

hzxuzhonghu: this is from `man splice`

howardjohn: hmm.. are any of those even possible for our usage?

src/socket.rs
hzxuzhonghu: good suggesttion

src/socket.rs
hzxuzhonghu: I am not sure, just in case

PR #178, opened 2022-11-23
Change the worker thread name for tracing easily

haiyuewa:
The default name fn is `|| "tokio-runtime-worker".into()`, it will make debug / log not to be friendly by the same thread name.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -16,6 +16,7 @@ extern crate core;
 #[cfg(feature = "gperftools")]
 extern crate gperftools;
 
+use std::sync::atomic::{AtomicUsize, Ordering};
 use ztunnel::*;
 
 // #[global_allocator]
@@ -29,6 +30,11 @@ fn main() -> anyhow::Result<()> {
     let config: config::Config = Default::default();
     tokio::runtime::Builder::new_multi_thread()
         .worker_threads(config.num_worker_threads)
+        .thread_name_fn(|| {
+            static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);
+            let id = ATOMIC_ID.fetch_add(1, Ordering::SeqCst);
+            format!("proxy-{}", id)
+        })
         .enable_all()
         .build()
         .unwrap()


src/main.rs
```diff
@@ -29,6 +30,11 @@ fn main() -> anyhow::Result<()> {
     let config: config::Config = Default::default();
     tokio::runtime::Builder::new_multi_thread()
         .worker_threads(config.num_worker_threads)
+        .thread_name_fn(|| {
+            static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);
+            let id = ATOMIC_ID.fetch_add(1, Ordering::SeqCst);
+            format!("proxy-{}", id)
```
PlatformLC, line 36: nit: add ztunnel signature like ztunnel-proxy?

haiyuewa: Yes, I originally added it. 

But I found that tokio thread_name_fn working scope is its OS thread group:

1. ps -aux | grep ztunnel
root      390796  0.0  0.0 298396 17660 pts/5    Sl+  14:30   0:00 ./out/rust/debug/ztunnel

2. ps -eLf | grep ztunnel
root      390796  352772  390796  0    3 14:30 pts/5    00:00:00 ./out/rust/debug/ztunnel
root      390796  352772  390797  0    3 14:30 pts/5    00:00:00 ./out/rust/debug/ztunnel
root      390796  352772  390798  0    3 14:30 pts/5    00:00:00 ./out/rust/debug/ztunnel

3. top -H -p 390796
PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
 390796 root      20   0  298396  17660  15296 S   0.0   0.1   0:00.02 ztunnel
 390797 root      20   0  298396  17660  15296 S   0.0   0.1   0:00.00 proxy-0
 390798 root      20   0  298396  17660  15296 S   0.0   0.1   0:00.00 proxy-1

In this TOP screen information, you will find "ztunnel-proxy" has some redundant informaiton, since
main thread name is "ztunnel", also it will only show 15 characters, if we have more worker threads,
'ztunnel-proxy-' will truncate the information. Like the original dump, just show 'tokio-runtime-w'.

So 'proxy-' should be enough, I think.

haiyuewa: 'proxy' is function, 'zproxy' sounds too Geek ? ;-)

howardjohn (approved):

PR #182, opened 2022-11-23
Change gateway to upstream workload waypoint address

ymesika:
I believe it was a typo for the case upstream server has enabled waypoint.


howardjohn (approved):

howardjohn (commented):
can we add some test cases? we alreadyt have a few for this function

PR #187, opened 2022-11-23
xds: use TLS (port 15012) to control plane

howardjohn:
Also refactors the node logic to only send the node on the first request (mirroring envoy; makes requests 50% smaller) since I already had to refactor the node logic anyways.


src/xds/client.rs
```diff
@@ -242,15 +244,13 @@ impl AdsClient {
...
         let outbound = async_stream::stream! {
             for request_type in watches {
                 let irv = irv.clone();
                 let initial = DeltaDiscoveryRequest {
                     type_url: request_type.clone(),
-                    node: Some(Node{
-                        id: node_id.clone(),
-                        ..Default::default()
-                    }),
+                    node: Some(node.clone()),
```
stevenctl, line 253: nit: seems like the only `node` use. Probably don't need to clone

stevenctl (approved):

PR #188, opened 2022-11-23
Simplified caching logic, changed expiration handling to use SystemTi…

briansonnenberg:
…me to allow for sub-second expiration in unit tests.


src/tls/boring.rs
```diff
@@ -93,52 +104,73 @@ impl CsrOptions {
...
-        let end: &Asn1TimeRef = self.cert.not_after();
-        let time_until_expired = current.diff(end).unwrap();
-        if time_until_expired.secs > 0 {
-            return false;
-        }
-        true
+        // duration_since returns an error if now() is later than not_after
+        self.cert
+            .not_after
+            .duration_since(SystemTime::now())
```
howardjohn, line 157: I think this can just be `self.cert.not_after >= SystemTime::now()`

src/tls/boring.rs
```diff
@@ -93,52 +104,73 @@ impl CsrOptions {
...
-        Duration::from_secs(u64::try_from(returnval).unwrap())
+        let halflife = self
+            .cert
+            .not_after
+            .duration_since(self.cert.not_before)
+            .unwrap()
+            / 2;
+        let elapsed = SystemTime::now()
+            .duration_since(self.cert.not_before)
+            .unwrap();
```
howardjohn, line 170: this doesn't seem safe, the cert could have arbitrary values

src/tls/boring.rs
```diff
@@ -93,52 +104,73 @@ impl CsrOptions {
...
-        }
-        true
+        // duration_since returns an error if now() is later than not_after
+        self.cert
+            .not_after
+            .duration_since(SystemTime::now())
+            .is_err()
     }
 
     pub fn get_duration_until_refresh(&self) -> Duration {
```
howardjohn, line 161: this seems like an easy+useful method to have unit tests for

src/tls/boring.rs
```diff
@@ -472,31 +506,16 @@ fn test_ca() -> Result<(x509::X509, PKey<Private>), Error> {
...
-    }
-
     #[test]
     fn is_fips_enabled() {
         assert!(boring::fips::enabled());
     }
 }
-
-#[test]
-#[cfg(not(feature = "fips"))]
```
howardjohn, line 499: I think we still want this (although should be inside the `mod tests`)

howardjohn (commented):
nice improvement to the test runtime

src/tls/boring.rs
```diff
@@ -472,31 +513,58 @@ fn test_ca() -> Result<(x509::X509, PKey<Private>), Error> {
...
+            Duration::from_secs(1000),
+            Duration::from_secs(expiry_seconds),
+        );
+        assert!(!future_certs.is_expired());
+        assert_eq!(future_certs.get_duration_until_refresh(), zero_dur);
+    }
+
+    #[test]
+    #[cfg(not(feature = "fips"))]
+    fn is_fips_disabled() {
```
howardjohn, line 567: nit: would make sense to keep this next to is_fips_enabled

howardjohn (approved):
LGTM, can youremove the Default in a followup?

PR #190, opened 2022-11-24
Xds readiness

keithmattix:
Add a `BlockReady` after the first ACK'd XDS response is received.

Part of #156


src/xds/client.rs
```diff
@@ -274,16 +276,28 @@ impl AdsClient {
             .into_inner();
 
         info!("Stream established");
+        let (tx, rx) = oneshot::channel();
+        let mut initial_xds_tx = Some(tx);
+        let initial_xds_rx = Some(rx);
```
hzxuzhonghu, line 281: Why do you wrap rx and tx

src/xds/client.rs
```diff
@@ -274,16 +276,28 @@ impl AdsClient {
...
+        let ready = mem::take(&mut self.block_ready);
+        tokio::spawn(async move {
+            match initial_xds_rx.unwrap().await {
+                Ok(_) => {
+                    drop(ready);
```
hzxuzhonghu, line 286: cannot drop self.block_ready?

src/xds/client.rs
keithmattix: The reason for wrapping `tx` is because I need to pass it into a function that's called in an infinite loop, but you're right I don't need to wrap `rx` here.

src/xds/client.rs
keithmattix: Unfortunately not; `run_internal` has a mutable reference for `self` so I can't drop a field on self

hzxuzhonghu: Maybe pass `block_ready` into the `run`

src/xds/client.rs
```diff
@@ -359,13 +373,19 @@ impl AdsClient {
...
+
         info!(
             type_url = type_url.clone(),
             none = response.nonce,
             "sending {}",
             if error_detail.is_some() {
                 "NACK"
             } else {
+                if let Some(tx) = initial_xds_recvd {
+                    if let Err(err) = tx.send(()) {
```
hzxuzhonghu, line 385: Is it safe to send multi times?

keithmattix: We never get to that point; `send()` won't even be invoked when `tx` is `None`, and it gets set to None on L297 when `take` is called

src/xds/client.rs
keithmattix: I thought about that, but the `block_ready` would have to be passed into `WorkloadManager::run` (which is a bit awkward from the caller's perspective). Furthermore, if we want to avoid `/admin/ready` flip-flopping (i.e. ready in one request and then not ready in a subsequent request), it's better to call `register_task`s around the same time. Waiting until `WorkloadManager::run` to create a `block_ready` for ADS when the LocalClient's already exists opens the door for that kind of inconsistent behavior.

src/xds/client.rs
hzxuzhonghu: good to know that

keithmattix: I'll add a comment explaining the mechanism; it's not super clear looking at it later

src/xds/client.rs
```diff
@@ -274,16 +276,27 @@ impl AdsClient {
             .into_inner();
 
         info!("Stream established");
+        let (tx, initial_xds_rx) = oneshot::channel();
+        let mut initial_xds_tx = Some(tx);
+        let ready = mem::take(&mut self.block_ready);
```
howardjohn, line 281: I think since we take it, if this function exits unsuccessfully we will still end up dropping it (?)

keithmattix: Ah, good call. So `ready` gets `move`d into the closure passed to `tokio::spawn`, so it'll `drop` when that closure exits. The only way for the closure to exit without a successful send is for the `Sender` to drop from scope. I'll change the code a bit so that the `tx` is within an infinite loop, guaranteeing that the only time the sender drops is after a message is passed.

src/xds/client.rs
```diff
@@ -287,16 +295,34 @@ impl AdsClient {
...
             tokio::select! {
                 _demand_event = self.demand.recv() => {
-                    self.handle_demand_event(None, &discovery_req_tx).await?;
+                    self.handle_demand_event(_demand_event, &discovery_req_tx).await?;
                 }
-                msg = response_stream.message() =>{
-                    self.handle_stream_event(msg?, &discovery_req_tx).await?;
+                msg = response_stream.message() => {
+                    if let XdsSignal::Ack = self.handle_stream_event(msg?, &discovery_req_tx).await? {
+                        let val = mem::take(&mut initial_xds_tx);
```
howardjohn, line 318: technically this only works because we have only 1 XDS type currently. Fine for now but somethign to keep in mind if we add others (RBAC)

keithmattix: Added a TODO and finished the comment above. It'll be included in my next PR

PR #197, opened 2022-11-25
Add option to use the builtin mock as a fake CA.

### LINKED ISSUE #140: Provide a better CA story for local development
Currently running locally is a bit painful, since you need a real ztunnel pod running in a kubernetes cluster to get the JWT token for authentication to the CA. This is worse than the normal state with Istio, where we still need a k8s cluster (technically can do without, but its not worth it), but have no requirements for pods in the cluster We should provide some easier way to run locally.

PiotrSikora:
This is useful for local development without Istio.

Fixes #140.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use std::net::SocketAddr;
+use std::time::Duration;
 
 use tokio::task::JoinHandle;
 use tokio::time;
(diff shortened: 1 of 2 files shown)


src/app.rs
```diff
@@ -75,8 +76,13 @@ pub async fn build_with_cert(
...
 pub async fn build(config: config::Config) -> anyhow::Result<Bound> {
-    let cert_manager = identity::SecretManager::new(config.clone());
-    build_with_cert(config, cert_manager).await
+    if config.fake_ca {
+        let cert_manager = identity::mock::MockCaClient::new(Duration::from_secs(86400));
+        build_with_cert(config, cert_manager).await
+    } else {
+        let cert_manager = identity::SecretManager::new(config.clone());
+        build_with_cert(config, cert_manager).await
+    }
```
PiotrSikora, line 85: There is probably some magic Rust syntax to simplify this to `let cert_manager = match config.fake_ca { ... }`, but I couldn't figure a way to fix the `match arms have incompatible types` error.

howardjohn (approved):

PR #198, opened 2022-11-25
Add option to configure address of the control plane.

PiotrSikora:
This allows using ztunnel without a control plane with XDS_ADDRESS= (empty),
which is useful for local development without Istio.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -53,7 +53,7 @@ const DEFAULT_WORKER_THREADS: usize = 2;
 impl Default for Config {
     fn default() -> Config {
         // TODO: copy JWT auth logic from CA client and use TLS here (port 15012)
-        let xds_address = match std::env::var("XDS").ok() {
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -53,11 +53,17 @@ const DEFAULT_WORKER_THREADS: usize = 2;
 impl Default for Config {
     fn default() -> Config {
         // TODO: copy JWT auth logic from CA client and use TLS here (port 15012)
-        let xds_address = Some(if std::env::var("KUBERNETES_SERVICE_HOST").is_ok() {
-            "https://istiod.istio-system:15012".to_string()
-        } else {
-            "https://localhost:15012".to_string()
-        });
+        let xds_address = match std::env::var("XDS").ok() {
```
hzxuzhonghu, line 56: how about `XDS_ADDRESS`

PiotrSikora: Done.

howardjohn (approved):

PR #200, opened 2022-11-25
Move the argv handling out of async context

haiyuewa:
The async context is expensive, like spawning worker threads, scheduling the task etc. So move the argv handling into the main thread.

Also, the CLI can be used to add more configuration into config::Config, which the proxy may need for run.


PiotrSikora (approved):
LGTM

howardjohn (approved):
We will probably have more async commands in the future but for now this works and we can refactor a bit as needed. Thanks

PR #201, opened 2022-11-25
eliminate Config clone in all places via using smart pointer

zhlsunshine:
Config is used and cloned by many places in ztunnel, so need to eliminate the really memory copy for Config struct based smart pointer `Arc`


src/workload.rs
```diff
@@ -254,11 +254,20 @@ impl WorkloadManager {
             None
         };
         let local_workloads = workloads.clone();
-        let local_client = config.local_xds_path.map(|path| LocalClient {
-            path,
-            workloads: local_workloads,
-            block_ready: awaiting_ready,
-        });
+        let local_xds_path = config.local_xds_path.clone();
```
howardjohn, line 257: why does this change? the logic seems simpler before

zhlsunshine: Hi @howardjohn,  there is an error for string not implement `Copy` trait, so I clone the `local_xds_path`.
I rebased the code and change it to borrow the `local_xds_path` from Arc config for elimination of such complication code.

hzxuzhonghu:
I donot see any reduction of clone, am i missing?

zhlsunshine:
> I donot see any reduction of clone, am i missing?

Hi @hzxuzhonghu, the purpose of this optimization is to eliminate the really heap copy for `config::Config` by using smart pointer, but not eliminate the `clone` calling.

zhlsunshine:
Hi @howardjohn & @hzxuzhonghu, I suggest that this PR can be merged because the config clone calling are increased. Any idea?

src/identity/manager.rs
```diff
@@ -71,8 +71,10 @@ where
 }
 
 impl SecretManager<CaClient> {
-    pub fn new(cfg: crate::config::Config) -> SecretManager<CaClient> {
-        let caclient = CaClient::new(cfg.ca_address.unwrap(), cfg.auth);
+    pub fn new(cfg: Arc<crate::config::Config>) -> SecretManager<CaClient> {
+        let ca_address = cfg.ca_address.clone().unwrap();
+        let auth = cfg.auth.clone();
```
hzxuzhonghu, line 76: Are these must be cloned?

hzxuzhonghu (commented):
LGTM, just one comment do we need to clone a Arc struct member field?

src/identity/manager.rs
zhlsunshine: Hi @hzxuzhonghu,  yes, the ownership to the field of `Config` can not be moved into `CaClient`, so need to clone here. the same as the `ca_address`.

howardjohn:
this has been implemented in another PR. Thanks!

PR #211, opened 2022-11-29
Handle the hbone copy error gracefully

haiyuewa:
Inspired by the io::copy error handling from:
https://github.com/tokio-rs/tokio/blob/master/examples/proxy.rs

        io::copy(&mut ri, &mut wo).await?;
        wo.shutdown().await

If the io::copy has error, just return error to the upper application to capture the real connection error like:

2022-11-29T08:35:32.680985Z ERROR ztunnel::proxy: tcp -> hbone: broken pipe
2022-11-29T08:35:32.680988Z ERROR ztunnel::proxy: hbone -> tcp: Connection reset by peer (os error 104)
2022-11-29T08:35:32.681038Z ERROR ztunnel::proxy::outbound: hbone client copy: broken pipe
2022-11-29T08:35:32.681045Z ERROR ztunnel::proxy::inbound: hbone server copy Err(Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" })

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -131,9 +131,12 @@ impl Inbound {
                         }
                         Hbone(req) => match hyper::upgrade::on(req).await {
                             Ok(mut upgraded) => {
-                                super::copy_hbone("hbone server", &mut upgraded, &mut stream)
-                                    .await
-                                    .expect("hbone server copy");
+                                if let Err(e) =
+                                    super::copy_hbone("hbone server", &mut upgraded, &mut stream)
+                                        .await
+                                {
+                                    error!("hbone server copy: {}", e);
+                                }
                             }
                             Err(e) => {
                                 // Not sure if this can even happen
--- a/src/proxy/mod.rs
+++ b/src/proxy/mod.rs
@@ -158,7 +158,6 @@ pub async fn copy_hbone(
         let mut wo = tokio::io::BufWriter::with_capacity(HBONE_BUFFER_SIZE, &mut wo);
         let res = tokio::io::copy(&mut ri, &mut wo).await;
         info!(?res, ?desc, "hbone -> tcp");
-        res.expect("");
         wo.shutdown().await
     };
 
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -206,14 +206,21 @@ impl OutboundConnection {
                 let code = response.status();
                 match hyper::upgrade::on(response).await {
                     Ok(mut upgraded) => {
-                        super::copy_hbone("hbone client", &mut upgraded, &mut stream)
-                            .await
-                            .expect("hbone client copy");
+                        if let Err(e) =
+                            super::copy_hbone("hbone client", &mut upgraded, &mut stream).await
+                        {
+                            error!("hbone client copy: {}", e);
+                            Err(Error::Io(e))
+                        } else {
+                            info!("request complete");
+                            Ok(())
+                        }
                     }
-                    Err(e) => error!("upgrade error: {}, {}", e, code),
+                    Err(e) => {
+                        error!("upgrade error: {}, {}", e, code);
+                        Err(Error::Http(e))
+                    }
                 }
-                info!("request complete");
-                Ok(())
             }
             Protocol::TCP => {
                 info!(


src/proxy/mod.rs
```diff
@@ -145,18 +145,31 @@ pub async fn copy_hbone(
...
         let mut ri = tokio::io::BufReader::with_capacity(HBONE_BUFFER_SIZE, &mut ri);
         let mut wo = tokio::io::BufWriter::with_capacity(HBONE_BUFFER_SIZE, &mut wo);
-        let res = tokio::io::copy(&mut ri, &mut wo).await;
-        info!(?res, ?desc, "hbone -> tcp");
-        res.expect("");
-        wo.shutdown().await
+        match tokio::io::copy(&mut ri, &mut wo).await {
+            Ok(len) => {
+                info!(len, ?desc, "hbone -> tcp");
+                wo.shutdown().await
```
hzxuzhonghu, line 151: This is changing, no shutdown when err

src/proxy/outbound.rs
```diff
@@ -200,14 +200,22 @@ impl OutboundConnection {
...
-                        super::copy_hbone("hbone client", &mut upgraded, &mut stream)
-                            .await
-                            .expect("hbone client copy");
+                        match super::copy_hbone("hbone client", &mut upgraded, &mut stream).await {
+                            Ok(_) => {
+                                info!("request complete");
+                                Ok(())
+                            }
+                            Err(e) => {
+                                error!("hbone client copy: {}", e);
```
hzxuzhonghu, line 209: shutdown?

src/proxy/mod.rs
```diff
@@ -145,18 +145,31 @@ pub async fn copy_hbone(
...
-        let res = tokio::io::copy(&mut ro, &mut wi).await;
-        info!(?res, ?desc, "tcp -> hbone");
-        wi.shutdown().await
+        match tokio::io::copy(&mut ro, &mut wi).await {
+            Ok(len) => {
+                info!(len, ?desc, "tcp -> hbone");
+                wi.shutdown().await
+            }
+            Err(e) => {
+                error!("tcp -> hbone: {e}");
```
howardjohn, line 169: Why do we need to log in all of these places when we already log at higher levels? Having multiple logs for 1 failure is confusing. Typically should either propogate the error or handle it, not both

howardjohn (changes requested):

src/proxy/mod.rs
haiyuewa: I kept the 'expect(msg)', will remove it.

src/proxy/outbound.rs
haiyuewa: From the code, shutdwown will not be called when copy error happens: https://github.com/tokio-rs/tokio/blob/master/examples/proxy.rs
And the shutdown will also return another error if called.

haiyuewa: The above reply should be for "This is changing, no shutdown when err" . ;-)

No need "shutdown" ? For the normal case, I didn't see 'shutdown', I missed it ?

src/proxy/mod.rs
PiotrSikora: There was no `shutdown` previously, since it would `panic` on `Err`.

haiyuewa: OK, I added it back, and just remove the "expect", and use "shutdown" as an error indication.

src/proxy/inbound.rs
```diff
@@ -132,9 +132,12 @@ impl Inbound {
...
                             Ok(mut upgraded) => {
-                                super::copy_hbone("hbone server", &mut upgraded, &mut stream)
-                                    .await
-                                    .expect("hbone server copy");
+                                let res =
+                                    super::copy_hbone("hbone server", &mut upgraded, &mut stream)
+                                        .await;
+                                if res.is_err() {
+                                    error!("hbone server copy: {:?}", res);
+                                }
```
howardjohn, line 140:
```suggestion
                               if let Err(e) = super::copy_hbone("hbone server", &mut upgraded, &mut stream)
                                    .await {
                                    error!("hbone server copy: {}", e);
                                }
```

Is probably preferred. Then we don't need to user debug formatter.

haiyuewa: Yes, this is very clean.

howardjohn (approved):

PR #215, opened 2022-11-29
unit test caclient directly

stevenctl:
Not deadset on merging this, but we should probably cover `CaClient`. 

I played around with a few ways to mock out the gRPC client but landed on making another `Box<dyn ...>`.

One approach was enabling the server codegen from tonic and then setting up a small local gRPC server. I couldn't find a nice way to elide the type information since we have `IstioCertificateServiceClient<T>` where `T` has information about the interceptor/connector stuff. Making `CaClient` generic over the same `T` would be very rough. 

The annoying part here is if we want another rpc from the same service, we have to wrap it up with our trait. There is probably some cool macro we could write for this if we really wanted. 

--- 

one actual codefix here is removing an unsafe unwrap.

### DIFF
--- a/src/identity/caclient.rs
+++ b/src/identity/caclient.rs
@@ -17,22 +17,45 @@ use std::collections::BTreeMap;
 use async_trait::*;
 use prost_types::value::Kind;
 use prost_types::Struct;
-use tonic::codegen::InterceptedService;
+
+use dyn_clone::DynClone;
 use tracing::{instrument, warn};
 
 use crate::identity::auth::AuthSource;
 use crate::identity::manager::Identity;
 use crate::identity::Error;
-use crate::tls::TlsGrpcChannel;
 use crate::tls::{self, SanChecker};
 use crate::xds::istio::ca::istio_certificate_service_client::IstioCertificateServiceClient;
-use crate::xds::istio::ca::IstioCertificateRequest;
+use crate::xds::istio::ca::{IstioCertificateRequest, IstioCertificateResponse};
 
-#[derive(Clone, Debug)]
-pub struct CaClient {
-    pub client: IstioCertificateServiceClient<InterceptedService<TlsGrpcChannel, AuthSource>>,
+#[async_trait]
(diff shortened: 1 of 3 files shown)


src/identity/caclient.rs
```diff
@@ -93,3 +125,97 @@ impl CertificateProvider for CaClient {
...
+    use crate::{
+        identity::{Error, Identity},
+        tls,
+        xds::istio::ca::{IstioCertificateRequest, IstioCertificateResponse},
+    };
+    use async_trait::async_trait;
+
+    #[derive(Clone, Default)]
+    struct MockCertService {
+        response: Option<IstioCertificateResponse>,
```
howardjohn, line 143: purely a nitpick but this could probably just be a single `Result`

src/identity/caclient.rs
```diff
@@ -93,3 +125,97 @@ impl CertificateProvider for CaClient {
...
+                Err(tonic::Status::not_found("response not setup yet"))
+            }
+        }
+    }
+
+    async fn test_ca_client_with_response(
+        res: IstioCertificateResponse,
+    ) -> Result<crate::tls::Certs, crate::identity::Error> {
+        let mut mock_service = Box::new(MockCertService::default());
+        mock_service.set_response(res);
```
howardjohn, line 180: we will eventually need to be able to retain the underlying mock_service probably so we can do something like `set err; fetch; set ok` or the opposite to test refreshes. but we can refactor as needed when we get to that

howardjohn (approved):
nice

src/identity/caclient.rs
stevenctl: That's what I wanted to do, but `Result` isn't clone and MockCertService must be clone :/

We could look at https://docs.rs/mockall/latest/mockall/ to give us an implementation instead of writing our own struct.

src/identity/caclient.rs
stevenctl: yeah just need RefCell on the inner part. But again, maybe that lib is worth it if we have more things to mock.

PR #216, opened 2022-11-29
metrics: refactor, expand, and test

howardjohn:
This handles a few things:

* Refactor how metrics are handled. Have a centralized place that handles all metrics; users consume this struct and just call `.record<T>()` with a variety of defined event types. This is a bit simpler to use IMO and avoids the current issue where we have to pass around a `mut` reference which only works because we only have 1 metric.
* Adds 1 more metric for outbound connections. This is not complete, more of a PoC
* A e2e test for metrics encoding


tests/proxy.rs
```diff
@@ -97,6 +97,22 @@ async fn test_tcp_request() {
...
+async fn test_stats() {
+    testapp::with_app(test_config(), |app| async move {
+        let body = app.admin_request_string("metrics").await;
+        for (metric, t) in &[
+            ("istio_build", "gauge"),
+            ("istio_connection_terminations", "counter"),
+            ("istio_connections_opened", "counter"),
+        ] {
+            let exp = format!("# TYPE {metric} {t}\n");
+            assert!(body.contains(&exp), "expected '{}' in:\n{}", exp, body);
```
zirain, line 110: `contains` is ok now, we can improve it later.
add a TODO for this?

zirain (approved):
lgtm, this's clearer than before

stevenctl (approved):

Issue #220, opened 2022-11-30
stackoverflow on cert provisioning


howardjohn:
I am working on a fix + ensuring we have proper coverage

PR #224, opened 2022-11-30
Re-introduces CA tests

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/220

This swaps out the mock implementation; this was broken since the trait just called itself instead of the underlying client.

Tonic has some example of doing a mock with an in memory stream (https://github.com/hyperium/tonic/blob/master/examples/src/mock/mock.rs) which is nice, but then to use it you need to be super generic over a massive trait. We also don't get as much test coverage.

Instead, this moves to making the CA actually bind to a real port and serve real TLS.


stevenctl (approved):

stevenctl:
Curious how the initial implementation passed on my machine and initially passed in CI..

PR #229, opened 2022-12-01
Fix cache_rx.changed() blocking

hzxuzhonghu:
Fix blocking if a cert is already there.

And also rename spawn to wait_termination

### DIFF
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -172,7 +172,7 @@ impl<T: CertificateProvider + Clone + Send + 'static> CertificateProvider for Se
                 return Err(Error::EmptyResponse(id.clone()));
             }
             Ok(_) => {
-                let current_cert = cache_rx.borrow_and_update().clone();
+                let current_cert = cache_rx.borrow().clone();
                 if let Some(cert) = current_cert {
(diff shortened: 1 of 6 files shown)


src/identity/manager.rs
```diff
@@ -172,7 +172,7 @@ impl<T: CertificateProvider + Clone + Send + 'static> CertificateProvider for Se
                 return Err(Error::EmptyResponse(id.clone()));
             }
             Ok(_) => {
-                let current_cert = cache_rx.borrow_and_update().clone();
+                let current_cert = cache_rx.borrow().clone();
```
hzxuzhonghu, line 175:
```
    /// Returns a reference to the most recently sent value.
    ///
    /// This method does not mark the returned value as seen, so future calls to
    /// [`changed`] may return immediately even if you have already seen the
    /// value with a call to `borrow`.
```

howardjohn: we should have a test to cover this

howardjohn (approved):

src/identity/manager.rs
hzxuzhonghu: Will try to do

PR #235, opened 2022-12-01
Use socket2 crate instead of reimplementing socket options.

PiotrSikora:
While there, fix original destination detection for IPv6 traffic.

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -26,19 +26,19 @@ pub fn set_transparent(l: &TcpListener) -> io::Result<()> {
 }
 
 pub fn orig_dst_addr_or_default(stream: &tokio::net::TcpStream) -> SocketAddr {
-    let sock = SockRef::from(stream);
-    match orig_dst_addr(&sock) {
+    match orig_dst_addr(stream) {
         Ok(Some(addr)) => addr,
         _ => stream.local_addr().expect("must get local address"),
     }
 }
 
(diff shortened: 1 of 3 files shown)


PiotrSikora:
Significant portion of this PR is pending review at https://github.com/rust-lang/socket2/pull/360, but I've decided to also copy it here, since it feels much safer than the code we're using right now.

hzxuzhonghu (approved):

src/socket.rs
```diff
@@ -73,99 +68,51 @@ pub fn set_transparent(_: &TcpListener) -> io::Result<()> {
...
-            let e = io::Error::last_os_error();
-            warn!("failed to read SO_ORIGINAL_DST: {:?}", e);
-            return Err(e);
+    use std::os::unix::io::AsRawFd;
+
+    use socket2::{SockAddr, SockRef};
+    use tokio::io;
+
+    // Replace with socket2's version once there is a release that contains
+    // https://github.com/rust-lang/socket2/pull/360
```
howardjohn, line 77: this is merged. I guess we may need to wait for a release though?

howardjohn (approved):

src/socket.rs
PiotrSikora: Yeah, especially because `master` started development for `v0.5.x` and they didn't make any releases of that yet.

PR #239, opened 2022-12-02
Integrate Codespace

therealmitchconnors:
The dockerfile is a temporary fix until we are confident enough to ship a development image and move config files into common...


docker/remote-env/Dockerfile
```diff
@@ -0,0 +1,80 @@
+# Remote docker environment for ztunnel development.
```
stevenctl, line 1: could move this to a makefile or markdown file

PR #240, opened 2022-12-02
add a test to expect failures with configured addresses

ZhengzheYang:


tests/proxy.rs
```diff
@@ -54,6 +54,39 @@ async fn test_shutdown_lifecycle() {
         .expect("app exits without error")
 }
 
+#[tokio::test]
+async fn test_configured_bound_error() {
```
howardjohn, line 58: The intent of the test was more something like

```
TcpListen(0.0.0.0:12345)
cfg.inbound_addr = 0.0.0.0:12345)
assert app::build().await fails
```

The intent being to validate that we actually fail on port conflict instead of silently run without binding (which we used to do)

ZhengzheYang: Thanks for the clarification! Updated. PTAL.

howardjohn (approved):

PR #243, opened 2022-12-05
Admin api improvements

### LINKED ISSUE #78: Admin API improvements
Split readiness probe and admin interface; admin interface should only be over localhost or UDS.

keithmattix:
Fixes #78 

Port numbers were picked based on a conversation I had with @howardjohn:
> I think we want
readiness: 15021
internal localhost one: 15000
stats: 15020
That should align with what istio is using

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -25,7 +25,7 @@ use tracing::{error, info, warn};
 
 use crate::identity::CertificateProvider;
 use crate::metrics::Metrics;
-use crate::{admin, config, identity, proxy, signal, workload};
+use crate::{admin, config, identity, proxy, readiness, signal, workload};
 
 pub async fn build_with_cert(
     config: config::Config,
@@ -41,7 +41,7 @@ pub async fn build_with_cert(
     // Note: there is still a hard timeout if the draining takes too long
     let (drain_tx, drain_rx) = drain::channel();
 
-    let ready = admin::Ready::new();
+    let ready = readiness::Ready::new();
     let proxy_task = ready.register_task("proxy listeners");
 
     let workload_manager = workload::WorkloadManager::new(
@@ -52,9 +52,13 @@ pub async fn build_with_cert(
     .await?;
 
     let shutdown_trigger = shutdown.trigger();
-    let admin = admin::Builder::new(config.clone(), workload_manager.workloads(), ready)
+    let admin = admin::Builder::new(config.clone(), workload_manager.workloads())
         .bind(registry, shutdown_trigger)
         .context("admin server starts")?;
+    let readiness_server = readiness::Builder::new(config.clone(), ready)
+        .bind(shutdown.trigger())
+        .context("error starting readiness server")?;
+    let readiness_address = readiness_server.address();
     let admin_address = admin.address();
 
     let drain_rx_admin = drain_rx.clone();
@@ -77,6 +81,7 @@ pub async fn build_with_cert(
     });
 
     let proxy_addresses = proxy.addresses();
+    let readiness_drain_rx = drain_rx.clone();
     thread::spawn(move || {
         let runtime = tokio::runtime::Builder::new_multi_thread()
             .worker_threads(config.num_worker_threads)
@@ -89,6 +94,7 @@ pub async fn build_with_cert(
             .build()
             .unwrap();
         runtime.block_on(async move {
+            readiness_server.spawn(readiness_drain_rx);
             proxy.run().await;
         });
     });
@@ -97,6 +103,7 @@ pub async fn build_with_cert(
         drain_tx,
         config,
         shutdown,
+        readiness_address,
         admin_address,
         proxy_addresses,
     })
@@ -115,6 +122,7 @@ pub async fn build(config: config::Config) -> anyhow::Result<Bound> {
 pub struct Bound {
     pub admin_address: SocketAddr,
     pub proxy_addresses: proxy::Addresses,
+    pub readiness_address: SocketAddr,
 
     pub shutdown: signal::Shutdown,
     config: config::Config,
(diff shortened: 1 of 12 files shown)


src/readiness/server.rs
```diff
@@ -0,0 +1,132 @@
...
+
+pub struct Builder {
+    addr: SocketAddr,
+    ready: readiness::Ready,
+}
+
+impl Builder {
+    pub fn new(config: config::Config, ready: readiness::Ready) -> Self {
+        Self {
+            addr: config.admin_addr,
```
howardjohn, line 18:
```suggestion
            addr: config.readiness_addr,
```

src/app.rs
```diff
@@ -51,9 +51,13 @@ pub async fn build_with_cert(
     .await?;
 
     let shutdown_trigger = shutdown.trigger();
-    let admin = admin::Builder::new(config.clone(), workload_manager.workloads(), ready)
+    let admin = admin::Builder::new(config.clone(), workload_manager.workloads())
         .bind(registry, shutdown_trigger)
         .expect("admin server starts");
+    let readiness_server = readiness::Builder::new(config.clone(), ready)
+        .bind(shutdown.trigger())
+        .expect("error starting readiness server");
```
howardjohn, line 59: I know we did it before but 
```suggestion
        .context("error starting readiness server")?;
```

would give better error messages

src/config.rs
```diff
@@ -133,7 +135,11 @@ pub fn parse_config() -> Result<Config, Error> {
         )?
         .0,
 
-        admin_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15021),
+        // admin API should only be accessible over localhost
+        // TODO: Unix domain socket?
+        admin_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::LOCALHOST), 15000),
+        readiness_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15021),
+        stats_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15020),
```
howardjohn, line 142: I don't think this is used yet.

src/config.rs
```diff
@@ -133,7 +135,11 @@ pub fn parse_config() -> Result<Config, Error> {
         )?
         .0,
 
-        admin_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15021),
+        // admin API should only be accessible over localhost
+        // TODO: Unix domain socket?
```
howardjohn, line 139: I think we probably won't do this since it makes `istioctl` usage a pain

src/config.rs
keithmattix: Ah good catch; I can remove it unless you want to use it for your stats work

keithmattix: Removing for now

tests/proxy.rs
```diff
@@ -161,6 +163,19 @@ async fn test_quit_lifecycle() {
         .expect("app exits without error");
 }
 
+#[tokio::test]
+async fn test_readiness() {
```
howardjohn, line 167: we already do this check in `TestApp.ready()` which is called in all other tests. If we do want a test _just_ for readiness we can do that but should re-use the code (which also has a backoff so we don't send a ton of requests in a hot loop)

tests/proxy.rs
```diff
@@ -37,7 +38,8 @@ fn test_config() -> config::Config {
         local_xds_path: Some("examples/localhost.yaml".to_string()),
         socks5_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 0),
         inbound_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 0),
-        admin_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 0),
+        admin_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 0), // Needs to bind to 0.0.0.0 due to the nature of testing environments
```
howardjohn, line 41: hmm, why is that?

howardjohn (approved):

src/config.rs
```diff
@@ -133,7 +134,9 @@ pub fn parse_config() -> Result<Config, Error> {
         )?
         .0,
 
-        admin_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15021),
+        // admin API should only be accessible over localhost
+        admin_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::LOCALHOST), 15000),
```
howardjohn, line 138: since you already have a merge conflcit: can we make this on 0.0.0.0 until we split stats off? otherwise stats are broken until we do.

tests/proxy.rs
keithmattix: I completely missed the whole `test_helpers` directory! 🤦🏾 I'll modify those tests to use the correct port and remove this

tests/proxy.rs
keithmattix: I was getting some strange AddressNotAvailable [errors](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_ztunnel/243/test_ztunnel/1600517946004738048) when I bound to localhost instead of 0.0.0.0. I assumed it was due to some docker bridge networking magic in the test environment

src/config.rs
keithmattix: Got it; will do

tests/proxy.rs
keithmattix: FYI - it looks like `ready()` doesn't check the status code because the changes in this PR should've made admin requests result in a 404

PR #244, opened 2022-12-05
metrics: implement connections metric with full set of labels

howardjohn:
Currently, we just have an empty Connections_opened metric. This introduces the full istio label set (almost - `cluster` is not implemented). It also adds the `closed` variant.

To test this and future metrics, support was added to the tests to parse prometheus metrics and run (trivial) queries over them.

### DIFF
--- a/src/metrics/traffic.rs
+++ b/src/metrics/traffic.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::identity::Identity;
 use prometheus_client::encoding::text::Encode;
 use prometheus_client::metrics::counter::Counter;
 use prometheus_client::metrics::family::Family;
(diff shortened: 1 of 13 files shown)


src/metrics/traffic.rs
```diff
@@ -12,37 +12,214 @@
...
+}
+
+#[derive(Default, Copy, Clone, Hash, PartialEq, Eq, Encode)]
+pub enum SecurityPolicy {
+    #[default]
+    unknown,
+}
+#[derive(Default, Hash, PartialEq, Eq, Clone, Debug)]
+// DefaultedUnknown is a wrapper around an Option that encodes as "unknown" when missing, rather than ""
+struct DefaultedUnknown<T>(Option<T>);
```
stevenctl, line 68: nit: maybe move this to mod.rs? Is it traffic metrics specific?

stevenctl (approved):

PR #247, opened 2022-12-06
stop loop when tokio runtime is shutdown

### LINKED ISSUE #246: Failed TCP handshake A Tokio 1.x context was found, but it is being shutdown.
When running `cargo bench` on `master` ([dbb276bce23a768a20c5bd7f72619e3936d7375b](https://github.com/istio/ztunnel/commit/dbb276bce23a768a20c5bd7f72619e3936d7375b)), I'm flooded with shutdown errors (most of the time, but not 100%):
```
2022-12-06T03:34:31.650700Z ERROR ztunnel::proxy::socks5: Failed TCP handshake A Tokio 1.x context was found, but it is being shutdown.  
2022-12-06T03:34:31.650666Z ERROR ztunnel::proxy::outbound: Failed TCP handshake A Tokio 1.x context was found, but it is being shutdown.
```
Bisecting suggests that it started with #161. cc @hzxuzhonghu

hzxuzhonghu:
fix #246


src/proxy/outbound.rs
```diff
@@ -104,7 +104,12 @@ impl Outbound {
                             };
                         });
                     }
-                    Err(e) => error!("Failed TCP handshake {}", e),
+                    Err(e) => {
+                        if e.get_ref().unwrap().to_string().contains("shutdown") {
```
PiotrSikora, line 108: I think you should compare against the complete error string (i.e. `"A Tokio 1.x context was found, but it is being shutdown."`.

Also, you might want to check if `e.kind() == std::io::ErrorKind::Other` to avoid unnecessary comparisons for other errors.

src/proxy/inbound_passthrough.rs
```diff
@@ -55,7 +55,12 @@ impl InboundPassthrough {
                         }
                     });
                 }
-                Err(e) => error!("Failed TCP handshake {}", e),
+                Err(e) => {
+                    if e.get_ref().unwrap().to_string().eq("shutdown") {
```
PiotrSikora, line 59: This should be `.contains()` instead of `.eq()`, like in the other places.

PiotrSikora (commented):
Thanks for the quick turnaround on this!
 
Do we also need to to cover this for `TcpStream::connect`?

src/proxy/outbound.rs
howardjohn: +1. Unfortunate we cannot use `RUNTIME_SHUTTING_DOWN_ERROR` constant directly since it isn't pub

hzxuzhonghu: I am worried this message could be changed, so I check ket word here

hzxuzhonghu: LMK if you insist on whole message check

src/proxy/inbound_passthrough.rs
hzxuzhonghu: this is my mistake

src/proxy/outbound.rs
howardjohn: or you can add a debug assert that does the slower/less precise check perhaps

hzxuzhonghu: SG

src/proxy/inbound_passthrough.rs
```diff
@@ -55,7 +56,17 @@ impl InboundPassthrough {
                         }
                     });
                 }
-                Err(e) => error!("Failed TCP handshake {}", e),
+                Err(e) => {
+                    debug_assert_eq!(e.get_ref().unwrap().to_string(), ERR_TOKIO_RUNTIME_SHUTDOWN);
```
PiotrSikora, line 60: Unless I'm missing something, this will `panic` in case of any other error, which I don't believe is something we should be doing.

src/proxy/socks5.rs
```diff
@@ -18,6 +18,7 @@ use crate::identity::CertificateProvider;
 use crate::metrics::Metrics;
 use crate::proxy::outbound::OutboundConnection;
 use crate::proxy::Error;
+use crate::proxy::ERR_TOKIO_RUNTIME_SHUTDOWN;
```
PiotrSikora, line 21: Nit: `use crate::proxy::{Error, ERR_TOKIO_RUNTIME_SHUTDOWN};` instead.

src/proxy/inbound_passthrough.rs
```diff
@@ -55,7 +56,17 @@ impl InboundPassthrough {
...
+                    if e.get_ref()
+                        .unwrap()
```
howardjohn, line 62: unwrap doesn't seem safe, not all errors have a ref. We can easily compare without unwrap

src/proxy/socks5.rs
```diff
@@ -91,7 +91,20 @@ impl Socks5 {
                             }
                         });
                     }
-                    Err(e) => error!("Failed TCP handshake {}", e),
+                    Err(e) => {
+                        debug_assert_eq!(
```
howardjohn, line 95: nit: might be nice to have a helper function

hzxuzhonghu:
>>Do we also need to to cover this for TcpStream::connect?

>@hzxuzhonghu any thoughts on that?

@PiotrSikora This maybe not needed, because when we end test no request will be sent. If no request is sent, the  TcpStream::connect will not be called any more

src/proxy/inbound_passthrough.rs
hzxuzhonghu: It will  not panic for release

/// Unlike [`assert_eq!`], `debug_assert_eq!` statements are only enabled in non
/// optimized builds by default. An optimized build will not execute
/// `debug_assert_eq!` statements unless `-C debug-assertions` is passed to the
/// compiler. This makes `debug_assert_eq!` useful for checks that are too
/// expensive to be present in a release build but may be helpful during
/// development. The result of expanding `debug_assert_eq!` is always type checked.

src/proxy/mod.rs
```diff
@@ -208,3 +208,5 @@ pub fn to_canonical_ip(ip: SocketAddr) -> IpAddr {
         },
     }
 }
+
+const ERR_TOKIO_RUNTIME_SHUTDOWN: &str = "A Tokio 1.x context was found, but it is being shutdown.";
```
howardjohn, line 212: nit: would expose this as `fn is_runtime_shutdown(e)`

howardjohn (approved):

src/proxy/mod.rs
hzxuzhonghu: SG

PR #252, opened 2022-12-07
Fix clusterIP access always do on demand request for workload

hzxuzhonghu:


hzxuzhonghu:
Will iterate on the function name, very confusing now

src/workload.rs
```diff
@@ -338,15 +338,41 @@ impl WorkloadInformation {
...
 
-    pub fn find_workload(&self, addr: &IpAddr) -> Option<Workload> {
+    pub async fn find_upstream(&self, addr: SocketAddr, hbone_port: u16) -> Option<Upstream> {
+        self.fetch_workload_on_demand(addr).await;
+        let wi = self.info.lock().unwrap();
+        wi.find_upstream(addr, hbone_port)
+    }
+
+    // It is to do on demand workload fetch if necessary, it handles both workload ip and clusterIP
+    async fn fetch_workload_on_demand(&self, addr: SocketAddr) {
```
howardjohn, line 348: why do we have fetch_workload_on_demand and fetch_workload? Seems like they should all use `fetch_workload_on_demand`?

howardjohn: Ah I guess it makes sense. One is for workload only, the other workload or VIP

it would be nice to have:

```rust
// Only support workload
pub fn fetch_workload() {
  self.find_workload(addr).unwrap_or_else(|| { fetch_on_demand(addr);   self.find_workload(addr))
}
// Support workload and VIP
pub fn fetch_address {
  self.find_workload(addr).or_else(self.workload_by_vip_exist(addr).or_else(|| fetch_on_demand(addr))
}
// internal logic for on demand
fn fetch_on_demand() {
            if let Some(demand) = &self.demand {
                info!("workload not found, sending on-demand request for {addr}");
                demand.demand(addr.ip().to_string()).await.recv().await;
                debug!("on demand ready: {addr}");
            }
}
```

Just a bit more intuitive to me

howardjohn (approved):

PR #254, opened 2022-12-07
xds: allow multiple types

howardjohn:
Currently, we actually only request `Workload`. In the near future, though, we will also fetch Authz policies. This improves the client to handle this.

Note: while we have logic for Authz in the code now, its essentially dead code. this is just a placeholder, in 1-2 weeks we will utilize it.

As part of this, a few things were refactored:

* Handler: instead of a HandlerContext that can be called to explicitly NACK, just have handlers return an Err to NACK. This is easier to use.
* Remove the `Workload(XdsResource<Workload>)` enum; just use `XdsResource<T>` types directly. The enum didn't provide any value and made things more complex

### DIFF
--- a/src/xds/client.rs
+++ b/src/xds/client.rs
@@ -13,18 +13,20 @@
 // limitations under the License.
 
 use std::collections::{HashMap, HashSet};
+use std::fmt::{Display, Formatter};
 use std::sync::Arc;
 use std::time::Duration;
 use std::{fmt, mem};
 
-use prost::Message;
+use prost::DecodeError;
+use thiserror::Error;
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 
 use crate::metrics::xds::*;
 use crate::metrics::{Metrics, Recorder};
-use crate::xds::istio::workload::Workload;
+use crate::xds::istio::workload::{Authorization, Workload};
 use crate::xds::service::discovery::v3::aggregated_discovery_service_client::AggregatedDiscoveryServiceClient;
 use crate::xds::service::discovery::v3::Resource as ProtoResource;
 use crate::xds::service::discovery::v3::*;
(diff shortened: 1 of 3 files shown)


src/xds/client.rs
```diff
@@ -38,50 +40,64 @@ pub struct ResourceKey {
...
 impl<T: prost::Message> Handler<T> for NopHandler {
-    fn handle(&self, _ctx: &mut HandlerContext, _res: Vec<XdsUpdate<T>>) {}
+    fn handle(&self, _res: Vec<XdsUpdate<T>>) -> Result<(), Vec<RejectedConfig>> {
+        Ok(())
+    }
+}
+
+/// handle_single_resource is a helper to process a set of updates with a closure that processes items one-by-one.
+/// It handles aggregating errors as NACKS.
+pub fn handle_single_resource<T: prost::Message, F: FnMut(XdsUpdate<T>) -> anyhow::Result<()>>(
```
stevenctl, line 74: Should this logic to do one-by-one processing and aggregating NACKS be moved into AdClient/wherever we call `handle?` 

The two reasons not to would be:
* Hold the same lock for the entire set of updates
* Some kind of interdependence between updates in a set. We'd have to handle cases where updates are not part of the same set, so I don't think this is a valid reason. 

Not sure of a good way to move this with a single lock unless the Handler interface also exposed some `HandleGuard` wrapping the lock but that indirection might be confusing.

howardjohn: IMO handlers should be _able_ to operate on the full set, but this helper allows them to effectively opt-in to handling 1-by-1. The motivation was the locking mostly.

src/xds/client.rs
```diff
@@ -324,103 +331,123 @@ impl AdsClient {
...
+            self.notify_on_demand(&key);
+            self.known_resources
+                .entry(key.type_url)
+                .or_default()
+                .insert(key.name);
+            r
+        });
+        // Due to lack of dynamic typing in Rust we have some code duplication here. In the future this could be a macro,
+        // but for now its easier to just have a bit of duplication.
+        let handler_response: Result<(), Vec<RejectedConfig>> = match type_url.as_str() {
```
stevenctl, line 405: Could reduce a bit with  `fn decode_and_handle<T>(raw) -> (Vec, Vec)`

stevenctl (approved):

src/xds/client.rs
stevenctl: My point was that operating on the full set is likely not the common case since some set of updates should be considered a small view of a larger continuous stream. 

The only things a handler has to worry about WRT a full set of updates are internal details like locking

Issue #227, opened 2022-12-01
Unit tests for xDS Client


howardjohn:
This is arguably some of the most critical code in the application. We should make sure it's well behaved. Might require some super-light local xDS server impl for tests.

linsun:
just started working on this, per Brian

hzxuzhonghu:
maybe write in golang?

keithmattix:
@howardjohn do we have a prioritized list of what tests are missing? Looks like some were added in https://github.com/istio/ztunnel/commit/f28a7fba761948a518ed79ee05e146feb765fe7b#diff-1e3ac018d4ce0a52fc178625893d303d35d0d57694ca25ad81273951ce218490R621

PR #261, opened 2022-12-08
Issue #227: Adds mock XDS server for unit tests.

briansonnenberg:
Can be merged for now and follow up with more tests.

### DIFF
--- a/src/test_helpers/xds.rs
+++ b/src/test_helpers/xds.rs
@@ -93,16 +93,20 @@ impl AdsServer {
 
         let workloads: Arc<Mutex<WorkloadStore>> = Arc::new(Mutex::new(WorkloadStore::default()));
         let xds_workloads = workloads.clone();
+        let xds_rbac = workloads.clone();
 
         let xds_client = xds::Config::new(cfg)
             .with_workload_handler(xds_workloads)
+            .with_authorization_handler(xds_rbac)
             .watch(xds::WORKLOAD_TYPE.into())
+            .watch(xds::AUTHORIZATION_TYPE.into())
             .build(metrics, ready.register_task("ads client"));
 
         let wi = WorkloadInformation {
             info: workloads,
             demand: None,
         };
+
         (tx, xds_client, wi)
     }
 }
@@ -135,7 +139,13 @@ impl AggregatedDiscoveryService for AdsServer {
                             Ok(_) => {
                                 let response = stream_rx.borrow().clone();
                                 info!("sending response...");
-                                tx.send(response).await.expect("working rx");
+                                match tx.send(response).await {
+                                    Ok(_) => {}
+                                    Err(e) => {
+                                        warn!("ads_server: send failed - {:?} ", e);
+                                        break;
+                                    }
+                                }
                             }
                             Err(_) => {
                                 warn!("ads_server: config update failed");
@@ -149,7 +159,7 @@ impl AggregatedDiscoveryService for AdsServer {
                     }
                 }
             }
-            info!("\tstream ended");
+            info!("stream ended");
         });
 
         let output_stream = ReceiverStream::new(rx);
--- a/src/test_helpers/xds.rs
+++ b/src/test_helpers/xds.rs
@@ -17,27 +17,29 @@ use std::pin::Pin;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
+use crate::config::RootCert;
+use crate::readiness::Ready;
+use crate::workload::{WorkloadInformation, WorkloadStore};
 use async_trait::async_trait;
 use futures::{future, Stream};
 use hyper::service::make_service_fn;
 use log::info;
 use prometheus_client::registry::Registry;
-use tokio::sync::{watch, mpsc};
+use tokio::sync::{mpsc, watch};
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::client::GrpcService;
-use tonic::{Streaming, Status, Response};
+use tonic::{Response, Status, Streaming};
 use tracing::warn;
-use crate::config::RootCert;
-use crate::readiness::Ready;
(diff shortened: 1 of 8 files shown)


src/test_helpers/xds.rs
```diff
@@ -0,0 +1,179 @@
...
+        AdsClient,
+    ) {
+        let delta_default = Err(tonic::Status::not_found("mock not set"));
+        let (delta_tx, delta_rx) = watch::channel(delta_default);
+
+        let default = Err(tonic::Status::not_found("mock not set"));
+        let (tx, rx) = watch::channel(default);
+
+        let server = AdsServer { response: rx, delta_response: delta_rx };
+        let listener = tokio::net::TcpListener::bind("127.0.0.1:15012").await.unwrap();
```
howardjohn, line 59:
```suggestion
        let listener = tokio::net::TcpListener::bind("127.0.0.1:0").await.unwrap();
```

src/test_helpers/xds.rs
```diff
@@ -0,0 +1,179 @@
...
+        (tx, delta_tx, xds_client)
+    }
+}
+
+
+#[async_trait]
+impl AggregatedDiscoveryService for AdsServer {
+    type DeltaAggregatedResourcesStream = Pin<Box<dyn Stream<Item = Result<DeltaDiscoveryResponse, Status>> + Send>>;
+    type StreamAggregatedResourcesStream = Pin<Box<dyn Stream<Item = Result<DiscoveryResponse, Status>> + Send>>;
+    async fn stream_aggregated_resources(
```
howardjohn, line 91: we don't use this in ztunnel, so we can just return an error or panic or something

src/test_helpers/xds.rs
```diff
@@ -0,0 +1,179 @@
...
+
+        let mut in_stream = request.into_inner();
+        let (tx, rx) = mpsc::channel(128);
+
+        tokio::spawn(async move {
+            while let Ok(result) = in_stream.message().await {
+                info!("processing result request: {:?}", result);
+                match result {
+                    Some(_) => tx
+                        .send(Ok(current_response.clone()))
```
howardjohn, line 162: I think we are going to need tobe a little more sophisticated than a static "current_response" for this one. Might need to allow tests to pass a closure or something that emits a response based on the request

briansonnenberg: Can you elaborate on your intuition here? The client should deterministically request resources based on the config, so how about a queue of responses for each case we want to test?  Then each test will essentially be a matter of defining the config for the client, and then writing the potential sequence of responses we want to make sure we handle. And finally validating the end state of our pulled resources on the client of course.

howardjohn: Defining just a strict sequence may be tricky. I guess it can work, it just means our tests need to be really precise/coupled on their expectations of the client behavior rather than more "black box" testing where we just make sure the final state is correct.

We can start simple and expand as needed though

src/test_helpers/xds.rs
```diff
@@ -0,0 +1,160 @@
...
+            80,
+            Some(listener_addr_string),
+            Some(root_cert),
+        );
+
+        let workloads: Arc<Mutex<WorkloadStore>> = Arc::new(Mutex::new(WorkloadStore::default()));
+        let xds_workloads = workloads.clone();
+
+        let xds_client = xds::Config::new(cfg)
+            .with_workload_handler(xds_workloads)
```
howardjohn, line 98: would be good to have the rbac handler now too?

src/test_helpers/xds.rs
```diff
@@ -0,0 +1,160 @@
...
+                            }
+                        };
+                    }
+                    None => {
+                        warn!("ads_server: stream failed");
+                        break;
+                    }
+                }
+            }
+            info!("\tstream ended");
```
howardjohn, line 152:
```suggestion
            info!("stream ended");
```

nit

src/workload.rs
```diff
@@ -97,6 +97,22 @@ pub struct Workload {
     pub authorization_policies: Vec<String>,
 }
 
+impl PartialEq<xds::istio::workload::Workload> for Workload {
+    fn eq(&self, other: &xds::istio::workload::Workload) -> bool {
```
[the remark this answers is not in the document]
briansonnenberg: To be clear, this is a comparison between our Workload struct and the prost Workload struct.  Does prost really allow derived implementations between prost and non-prost structs?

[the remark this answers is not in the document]
howardjohn: oops, missed that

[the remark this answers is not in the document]
howardjohn: what is it for? I would expect someone to convert XdsWorkload to Workload (From) and then compare eq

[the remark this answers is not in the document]
briansonnenberg: It was just for convenient, concise comparison of the input server responses and the output workload state.  Converting should work too if you think this is overkill.

src/xds/client.rs
```diff
@@ -608,4 +608,127 @@ pub enum AdsError {
...
+            .expect("failed to send server response");
+        verify_workload(
+            std::net::IpAddr::V4(ip),
+            Some(workloads[0].clone()),
+            &workload_store,
+        )
+        .await;
+        tx.send(abort_response)
+            .expect("failed to send server response");
+        sleep(Duration::from_millis(50)).await;
```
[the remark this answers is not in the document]
briansonnenberg: So there needs to be some time for the abort server response to actually be sent and processed, but there's no workload changes to wait for.

src/workload.rs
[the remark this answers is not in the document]
howardjohn: I am a bit worried about this not actually checking everythign and getting used in production code when its really only meant for tests

howardjohn (approved):

PR #263, opened 2022-12-09
Abstract tokio runtime shut down check

hzxuzhonghu:


src/proxy/util.rs
```diff
@@ -0,0 +1,24 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::io::{Error, ErrorKind};
+
+pub fn is_runtime_shutdown(e: &Error) -> bool {
+    match e.kind() {
+        ErrorKind::Other => {
```
hzxuzhonghu, line 19: I am not sure which is the idiotic way for rust

`match vs if`

PiotrSikora: Since we care about both conditions, I'd go with `if ... && ...` instead of this `match` and `==` mix.

src/workload.rs
```diff
@@ -309,13 +309,12 @@ impl WorkloadInformation {
     pub async fn fetch_workload(&self, addr: &IpAddr) -> Option<Workload> {
```
PiotrSikora, line 309: All those changes in `workload.rs` seem unrelated.

hzxuzhonghu: These are meant to be commited to https://github.com/istio/ztunnel/pull/262

src/proxy/util.rs
hzxuzhonghu: ok

howardjohn (approved):

PR #264, opened 2022-12-09
support original source proxy

PlatformLC:
Related to #241 
support original source proxy

-  HBONE connection:  use XFF header to encode the original source information.
-  local shortcut or direct tcp stream:  retrieve original source information from socket directly.

```shell
lichun@lichun-ws:~$ kubectl get pods -o wide 
NAME                                   READY   STATUS    RESTARTS      AGE   IP            NODE            NOMINATED NODE   READINESS GATES
httpbin-no-mesh-654476d4ff-pfgxp       1/1     Running   0             35m   10.101.2.14   kind1-worker3   <none>           <none>
httpbin-worker1-5d5748d687-pf4cb       1/1     Running   0             77m   10.101.3.15   kind1-worker    <none>           <none>
httpbin-worker2-f4b4788cb-vw2hz        1/1     Running   0             77m   10.101.1.9    kind1-worker2   <none>           <none>
httpbin-worker3-6989c5fb68-46g6r       1/1     Running   0             77m   10.101.2.13   kind1-worker3   <none>           <none>
notsleep-7dc8b55755-trgqk              1/1     Running   4 (29h ago)   3d    10.101.2.2    kind1-worker3   <none>           <none>
purchase-history-v1-6cdb997954-6rhfd   1/1     Running   4 (29h ago)   3d    10.101.3.2    kind1-worker    <none>           <none>
recommendation-79f7844ff-scmjq         1/1     Running   4 (29h ago)   3d    10.101.3.3    kind1-worker    <none>           <none>
sleep-5f5d4b5bfb-cdtrr                 1/1     Running   4 (29h ago)   24d   10.101.2.5    kind1-worker3   <none>           <none>
web-api-6c95c8d759-92d42               1/1     Running   4 (29h ago)   3d    10.101.3.4    kind1-worker    <none>           <none>
lichun@lichun-ws:~$ kubectl exec deploy/sleep -- curl -s http://httpbin-worker1:8000/ip
{
  "origin": "10.101.2.5"
}
lichun@lichun-ws:~$ kubectl exec deploy/sleep -- curl -s http://httpbin-worker2:8000/ip
{
  "origin": "10.101.2.5"
}
lichun@lichun-ws:~$ kubectl exec deploy/sleep -- curl -s http://httpbin-worker3:8000/ip
{
  "origin": "10.101.2.5"
}

```

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -138,7 +138,7 @@ impl Inbound {
                 Hbone(req) => {
                     let origin_src = if origin_src.is_some() {
                         // overwrite the original source if encoded in fwded from waypoint
-                        super::get_original_src_from_fwded(req)
+                        super::get_original_src_from_fwded(req).map_or(origin_src, Some)
                     } else {
                         origin_src
                     };
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -28,7 +28,6 @@ use crate::config::Config;
 use crate::identity::CertificateProvider;
 use crate::proxy::inbound::InboundConnect::{DirectPath, Hbone};
 use crate::proxy::{ProxyInputs, TraceParent, TRACEPARENT_HEADER};
-use crate::socket;
 use crate::rbac;
 use crate::socket::{relay, to_canonical};
 use crate::tls::TlsError;
@@ -83,9 +82,10 @@ impl Inbound {
                 dst,
             };
             let workloads = self.workloads.clone();
+            let enable_original_source = self.cfg.enable_original_source;
             async move {
                 Ok::<_, hyper::Error>(service_fn(move |req| {
-                    Self::serve_connect(workloads.clone(), conn.clone(), req)
+                    Self::serve_connect(workloads.clone(), conn.clone(), enable_original_source, req)
                 }))
             }
         });
@@ -134,12 +134,14 @@ impl Inbound {
         let start = Instant::now();
         let stream = {
             match &request_type {
-                DirectPath(_) => {
-                    // let org_src = super::get_original_src_from_stream(stream);
-                    super::freebind_connect(origin_src, addr).await
-                }
-                Hbone(_) => {
-                    // let org_src = super::get_original_src_from_fwded(req);
+                DirectPath(_) => super::freebind_connect(origin_src, addr).await,
+                Hbone(req) => {
+                    let origin_src = if origin_src.is_some() {
+                        // overwrite the original source if encoded in fwded from waypoint
+                        super::get_original_src_from_fwded(req)
+                    } else {
+                        origin_src
+                    };
                     super::freebind_connect(origin_src, addr).await
                 }
             }
@@ -205,6 +207,7 @@ impl Inbound {
     async fn serve_connect(
         workloads: WorkloadInformation,
         conn: rbac::Connection,
+        enable_original_source: bool,
         req: Request<Body>,
     ) -> Result<Response<Body>, hyper::Error> {
         match req.method() {
@@ -237,6 +240,11 @@ impl Inbound {
                         .body(Body::empty())
                         .unwrap());
                 }
+                let org_src = if enable_original_source {
+                    Some(conn.src_ip.clone())
+                } else {
+                    None
+                };
 
                 let Some(upstream) = workloads.fetch_workload(&addr.ip()).await else {
                     info!(%conn, "unknown destination");
@@ -259,7 +267,7 @@ impl Inbound {
                             .unwrap());
                     }
                 }
-                let status_code = match Self::handle_inbound(Hbone(req), Some(conn.src_ip), addr)
+                let status_code = match Self::handle_inbound(Hbone(req), org_src, addr)
                     .instrument(tracing::span::Span::current())
                     .await
                 {
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -14,7 +14,7 @@
 
 use std::fmt;
 use std::fmt::{Display, Formatter};
-use std::net::SocketAddr;
+use std::net::{IpAddr, SocketAddr};
 use std::time::Instant;
 
 use drain::Watch;
@@ -26,8 +26,9 @@ use tracing::{debug, error, info, instrument, trace, trace_span, warn, Instrumen
 
 use crate::config::Config;
 use crate::identity::CertificateProvider;
-use crate::proxy::inbound::InboundConnect::Hbone;
+use crate::proxy::inbound::InboundConnect::{DirectPath, Hbone};
 use crate::proxy::{ProxyInputs, TraceParent, TRACEPARENT_HEADER};
+use crate::socket;
 use crate::rbac;
 use crate::socket::{relay, to_canonical};
 use crate::tls::TlsError;
@@ -127,18 +128,19 @@ impl Inbound {
     /// handle_inbound serves an inbound connection with a target address `addr`.
     pub(super) async fn handle_inbound(
         request_type: InboundConnect,
+        origin_src: Option<IpAddr>,
         addr: SocketAddr,
     ) -> Result<(), std::io::Error> {
         let start = Instant::now();
         let stream = {
             match &request_type {
-                InboundConnect::DirectPath(stream) => {
-                    let org_src = super::get_original_src_from_stream(stream);
-                    super::freebind_connect(org_src, addr).await
+                DirectPath(_) => {
+                    // let org_src = super::get_original_src_from_stream(stream);
+                    super::freebind_connect(origin_src, addr).await
                 }
-                Hbone(req) => {
-                    let org_src = super::get_original_src_from_fwded(req);
-                    super::freebind_connect(org_src, addr).await
+                Hbone(_) => {
+                    // let org_src = super::get_original_src_from_fwded(req);
+                    super::freebind_connect(origin_src, addr).await
                 }
             }
         };
@@ -257,7 +259,7 @@ impl Inbound {
                             .unwrap());
                     }
                 }
-                let status_code = match Self::handle_inbound(Hbone(req), addr)
+                let status_code = match Self::handle_inbound(Hbone(req), Some(conn.src_ip), addr)
                     .instrument(tracing::span::Span::current())
                     .await
                 {
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -130,7 +130,18 @@ impl Inbound {
         addr: SocketAddr,
     ) -> Result<(), std::io::Error> {
         let start = Instant::now();
-        let stream = TcpStream::connect(addr).await;
+        let stream = {
+            match &request_type {
+                InboundConnect::DirectPath(stream) => {
+                    let org_src = super::get_original_src_from_stream(stream);
+                    super::freebind_connect(org_src, addr).await
+                },
+                Hbone(req) => {
+                    let org_src = super::get_original_src_from_xff(req);
+                    super::freebind_connect(org_src, addr).await
+                }
+            }
+        };
         match stream {
             Err(err) => {
                 warn!(dur=?start.elapsed(), "connection to {} failed: {}", addr, err);
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -135,7 +135,7 @@ impl Inbound {
                 InboundConnect::DirectPath(stream) => {
                     let org_src = super::get_original_src_from_stream(stream);
                     super::freebind_connect(org_src, addr).await
-                },
+                }
                 Hbone(req) => {
                     let org_src = super::get_original_src_from_xff(req);
                     super::freebind_connect(org_src, addr).await
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -274,12 +274,12 @@ impl Inbound {
                     }
                 }
                 let status_code = match Self::handle_inbound(Hbone(req), org_src, addr)
-                        .instrument(tracing::span::Span::current())
-                        .await
-                    {
-                        Ok(_) => StatusCode::OK,
-                        Err(_) => StatusCode::SERVICE_UNAVAILABLE,
-                    };
+                    .instrument(tracing::span::Span::current())
+                    .await
+                {
+                    Ok(_) => StatusCode::OK,
+                    Err(_) => StatusCode::SERVICE_UNAVAILABLE,
+                };
 
                 Ok(Response::builder()
                     .status(status_code)
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -137,7 +137,7 @@ impl Inbound {
                     super::freebind_connect(org_src, addr).await
                 }
                 Hbone(req) => {
-                    let org_src = super::get_original_src_from_xff(req);
+                    let org_src = super::get_original_src_from_fwded(req);
                     super::freebind_connect(org_src, addr).await
                 }
             }
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -85,7 +85,12 @@ impl Inbound {
             let enable_original_source = self.cfg.enable_original_source;
             async move {
                 Ok::<_, hyper::Error>(service_fn(move |req| {
-                    Self::serve_connect(workloads.clone(), conn.clone(), enable_original_source, req)
+                    Self::serve_connect(
+                        workloads.clone(),
+                        conn.clone(),
+                        enable_original_source,
+                        req,
+                    )
                 }))
             }
         });
@@ -241,13 +246,14 @@ impl Inbound {
                         .unwrap());
                 }
                 let org_src = if enable_original_source {
-                    Some(conn.src_ip.clone())
+                    Some(conn.src_ip)
                 } else {
                     None
                 };
 
                 let Some(upstream) = workloads.fetch_workload(&addr.ip()).await else {
-                    info!(%conn, "unknown destination");
+
+                                       info!(%conn, "unknown destination");
                     return Ok(Response::builder()
                         .status(StatusCode::NOT_FOUND)
                         .body(Body::empty())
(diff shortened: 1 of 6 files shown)


src/proxy/inbound.rs
```diff
@@ -113,7 +113,18 @@ impl Inbound {
...
-        let stream = TcpStream::connect(addr).await;
+        let stream = {
+            match &request_type {
+                InboundConnect::DirectPath(stream) => {
+                    let org_src = super::get_original_src_from_stream(stream);
+                    super::freebind_connect(org_src, addr).await
+                },
+                Hbone(req) => {
+                    let org_src = super::get_original_src_from_xff(req);
+                    super::freebind_connect(org_src, addr).await
```
hzxuzhonghu, line 124: Curious if we bind to `org_src`, how could the response packet from the endpoint routed back to this ztunnel

PlatformLC: In redirection part(i.e. iptables configured by isito-cni) take care of it, it will use conntrack mark and policy route to make the rsp back to ztunnel.

src/socket.rs
```diff
@@ -73,6 +89,23 @@ mod linux {
     use socket2::{SockAddr, SockRef};
     use tokio::io;
 
+    pub fn set_ip_freebind(sock: &SockRef) -> io::Result<()> {
```
PiotrSikora, line 92: You should use [`socket2::set_freebind`](https://docs.rs/socket2/latest/socket2/struct.Socket.html#method.set_freebind) instead of reimplementing (see: #235).

Also, you need to use[`socket2::set_freebind_ipv6`](https://docs.rs/socket2/latest/socket2/struct.Socket.html#method.set_freebind_ipv6) for IPv6.

src/socket.rs
```diff
@@ -19,13 +19,21 @@ use realm_io;
...
 
 #[cfg(target_os = "linux")]
 pub fn set_transparent(l: &TcpListener) -> io::Result<()> {
     SockRef::from(l).set_ip_transparent(true)
 }
 
+#[cfg(target_os = "linux")]
+pub fn set_freebind(sock: &TcpSocket) -> io::Result<()> {
+    let sock = SockRef::from(sock);
+    sock.set_ip_transparent(true)?;
```
PiotrSikora, line 33: With this change, ztunnel must be running with `CAP_NET_ADMIN` privileges, whereas previously that wasn't necessary.

I think we need to make a decision on what behavior do we want, since right now it will fail at runtime.

If we want to require `CAP_NET_ADMIN` capabilities, then we should check if `IP_TRANSPARENT` works at startup, and bail out if it's not. We should also modify existing code paths that tries to enable transparent mode to fail then.

If we want to proxy traffic even when running in non-transparent mode, then we need to ignore errors when setting `IP_TRANSPARENT` and `IP_FREEBIND`.

PiotrSikora (changes requested):

src/socket.rs
howardjohn: IMO we should allow either mode. We already do this for setting IP_TRANSPARENT on the listener

PiotrSikora: Should we make this behavior configurable instead of opportunistic?

src/proxy/outbound.rs
```diff
@@ -201,6 +201,7 @@ impl OutboundConnection {
                     .method(hyper::Method::CONNECT)
                     .version(hyper::Version::HTTP_2)
                     .header("baggage", baggage(&req))
+                    .header("X-Forwarded-For", remote_addr.to_string())
```
PiotrSikora, line 204: I think we need to use `Forwarded` header here, since presumably we want to bind to the source IP and port.

src/proxy/inbound.rs
hzxuzhonghu: Donot we need to setup an additional route table?

hzxuzhonghu: I remember we do that for original src address support in istio.

hzxuzhonghu: https://www.envoyproxy.io/docs/envoy/latest/configuration/listeners/listener_filters/original_src_filter#extra-setup

src/proxy/mod.rs
```diff
@@ -209,4 +211,55 @@ pub fn to_canonical_ip(ip: SocketAddr) -> IpAddr {
...
+pub fn get_original_src_from_xff(req: &Request<Body>) -> Option<IpAddr> {
+    let value = req.headers().get("X-Forwarded-For")?;
+    let ip = value
+        .to_str()
+        .unwrap_or("")
+        .to_string()
+        .split(',')
+        .next()?
+        .trim()
+        .parse::<IpAddr>();
```
KfreeZ, line 223: maybe you can try map_or() here. for simplicity

PlatformLC: Thanks for sharing the info. Seems `map_or()` and `ok()` convenient ways to simplify the match blocks.

src/socket.rs
PlatformLC: Thanks for sharing the info. Updated, but set IPV6_TRANPARENT option is not available currently, so keep its implementation in our mod.

src/socket.rs
PlatformLC: > IMO we should allow either mode. We already do this for setting IP_TRANSPARENT on the listener

Should we have a configurable flag to indicate we use transparent mode or not?

src/proxy/inbound.rs
PlatformLC: > I remember we do that for original src address support in istio.

Yes, for ambient ztunnel, these works have been covered  by istio-cni && ztunnel init container.
i.e. : 
In host:
```
-A ztunnel-PREROUTING ! -i vethd7b59a4c -m connmark --mark 0x210/0x210 -j MARK --set-xmark 0x40/0x40
-A ztunnel-PREROUTING -m mark --mark 0x40/0x40 -j RETURN
-A ztunnel-PREROUTING ! -s 10.101.3.14/32 -i vethd7b59a4c -j MARK --set-xmark 0x210/0x210  # ip for ztunnel pod ip, veth is the veth name of ztunnel

# ip rule
0:	from all lookup local
100:	from all fwmark 0x200/0x200 goto 32766
101:	from all fwmark 0x100/0x100 lookup 101
**102:	from all fwmark 0x40/0x40 lookup 102**

# ip r show table 102
default via 10.101.3.14 dev vethd7b59a4c onlink 
10.101.3.14 dev vethd7b59a4c scope link
```

In ztunnel:
```
-A PREROUTING ! -d 10.101.3.14/32 -i eth0 -p tcp -j MARK --set-xmark 0x4d3/0xfff

# ip rule 
0:	from all lookup local
20000:	from all fwmark 0x400/0xfff lookup 100
20001:	from all fwmark 0x401/0xfff lookup 101
20002:	from all fwmark 0x402/0xfff lookup 102
20003:	from all fwmark 0x4d3/0xfff lookup 100
32766:	from all lookup main
32767:	from all lookup default
# ip r show table 100
local default dev lo scope host 
```

howardjohn (changes requested):
This is a great start but I have some concerns. I am going to do some more research on the header part.

src/proxy/mod.rs
```diff
@@ -244,3 +246,64 @@ impl TryFrom<&str> for TraceParent {
         })
     }
 }
+pub fn get_original_src_from_fwded(req: &Request<Body>) -> Option<IpAddr> {
```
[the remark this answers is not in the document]
howardjohn: We don't necessarily know its from another ztunnel, we just know its from someone that has a valid mTLS certificate. So its not "public internet" untrusted, but its also a source of risk.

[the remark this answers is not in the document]
PlatformLC: The traffic transmitted should be safe due mTLS certificate guaranteed(i.e. from ztunnel or waypoints) , am I right?
So the security  concern is 

1. at encode side, guarantee encoded IP is the trusted one? 
2. at decode side, guarantee decoded IP is also the tursted one?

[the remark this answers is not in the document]
howardjohn: mTls doesn't encode which IP it's allowed to impersonate unless we also lookup the identity of the IP address. However, we may not know all info about all clients, so that doesn't work.

A header in hbone is essentially untrusted entirely. The only thing we know is they have an mTLS cert.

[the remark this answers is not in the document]
howardjohn: sorry I think my comments got out of order since I forgot to hit "send"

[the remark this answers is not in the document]
PlatformLC: > 

Briefly went through the tweaked code,  we use HBONE tcp/ip layer to transfer to the original source info instead of http header. This could simplify and eliminate the security concern for forwarded head.

However spoofing the HBONE src_ip should be the same concern as non-meshed APP spoofing the src_ip?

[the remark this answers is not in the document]
howardjohn: Yes but I think that is ok - we are preserving the semantics of Kubernetes here. If users chose to add source-based checks to their apps (which are not necessarily secure) they get the same behavior with and without ztunnel, making it transparent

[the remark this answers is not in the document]
howardjohn: Not quite.

1. FORWARDED parsing is sensitive. We can make it safe though, so this is minor issue
2. FORWARDED fundamentally is unsafe because we are putting highly trusted data (source IP to spoof) into a low(er) trust payload (HBONE headers). This allows privilege escalation
3. A safer mechanism is to make ztunnel outbound just spoof the IP and then propogate the SRC IP from that rather than read headers; the diff I pasted above does that and I have tested it works.

src/proxy/inbound.rs
```diff
@@ -73,8 +74,14 @@ impl Inbound {
```
howardjohn, line 74: can we also add proxy_inbound_plaintext
```

        let mut outbound = super::freebind_connect(Some(source.ip()), orig).await?;
```

src/proxy/mod.rs
```diff
@@ -244,3 +246,64 @@ impl TryFrom<&str> for TraceParent {
         })
     }
```
howardjohn, line 247: Once we merge this we should add test to istio/istio:

```diff
diff --git a/tests/integration/ambient/baseline_test.go b/tests/integration/ambient/baseline_test.go
index 9d85cb4eac..b9aae41adc 100644
--- a/tests/integration/ambient/baseline_test.go
+++ b/tests/integration/ambient/baseline_test.go
@@ -116,6 +116,17 @@ func supportsL7(opt echo.CallOptions, instances ...echo.Instance) bool {
 	return (hasWaypoint || hasSidecar) && isL7Scheme
 }
 
+func OriginalSourceCheck(t framework.TestContext, src echo.Instance) echo.Checker{
+		// Check that each response saw one of the workload IPs for the src echo instance
+			addresses := sets.New(src.WorkloadsOrFail(t).Addresses()...)
+		return check.Each(func(response echot.Response) error {
+			if !addresses.Contains(response.IP) {
+				return fmt.Errorf("expected original source (%v) to be propogated, but got %v", addresses.UnsortedList(), response.IP)
+			}
+			return nil
+		})
+}
+
 func TestServices(t *testing.T) {
 	runTest(t, func(t framework.TestContext, src echo.Instance, dst echo.Instance, opt echo.CallOptions) {
 		if supportsL7(opt, src, dst) {
@@ -130,6 +141,17 @@ func TestServices(t *testing.T) {
 			// TODO: fix this and remove this skip
 			opt.Check = check.OK()
 		}
+
+		if !dst.Config().HasWaypointProxy() &&
+			!src.Config().HasWaypointProxy() &&
+			(src.Config().Service != dst.Config().Service) &&
+			!dst.Config().HasSidecar() {
+			// Check original source, unless there is a waypoint in the path. For waypoint, we don't (yet?) propagate original src.
+			// Self call is also (temporarily) broken
+			// Sidecars lose the original src
+			opt.Check = check.And(opt.Check, OriginalSourceCheck(t, src))
+		}
+
 		// TODO test from all source workloads as well
 		src.CallOrFail(t, opt)
 	})
```

howardjohn (commented):
@PiotrSikora any other comments ? I had a few small ones but don't mind doing them in follow-ups

src/proxy/mod.rs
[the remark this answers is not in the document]
PiotrSikora: Regarding spoofing on the outbound - I don't believe this works for the general case, since it requires the inbound ztunnel to have a routing table configured to send traffic to spoofed IP via outbound ztunnel.

src/socket.rs
PiotrSikora: I believe that `IP_TRANSPARENT` works for both IPv4 and IPv6, so `IPV6_TRANPARENT` might not be necessary. But it would be good to verify.

src/proxy/mod.rs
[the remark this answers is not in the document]
howardjohn: It seems to work... not sure I understand exactly how though, the redirection/routing table configuration we have is pretty complex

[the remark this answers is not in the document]
PiotrSikora: Some more thoughts on this:
1. Could it be the case that in tests workload IP == outbound ztunnel IP, so we're not really spoofing the IP address?
2. I don't believe it will work with waypoints, since the spoofed IP will be lost between waypoint and inbound ztunnel.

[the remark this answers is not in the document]
PiotrSikora: ...and some more:

3. The advantage of `Forwarded` header is that it works end-to-end without complex network routing setup.
4. If we're allowing peers to connect using spoofed IP address, then it seems that the vulnerability is already there.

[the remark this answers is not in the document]
howardjohn: 1. I agree we probably don't have adequate test coverage here. Its a bit tricky to begin with since CI happens to run as root (probably not a good idea... but that is how it is configured across the project today) but locally folks probably aren't.
2. is definitely true. In theory the waypoint could do it as well although I don't know that we want to give it NET_ADMIN. Generally the idea has been that ztunnel should be as transparent as possible while waypoint is essentially opting in to behavioral changes, so losing orig src with waypoint doesn't seem as bad (and in many cases could be sent with XFF header or similar?).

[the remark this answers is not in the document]
PiotrSikora: 2. I don't think that we should provide an inconsistent behavior of having/not having the original source depending on whether waypoint is used or not. Also, sending `X-Forwaded-For` (which is `Forwarded` without a source port) requires inbound ztunnel to parse IP from the header, so we're back at square one.
4. Agreed, except that the thing we're providing instead might not actually work ;)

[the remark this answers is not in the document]
howardjohn: re re 2: XFF or Forwarded was meant for user application, I didn't mean ztunnel would do anything

4. Aside from waypoints (which, IMO, are a different story) is there case we know it doesn't work?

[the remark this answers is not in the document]
PiotrSikora: re 4. I'm pretty sure it wouldn't work if workload IP != outbound ztunnel IP, but I haven't tested it.

I think one solution to do this properly and securely would be to map outbound ztunnel to a list of workloads it serves (either via xDS or maybe even in the TLS client certificate, since the outbound ztunnel or something mimicking it could technically spoof its source IP and pretend to be a different workload), so that we know that the connecting IP is allowed to forward traffic for the IP address from the `Forwarded` header.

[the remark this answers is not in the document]
howardjohn: In our k8s test (https://github.com/istio/ztunnel/pull/264#discussion_r1047475451) it works if: there is no waypoint (as noted, this breaks the source), the server doesn't have a sidecar (this also breaks the source, unless TPROXY mode is enabled), and the src!=dst (self call; I think this is just an issue in our redirection logic that can be fixed).

The mapping was something I had considered but it seems concerning for an inbound ztunnel to need to know info about ALL possible clients. We don't always have global knowledge.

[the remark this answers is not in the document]
PiotrSikora: What if we embed this information in the TLS certificate? Then we wouldn't need to have the global knowledge.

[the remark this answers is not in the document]
howardjohn: That is an interesting idea... there have been a variety of other discussions that settled on "Put the IP in the cert" as well. My initial concern had been that we integrate with a variety of third party CAs which would all need to be updated to encode the IP (which may not be feasible for some), but it may be worth considering adding this requirement regardless given the possible benefits.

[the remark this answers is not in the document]
PiotrSikora: Yeah, I think that's the right solution, especially if the IP address can be spoofed on the network...

But that's a much larger change, so in the interest of unblocking this PR, could we make this configurable for now?
1. non-transparent,
2. spoof outbound IP,
3. use `Forwarded` header (needed for waypoints and when IP address cannot be spoofed on the network).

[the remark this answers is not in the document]
howardjohn: I am good with a short term approach for now. I started a doc on the options discussed here https://docs.google.com/document/d/17fDTMLg-22UPfh62d5wumAOIpbeoOyWix2WTE54hDNc/ so its not just in a github comment and asked some other folks to review. If the main concern is the waypoint hop I think we can handle it with a hybrid approach of using `Forwarded` in trusted cases but not all cases. I am not sure that handles all cases you are interested in, though.

howardjohn:
@yuval-k see https://docs.google.com/document/d/17fDTMLg-22UPfh62d5wumAOIpbeoOyWix2WTE54hDNc/edit# and/or https://github.com/istio/ztunnel/pull/264#discussion_r1046384840 for more context, would appreciate your feedback on the doc

src/proxy/inbound.rs
```diff
@@ -114,10 +123,24 @@ impl Inbound {
...
+        origin_src: Option<IpAddr>,
         addr: SocketAddr,
     ) -> Result<(), std::io::Error> {
         let start = Instant::now();
-        let stream = TcpStream::connect(addr).await;
+        let stream = {
+            match &request_type {
+                DirectPath(_) => super::freebind_connect(origin_src, addr).await,
+                Hbone(req) => {
+                    let origin_src = if origin_src.is_some() {
```
howardjohn, line 134: why only if it is_some()?

src/proxy/inbound.rs
```diff
@@ -73,8 +74,16 @@ impl Inbound {
 
     pub(super) async fn run(self) {
         let (tx, rx) = oneshot::channel();
-        let service =
-            make_service_fn(|_| async { Ok::<_, hyper::Error>(service_fn(Self::serve_connect)) });
+        let service = make_service_fn(|socket: &tokio_boring::SslStream<tokio::net::TcpStream>| {
+            let peer_addr = if self.cfg.enable_original_source {
```
howardjohn, line 78: this seems a bit unexpected to me, I would expect `enable_original_source` to maybe only be checked in freebind_connect? I guess that would require passing `cfg` all the way through?

PlatformLC:
Just rebase to the latest master branch. @howardjohn  Would you help to review it again?

src/proxy/inbound.rs
howardjohn: answer myself: None is if its not enabled; we always get original src if its enabled

howardjohn (approved):

PR #268, opened 2022-12-09
waypoints: disable consumer waypoints

howardjohn:
Per
https://docs.google.com/document/d/1QdpTl0y2snASOPo8GEKapuwDD-Mfoy2zz8hPVbgRexA

### DIFF
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -280,27 +280,6 @@ impl OutboundConnection {
             });
         }
 
-        // For case source client has enabled waypoint
-        if !source_workload.waypoint_addresses.is_empty() {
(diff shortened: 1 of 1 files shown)


src/proxy/outbound.rs
```diff
@@ -280,27 +280,6 @@ impl OutboundConnection {
             });
         }
 
-        // For case source client has enabled waypoint
-        if !source_workload.waypoint_addresses.is_empty() {
```
stevenctl, line 284: Should this be a feature flag? And do any of the istio/istio tests have behavior that require client waypoints?

howardjohn: @dhawton is working on the istio/istio side, I should probably mark this WIP until those changes are ready

stevenctl (approved):

PR #269, opened 2022-12-12
logging: cleanup and expand scope

howardjohn:
This adds a variety of improvements to the recent dynamic logging:
* use Result instead of Option and bool when appopriate
* DRY a bit of the hyper responses
* Add a `reset` parameter, otherwise its not possible to reset logs back to normal. Envoy doesn't really have this issue since setting the level to "info" means "set all loggers to info"; in `tracing` this means "set all *unset* loggers to info". So setting it doesn't imply resetting
* Simplify some of the code
* Properly parse query params

### DIFF
--- a/src/admin/mod.rs
+++ b/src/admin/mod.rs
@@ -14,7 +14,7 @@
 
 // Forked from https://github.com/olix0r/kubert/blob/main/kubert/src/admin.rs
 
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::sync::Mutex;
 use std::{net::SocketAddr, sync::Arc, time::Duration};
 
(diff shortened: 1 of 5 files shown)


src/admin/mod.rs
```diff
@@ -389,74 +373,78 @@ async fn handle_metrics(reg: Arc<Mutex<Registry>>, _req: Request<Body>) -> Respo
 }
 
 //mirror envoy's behavior: https://www.envoyproxy.io/docs/envoy/latest/operations/admin#post--logging
-//NOTE: mutilple query parameters is not supported, for example
+//NOTE: multiple query parameters is not supported, for example
 //curl -X POST http://127.0.0.1:15021/logging?"tap=debug&router=debug"
 static HELP_STRING: &str = "
 usage: POST /logging\t\t\t\t\t\t(To list current level)
 usage: POST /logging?level=<level>\t\t\t\t(To change global levels)
 usage: POST /logging?level={mod1}:{level1},{mod2}:{level2}\t(To change specific mods' logging level)
```
KfreeZ, line 382: add a /reset here?

stevenctl (approved):

PR #272, opened 2022-12-12
fix two e2e test flakes

howardjohn:
* Bind test always used static port. make it dynamic to avoid conflicts on test. Also test all 5 ports. Found inbound_passthrough and admin are broken, fixed them
* Metrics test fetches metrics 1x then checks them repeatedly. Need to fetch+check in the assert.

### DIFF
--- a/tests/proxy.rs
+++ b/tests/proxy.rs
@@ -13,7 +13,8 @@
 // limitations under the License.
 
 use std::fmt::Debug;
-use std::net::{IpAddr, Ipv6Addr, SocketAddr};
+use std::future::Future;
+use std::net::SocketAddr;
 use std::ops::Add;
 use std::str::FromStr;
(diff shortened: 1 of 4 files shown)


tests/proxy.rs
```diff
@@ -45,15 +46,40 @@ async fn test_shutdown_lifecycle() {
         .expect("app exits without error")
 }
 
-#[tokio::test]
-async fn test_conflicting_bind_error() {
+// Check that port conflicts on any address results in the app failing instead of silently failing
+async fn test_bind_conflict<F: FnOnce(&mut ztunnel::config::Config) -> &mut SocketAddr>(f: F) {
```
hzxuzhonghu, line 50: Would make it more fine granularity by splitting into two functions, first allocate a idle port and second set to Config

stevenctl (approved):

PR #273, opened 2022-12-12
Moving boringssl to vendor/ folder and add to .gitattributes

howardjohn:
This will stop making it show ztunnel as a "C project"

note: .gitattribute change will be copied to common-files after it merges

```
$ github-linguist
74.45%  1154827    C
15.50%  240415     Rust
6.44%   99913      C++
2.87%   44537      Shell
0.55%   8572       Makefile
0.16%   2557       Dockerfile
0.02%   317        Ruby
```

after:
```
$ github-linguist
81.11%  240415     Rust
15.03%  44537      Shell
2.89%   8572       Makefile
0.86%   2557       Dockerfile
0.11%   317        Ruby
```
```

### CODE REVIEW

File: .gitattributes [hunk 1]
```diff
@@ -9,3 +9,4 @@
 *.gen.json linguist-generated=true
 *_pb2.py linguist-generated=true
 go.sum merge=union
+vendor/**  linguist-vendored
```
PiotrSikora, line 12: I don't particularly care, but this PR could be simplified to adding `boringssl-fips/** linguist-vendored` instead.

howardjohn: That was my first idea but the .gitattributes is vendored from https://github.com/istio/common-files itself, so while it isn't impossible to have ztunnel-specific things in it, its a bit nicer to have it generic

stevenctl (approved):

Issue #270, opened 2022-12-12
`console` feature is broken


PiotrSikora:
```sh
$ RUSTFLAGS="--cfg tokio_unstable" cargo build --features console
```
```
   Compiling ztunnel v0.0.0 (.../Aviatrix/ztunnel)
error[E0277]: the trait bound `impl tracing_subscriber::Layer<tracing_subscriber::Registry>: tracing_subscriber::Layer<Layered<impl tracing_subscriber::Layer<tracing_subscriber::Registry>, tracing_subscriber::Registry>>` is not satisfied
    --> src/telemetry.rs:28:15
     |
28   |         .with(fmt_layer())
     |          ---- ^^^^^^^^^^^ the trait `tracing_subscriber::Layer<Layered<impl tracing_subscriber::Layer<tracing_subscriber::Registry>, tracing_subscriber::Registry>>` is not implemented for `impl tracing_subscriber::Layer<tracing_subscriber::Registry>`
     |          |
     |          required by a bound introduced by this call
     |
     = help: the following other types implement trait `tracing_subscriber::Layer<S>`:
               Box<(dyn tracing_subscriber::Layer<S> + Sync + std::marker::Send + 'static)>
               Box<L>
               ConsoleLayer
               DynFilterFn<S, F, R>
               EnvFilter
               FilterFn<F>
               Filtered<L, F, S>
               Layered<A, B, S>
             and 7 others
note: required by a bound in `tracing_subscriber::prelude::__tracing_subscriber_SubscriberExt::with`
    --> .../.cargo/registry/src/github.com-1ecc6299db9ec823/tracing-subscriber-0.3.16/src/layer/mod.rs:1485:12
     |
1485 |         L: Layer<Self>,
     |            ^^^^^^^^^^^ required by this bound in `tracing_subscriber::prelude::__tracing_subscriber_SubscriberExt::with`

error[E0599]: the method `init` exists for struct `Layered<impl tracing_subscriber::Layer<tracing_subscriber::Registry>, Layered<impl tracing_subscriber::Layer<tracing_subscriber::Registry>, tracing_subscriber::Registry>>`, but its trait bounds were not satisfied
  --> src/telemetry.rs:29:10
   |
29 |         .init();
[...]

PR #275, opened 2022-12-13
Add check for features building

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/270

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -6,6 +6,10 @@ test:
 build:
 	cargo build
 
+# Test that all important features build
+build-features:
+	cargo build --features console
+
 # target in common/Makefile.common.mk doesn't handle our third party vendored files; only check golang and rust codes
 lint-copyright:
 	@${FINDFILES} \( -name '*.go' -o -name '*.rs' \) \( ! \( -name '*.gen.go' -o -name '*.pb.go' -o -name '*_pb2.py' \) \) -print0 |\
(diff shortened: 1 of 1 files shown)


Makefile.core.mk
```diff
@@ -6,6 +6,10 @@ test:
 build:
 	cargo build
 
+# Test that all important features build
+build-features:
+	cargo build --features console
```
PiotrSikora, line 11:
```suggestion
	RUSTFLAGS="${RUSTFLAGS} --cfg tokio_unstable" cargo build --features console
```

PiotrSikora (commented):
This might build without `--cfg tokio_unstable`, but the binary aborts at startup, so it's not exactly a working build.

Makefile.core.mk
howardjohn: I was trying to avoid this since it needs to rebuild everything again I think. Since we don't actually run it does it matter or do you think there are cases where the RUSTFLAGS may impact whether it builds?

Unfortunately we don't have caching of builds, so the build is fairly slow

PiotrSikora: Yes, it would rebuild everything, which is unfortunate.

However, without `--cfg tokio_unstable`, the code in `tokio`, `tokio-tracing` and `tokio-console` that's guarded by `#[cfg(tokio_unstable)]` (and there is a lot of it) won't be compiled, so a check without it might result in some false positives.

I guess we could leave it as-is for now, add revisit if there is a false positive in the future.

PiotrSikora (approved):

stevenctl (approved):

PR #280, opened 2022-12-15
(feat) Add `cargo deny` to check for CVEs & allowed licenses

hypernovasunnix:
Creates two new targets (`cve-check` & `license-check`) that utilizes `cargo-deny`.
`cve-check` will check for any CVEs in all deps (currently there in one)
`license-check` uses the `deny.toml` to scan for allowed licenses

### DIFF
--- a/Makefile.core.mk
+++ b/Makefile.core.mk
@@ -21,6 +21,12 @@ lint: lint-scripts lint-yaml lint-markdown lint-licenses lint-copyright
 check:
 	cargo check
 
+cve-check:
+	cargo deny check advisories
+
+license-check:
+	cargo deny check licenses
+
 fix: fix-copyright-banner
 	cargo clippy --fix --allow-staged --allow-dirty
 	cargo fmt
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,27 @@
+[licenses]
+allow = ["Apache-2.0",
+"CC-BY-3.0",
+"ISC",
+"AFL-2.1",
+"AFL-3.0",
+"Artistic-1.0",
+"Artistic-2.0",
+"Apache-1.1",
+"BSD-1-Clause",
+"BSD-2-Clause",
+"BSD-3-Clause",
+"0BSD",
+"FTL",
+"LPL-1.02",
+"MS-PL",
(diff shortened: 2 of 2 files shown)


Makefile.core.mk
```diff
@@ -37,3 +37,4 @@ gen-check: gen check-clean-repo
 
 presubmit: export RUSTFLAGS = -D warnings
 presubmit: build test lint gen-check build-features
+presubmit: cargo install cargo-audit && cargo audit
```
howardjohn, line 40: One minor thing - I think we will want to do the CVE checking only in postsubmit. Right now the postsubmit job (confusingly) also runs `make presubmit` but we could make a new target that runs the normal presubmit + CVE check and move the job to run that. Or even a periodic job for the CVE scanning

The reason is we generally try to keep the presubmit hermetic, so external changes (new CVE published) don't suddenly break all PRs

deny.toml
```diff
@@ -0,0 +1,9 @@
+[licenses]
+allow = ["Apache-2.0", "CC-BY-3.0", "ISC", "AFL-2.1", "AFL-3.0", "Artistic-1.0",
+"Artistic-2.0", "Apache-1.1", "BSD-1-Clause", "BSD-2-Clause", "BSD-3-Clause",
+"0BSD", "FTL", "LPL-1.02", "MS-PL", "MIT", "NCSA", "OpenSSL", "PHP-3.0", "TCP-wrappers", 
+"W3C", "Xnet", "Zlib", "CC0-1.0", "APSL-2.0", "CDDL-1.0", "CDDL-1.1", "CPL-1.0", 
+"EPL-1.0", "IPL-1.0", "MPL-1.0", "MPL-1.1", "MPL-2.0", "MPL-2.0-no-copyleft-exception",
```
howardjohn, line 6: Sorry wasn't clear: the `reciprocal_licenses` are also banned. Not sure why we have them split out from `restricted_licenses`...

The being said, none of them are used so we are in the clear.

Also, nitpick:  it might be nice to have these 1-per-line

howardjohn (approved):

howardjohn (approved):

Issue #287, opened 2022-12-16
ztunnel does not report ready


dhawton:
Beginning in commit https://github.com/istio/ztunnel/commit/bd72b58e86fd4f2434c4e1376d34f95dd0c6b0f0 the /healthz/ready endpoint never reports a 200. It stays on 500 and indicates it's waiting for the `proxy listeners` task to complete.

```
< HTTP/1.1 500 Internal Server Error
< content-type: text/plain
< content-length: 36
< date: Fri, 16 Dec 2022 15:46:00 GMT
< 
not ready, pending: proxy listeners
```

The logs also back this up:

```2022-12-16T17:07:45.483499Z  INFO ztunnel::admin: Readiness blocker complete, still awaiting 2 tasks task="workload manager" dur=968.62µs
2022-12-16T17:07:45.483522Z  INFO ztunnel::proxy::inbound_passthrough: listener established address=[::]:15006 component="inbound plaintext" transparent=true
2022-12-16T17:07:45.483530Z  INFO ztunnel::proxy::inbound: listener established address=[::]:15008 component="inbound" transparent=true
2022-12-16T17:07:45.483536Z  INFO ztunnel::proxy::outbound: listener established address=[::]:15001 component="outbound" transparent=true
2022-12-16T17:07:45.483542Z  INFO ztunnel::proxy::socks5: listener established address=[::]:15080 component="socks5"
2022-12-16T17:07:45.483549Z  INFO ztunnel::admin: listener established address=[::]:15021 component="admin"
2022-12-16T17:07:45.485905Z  INFO ztunnel::xds::client: sending initial request resources=0 type_url="type.googleapis.com/istio.workload.Workload"
2022-12-16T17:07:45.487912Z  INFO xds{id=1}: ztunnel::xds::client: received response type_url="type.googleapis.com/istio.workload.Workload" size=1
2022-12-16T17:07:45.487938Z  INFO ztunnel::admin: Readiness blocker complete, still awaiting 1 tasks task="ads client" dur=5.409414ms
2022-12-16T17:07:51.822691Z  INFO xds{id=1}: ztunnel::xds::client: received response type_url="type.googleapis.com/istio.workload.Workload" size=1
2022-12-16T17:07:52.770151Z  INFO xds{id=1}: ztunnel::xds::client: received response type_url="type.googleapis.com/istio.workload.Workload" size=2
```

[...]

PR #289, opened 2022-12-16
readiness: fix never getting ready

howardjohn:
This was broken *intentionally* as part of local debugging.. but accidentally committed due to lack of proper test coverage. We tested readiness, but didn't actually fail if it didn't become ready.

This fixes the mistake and ensures tests cover it.

closes https://github.com/istio/ztunnel/issues/287


dhawton:
Closes #287

stevenctl (approved):

Issue #15, opened 2022-10-19
Implement connection pooling


howardjohn:

stevenctl:
HBONE Connections between zTunnels from the same source and destination Pods/IPs should be reusable across multiple underlying client connections (streams inside HBONE).

stevenctl:
HBONE Connections between zTunnels from the same source and destination Pods/IPs should be reusable across multiple underlying client connections (streams inside HBONE).

linsun:
Speaking with Eitan, he is interested in this.

howardjohn:
@EItanya is working on this I think, but cannot assign since they are not in the org yet

EItanya:
I am indeed!

SkyfireFrancisZ:
@@EItanya, please update the estimated Size for this workitem.

EItanya:
I am off this week, but I will try and have one as soon as I'm back

SkyfireFrancisZ:
Dev environment setup, start digging into actual connection pooling today.

EItanya:
After quite a bit of digging I have found a number of paths forward. First some background info.

The http library that `ztunnel` uses is called `hyper`. Hyper has it's own connection pooling mechanism build into it's http client: https://github.com/hyperium/hyper/blob/0.14.x/src/client/pool.rs

However, this implementation doesn't work in our case for a number of reasons:
1. We cannot use the client impl they have due to the issue @howardjohn found https://github.com/hyperium/hyper/issues/2863.
2. The pooling mechanism only uses protocol/authority, but we need the client id as well. https://github.com/hyperium/hyper/blob/d32beb3843179acdaf0e64f3e708b9dbcd0c29e4/src/client/pool.rs#LL58

The three paths forward are as follows:
1. Write an entirely new pooling mechanism from scratch. I do not recommend this as the code is quite complex, and `hyper` has presumably been fixing and enhancing this for quite a long time.
2. Copy/Paste their pooling implementation. At first I thought this would be the best path forward, but unfortunately a lot of the impl is private, so copying the pool impl would require copying their connection impl as well.
3. Work with upstream to fix the few aspects of their impl that don't work for us. This is my favorite option because the number of changes are low, and the maintainer seems very willing to work with us. See [this issue](https://github.com/hyperium/hyper/issues/3064) for reference. Just to test this hypothesis, I put the generic key impl up for review: https://github.com/hyperium/hyper-util/pull/12

[...]

PR #293, opened 2022-12-19
Implement connection pooling

howardjohn:
Blocked by some upstream PRs

Fixes https://github.com/istio/ztunnel/issues/15

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,8 +39,8 @@ drain = "0.1.1"
 futures = "0.3.12"
 gperftools = { version = "0.2.0", features = ["heap"], optional = true }
 hyper = { version = "1.0.0-rc.3", features = ["full"] }
-# Pending https://github.com/hyperium/hyper-util/pull/25
-hyper-util = { git = "https://github.com/howardjohn/hyper-util", branch = "client/h2-timer", features = ["full"] }
+# Pending https://github.com/hyperium/hyper-util/pull/25, https://github.com/hyperium/hyper-util/pull/24
+hyper-util = { git = "https://github.com/howardjohn/hyper-util", branch = "h2-timer-expose-exec", features = ["full"] }
 libc = "0.2.126"
 log = "0.4"
 once_cell = "1.16.0"
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -235,8 +235,9 @@ impl OutboundConnection {
                     .expect("hbone requires destination workload");
 
                 let pool_key = pool::Key {
-                    src: req.source.identity(),
-                    dst: dst_identity.clone(),
+                    src_id: req.source.identity(),
+                    dst_id: dst_identity.clone(),
+                    dst: req.gateway,
                 };
 
                 // Setup our connection future. This won't always run if we have an existing connection
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -18,6 +18,7 @@ use std::time::Instant;
 use boring::ssl::ConnectConfiguration;
 use bytes::Bytes;
 use drain::Watch;
+use http_body_util::Empty;
 use hyper::header::FORWARDED;
 use hyper::StatusCode;
 use tokio::net::{TcpListener, TcpStream};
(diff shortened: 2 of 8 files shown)


src/proxy/outbound.rs
```diff
@@ -136,6 +159,33 @@ pub struct OutboundConnection {
...
     pub id: TraceParent,
+    pub pool: Pool<PoolValue, PoolKey>,
+}
+
+#[derive(Debug, Clone)]
+pub struct PoolClient(http2::SendRequest<Body>);
+
+impl Poolable for PoolClient {
+    fn is_open(&self) -> bool {
+        // TODO
```
stevenctl, line 170: What's the implication of always giving true? Just that something always goes back into the pool?

howardjohn: IDK, that is the main thing left to do for this PR aside from waiting for upstream

src/proxy/outbound.rs
```diff
@@ -136,6 +159,33 @@ pub struct OutboundConnection {
...
+
+    fn can_share(&self) -> bool {
+        true // http2 always shares
+    }
+}
+
+#[derive(PartialEq, Eq, Hash, Clone, Debug)]
+pub struct PoolKey {
+    src: Identity,
+    dst: Identity,
```
hzxuzhonghu, line 188: With src & dst identity, it could not perfectly reflect the relationship of connections number and the workload numbers. 

It is very common to have multi pods within a node with same identity

hzxuzhonghu (commented):
It is better to have a separate module for connection pool

src/proxy/outbound.rs
howardjohn: The original intent was to explicitly pool those. What is the motivation to not pool them?

hzxuzhonghu: This is not a big problem as there is no connection pool size limit, it is max_idle_per_host.

KfreeZ: wondering how to map the pods with same identity to the same connection? In the HTTP2 stream?

howardjohn: Yes, they are multiplexed. Each user connection is a H2 stream

KfreeZ: Thank you, John, I think this pool would significantly improve the performance, especially for some benchmarks, looking forward to it.

src/proxy/pool.rs
```diff
@@ -0,0 +1,205 @@
...
+                    _ => return Err(Error::Dropped),
+                },
+            };
+
+        Ok(Connection(request_sender))
+    }
+}
+#[cfg(test)]
+mod test {
+    /*
```
howardjohn, line 132: just need to fix up this test after hyper 1.0 update then this is good to go

src/proxy/pool.rs
```diff
@@ -0,0 +1,222 @@
...
+        self.0 .0.send_request(req)
+    }
+}
+
+impl Pool {
+    pub async fn connect<F>(&self, key: Key, connect: F) -> Result<Connection, Error>
+    where
+        F: Future<Output = Result<http2::SendRequest<Empty<Bytes>>, Error>>,
+    {
+        let reuse_connection = self.pool.checkout(key.clone());
```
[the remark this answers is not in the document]
howardjohn: When it is dropped it does so automatically (RAII)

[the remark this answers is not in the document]
wangfakang: got it, thanks.

src/proxy/pool.rs
```diff
@@ -0,0 +1,222 @@
...
+
+        let connect_pool = async {
+            let ver = hyper_util::client::pool::Ver::Http2;
+            let Some(connecting) = self.pool.connecting(&key, ver) else {
+                // TODO: this is obscure, make it more explicit
+                return Err(Error::Dropped)
```
stevenctl, line 112: IIUC this is `None` if we're already in a "connecting" state elsewhere. Why do we drop? Do we assume that `reuse_connection` would win?

howardjohn: Made this more clear. As you noted this means we are already connecting somewhere else, so we cancel this connecting and use the in-flight one. We later match on the error.

`Dropped` is a bad name though

stevenctl (approved):

Cargo.toml
```diff
@@ -39,8 +39,8 @@ drain = "0.1.1"
 futures = "0.3.12"
 gperftools = { version = "0.2.0", features = ["heap"], optional = true }
 hyper = { version = "1.0.0-rc.3", features = ["full"] }
-# Pending https://github.com/hyperium/hyper-util/pull/25
-hyper-util = { git = "https://github.com/howardjohn/hyper-util", branch = "client/h2-timer", features = ["full"] }
+# Pending https://github.com/hyperium/hyper-util/pull/25, https://github.com/hyperium/hyper-util/pull/24
+hyper-util = { git = "https://github.com/howardjohn/hyper-util", branch = "h2-timer-expose-exec", features = ["full"] }
```
hzxuzhonghu, line 43: ?

PR #294, opened 2022-12-19
parse proxy config

stevenctl:
Allows using `ProxyConfig` from either meshconfig file (hardcoded location) or PROXY_CONFIG env var.

Instead of generating structs with tonic/prost I just defined one directly in Rust for these reasons:
* tonic can't generate a reasonable `default` impl; it also won't make everything optional
* serde requires a `default` impl or it won't allow missing fields
* serde can't  deserialize prost::Duration. I was able to workaround this

On the commit before "abandon prost" you can see all the extra config required just to generate the struct with needed annotations. I don't think we plan on fetching proxy config over xds in the near term, and if we do we should just use a separate struct and manually merge that. 

Alternatively, we can switch back to the yaml merge approach + special code for some fields It did work, and it's similar enough to https://github.com/istio/istio/blob/master/pkg/config/mesh/mesh.go#L151


src/config.rs
```diff
@@ -180,12 +210,142 @@ pub fn parse_config() -> Result<Config, Error> {
...
+                .clone();
+            merge_yaml(&mut mc, pc_env);
+            Ok(mc)
+        }
+        (Ok(mc), Err(_)) => Ok(mc
+            .get("defaultConfig")
+            .unwrap_or(&serde_yaml::Value::Null)
+            .clone()),
+        (Err(_), Ok(pc_env)) => Ok(pc_env),
+        (Err(mc), Err(pc_env)) => Err(anyhow!("neither parsed: {mc}; {pc_env}")),
```
howardjohn, line 269: This gives a confusing message: `WARN ztunnel::config: neither parsed: No such file or directory (os error 2); EOF while parsing a value`

There isn't context on what we are trying to do, and "EOF ..." is confusing - maybe we should not have a default for PROXY_CONFIG and just only process it if it is set?

src/config.rs
```diff
@@ -180,12 +210,142 @@ pub fn parse_config() -> Result<Config, Error> {
...
+        let mesh_config_path = "./src/test_helpers/mesh_config.yaml";
+        let pc = construct_proxy_config(mesh_config_path).unwrap();
+        let cfg = construct_config(pc).unwrap();
+        assert_eq!(cfg.readiness_addr.port(), 15888);
+        assert_eq!(cfg.admin_addr.port(), 15099);
+        // TODO remove prefix
+        assert_eq!(cfg.proxy_metadata["ISTIO_META_FOO"], "bar");
+
+        // env only
+        let pc_env = "{ \
```
howardjohn, line 325: nit: could use a raw string like `r#"{`

src/config.rs
```diff
@@ -180,12 +210,142 @@ pub fn parse_config() -> Result<Config, Error> {
...
+            \"discoveryAddress\": \"istiod-rev0.istio-system-2:15012\", \
+            \"proxyAdminPort\": 15999 \
+        }";
+        std::env::set_var("PROXY_CONFIG", pc_env);
```
howardjohn, line 329: do we need to unset it after?

src/xds/client.rs
```diff
@@ -603,3 +610,25 @@ pub enum AdsError {
     #[error("XDS payload without resource")]
     MissingResource(),
 }
+
+// struct StructWrapper(Struct);
```
howardjohn, line 614: can be cleaned up

src/config.rs
```diff
@@ -180,12 +210,142 @@ pub fn parse_config() -> Result<Config, Error> {
...
+            .map_err(anyhow::Error::new)
+            .and_then(|env_str| serde_yaml::from_str(&env_str).map_err(anyhow::Error::new));
+
+    let yaml = match (mesh_config, proxy_config_env) {
+        (Ok(mc), Ok(pc_env)) => {
+            let mut mc = mc
+                .get("defaultConfig")
+                .unwrap_or(&serde_yaml::Value::Null)
+                .clone();
+            merge_yaml(&mut mc, pc_env);
```
howardjohn, line 261: I am not sure about the usage of `Value` and then a generic `merge_yaml` on it.

ProxyConfig merging, as done in istiod, is not actually a generic merge at all, and has special logic for fields in ApplyProxyConfig. In particular, proxy_metadata behaves a bit differently, at least with `PROXY_CONFIG='{"proxyMetadata": null}'`.

Would it make sense to just implement our own MergeProxyConfig function? There are only a few fields so it wouldn't be too hard I think?

src/config.rs
```diff
@@ -104,6 +115,8 @@ pub struct Config {
 pub enum Error {
     #[error("invalid env var {0}={1}")]
     EnvVar(String, String),
+    #[error("error occurred: {0}")]
```
howardjohn, line 118: This gives a confusing error: `Error: error occurred: invalid type: integer `1`, expected a map`; without context its not clear what the issue is

src/config.rs
stevenctl: Will do

src/config.rs
stevenctl: I'll just change the fn to accept both inputs as params to make it more testable

src/config.rs
```diff
@@ -180,12 +212,154 @@ pub fn parse_config() -> Result<Config, Error> {
...
+    // only include ISTIO_META_ prefixed fields in this map
+    // TODO we could use any other items here for the various env vars for construct_config?
+    // https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#:~:text=Additional%20environment%20variables
+    pc.proxy_metadata = pc
+        .proxy_metadata
+        .into_iter()
+        .filter_map(|(k, v)| Some((k.strip_prefix("")?.to_string(), v)))
+        .collect();
+    
+    // TODO if certain fields like trustDomainAliases are added, make sure they merge like:
```
howardjohn, line 305: fwiw this one is in mesh config. But we may need this and trust_domain eventually actually. Or send it over XDS

howardjohn (approved):

src/config.rs
```diff
@@ -31,17 +34,21 @@ const CA_ADDRESS: &str = "CA_ADDRESS";
 const TERMINATION_GRACE_PERIOD: &str = "TERMINATION_GRACE_PERIOD";
 const FAKE_CA: &str = "FAKE_CA";
 const ZTUNNEL_WORKER_THREADS: &str = "ZTUNNEL_WORKER_THREADS";
+const PROXY_CONFIG: &str = "PROXY_CONFIG";
 
-const DEFAULT_WORKER_THREADS: usize = 2;
+const DEFAULT_WORKER_THREADS: u16 = 2;
+const DEFAULT_ADMIN_PORT: u16 = 15021;
+const DEFAULT_STATUS_PORT: u16 = 15020;
```
shashankram, line 41: Admin was 15000 and status/readiness was 15021 before. Was this incorrectly changed here or is there a corresponding change for the manifests in the Istio repo missing?
/cc @howardjohn @stevenctl

shashankram: Fix: https://github.com/istio/ztunnel/pull/318

PR #295, opened 2022-12-20
WorkloadStore makes use of RwLock

hzxuzhonghu:


howardjohn (commented):
Do the benchmarks show improvements from this change?

hzxuzhonghu:
No obvious improve, looks there is no race in the benchmark test

PR #296, opened 2022-12-20
add AMBIENT_TYPE to connection metadata

GregHanson:
add missing `AMBIENT_TYPE` value to connection metadata so it can be properly used in Istiod logic:
https://github.com/istio/istio/blob/experimental-ambient/pilot/pkg/model/context.go#L716

### DIFF
--- a/src/xds/client.rs
+++ b/src/xds/client.rs
@@ -304,6 +304,8 @@ impl AdsClient {
         let ns = ns.as_deref().unwrap_or("");
         let node = std::env::var("NODE_NAME");
         let node = node.as_deref().unwrap_or("");
+        let ambient_type = std::env::var("AMBIENT_TYPE");
+        let ambient_type = ambient_type.as_deref().unwrap_or("");
         Node {
             id: format!("sidecar~{ip}~{pod_name}.{ns}~{ns}.svc.cluster.local"),
             metadata: Some(Self::build_struct([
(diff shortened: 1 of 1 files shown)


src/xds/client.rs
```diff
@@ -304,13 +304,16 @@ impl AdsClient {
...
+        let ambient_type = std::env::var("AMBIENT_TYPE");
+        let ambient_type = ambient_type.as_deref().unwrap_or("");
         Node {
             id: format!("sidecar~{ip}~{pod_name}.{ns}~{ns}.svc.cluster.local"),
             metadata: Some(Self::build_struct([
                 ("POD_NAME", pod_name),
                 ("POD_NAMESPACE", ns),
                 ("INSTANCE_IP", ip),
                 ("NODE_NAME", node),
+                ("AMBIENT_TYPE", ambient_type),
```
howardjohn, line 316: maybe should just be a constant? ztunnel

howardjohn (approved):

src/xds/client.rs
GregHanson: or I can add a default value on the env var lookup similar to what is done for `INSTANCE_IP` above.

GregHanson: I had a PR in the ztunnel charts to include the env var, but I can also just add the default value: https://github.com/istio/istio/pull/42527

howardjohn: I don't think there is any use case to not have it hardcoded. I think we should do https://github.com/istio/istio/issues/42313 actually, eventually

stevenctl: Is there any chance in the future a "sandwich" ztunnel might need different config?

howardjohn: Maybe but we can probably define that as needed?

Issue #259, opened 2022-12-08
Implement the RBAC DS API in Istiod


ymesika:
This issue a shaddowed Istio issue for the implementation of the RBAC API in Istiod's XDS.
The API design is covered by #205 .

ymesika:
Still WIP.
Currently have an implementation with unit-test that on `AuthorizationPolicy` addition the RBAC DS generator sends a delta XDS with an RBAC resource.

Issue #205, opened 2022-11-28
Design for (L4) RBAC


ymesika:
Before implementing #29 we should come up with an agreeable design for this feature.

SkyfireFrancisZ:
design doc started, will make public this week.

ymesika:
Updated the [design document](https://docs.google.com/document/d/17mRVzXe8PS7VoligvIx52T10tOP7xPQ9noeOzoLO2cY) based on the discussion we had in the ambient weekly meeting.
It's ready for re-review.

ymesika:
We had another discussion in the Dec 14 ambient weekly meeting.
RBAC proto in the design doc is still not the best fit and we iterate over its desired model.

ymesika:
I will be away until next year so allowing others to jump in if they wish to.
Anyway, will be glad to re-assign when I'm back if it's still relevant.

howardjohn:
@ymesika if you had done any coding work beyond the doc it would be helpful to push a WIP. No problem if not

Issue #29, opened 2022-10-20
Implement (L4) RBAC


howardjohn:
This was an initial attempt in Go:

```go
unc (p *Proxy) AssertRBAC(r *http.Request) error {
	ip, dport, err := net.SplitHostPort(r.Host)
	if err != nil {
		return err
	}
	pip, err := netip.ParseAddr(ip)
	if err != nil {
		return err
	}
	wl := p.ConnectionManager.FindWorkloadByAddr(pip)
	var identity string
	if r.TLS != nil && len(r.TLS.PeerCertificates) > 0 {
		n, err := util.ExtractIDs(r.TLS.PeerCertificates[0].Extensions)
		if err != nil {
			return err
		}
		if len(n) > 0 {
			identity = n[0]
		}
	}

	if wl.RBAC == nil {
		return nil
	}
	var namespace string
	if identity != "" {
		s, err := spiffe.ParseIdentity(identity)
		if err != nil {
			return err
		}
		namespace = s.Namespace
	}

	log := log.WithLabels("ip", pip.String(), "ident", identity, "port", dport)

	// This is from the PEP, which handles this already
	// TODO: make this check more robust
	if wl.RemoteProxy != (netip.Addr{}) && identity == wl.Identity {
		if len(wl.RBAC.Allow) == 0 {
			log.Debugf("allow (no policies)")
		} else {
			log.Infof("allow (from remote)")
		}
		return nil
	}
	deny := rbacMatch(wl.RBAC.Deny, namespace, identity, dport)
	if deny {
		// TODO context
		return fmt.Errorf("RBAC: deny")
	}
	if len(wl.RBAC.Allow) == 0 {
		log.Debug("allow (no policies)")
		return nil
	}
	allow := rbacMatch(wl.RBAC.Allow, namespace, identity, dport)
	if allow {
		log.Info("allow")
		return nil
	}
	return fmt.Errorf("RBAC: no allow matched")
}

func rbacMatch(pol []*uproxyapi.Policy, namespace string, identity string, dports string) bool {
	dport, _ := strconv.Atoi(dports)
	for _, pol := range pol {
		// TODO pol.When
		ruleMatch := false
		for _, rule := range pol.Rule {
			rmatch := true
			if rule.Namespace != "" && rule.Namespace != namespace {
				rmatch = false
			}
			if rule.Identity != "" && "spiffe://"+rule.Identity != identity {
				rmatch = false
			}
			if rule.Invert {
				rmatch = !rmatch
			}
[...]

PR #297, opened 2022-12-20
implement L4 AuthorizationPolicy

howardjohn:
This commit introduces the first iteration of L4 AuthorizationPolicy
support.

* Policies are read over XDS and converted to a new internal type. These
  are indexed by namespace
* Inbound requests check against applicable policies and send 401 in the
  HBONE response when they fail to match
* Current matching algorithm is fairly efficient (TODO: benchmarks), but
  not optimized (one could imagine a trie like implementation; today it
is a linear search).

Fixes https://github.com/istio/ztunnel/issues/259
Fixes https://github.com/istio/ztunnel/issues/205
Fixes https://github.com/istio/ztunnel/issues/29

### DIFF
--- a/proto/networkpolicy.proto
+++ b/proto/networkpolicy.proto
@@ -1,0 +1,107 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package istio.workload;
+option go_package="pkg/workloadapi";
+
+import "google/protobuf/empty.proto";
+
+message RBAC {
+  string name = 1;
+  string namespace = 2;
+
+  // Determine the scope of this RBAC policy.
+  // If set to NAMESPACE, the 'namespace' field value will be used.
+  RBACScope scope = 3;
+  // The action to take if the request is matched with the rules.
+  // Default is ALLOW if not specified.
+  RBACPolicyAction action = 4;
+  // Set of RBAC policy groups each containing its rules.
+  // If at least one of the groups is matched the policy action will
+  // take place.
+  // Groups are OR-ed.
+  repeated RBACPolicyRulesGroup groups = 5;
+}
+
+message RBACPolicyRulesGroup {
+  // Rules are AND-ed
+  // This is a generic form of the authz policy's to, from and when
+  repeated RBACPolicyRules rules = 1;
+}
+
+message RBACPolicyRules {
+  // The logical behavior between the matches (if there are more than one)
+  //  MatchBehavior match_behavior = 1;
+  repeated RBACPolicyRuleMatch matches = 2;
+}
+
+message RBACPolicyRuleMatch {
+  // Values of specific type are ORed
+  // If multiple types are set, they are ANDed
+
+  repeated StringMatch namespaces = 1;
+  repeated StringMatch not_namespaces = 2;
+
+  repeated StringMatch principals = 3;
+  repeated StringMatch not_principals = 4;
+
+  repeated Address source_ips = 5;
+  repeated Address not_source_ips = 6;
+
+  repeated Address destination_ips = 7;
+  repeated Address not_destination_ips = 8;
+
+  repeated uint32 destination_ports = 9;
+  repeated uint32 not_destination_ports = 10;
+}
+
+message Address {
+  bytes address = 1;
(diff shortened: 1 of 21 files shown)


src/proxy/inbound.rs
```diff
@@ -73,8 +76,25 @@ impl Inbound {
...
+                src_identity: socket
+                    .ssl()
+                    .peer_certificate()
+                    .and_then(|x| crate::tls::boring::extract_sans(&x).first().cloned()),
+                src_ip: to_canonical(socket.get_ref().peer_addr().unwrap()).ip(),
+                dst,
+            };
+            let workloads = self.workloads.clone();
+            async move {
+                let conn = conn.clone();
```
stevenctl, line 91: redundant clones?

src/proxy/inbound.rs
```diff
@@ -187,6 +215,22 @@ impl Inbound {
...
+                if addr.ip() != conn.dst.ip() {
+                    info!("Sending 400, ip mismatch {addr} != {}", conn.dst);
+                    return Ok(Response::builder()
+                        .status(StatusCode::BAD_REQUEST)
+                        .body(Body::empty())
+                        .unwrap());
+                }
+                // Orig has 15008, swap with the real port
+                let conn = rbac::Connection { dst: addr, ..conn };
+                if !workloads.assert_rbac(&conn).await {
```
stevenctl, line 227: Just tested locally with the local redirect method and I'm getting:

```
2022-12-28T16:48:39.049473Z  INFO inbound{id=2f63be3786443ac0803b0dd43a307b6a peer_ip=127.0.0.1 peer_id=spiffe://cluster.local/ns/default/sa/default}: ztunnel::proxy::inbound: RBAC rejected conn=127.0.0.1(spiffe://cluster.local/ns/default/sa/default)->127.0.0.1:8000
```

stevenctl: `no allow policies matched`

howardjohn: do you have policies defined?

src/rbac.rs
```diff
@@ -0,0 +1,385 @@
...
+        } else {
+            let matches = positive.iter().any(&mut predicate);
+            trace!(%matches, "type"="positive", "{positive:?}");
+            matches
+        };
+        let nm = if negative.is_empty() {
+            trace!(matches = true, "type" = "negative", "no match declared");
+            true
+        } else {
+            let matches = negative.iter().any(&mut predicate);
```
stevenctl, line 169: Negate predicate this or change `any` to `none`

src/proxy/inbound.rs
stevenctl: There is an `allow: not 9999` in local.yaml but I don't think it's matching correctly. See the other comment.

src/rbac.rs
stevenctl: or I guess just make it check `pm && !nm` is cleaner

howardjohn: good catch - will add a bunch of tests before we merge this

stevenctl (approved):

proto/networkpolicy.proto
```diff
@@ -0,0 +1,107 @@
...
+  //  MatchBehavior match_behavior = 1;
+  repeated RBACPolicyRuleMatch matches = 2;
+}
+
+message RBACPolicyRuleMatch {
+  // Values of specific type are ORed
+  // If multiple types are set, they are ANDed
+
+  repeated StringMatch namespaces = 1;
+  repeated StringMatch not_namespaces = 2;
```
hzxuzhonghu, line 56: No need using stringMatch for namespaces, makes things a little more complex

proto/networkpolicy.proto
```diff
@@ -0,0 +1,107 @@
...
+    string suffix = 3;
+
+    google.protobuf.Empty presence = 4;
+  }
+}
+
+enum RBACScope {
+  // ALL means that the authorization policy will be applied to all workloads
+  // in the mesh (any namespace).
+  GLOBAL = 0;
```
hzxuzhonghu, line 93: Whyt not use Namespace="" to represent all namespaces

proto/networkpolicy.proto
howardjohn: It's part of AurhorizationPolicy unfortunately. I agree but we can't really change it at this point

proto/networkpolicy.proto
```diff
@@ -0,0 +1,107 @@
...
+  // NAMESPACE means that the policy will only be applied to workloads in a
+  // specific namespace.
+  NAMESPACE = 1;
+  // WORKLOAD_SELECTOR means that the policy will only be applied to specific
+  // workloads that were selected by their labels.
+  WORKLOAD_SELECTOR = 2;
```
hzxuzhonghu, line 99: Cross namespaces selector seems very weird

proto/networkpolicy.proto
howardjohn: It's still useful to know which namespace it came from even if it's a global rule, IMO

proto/networkpolicy.proto
howardjohn: It's part of the authz v1 API unfortunately

PR #298, opened 2022-12-21
prefetch certs on workload discovery

adiprerepa:


src/app.rs
```diff
@@ -48,9 +49,9 @@ pub async fn build_with_cert(
         config.clone(),
         metrics.clone(),
         ready.register_task("workload manager"),
+        cert_manager.borrow(),
```
howardjohn, line 52:
```suggestion
        &cert_manager,
```

src/workload.rs
```diff
@@ -213,7 +217,16 @@ impl xds::Handler<XdsWorkload> for Arc<Mutex<WorkloadStore>> {
         let mut wli = self.lock().unwrap();
         let handle = |res: XdsUpdate<XdsWorkload>| {
             match res {
-                XdsUpdate::Update(w) => wli.insert_xds_workload(w.resource)?,
+                XdsUpdate::Update(w) => {
+                    // prefetch CA cert on discovery
+                    let identity = &Identity::Spiffe {
```
howardjohn, line 222: my be better to do it in insert_xds_workload, then we have already converted XdsWorkload to Workload can can do `workload.identity()`. Notably, the logic here differs from `identity()`

src/workload.rs
```diff
@@ -387,13 +402,15 @@ impl WorkloadInformation {
...
-pub struct WorkloadStore {
+pub struct WorkloadStore<'a> {
     workloads: HashMap<IpAddr, Workload>,
     // workload_to_vip maintains a mapping of workload IP to VIP. This is used only to handle removals.
     workload_to_vip: HashMap<IpAddr, HashSet<(SocketAddr, u16)>>,
     // vips maintains a mapping of socket address with service port to workload ip and socket address
     // with target ports in hashset.
     vips: HashMap<SocketAddr, HashSet<(IpAddr, u16)>>,
+    // for fetching certificates on workload discovery
+    prefetch: &'a dyn Fn(&Identity) -> dyn Future<Output=Result<Certs, Error>>
```
howardjohn, line 413: might consider wrapping this in an Arc so we don't need all the lifetimes.

howardjohn: And pass a `Box<dyn CertificateProvider>` directly instead of a closure will simplify. Note we already do this in other places

src/proxy.rs
```diff
@@ -50,7 +51,7 @@ impl Proxy {
     pub async fn new(
         cfg: config::Config,
         workloads: WorkloadInformation,
-        cert_manager: Box<dyn CertificateProvider>,
+        cert_manager: Arc<Mutex<Box<dyn CertificateProvider>>>,
```
howardjohn, line 54: I don't think we need Box if its already in an arc mutex

howardjohn: But I am not sure why we need the mutex anyways, the CertificateProvider already has one as part of it.

Maybe https://github.com/istio/ztunnel/pull/325 resolves this?

src/workload.rs
```diff
@@ -487,6 +497,18 @@ impl WorkloadInformation {
     }
 }
 
+impl Default for Box<dyn CertificateProvider> {
```
howardjohn, line 500: We shouldn't have fake test things in prod code, makes it possible to accidentally use this

src/workload.rs
```diff
@@ -487,6 +497,18 @@ impl WorkloadInformation {
...
+    fn default() -> Self {
+        Box::new(MockCaClient::new(Duration::from_secs(5)))
+    }
+}
+
+impl std::fmt::Debug for Box<dyn CertificateProvider> {
```
howardjohn, line 506: This seems suspicious? why are we emitting just `""`?

src/workload.rs
```diff
@@ -531,6 +563,22 @@ impl WorkloadStore {
                     .insert((service_sock_addr, port.target_port as u16));
             }
         }
+        // prefetch certificate - get the handle of the current runtime so we can
+        // block on fetch_certificate
+        let handle = Handle::current();
+        let _ = handle.enter();
+        // we just need to call fetch_certificate to cache, don't care about the actual cert
+        match block_on(
```
howardjohn, line 571: I don't think we want to block XDS for certs. We just want to pre-warm things asyncronously. I would imagine this would mean we push the request onto a channel or queue and do not wait for completion.

This will also allow the secret manager to rate limit requests. On demand (request blocking) requests can go ASAP or high priority,and these background ones can slowly trickle in when there is no work.

src/proxy.rs
adiprerepa: would just an `Arc<dyn CertificateProvider>` suffice then?

howardjohn: You may just want to call .clone() on it instead? The CertificateProviders internally use Arc already so they are cheap to clone

adiprerepa: makes sense

src/workload.rs
adiprerepa: yeah, changed to watch channel

src/workload.rs
adiprerepa: removed, not sure why i put that there

src/workload.rs
adiprerepa: done

src/workload.rs
```diff
@@ -257,8 +259,26 @@ impl WorkloadManager {
         config: config::Config,
         metrics: Arc<Metrics>,
         awaiting_ready: readiness::BlockReady,
+        cert_manager: Box<dyn CertificateProvider>,
     ) -> anyhow::Result<WorkloadManager> {
-        let workloads: Arc<Mutex<WorkloadStore>> = Arc::new(Mutex::new(WorkloadStore::default()));
+        let (tx, mut rx) = watch::channel(Identity::default());
```
howardjohn, line 264: `watch` is wrong here. This is for storing a single latest value. The current code will actually only trigger once with the last identity requested, instead of on each request. a mpsc is probably more appropriate (basically the same as Go channel)

src/workload.rs
```diff
@@ -257,8 +259,26 @@ impl WorkloadManager {
...
+                    Err(e) => error!(
+                        "unable to prefetch cert for {:?}, skipping, {:?}",
+                        workload_identity.to_string(),
+                        e
+                    ),
+                }
+            }
+        });
+        let workloads: Arc<Mutex<WorkloadStore>> = Arc::new(Mutex::new(WorkloadStore {
+            cert_tx: Some(tx),
```
howardjohn, line 279: Do we need to make it an Option? Or can we just store directly?

src/workload.rs
```diff
@@ -257,8 +259,26 @@ impl WorkloadManager {
         config: config::Config,
         metrics: Arc<Metrics>,
         awaiting_ready: readiness::BlockReady,
+        cert_manager: Box<dyn CertificateProvider>,
     ) -> anyhow::Result<WorkloadManager> {
-        let workloads: Arc<Mutex<WorkloadStore>> = Arc::new(Mutex::new(WorkloadStore::default()));
+        let (tx, mut rx) = watch::channel(Identity::default());
+        tokio::spawn(async move {
+            while rx.changed().await.is_ok() {
```
howardjohn, line 266: We should add a TODO comment to rate limit these. OK for now though. I am looking into what the limit should be -- I think we definitely do not want ~255 requests at once, but not sure what is reasonable.

src/workload.rs
adiprerepa: I made it an option so I didn't have to deal with the default macro, lmk if theres a workaround

src/workload.rs
adiprerepa: ah, fixed.

src/workload.rs
```diff
@@ -531,6 +555,11 @@ impl WorkloadStore {
                     .insert((service_sock_addr, port.target_port as u16));
             }
         }
+        if let Some(tx) = self.cert_tx.as_mut() {
+            if tx.blocking_send(widentity).is_err() {
```
howardjohn, line 559: blocking_send is only when not in async runtime. This panics. We should use try_send instead and ignore the error. This will fail if we have 100 identities queued up - no problem, this is prefetch anyways, not required just an optimization. Without try_send once we hit 100 identities we would block xds on CA - not good

src/workload.rs
```diff
@@ -257,8 +259,26 @@ impl WorkloadManager {
         config: config::Config,
         metrics: Arc<Metrics>,
         awaiting_ready: readiness::BlockReady,
+        cert_manager: Box<dyn CertificateProvider>,
     ) -> anyhow::Result<WorkloadManager> {
-        let workloads: Arc<Mutex<WorkloadStore>> = Arc::new(Mutex::new(WorkloadStore::default()));
+        let (tx, mut rx) = mpsc::channel::<Identity>(100);
```
howardjohn, line 264: lets make it 256, this is the max pods per node on typical k8s cluster.

src/workload.rs
howardjohn: Alternative would be to drop Default, make a `WorkloadStore::new() -> (WorkloadStore, rx)`. In new manually set the ~5 hashmaps. But this is ok too

src/workload.rs
adiprerepa: good idea

src/workload.rs
adiprerepa: done

src/workload.rs
```diff
@@ -531,6 +555,11 @@ impl WorkloadStore {
                     .insert((service_sock_addr, port.target_port as u16));
             }
         }
+        if let Some(tx) = self.cert_tx.as_mut() {
+            if let Err(e) = tx.try_send(widentity) {
```
howardjohn, line 559: future improvement would be to store a list of identities we already warmed and do not re-warm them. Re-warm is fairly cheap (we have a cache), but not as cheap as a set lookup.

src/workload.rs
```diff
@@ -516,6 +539,7 @@ impl WorkloadStore {
     fn insert_xds_workload(&mut self, w: XdsWorkload) -> anyhow::Result<()> {
         let workload = Workload::try_from(&w)?;
         let wip = workload.workload_ip;
+        let widentity = workload.identity();
```
howardjohn, line 542: Forgot something import - we can only do this when workload node == current node (cfg.local_node). Otherwise it will always fail

howardjohn: can you followup with a fix?

howardjohn (approved):

src/workload.rs
adiprerepa: will do

Issue #183, opened 2022-11-23
Add fuzz testing


howardjohn:
Istio and Envoy both integrate with OSS fuzz. Figure out how we can, as well. This is likely a bit different since we use rust, but I think there are other rust projects fuzzing with oss fuzz. Once we have that hooked up, figure out which areas we can/want to fuzz and implement those tests.

howardjohn:
I am starting work on this

howardjohn:
https://github.com/google/oss-fuzz/pull/9343

howardjohn:
Initial fuzzing is done, changing this to expanding fuzz testing

PR #302, opened 2023-01-03
fuzz: introduce initial fuzz targets

howardjohn:
These take raw bytes, and convert to Proto types then internal types.
For https://github.com/istio/ztunnel/issues/183


fuzz/.gitignore
```diff
@@ -0,0 +1,4 @@
+target
```
hzxuzhonghu, line 1: shouldn't this go to the root gitignore

hzxuzhonghu (commented):
LGTM except the gitignore

fuzz/.gitignore
howardjohn: it's for fuzz/target. This file is generated by `cargo fuzz`

hzxuzhonghu: good to know

hzxuzhonghu (approved):

Issue #73, opened 2022-11-09
Implement Hairpinning


stevenctl:
Traffic sent from outside ambient to a Pod in ambient should use the Waypoint if available.

There are some cases guarded with Skip in existing tests, but that might not be comprehensive.

Example: 
Traffic arrives on the inbound-plaintext port of a uProxy, but the destination is part of an ambient mesh; we should either deny the traffic, or initiate a tunnel. 

For this milestone, we likely just want to deny it. Well-behaved traffic should use proper ingress.

SkyfireFrancisZ:
@linsun @howardjohn do you think this is required for OSS Alpha?

howardjohn:
Yes, we cannot securely implement waypoint proxies without it.

linsun:
Yep what John said.

howardjohn:
Need to make sure we also skip the node fastpath when there is a waypoint.

howardjohn:
Current plan is to implement this for plaintext only and deny HBONE requests instead of hairpinning, with the assumption any HBONE clients are waypoint-aware. We can loosen this restriction, if needed, in the future.

SkyfireFrancisZ:
PR out for review

SkyfireFrancisZ:
waiting on Steven's feedback on the PR

SkyfireFrancisZ:
code merged. waiting on testing PR

howardjohn:
Just need https://github.com/istio/ztunnel/pull/348

PR #304, opened 2023-01-03
Implement inbound passthrough hairpin

howardjohn:
For https://github.com/istio/ztunnel/issues/73

This is a pretty large change:
* Outbound is refactored to take in a ProxyInputs instead of ~5 fields. This is useful since we use this now from all the proxy types (outbound, socks5, inbound), since any of those can result in outbound requests now.
* Add a few debugging logs that were useful when testing this.
* Add new testing for "waypoints"
* The core logic: new handling for inbound and inbound passthrough
  * Both must be sending to known destinations, no arbitrary passtrhough
  * Inbound passthrough CAN be sent directly... but if the dest has a waypoint, we hairpin it. This is to support non-mesh clients
  * Inbound HBONE must send to the waypoint first, if there is one. We could hairpin this in the future, if we want, but for now we are stricter.

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -27,7 +27,7 @@ use tracing::{debug, error, info, instrument, trace, trace_span, warn, Instrumen
 use crate::config::Config;
 use crate::identity::CertificateProvider;
 use crate::proxy::inbound::InboundConnect::Hbone;
-use crate::proxy::{TraceParent, TRACEPARENT_HEADER};
+use crate::proxy::{ProxyInputs, TraceParent, TRACEPARENT_HEADER};
 use crate::rbac;
 use crate::socket::{relay, to_canonical};
 use crate::tls::TlsError;
@@ -35,7 +35,7 @@ use crate::workload::WorkloadInformation;
 
 use super::Error;
 
-pub struct Inbound {
+pub(super) struct Inbound {
     cfg: Config,
     listener: TcpListener,
     cert_manager: Box<dyn CertificateProvider>,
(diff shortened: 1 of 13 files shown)


src/proxy/inbound_passthrough.rs
```diff
@@ -69,13 +76,43 @@ impl InboundPassthrough {
...
+        mut inbound: TcpStream,
     ) -> Result<(), Error> {
-        let orig = socket::orig_dst_addr_or_default(inbound);
+        let orig = socket::orig_dst_addr_or_default(&inbound);
+        let Some(upstream) = pi.workloads.fetch_workload(&orig.ip()).await else {
+            return Err(Error::UnknownDestination(orig.ip()))
+        };
+        if !upstream.waypoint_addresses.is_empty() {
+            // This is an inbound request not over HBONE, but we have a waypoint.
+            // The request needs to go through the waypoint for policy enforcement.
```
hzxuzhonghu, line 89: Will L4 traffic goes to waypoint here？

hzxuzhonghu: Then could i understand that a single Authz policy can be enforced by both waypoint and ztunnel?

howardjohn: Yes, but not on a single connection. A given connection will be enforced at waypoint or ztunnel, not both.

This is important because a policy like "allow GET from spiffe://foo" will always deny at ztunnel since we cannot parse L7

hzxuzhonghu: >Yes, I'd they have one. Note waypoint is never intended to only handle L7 policies. There are plenty of configs that are L4 that only waypoint can handle.

I am concerned for traffic to some L4 ports, it is redirected to waypoint. Why in this case ztunnel can not handle it? Imaging if there are no waypoint like below processing, so I mean could separate the traffic with protocol, not only by ip

hzxuzhonghu: Understood that.

But as i know, WASM, Ext Auth are mostly L7. We can get the protocol from port name before. If we must support protocol sniffer, this comment can be ignored.

howardjohn:
More testing is in https://github.com/istio/ztunnel/pull/319

src/proxy/inbound.rs
```diff
@@ -229,7 +224,29 @@ impl Inbound {
...
+
+                let Some(upstream) = workloads.fetch_workload(&addr.ip()).await else {
+                    info!(%conn, "unknown destination");
+                    return Ok(Response::builder()
+                        .status(StatusCode::NOT_FOUND)
+                        .body(Body::empty())
+                        .unwrap());
+                };
+                if !upstream.waypoint_addresses.is_empty() {
+                    let from_waypoint = conn
```
stevenctl, line 236: `from_waypoint` is slightly confusing.  Could also be a call from same SA.

stevenctl (commented):
LGTM

PR #306, opened 2023-01-04
POD_NAME should be changed to NAME

hzxuzhonghu:


src/xds/client.rs
```diff
@@ -309,7 +309,7 @@ impl AdsClient {
         Node {
             id: format!("sidecar~{ip}~{pod_name}.{ns}~{ns}.svc.cluster.local"),
             metadata: Some(Self::build_struct([
-                ("POD_NAME", pod_name),
+                ("NAME", pod_name),
```
hzxuzhonghu, line 312: We should link to the right struct

howardjohn (approved):

Issue #307, opened 2023-01-04
build failed without using docker


hzxuzhonghu:
cargo build
   Compiling boring-sys v2.1.0
   Compiling thiserror v1.0.37
   Compiling num-integer v0.1.45
   Compiling rayon v1.6.0
error[E0554]: `#![feature]` may not be used on the stable release channel
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/thiserror-1.0.37/src/lib.rs:238:34
    |
238 | #![cfg_attr(provide_any, feature(provide_any))]
    |                                  ^^^^^^^^^^^

For more information about this error, try `rustc --explain E0554`.
error: could not compile `thiserror` due to previous error
warning: build failed, waiting for other jobs to finish...

howardjohn:
This happens when building with nightly compiler then later without in my experience. `cargo clean -p thiserror` resolves.. not sure why its cached

PR #308, opened 2023-01-04
bump thiserror to 1.0.38

hzxuzhonghu:
Fix https://github.com/istio/ztunnel/issues/307


howardjohn (approved):
LGTM but I am not sure it fixes it? Or I don't understand why at least (see comment on issue)

hzxuzhonghu:
The build complains thiserror

howardjohn:
But updating is just clearing the build cache I suspect

howardjohn:
The real issue seems to be a cache invalidation issue in cargo

PR #315, opened 2023-01-05
refactor modules to remove mod.rs files

ilrudie:
refactors modules to remove mod.rs files in favor of more descriptive file names


howardjohn (approved):

hzxuzhonghu:
According to http://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/second-edition/ch07-01-mod-and-the-filesystem.html#rules-of-module-filesystems, this maybe not the recommended practice

ilrudie:
I don't see a similar prescription for _how_ to organize code in the recent rust documentation. When looking at the source for the `second-edition` of the book presently it just says:

> The second edition of the book is no longer distributed with Rust's documentation.

It goes on to direct the reader to a section in the newer book. The newer book now leans towards the descriptively named  implementation and it does mention that the `mod.rs` style is still valid.

I mention this to suggest that I believe today this is really a style decision. I do see the authors' point about having many mod.rs files open at the same time in a larger project so I was leaning a bit towards not using them. That being said if we have reason to use `mod.rs` or want to discuss options to mix and match let's chat about it. Either way a decision lands about how we want to structure our project if it would help I would be happy to write some markdown about why this project has chosen one style over the others.

ilrudie:
I'll poke around a little in the source for the book and see if the authors' had a relevant discussion when they removed the suggestion to use `mod.rs` from the book. If I find anything pertinent I'll link it back here as a reference.

hzxuzhonghu:
Thanks, mixed use is not encouraged.

ilrudie:
Yeah. Mixing is possible but it seems less than ideal. At least it would be a compile error straight away if a mixed use mistake was made but I agree it's probably best to simply avoid.

ilrudie:
[This PR](https://github.com/rust-lang/book/pull/1535) looks to be where they moved from using mod.rs as the primary style in the book to using the "descriptive names" style. I don't see discussion about why the implicit recommendation shifted though.

Issue #310, opened 2023-01-04
Add bytes sent and received metrics


ymesika:
https://github.com/istio/ztunnel/pull/244 added the infrastructure for metrics and the connection opened / closed with relevant labels.
We need to also add the bytes sent and received to match the TCP metrics we have in sidecars.

PR #317, opened 2023-01-09
Add bytes sent and received metrics

ymesika:
PR for #310 

Adding support for the `istio_tcp_received_bytes_total` and `istio_tcp_sent_bytes_total` with their labels.
Currently only supports relay without zero copy as this hides the information about transferred bytes.


src/proxy/outbound.rs
```diff
@@ -260,9 +264,18 @@ impl OutboundConnection {
                 // Create a TCP connection to upstream
                 let mut outbound = TcpStream::connect(req.gateway).await?;
                 // Proxying data between downstrean and upstream
-                relay(&mut stream, &mut outbound).await?;
-
-                Ok(())
+                match relay(&mut stream, &mut outbound, self.cfg.zero_copy_enabled).await {
```
howardjohn, line 267: We will need this in other places as well, but this is a good start

howardjohn (approved):

src/proxy/outbound.rs
```diff
@@ -260,9 +264,18 @@ impl OutboundConnection {
...
+                    // Connection closed with count of bytes transferred between streams
+                    Ok(Some(r)) => {
```
howardjohn, line 269:
```suggestion
                    Ok(Some((sent, recv))) => {
```

instead of r.0 and r.1 is more clear

howardjohn:
I added a hold to prevent auto-merge since there were some nits, feel free to remove

src/metrics/traffic.rs
```diff
@@ -100,6 +102,10 @@ impl<T: Encode> Encode for DefaultedUnknown<T> {
 
 pub struct ConnectionClose<'a>(&'a ConnectionOpen);
 
+pub struct ReceivedBytes<'a>(pub &'a ConnectionOpen);
```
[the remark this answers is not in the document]
ymesika: Without the pub I get an error that the constructor is private when creating an instance here: https://github.com/istio/ztunnel/blob/ab5b28b4ee74f8063ee11050c8a636f392fc360b/src/proxy/outbound.rs#L175

Not sure how to overcome it without pub.

PR #318, opened 2023-01-09
config: correct admin and status ports

shashankram:
Fixes the ports used for admin and status endpoints. 65225bb incorrectly updated the port numbers.


howardjohn (approved):

linsun:
Thanks @shashankram !!

Issue #250, opened 2022-12-06
Make e2e test coverage more realistic


howardjohn:
Currently, our e2e tests use socks5 to connect to the waypoint. This removes the need for outbound redirection.
Eventually it may decide to connect to `<workloadIP>:15008`. Because the ztunnel and workload run on the same IP (localhost),
this *happens* to work without redirection (combined with the fact ORIG_DST lookup has a fallback).

While this works and lets us test most of the flow, it is not entirely realistic.

This issue tracks improving the situation to give us more real world coverage.

Some ideas:
* Require root; setup iptables rules on the host. This is not ideal due to root and messy management of host rules
* Require root; setup iptables rules in netns's for the test. This improves the rule management but is a bit more complex
* Run in docker; setup iptable rules in docker containers. This is like above but drops root requirement (daemon has root, though), removes manual netns setup. But does add docker dependency, complexity, and poor performance.

So far I am not a huge fan of these ideas, so its worth looking into some others as well

type: implementation

PR #319, opened 2023-01-10
tests: use network namespaces

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/250

This PR introduces using network namespaces to deploy ""workloads"", which run in their own network namespace. For example, a test could deploy a client, ztunnel, and server. Redirection is configured to simulate a real k8s environment. This allows us to test a LOT more realistically than we do today, at the cost of requiring root (and complexity)

### DIFF
--- a/scripts/node-redirect.sh
+++ b/scripts/node-redirect.sh
@@ -2,16 +2,14 @@
 
 set -ex
 
-# TODO: use a loop instead of ipset maybe
-ipset create ztunnel-pods-ips hash:ip
-ipset add ztunnel-pods-ips 10.0.1.1
-ipset add ztunnel-pods-ips 10.0.2.1
-ipset add ztunnel-pods-ips 10.0.3.1
-
-
 HOST_IP="10.0.0.1"
-ZTUNNEL_INTERFACE=veth3
-ZTUNNEL_IP="10.0.3.1"
+
+ZTUNNEL_IP="${1:?ztunnel IP}"
+ZTUNNEL_INTERFACE="${2:?ztunnel interface}"
+ipset create ztunnel-pods-ips hash:ip
+for i in `seq 1 ${3:?ip count}`; do
+  ipset add ztunnel-pods-ips 10.0."${i}".1
+done
 
 # Setup interfaces
 ip link add name istioin type geneve id 1000 remote "${ZTUNNEL_IP}"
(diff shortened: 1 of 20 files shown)


hzxuzhonghu:
Instead of manually setting the namespace, why not use docker to setup?

howardjohn:
That would require us to build many containers. This gives us the benefits
docker provides but without needing to build so many containers (we would
need multiple containers *per test* to handle the specific details of the
tests). Additionally, to handle the nested nature we would need
docker-in-docker which is complex. Docker containers are also orders of
magnitude slower to run (ignoring the build time which is 1000x slower)

hzxuzhonghu:
docker-in-docker is not a new concept, not very complex. Compared to the script, the complexity can be less

howardjohn:
We still need the 2 complex scripts (these come from istio/istio) even with
docker, we just save ~8 likes of `ip` commands.

Docker is not viable due to needing to build images, I spent a lot of time
trying to make that work first.

scripts/ztunnel-redirect.sh
```diff
@@ -0,0 +1,116 @@
+#!/bin/bash
+# shellcheck disable=SC2086
+# This script sets up redirection in the ztunnel network namespace for namespaced tests (tests/README.md)
+set -ex
+
+INSTANCE_IP="${1:?INSTANCE_IP}"
+shift
+
+
+# START https://github.com/solo-io/istio-sidecarless/blob/master/redirect-worker.sh#L198-L205
```
stevenctl, line 10: this link 404s (for me)

src/proxy/outbound.rs
```diff
@@ -338,6 +338,12 @@ impl OutboundConnection {
             && self.pi.cfg.local_node == Some(us.workload.node.clone())
             && us.workload.protocol == Protocol::HBONE
         {
+            trace!(
```
stevenctl, line 341: Is it intended to keep the trace! here only for same node?

scripts/node-redirect.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
stevenctl, line 1: My confidence as a reviewer of the redirect scripts is pretty low.. can Yuval or someone with more expertise step in?

tests/namespaced.rs
```diff
@@ -0,0 +1,357 @@
...
+            std::any::type_name::<T>()
+        }
+        let name = type_name_of(f);
+        &name[..name.len() - 3]
+    }};
+}
+
+/// setup_netns_test prepares a test using network namespaces. This checks we have root,
+/// and automatically setups up a namespace based on the test name (to avoid conflicts).
+macro_rules! setup_netns_test {
```
stevenctl, line 42: why does this have to be a macro?

stevenctl (approved):
lgtm, but again I probably wouldn't have caught subtle issues with the redirect stuff.

tests/namespaced.rs
howardjohn: It was intended to do more... but it does `return` which is a bit nice

PR #323, opened 2023-01-10
log version we are running as

howardjohn:


adiprerepa (approved):

src/main.rs
```diff
@@ -52,6 +52,7 @@ fn version() -> anyhow::Result<()> {
 }
 
 async fn proxy(cfg: config::Config) -> anyhow::Result<()> {
+    info!("version: {}", version::BuildInfo::new());
```
hzxuzhonghu, line 55: there is a function `version` can be called

howardjohn: It's useful to have it in the logs so when someone posts a log in an issue we know what version was ran. We have this for all other Istio components, and it prevented me from debugging an issue today because it was missing

howardjohn: that does print, not log

PR #325, opened 2023-01-10
Make caclient not require `mut`

howardjohn:
No need for exclusive access here

### DIFF
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -97,6 +97,8 @@ impl Default for Identity {
     }
 }
 
+/// SecretManager provides a wrapper around a CertificateProvider with caching.
+/// It is designed to be cheap to clone.
 #[derive(Clone)]
 pub struct SecretManager<C: CertificateProvider>
 where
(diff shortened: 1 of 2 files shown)


adiprerepa (approved):
lgtm

src/identity/manager.rs
```diff
@@ -184,12 +181,9 @@ impl<T: CertificateProvider + Clone + Send + 'static> CertificateProvider for Se
                             let (tx, rx) = watch::channel(None);
                             write_locked_cache.insert(id.clone(), rx.clone());
                             drop(write_locked_cache);
-                            tokio::spawn(SecretManager::refresh_handler(
-                                id.clone(),
-                                self.cache.clone(),
-                                self.client.clone(),
-                                tx,
-                            ));
+                            let s = self.clone();
```
hzxuzhonghu, line 184: Doesn't this clone bring much cost

howardjohn: No, it is designed to be cheap to clone (I also added a comment in the code to clarify this). It is just incrementing a reference counter

howardjohn: Also what tonic recommends

howardjohn: Client is also an Arc under the hood and designed (by tonic) to be cheap to clone

hzxuzhonghu: Thanks for clarify

PR #328, opened 2023-01-11
Remove warnings when building on non-linux env

ymesika:


src/socket.rs
```diff
@@ -15,10 +15,13 @@
 use std::io::Error;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 
+#[cfg(all(target_os = "linux"))]
 use realm_io;
+#[cfg(all(target_os = "linux"))]
 use socket2::SockRef;
```
howardjohn, line 21: maybe move import to something already covered by a `cfg` - you can import within a function

src/socket.rs
```diff
@@ -138,15 +141,16 @@ mod linux {
     }
 }
 
+#[cfg(all(target_os = "linux"))]
 const EINVAL: i32 = 22;
 pub async fn relay(
     downstream: &mut tokio::net::TcpStream,
     upstream: &mut tokio::net::TcpStream,
-    zero_copy_enabled: bool,
+    _zero_copy_enabled: bool,
```
howardjohn, line 149: underscore seems odd given it *is* used. Maybe we cna make the entire function differ based on OS to avoid?

src/socket.rs
ymesika: nice

howardjohn (approved):

src/socket.rs
```diff
@@ -50,6 +48,9 @@ pub fn orig_dst_addr_or_default(stream: &tokio::net::TcpStream) -> SocketAddr {
 
 #[cfg(target_os = "linux")]
 fn orig_dst_addr(stream: &tokio::net::TcpStream) -> io::Result<SocketAddr> {
+    use socket2::SockRef;
+    use tracing::warn;
```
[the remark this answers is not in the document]
ymesika: It's allowed in rust but anyway I improved it to have a use block for linux

PR #329, opened 2023-01-11
Fix regression with setting empty environment variable

howardjohn:
This was broken with ProxyConfig supported added

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -170,17 +170,18 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
     } else {
         "https://localhost:15012".to_string()
     };
-
-    let xds_address = parse(XDS_ADDRESS)?
-        .or(pc.discovery_address)
-        .or_else(|| Some(default_istiod_address.clone()));
+    let xds_address = empty_to_none(
+        parse(XDS_ADDRESS)?
+            .or(pc.discovery_address)
+            .or_else(|| Some(default_istiod_address.clone())),
+    );
 
     let fake_ca = parse_default(FAKE_CA, false)?;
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -170,17 +170,18 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
     } else {
         "https://localhost:15012".to_string()
     };
-
-    let xds_address = parse(XDS_ADDRESS)?
-        .or(pc.discovery_address)
-        .or_else(|| Some(default_istiod_address.clone()));
+    let xds_address = empty_to_none(
+        parse(XDS_ADDRESS)?
```
hzxuzhonghu, line 174: why not fix the parse and parse default?

howardjohn: parse distinguishes between unset (None) and set to "" (Some("")). I don't think that's a bug, just different use case

hzxuzhonghu: But we do not need to distinguish unset and empty "".

howardjohn: XDS is disabled by setting to "". Unset means to use the default address

src/config.rs
```diff
@@ -314,6 +315,15 @@ fn construct_proxy_config(mc_path: &str, pc_env: Option<&str>) -> anyhow::Result
     Ok(pc)
 }
 
+pub fn empty_to_none<A: AsRef<str>>(inp: Option<A>) -> Option<A> {
```
hzxuzhonghu, line 318: nit since it is Option<string>, can rewrite as 
pub fn empty_to_none (inp: Option<String>) -> Option<String> {

stevenctl (approved):

PR #333, opened 2023-01-12
orig src: cleanups from initial impl

howardjohn:
* Switch to a library impl of Forwarded. The one we wrote does not handle edge cases. Added tests to cover these.
* Unconditionally append Forwarded (still useful, even if its not used by the ztunnel on the other end)
* Unconditionally spoof the outbound IP
* Only use `Forwarded` from trusted waypoint
* Allow explicitly opt in or opt out. By default, auto detect support

This aligns with https://docs.google.com/document/d/17fDTMLg-22UPfh62d5wumAOIpbeoOyWix2WTE54hDNc/edit#

I also wrote e2e tests in istio/istio; they pass with and without this PR

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -19,9 +19,9 @@ use std::{fmt, io};
 
 use boring::error::ErrorStack;
 use drain::Watch;
+use hyper::{header, Body, Request};
 use rand::Rng;
-use tokio::net::{TcpSocket, TcpStream};
-
+use tokio::net::{TcpListener, TcpSocket, TcpStream};
 use tracing::{error, trace, warn};
 
 use inbound::Inbound;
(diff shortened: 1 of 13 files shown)


src/proxy.rs
```diff
@@ -244,35 +243,50 @@ impl TryFrom<&str> for TraceParent {
...
-        first
-            .parse::<IpAddr>()
-            .map_or_else(|_| first.split(':').next()?.parse::<IpAddr>().ok(), Some)
-    }
+    req.headers()
+        .get(header::FORWARDED)
+        .and_then(|rh| rh.to_str().ok())
+        .and_then(|rh| http_types::proxies::Forwarded::parse(rh).ok())
+        .and_then(|ph| {
+            println!("got {:?}", ph.forwarded_for().last());
```
stevenctl, line 285: remove before merging

stevenctl (approved):

PR #335, opened 2023-01-13
check node before prefetching

adiprerepa:
follow-up #298

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -563,12 +563,10 @@ impl WorkloadStore {
                     .insert((service_sock_addr, port.target_port as u16));
             }
         }
-        if let Some(node) = self.local_node.clone() {
-            if node == w.clone().node {
-                if let Some(tx) = self.cert_tx.as_mut() {
-                    if let Err(e) = tx.try_send(widentity) {
-                        info!("couldn't prefetch: {:?}", e)
-                    }
+        if Some(w.node) == self.local_node.clone() {
+            if let Some(tx) = self.cert_tx.as_mut() {
+                if let Err(e) = tx.try_send(widentity) {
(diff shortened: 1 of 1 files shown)


adiprerepa:
I really really really really don't like the four nested if statements but Rust won't me chain `if let` and `if <expr>` together.

src/workload.rs
```diff
@@ -559,9 +563,13 @@ impl WorkloadStore {
                     .insert((service_sock_addr, port.target_port as u16));
             }
         }
-        if let Some(tx) = self.cert_tx.as_mut() {
-            if let Err(e) = tx.try_send(widentity) {
-                info!("couldn't prefetch: {:?}", e)
+        if let Some(node) = self.local_node.clone() {
+            if node == w.clone().node {
```
howardjohn, line 567: Replace with `if Some(w.node) == self.local_node`. There should be clones needed

adiprerepa: done

src/workload.rs
```diff
@@ -559,9 +563,11 @@ impl WorkloadStore {
                     .insert((service_sock_addr, port.target_port as u16));
             }
         }
-        if let Some(tx) = self.cert_tx.as_mut() {
-            if let Err(e) = tx.try_send(widentity) {
-                info!("couldn't prefetch: {:?}", e)
+        if Some(w.node) == self.local_node.clone() {
```
howardjohn, line 566: why do we need to clone? if its needed can we do `Some(&w.node) == self.local_node.as_ref`

adiprerepa: if I just did `self.local_node` it would error out with `cannot move`. Changed it to that though.

howardjohn (approved):

PR #338, opened 2023-01-16
Add bytes received and sent to hbone connection

ymesika:
Noticed that TCP connections opened/closed are also counted with HBONE connections.
#310 only covered read/write bytes for TCP connections.


howardjohn (approved):
Nice

PR #339, opened 2023-01-16
Replace wrong IPs with constants

ymesika:
@howardjohn keep me honest here but I think current benches for HBONE are going to TCP.
At least that's what I saw in the benchmarks I ran.
Suggested is a correction for the IPs based on the constants.


howardjohn (approved):
ty

PR #348, opened 2023-01-20
Fix hairpin when client and waypoint are remote

howardjohn:
Also add test and timeout to make things more robust

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -15,6 +15,7 @@
 use std::fmt::Debug;
 use std::net::{IpAddr, SocketAddr};
 use std::sync::Arc;
+use std::time::Duration;
 use std::{fmt, io};
 
 use boring::error::ErrorStack;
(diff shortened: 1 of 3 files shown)


src/proxy.rs
```diff
@@ -298,36 +300,52 @@ pub fn get_original_src_from_stream(stream: &TcpStream) -> Option<IpAddr> {
...
-            };
-            Ok(socket.connect(addr).await?)
+                };
+                trace!(%src, dest=%addr, "connect with source IP");
+                Ok(socket.connect(addr).await?)
+            }
         }
     }
+    // Wrap the entire connect function in a timeout
+    timeout(CONNECTION_TIMEOUT, connect(local, addr))
```
ilrudie, line 340: **question (non-blocking)**: should the timeout value be something configurable and it defaults to the constant if not configured?

**suggestion**: (if we do want this to be configurable) open an issue to track implementing a way to set the timeout

stevenctl (approved):

Issue #251, opened 2022-12-06
Define complete set of integration points


howardjohn:
A concern for ztunnel is that it expands to support an unbounded number of integrations, leading to similar issues we have with Envoy. A goal of ztunnel is that it is light and minimal.

I would propose we limit integration points to the following:

* Metrics: Prometheus and OTLP
* Access Logs: stdout and OTLP
* Traces: OTLP (if we implement tracing)
* CA: Citadel gRPC and Kubernetes CSR (we can actually drop Citadel if Istiod supports CSR, but we might want it for VMs)

type: design

bleggett:
One thing I think it is important for ambient to continue to support is [support for alternative SPIFFE-compliant workload CAs that implement SDS](https://istio.io/latest/docs/ops/integrations/spire/)

We already support that in sidecar so parity is good/important, but in general I think having as few opinions about how certs are minted and rotated and what the CA is attesting as possible is good for constraining ztunnel's complexity.

howardjohn:
Any option would be a tradeoff between supported integrations and complexity. SDS might make the cut. The Kubernetes CSR is a core K8s API intended for exactly that purpose though, so I would be curious if existing SDS implementations could/would implement the K8s CSR API - and if not, their reasoning may be interesting

howardjohn:
Hmm I guess SDS probably gets us a full cert+priv key, while CSR simply signs them.

bleggett:
@howardjohn does ztunnel currently mint workload certs or sign anything related to workloads itself, or does it largely work like sidecars do and simply expect $some_workload_cert to exist and be validatable against some authority?

howardjohn:
We have a certificate per workload identity on the node, not just 1 cert

bleggett:
> We have a certificate per workload identity on the node, not just 1 cert

Right - but ztunnel shouldn't ever need to mint any of those, it should just expect to obtain them, and validate them against some authority, correct.

howardjohn:
If there was some way to obtain them, yes. Currently there isn't afaik

bleggett:
Linking https://github.com/istio/istio/issues/42339 as potentially related - as currently sidecar proxy SDS support is how we integrate SPIRE with Sidecar and presumably we'd need a semi-compatible model for ambient.

howardjohn:
[...]

PR #353, opened 2023-01-24
proposal: extensions

yuval-k:
Allow extending the behaviour of ztunnel by invoking it as a library with extensions.


examples/extension.rs
```diff
@@ -0,0 +1,33 @@
...
+
+impl ExampleExtension {
+    fn new() -> Self {
+        Self {}
+    }
+}
+
+impl Extension for ExampleExtension {
+    fn on_listen(&self, l: &tokio::net::TcpListener, _: ListenerType) {
+        print!("ExampleExtension: Listening on {}", l.local_addr().unwrap());
```
ymesika, line 27: You might want to use the logging macros instead

src/lib.rs
```diff
@@ -31,3 +32,47 @@ pub mod xds;
...
+
+// #[global_allocator]
+// static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
+
+// #[global_allocator]
+// static GLOBAL: tcmalloc::TCMalloc = tcmalloc::TCMalloc;
+
+pub fn entry(extension: Option<Box<dyn Extension + Sync + Send>>) -> anyhow::Result<()> {
+    telemetry::setup_logging();
+    let config: config::Config = config::parse_config(extension)?;
```
ymesika, line 50: I think it should be placed after the match below

yuval-k: this part of was moved as is from main.rs

howardjohn (commented):
I think with any extensibility mechanism, we need:
* Very very clear and explicit bounds, to avoid this creeping from 4 callbacks to 400
* Multiple usages by different vendors/users that validate the abstraction picked is correct

So far I have chatted with 3 folks about this (including you and myself), and 2/3 of them would not be supported by this model. Its perfectly fine to have folks need customization beyond what extensibility we provide out of the box - there will always be folks with super bespoke use cases that we can't and should not bend over backwards to meet. 

However, I would like to at least see some other user that could implement there desired extensions via this mechanism. If not, it may make sense to wait until we have reached that point before prematurely merging anything.

I do think in general this seems like a plausible extension mechanism, though

linsun:
@howardjohn can you explain the scenarios of extensions from 2/3 you mentioned that would not fit?

howardjohn:
More invasive changes like different transport protocols, custom tweaks to how we do HBONE, different copying mechanisms, etc. Its a bit vague, though, for all involved I think - which is another reason not to rush things

kyessenov:
Is the extended featureset so broad that the waypoints cannot capture all datapaths? Normally, network functions could be chained, which is essentially what waypoints do to enhance L7 processing. Why not adopt the similar model for ztunnels and run traffic through a chain of coprocessors? The protocol is sufficient to express it, and the running out of process has many flexibility benefits.

yuval-k:
I think that some extensions (like tweaks to HBONE) will require ztunnel extensions vs out of process ones.

As far as the API itself, these extensions are not an istio user-facing api, but rather an istio vendor-facing api (similar to envoy filters in that regard).

As such, I think it's ok  to treat extensions as an internal API, and allow it to be re-factored/broken without notice (as long as the same use-cases are still enabled, i.e. downstream code will still work after it adapts to the new API).

This is similar to envoy's extension development model (i.e. the stable config is the `XDS` protos, not `Http::StreamDecoderFilter`).

kyessenov:
HBONE is the only stable specification we have here, so I'm hoping we don't make vendor-specific customizations to it.
Otherwise, well-documented internal interfaces seems like a good practice, as long as they are not treated as an extension model.

bleggett:
1. I think to @kyessenov 's point we will need 2 kinds of integrations

- Documented, end-user facing integrations - "bring your own X and we will play with it" - these are the kinds of things John is mentioning in #251 - the list John has there makes sense, and I agree these need to be limited and formalized, as we will have to document and support them here in this repo -> _we_ (this repo/Istio) bear the cost.

- Internal interfaces/extension points around typical TCP proxy lifecycle hooks that make it easier to hack/extend/experiment with ztunnel proper, but which Istio doesn't make stability, compatibility, or functionality guarantees about and which we don't support as user-facing - if you want to use one of these you either need to fork the repo (and bear that cost yourself) or wrap ztunnel with your own code (and bear the cost of keeping in sync with upstream) -> _you_ (vendor/entity hacking on ztunnel) bear the cost.

2. I think the best way to constrain ourselves is to keep `ztunnel` a relatively ignorant TCP proxy that routes packets and ignores protocols - if that is all it does, then that is a pretty effective hard limit on how it can be extended, and what  we choose to allow as far as upstreamed extension points.

costinm:
A different perspective:    I think we have close to consensus that  we should keep ztunnel minimal in the Istio distribution and have a stable common protocol. This will allow integrations and interop - including other implementations of the  protocol, for example in proxyless or  native implementations in different languages or non-istio implementations ( and outside of K8S ).

While I don't think a hook would help - having a relatively stable API in ztunnel that allows different parts to be reused is not bad. For example a rust application outside of k8s ( or on a k8s cluster where user doesn't have permissions to set ztunnel/cni) should be able to reuse the ztunnel code for connecting to XDS and getting PTR-DS, as well as the code to 
add metadata.

It is not clear if such code should be ztunnel or another crate that is focused on providing a library-like API and used 
in rust-based applications using HBONE and/or PTR-DS, but I don't see why someone would need to duplicate the 
code in ztunnel just to provide a stable API. For other languages - separate libraries will likely be needed for full 
proxyless/native ambient support. 

With a stable API - it should be possible to customize and hook with far more flexibility and still use common code
for integration with the XDS server.

bleggett:
> A different perspective: I think we have close to consensus that we should keep ztunnel minimal in the Istio distribution and have a stable common protocol. This will allow integrations and interop - including other implementations of the protocol, for example in proxyless or native implementations in different languages or non-istio implementations ( and outside of K8S ).
> 
> While I don't think a hook would help - having a relatively stable API in ztunnel that allows different parts to be reused is not bad. For example a rust application outside of k8s ( or on a k8s cluster where user doesn't have permissions to set ztunnel/cni) should be able to reuse the ztunnel code for connecting to XDS and getting PTR-DS, as well as the code to add metadata.

One thing I would note is that we are already _pretty_ near to hitting the `clippy` `too_many_arguments` lint flag in e.g. `handle_inbound` (one of the functions we proposed adding an extension hook to) on `master` already, due to dependency injection: https://github.com/istio/ztunnel/blob/master/src/proxy/inbound.rs#L144

There are ways around that of course, but I think a hook API there makes a lot more sense than piling more stuff into those same handlers for whatever the next feature is that comes along.

Arguably L4 metrics _could_ have been implemented with some variant of the extension API proposed here. If more things _like_ the L4 metrics implementation come along, I think we might need to revisit the "extension hooks don't simplify code" argument before we jam more things into the core TCP connection handlers.

PR #355, opened 2023-01-24
fix dockerfile for remote-env lint issues

ilrudie:
- adds `--no-install-recommends` to apt-get at suggestion of linter; avoids adding unnecessary packages
- disable the linter warning about `pipefail` since the shell doesn't support it

### DIFF
--- a/docker/remote-env/Dockerfile
+++ b/docker/remote-env/Dockerfile
@@ -27,7 +27,7 @@ FROM gcr.io/istio-testing/build-tools:master-65b95c3425a26e633081b2d0834cc0df6e8
 # - sudo, while not required, is recommended to be installed, since the
 #   workspace user (`gitpod`) is non-root and won't be able to install
 #   and use `sudo` to install any other tools in a live workspace.
-RUN apt-get update && apt-get install -yq \
+RUN apt-get update && apt-get install --no-install-recommends -yq \
     net-tools \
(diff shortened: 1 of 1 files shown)


docker/remote-env/Dockerfile
```diff
@@ -62,6 +62,8 @@ ARG REMOTE_USER_SHELL=/bin/bash
 ARG REMOTE_USER_LOGIN_SCRIPT=$REMOTE_USER_HOME/.bashrc
 ARG REMOTE_USER_ID=3333
 ARG REMOTE_USER_GROUPS=sudo
+# ignoring because the current shell doesn't support pipefail; likely want to have a better fix long term
+# hadolint ignore=DL4006
 RUN useradd -lm \
```
ymesika, line 67: Maybe as long term fix exec it?
```sh
RUN ["/bin/bash", "-c", "set -o pipefail && useradd ..."]
```

Anyway, nit

ilrudie: https://github.com/istio/ztunnel/issues/360

ilrudie: I like the suggestion. I opened an issue to track on it. That exact syntax doesn't satisfy the linter but it may be worth using anyway to at least get the pipefail working.

howardjohn (approved):

PR #367, opened 2023-01-31
Fix loopback connections with direct inbound proxy and enable transparent socket: true

kdorosh:
Fix loopback connections with direct inbound proxy and enable transparent socket: true

in https://github.com/istio/ztunnel/pull/333 the default behavior is now to unconditionally spoof the outbound IP. this does not work if the outbound ip is loopback, like it used to before this default was changed. this PR fixes that scenario so ztunnel can proxy to itself if there was a good reason to (e.g. in integration tests)


kdorosh:
i actually think that if the VIP resolved to loopback/local we should error on self call here. this is already failing on latest versions of ztunnel (appropriately) so this can be closed

PR #371, opened 2023-02-03
More TCP metrics support

howardjohn:
* Add metrics with splice
* Give telemetry at all points we write
* Test and validate the telemetry

With this, Kiali shows the mesh with just ztunnel.

Not yet implemented is looking up source attributes for baggage or XDS, but can be a followup

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ tower = { version = "0.4.12", features = ["full"] }
 tracing = "0.1.34"
 tracing-subscriber = { version = "0.3.16" , features = ["registry", "env-filter"]}
 dyn-clone = "1.0.9"
-realm_io = "0.3.5"
+realm_io = { git = "https://github.com/howardjohn/realm", branch = "io/copy-bytes-sent" }
 go-parse-duration = "0.1.1"
 prometheus-parse = "0.2.3"
 url = "2.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,7 @@ tower = { version = "0.4.12", features = ["full"] }
 tracing = "0.1.34"
 tracing-subscriber = { version = "0.3.16" , features = ["registry", "env-filter"]}
 dyn-clone = "1.0.9"
-realm_io = { git = "https://github.com/howardjohn/realm", branch = "io/copy-bytes-sent" }
+realm_io = "0.4"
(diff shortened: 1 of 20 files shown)


Cargo.toml
```diff
@@ -59,7 +59,7 @@ tower = { version = "0.4.12", features = ["full"] }
 tracing = "0.1.34"
 tracing-subscriber = { version = "0.3.16" , features = ["registry", "env-filter"]}
 dyn-clone = "1.0.9"
-realm_io = "0.3.5"
+realm_io = { git = "https://github.com/howardjohn/realm", branch = "io/copy-bytes-sent" }
```
hzxuzhonghu, line 62: Will you file a pr to realm?

howardjohn: https://github.com/zhboner/realm/pull/95. We can see what they say before we merge this, hopefully they are ok with it

howardjohn: Merged, switched to real dep

src/test_helpers/tcp.rs
```diff
@@ -121,11 +127,28 @@ where
...
+            loop {
+                let read = r.read(&mut buffer).await.expect("tcp ready");
+                if read == 0 {
+                    break;
+                }
+                let wrote = w.write(&buffer[..read]).await.expect("tcp ready");
+                if wrote == 0 {
+                    break;
+                }
+                let wrote = w.write(&buffer[..read]).await.expect("tcp ready");
```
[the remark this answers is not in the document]
howardjohn: this is is just for testing so we can make sure we record the "read" and "wrote" metrics correctly. Before we always read and write same data so we cannot actually test the 2 metrics we're not swapped accidentally

tests/namespaced.rs
```diff
@@ -136,16 +170,70 @@ async fn test_hbone_request() -> anyhow::Result<()> {
...
+fn source_labels() -> HashMap<String, String> {
+    HashMap::from([("reporter".to_string(), "source".to_string())])
+}
+
+async fn verify_metrics(
+    ztunnel: TestApp,
+    assertions: &[(&str, u64)],
+    labels: &HashMap<String, String>,
+) {
+    // Wait form metrics to populate...
```
ymesika, line 204: for

ymesika (approved):
Nice. lgtm.

PR #374, opened 2023-02-03
Add health status

howardjohn:
Replaces https://github.com/istio/istio/pull/43085

Note: also fixes a bug where removing a VIP is not respected

### DIFF
--- a/proto/workload.proto
+++ b/proto/workload.proto
@@ -71,6 +71,15 @@ message Workload {
   // NOTE: this *only* includes Selector based policies. Namespace and global polices
   // are returned out of band.
   repeated string authorization_policies = 16;
+
+  WorkloadStatus status = 17;
+}
+
+enum WorkloadStatus {
+  // Workload is healthy and ready to serve traffic.
+  HEALTHY = 0;
+  // Workload is unhealthy and NOT ready to serve traffic.
+  UNHEALTHY = 1;
 }
 
 enum WorkloadType {
--- a/src/test_helpers.rs
+++ b/src/test_helpers.rs
@@ -97,6 +97,7 @@ pub fn test_default_workload() -> Workload {
         canonical_name: "".to_string(),
         canonical_revision: "".to_string(),
         node: "".to_string(),
+        status: Default::default(),
 
         authorization_policies: Vec::new(),
         native_hbone: false,
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -14,6 +14,7 @@
 
 use std::collections::{HashMap, HashSet};
 use std::convert::Into;
+use std::default::Default;
 use std::net::{IpAddr, SocketAddr};
 use std::ops::Deref;
 use std::sync::{Arc, Mutex};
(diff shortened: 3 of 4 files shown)


bleggett (commented):
1. Right now, nothing in `ztunnel` needs to know this information - it is blissfully ignorant of Pod lifecycles or "health" as a concept. Is there a specific reason `ztunnel` needs to know this, or a desired feature it enables, or are we anticipating such a feature?


2. I have some concerns about letting K8S Pod lifecycle state machine stuff leak out of `istiod` into ztunnel - ideally `ztunnel` shouldn't care about a high-level concept like endpoint "health", and the fact that it doesn't is a good thing, given how specific and bug-prone k8s lifecycle state machine stuff has historically been for us.

Right now it's two simple states - A workload exists in the workload API, or it doesn't. If it exists, it is a valid source of traffic. If it exists and has VIPs, it is a valid destination for traffic - if it has neither, it shouldn't be in the workload API and `ztunnel` doesn't care or need to know about it.

Is there a scenario where a workload might not have a VIP but should be a valid routable destination?

howardjohn:
Yes, see ProxyTerminatingEndpoints KEP in Kubernetes. There are cases where we will legitimately send traffic to endpoints that are not ready, even through a VIP. Having state of an endpoint is standard in proxies as well, we will likely have behavior around "STARTING" and "TERMINATING" at some point for example

bleggett:
@howardjohn OK, then I'm fine with it as long as we stick to the two-state model here. TY!

bleggett (approved):
LGTM

proto/workload.proto
```diff
@@ -71,6 +71,15 @@ message Workload {
...
   repeated string authorization_policies = 16;
+
+  WorkloadStatus status = 17;
+}
+
+enum WorkloadStatus {
+  // Workload is healthy and ready to serve traffic.
+  HEALTHY = 0;
+  // Workload is unhealthy and NOT ready to serve traffic.
+  UNHEALTHY = 1;
```
EItanya, line 82: Thinking out loud about the ordering here, should the default Workload Status be UNHEALTHY or HEALTHY? It feels to me that `UNHEALTHY` might be a better default. UNHEALTHY until proven healthy maybe.

howardjohn: I mostly picked this to align with Envoy's API - but not sure if that is good.

In practice this should always always be explicitly set. In a hypothetical broken control plane that never set this it does seem like we would prefer 100% healthy than 100% unhealthy, but we should also not have broken control planes...

hzxuzhonghu (approved):
lgtm

proto/workload.proto
EItanya: Yea, I think if Envoy is this way, then maybe it's better to keep it in line, but I do think the other way makes more sense to me, with the addition of some panic threshold.

bleggett: The states are not set in stone anyway, we will likely have to extend or rework the states when we get to actually implementing e.g. the [ProxyTerminatingEndpoint KEP](https://github.com/istio/ztunnel/issues/375) in ZTunnel. We can revisit then if needed.

PR #380, opened 2023-02-07
Issue #9: Handled some expects()/unwraps()

briansonnenberg:
I think the rest of the unwraps/expects are at bootup or in test code.

### DIFF
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -367,6 +367,9 @@ impl OutboundConnection {
                 request_type: RequestType::ToServerWaypoint,
             });
         }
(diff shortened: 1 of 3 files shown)


src/proxy/outbound.rs
```diff
@@ -367,6 +367,9 @@ impl OutboundConnection {
                 request_type: RequestType::ToServerWaypoint,
             });
         }
+        if us.workload.gateway_address.is_none() {
```
howardjohn, line 370: I think this is actually impossible to get with how find_upstream works, which may suggest that we should have a debug_assert or refactor to make the impossible state not representable

howardjohn (approved):

PR #384, opened 2023-02-08
Few fixes for build on non-linux os

ymesika:
Also consider removing the non-linux support for easier maintenance?


howardjohn (approved):

bleggett:
I don't think building on non-Linux is a huge priority - but cleanly building on Linux across the arches we currently already publish Istio artifacts for is.

PR #385, opened 2023-02-08
Support setting the ports in environment variables

ymesika:
Environment variable with highest priority. If not set, same as before.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -37,6 +37,9 @@ const FAKE_CA: &str = "FAKE_CA";
 const ZTUNNEL_WORKER_THREADS: &str = "ZTUNNEL_WORKER_THREADS";
 const ENABLE_ORIG_SRC: &str = "ENABLE_ORIG_SRC";
 const PROXY_CONFIG: &str = "PROXY_CONFIG";
+const ZTUNNEL_ADMIN_PORT: &str = "ZTUNNEL_ADMIN_PORT";
+const ZTUNNEL_READINESS_PORT: &str = "ZTUNNEL_READINESS_PORT";
+const ZTUNNEL_STATS_PORT: &str = "ZTUNNEL_STATS_PORT";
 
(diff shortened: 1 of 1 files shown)


shashankram:
Thanks @ymesika for this change. This is a useful customization in environments needing better control over the proxy ports.

howardjohn (commented):
Can you expand on the use case a bit? I am mostly curious why we only set these 3 ports?

Also why we need env var for stats and admin, which already have a setting in proxy config. AFAIK the readiness port is the same Envoy uses which is also hardcoded, so why does ztunnel have more flexibility than envoy?

shashankram:
> Can you expand on the use case a bit? I am mostly curious why we only set these 3 ports?
> 
> Also why we need env var for stats and admin, which already have a setting in proxy config. AFAIK the readiness port is the same Envoy uses which is also hardcoded, so why does ztunnel have more flexibility than envoy?

I don't think we need overrides for the admin port, but it would make sense to provide the ability to configure the stats port so pilot-agent and ztunnel can co-exist on the same pod, primarily for users relying on the existing istio-dns proxy capability for custom DNS resolution. For this specific scenario, to use the mesh config yaml we'd need 2 separate copies of it with different ports. Having an env var to override the default port seems more flexible at the cost of reduced complexity of needing multiple mesh config files.

howardjohn:
I am not sure why having 2 configs for 2 different things is a problem.

shashankram:
> I am not sure why having 2 configs for 2 different things is a problem.

It's not a problem, but rather relatively inconvenient compared to updating the env var without needing to maintain multiple config files. If you prefer having 2 separate config files for this scenario, I don't envision needing to configure these ports via env vars.

howardjohn:
IMO proxy config is the standard way to configure things in Istio and we have used it here, so we should be consistent.

ymesika:
Closing in favor of having different configs.

PR #387, opened 2023-02-08
K8 CSR support

aryan16:
Added support for K8 CSR and custom signer

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,6 +69,9 @@ http-types = { version = "2.12.0", default-features = false }
 netns-rs = "0.1.0"
 textnonce = { version = "1.0.0" }
 priority-queue = "1.3.0"
+kube = { version = "0.78.0", default-features = false, features = ["runtime","derive","client"] }
+k8s-openapi = { version = "0.17.0", default-features = false, features = ["v1_22"] }
+bytestring = "1.2.0"
 
 [build-dependencies]
 tonic-build = { version = "0.8", default-features=false, features = ["prost"] }
--- a/src/app.rs
+++ b/src/app.rs
@@ -130,10 +130,16 @@ pub async fn build_with_cert(
     })
 }
 
-pub async fn build(config: config::Config) -> anyhow::Result<Bound> {
+pub async fn build(config: config::Config) -> anyhow::Result<Bound, anyhow::Error> {
     if config.fake_ca {
         let cert_manager = identity::mock::new_secret_manager(Duration::from_secs(86400));
         build_with_cert(config, cert_manager).await
+    } else if config.k8s_custom_signer.is_some() {
+        let cert_manager = identity::SecretManager::origin(config.clone()).await;
+        match cert_manager {
+            Ok(cert_mgr) => build_with_cert(config, cert_mgr).await,
+            Err(_) => Err(anyhow::Error::msg("failed to build the client")),
+        }
     } else {
         let cert_manager = identity::SecretManager::new(config.clone())?;
         build_with_cert(config, cert_manager).await
--- a/src/config.rs
+++ b/src/config.rs
@@ -34,6 +34,7 @@ const XDS_ADDRESS: &str = "XDS_ADDRESS";
 const CA_ADDRESS: &str = "CA_ADDRESS";
 const TERMINATION_GRACE_PERIOD: &str = "TERMINATION_GRACE_PERIOD";
 const FAKE_CA: &str = "FAKE_CA";
+const K8S_CUSTOM_SIGNER: &str = "K8S_CUSTOM_SIGNER";
 const ZTUNNEL_WORKER_THREADS: &str = "ZTUNNEL_WORKER_THREADS";
 const ENABLE_ORIG_SRC: &str = "ENABLE_ORIG_SRC";
 const PROXY_CONFIG: &str = "PROXY_CONFIG";
@@ -88,8 +89,12 @@ pub struct Config {
     /// CA address to use. If fake_ca is set, this will be None.
     /// Note: we do not implicitly use None when set to "" since using the fake_ca is not secure.
     pub ca_address: Option<String>,
+    // Custom Signer
+    pub k8s_custom_signer: Option<String>,
     /// Root cert for CA TLS verification.
     pub ca_root_cert: RootCert,
+    /// K8s CA cert
+    pub k8s_ca_crt: RootCert,
     /// XDS address to use. If unset, XDS will not be used.
     pub xds_address: Option<String>,
     /// Root cert for XDS TLS verification.
@@ -186,6 +191,7 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
     } else {
         Some(parse_default(CA_ADDRESS, default_istiod_address)?)
     });
+    let k8s_custom_signer = Some(parse_default(K8S_CUSTOM_SIGNER, "".to_owned())?);
 
     Ok(Config {
         window_size: 4 * 1024 * 1024,
@@ -224,8 +230,14 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
         // TODO: full FindRootCAForXDS logic like in Istio
         xds_root_cert: RootCert::File("./var/run/secrets/istio/root-cert.pem".parse().unwrap()),
         ca_address,
+        k8s_custom_signer,
         // TODO: full FindRootCAForCA logic like in Istio
         ca_root_cert: RootCert::File("./var/run/secrets/istio/root-cert.pem".parse().unwrap()),
+        k8s_ca_crt: RootCert::File(
+            "./var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+                .parse()
+                .unwrap(),
+        ),
         local_xds_config: parse::<PathBuf>(LOCAL_XDS_PATH)?.map(ConfigSource::File),
         xds_on_demand: parse_default(XDS_ON_DEMAND, false)?,
         proxy_metadata: pc.proxy_metadata,
--- a/src/identity/customca.rs
+++ b/src/identity/customca.rs
@@ -1,0 +1,154 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::CertificateProvider;
+use crate::identity::Error;
+use crate::tls;
+use crate::{identity::manager::Identity, tls::SanChecker};
+use async_trait::async_trait;
+use futures::{StreamExt, TryStreamExt};
+use k8s_openapi::api::certificates::v1::{
+    CertificateSigningRequest, CertificateSigningRequestSpec,
+};
+use kube::{
+    api::{Api, ListParams, PostParams, WatchEvent},
+    client::Client,
+};
+use std::str;
+use tracing::{info, instrument, warn};
+
+const TTL: i32 = 86400;
+
+#[derive(Clone)]
+pub struct CustomSigner {
+    pub signer: String,
+    pub client: Client,
+}
+
+impl CustomSigner {
+    pub fn new(signer: String, client: Client) -> Result<CustomSigner, Error> {
+        Ok(CustomSigner { signer, client })
+    }
+}
+
+#[async_trait]
+impl CertificateProvider for CustomSigner {
+    #[instrument(skip_all)]
+    async fn fetch_certificate(&self, id: &Identity) -> Result<tls::Certs, Error> {
+        let client = self.client.clone();
+        let csr: Api<CertificateSigningRequest> = Api::all(client);
+        let pp = PostParams::default();
+        let cs = tls::CsrOptions {
+            san: id.to_string(),
+        }
+        .generate()?;
+        let cert: Vec<u8> = cs.csr;
+        let pkey = cs.pkey;
+        let signer_name = &self.signer;
+        let usage = vec!["client auth".to_string()];
+
+        let obj_metadata = k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta {
+            generate_name: Some("csr-ztunnel-".to_string()),
+            ..Default::default()
+        };
+        let csr_spec = CertificateSigningRequestSpec {
+            expiration_seconds: Some(TTL),
+            request: k8s_openapi::ByteString(cert),
+            signer_name: signer_name.to_string(),
+            usages: Some(usage),
+            ..Default::default()
+        };
+
+        let csr_req = CertificateSigningRequest {
+            metadata: obj_metadata,
+            spec: csr_spec,
+            ..Default::default()
+        };
+
+        // creating CSR
+        match csr.create(&pp, &csr_req).await {
+            Ok(o) => {
+                let csr_name = o.metadata.name.unwrap_or_default();
+                let final_certs = check_signed_cert(&csr, &csr_name).await;
+                match final_certs {
+                    Ok(cert_chain) => {
+                        let cert_chain_vec = cert_chain.0;
+                        let final_cert_chain = vec![str::from_utf8(&cert_chain_vec).unwrap()];
+                        let leaf = final_cert_chain[0].as_bytes();
+                        let ca_cert_bundle = if final_cert_chain.len() > 1 {
+                            final_cert_chain[1..]
+                                .iter()
+                                .map(|final_cert_chain| final_cert_chain.as_bytes())
+                                .collect()
+                        } else {
+                            warn!("no chain certs for: {}", id);
+                            vec![]
+                        };
+                        let res_certs = tls::cert_from(&pkey, leaf, ca_cert_bundle);
+                        res_certs
+                            .verify_san(id)
+                            .map_err(|_| Error::SanError(id.clone()))?;
+                        return Ok(res_certs);
+                    }
+                    Err(_) => return Err(Error::EmptyResponse(id.clone()))?,
+                }
+            }
+            Err(kube::Error::Api(ae)) => warn!("CSR creation failed with an error {}", ae.message),
+            Err(_) => return Err(Error::EmptyResponse(id.clone()))?,
+        };
+        Err(Error::EmptyResponse(id.clone()))
+    }
+}
+async fn check_signed_cert(
+    csr: &Api<CertificateSigningRequest>,
+    csr_name: &String,
+) -> Result<k8s_openapi::ByteString, anyhow::Error> {
+    let lp = ListParams::default()
+        .fields(&format!("{}{}", "metadata.name=", csr_name))
+        .timeout(60);
+    let l = csr.list(&lp).await?;
+    if !l.items.is_empty() {
+        let req_signed = l.items[0].clone();
+        match req_signed.status {
+            Some(status) => {
+                if let Some(cert) = status.certificate {
+                    return Ok(cert);
+                }
+            }
+            None => info!("No status field in the CSR: {} yet", csr_name),
+        };
+    }
+    if let Some(rv) = l.metadata.resource_version {
+        let mut stream = csr.watch(&lp, &rv).await?.boxed();
+        while let Some(status) = stream.try_next().await? {
+            match status {
+                WatchEvent::Added(o) => {
+                    info!("Added {}", o.metadata.name.unwrap_or_default());
+                }
+                WatchEvent::Modified(o) => match o.status {
+                    Some(status) => {
+                        if let Some(cert) = status.certificate {
+                            return Ok(cert);
+                        }
+                    }
+                    None => warn!("CSR is not yet signed by the signer"),
+                },
+                _ => {}
+            }
+        }
+    }
+    Err(anyhow::Error::msg(
+        "Signer is unable to sign the certificate",
+    ))
+}
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -26,6 +26,7 @@ use tokio::time::{sleep_until, Duration, Instant};
 
 use crate::tls;
 
+use super::CustomSigner;
 use super::Error::{self, Spiffe};
 use super::{CaClient, CertificateProvider};
 
@@ -411,6 +412,20 @@ impl SecretManager<CaClient> {
     pub fn new(cfg: crate::config::Config) -> Result<Self, Error> {
         let caclient = CaClient::new(cfg.ca_address.unwrap(), cfg.ca_root_cert.clone(), cfg.auth)?;
         Ok(Self::new_with_client(caclient))
+    }
+}
+
+impl SecretManager<CustomSigner> {
+    pub async fn origin(cfg: crate::config::Config) -> Result<Self, anyhow::Error> {
+        let k8s_client = tls::create_k8s_client(cfg.k8s_ca_crt).await;
+        match k8s_client {
+            Ok(client) => {
+                let custom_signer =
+                    CustomSigner::new(cfg.k8s_custom_signer.unwrap_or_default(), client)?;
+                Ok(Self::new_with_client(custom_signer))
+            }
+            Err(_) => Err(anyhow::Error::msg("K8s client creation failed")),
+        }
     }
 }
 
--- a/src/tls/boring.rs
+++ b/src/tls/boring.rs
@@ -45,6 +45,8 @@ use crate::config::RootCert;
 use crate::identity::{self, Identity};
 
 use super::Error;
+use kube::{client::ConfigExt, Client, Config};
+use tower::ServiceBuilder;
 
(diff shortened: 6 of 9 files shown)


aryan16:
@howardjohn I will move the client creation to the common level (probably when we create the configs) ?

src/identity/customca.rs
```diff
@@ -0,0 +1,129 @@
...
+            ..Default::default()
+        };
+
+        let csr_req = CertificateSigningRequest {
+            metadata: obj_metadata,
+            spec: csr_spec,
+            ..Default::default()
+        };
+
+        // creating CSR
```
hzxuzhonghu, line 66: Who is going to approve this csr?

hzxuzhonghu (commented):
Why do we have a different k8s csr support from sidecar? We send request to istiod  and istiod will do this for sidecar

src/identity/customca.rs
```diff
@@ -0,0 +1,129 @@
...
+            Ok(o) => {
+                let name = o.metadata.name.unwrap_or_default();
+                assert_eq!(csr_req.metadata.name.unwrap_or_default(), name);
+            }
+            Err(kube::Error::Api(ae)) => assert_eq!(ae.code, 409),
+            Err(_) => return Err(Error::EmptyResponse(id.clone()))?
+        };
+
+        let final_certs = check_signed_cert(&csr, &stored_csr_name).await;
+        let res_certs : tls::Certs;
```
briansonnenberg, line 77: nit: Generally want to avoid uninitialized declarations

src/identity/customca.rs
aryan16: users need to set their custom signer - https://github.com/istio/ztunnel/blob/4e5667f8ed0ef7ccd3cdfe8b6c67774fc0a62fbd/src/config.rs#L93

aryan16:
> Why do we have a different k8s csr support from sidecar? We send request to istiod and istiod will do this for sidecar

IMO Ztunnel has all the capabilities to make a CSR request directly. Why to add an extra complexity of sending req to istiod.
cc : @howardjohn

Cargo.toml
```diff
@@ -68,6 +68,10 @@ ipnet = { version = "2.7.0", features = ["serde"] }
 http-types = { version = "2.12.0", default-features = false }
 netns-rs = "0.1.0"
 textnonce = { version = "1.0.0" }
+boringssl = "0.0.5"
```
howardjohn, line 71: need to use the existing `boring`, we already have a library for this

Cargo.toml
```diff
@@ -68,6 +68,10 @@ ipnet = { version = "2.7.0", features = ["serde"] }
...
+kube = { version = "0.78.0", default-features = false, features = ["runtime","derive","rustls-tls", "client"] }
```
howardjohn, line 72: can we turn off rustls

src/identity/customca.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub fn new(signer: Option<std::string::String>) -> Result<CustomSigner, Error> {
+        Ok(CustomSigner { signer })
+    }
+}
+
+#[async_trait]
+impl CertificateProvider for CustomSigner {
+    #[instrument(skip_all)]
+    async fn fetch_certificate(&self, id: &Identity) -> Result<tls::Certs, Error> {
+        let client: Client = Client::try_default()
```
howardjohn, line 31: we should build client once and use for all fetches

src/identity/manager.rs
```diff
@@ -122,6 +123,17 @@ impl SecretManager<CaClient> {
     }
 }
 
+impl SecretManager<CustomSigner> {
+    pub fn origin(cfg: crate::config::Config) -> Result<Self, Error> {
+        let cache: HashMap<Identity, watch::Receiver<Option<tls::Certs>>> = Default::default();
```
liminw, line 128: How often is the cache refreshed?

src/identity/customca.rs
```diff
@@ -0,0 +1,129 @@
...
+impl CustomSigner {
+    pub fn new(signer: Option<std::string::String>) -> Result<CustomSigner, Error> {
+        Ok(CustomSigner { signer })
+    }
+}
+
+#[async_trait]
+impl CertificateProvider for CustomSigner {
+    #[instrument(skip_all)]
+    async fn fetch_certificate(&self, id: &Identity) -> Result<tls::Certs, Error> {
```
liminw, line 30: Can you document what is trigger fetch_certificate()? And how is certificate being refreshed?

src/identity/manager.rs
```diff
@@ -122,6 +123,17 @@ impl SecretManager<CaClient> {
...
+        let custom_signer = CustomSigner::new(cfg.custom_signer)?;
+        Ok(Self {
+            client: custom_signer,
+            cache: Arc::new(RwLock::new(cache)),
```
ymesika, line 132: I can't find any place in this PR that uses this cache.
Is it on just a place holder for something?

hzxuzhonghu:
@aryan16 How can we support VM installation using this mode?

src/identity/manager.rs
aryan16: The cache is refreshed after half of the cert life is over.

src/identity/customca.rs
aryan16: If cert is already there in cache we don't need to create a CSR but if it's not there, we refresh the cache by creating a new K8s CSR. It can be triggered in 3 ways - 
1) Pod start-up - pre-load them so that first req shouldn't be slow.
2) Inbound Req
3) Outbound Req
@howardjohn please correct me if I missed anything here.

src/identity/manager.rs
aryan16: https://github.com/istio/ztunnel/blob/4e5667f8ed0ef7ccd3cdfe8b6c67774fc0a62fbd/src/app.rs#L138
one of the examples, how we call fetch_certificate() - https://github.com/istio/ztunnel/blob/master/src/proxy/inbound.rs#L413

Cargo.toml
aryan16: We need to put something in features, otherwise it will use default one which is openssl.

Cargo.toml
aryan16: sure, will check this.

Cargo.toml
howardjohn: We can ONLY using `boring` for TLS. No openssl, rustls, etc. This is a hard requirement that no other libraries can do crypto operations.

aryan16:
> @aryan16 How can we support VM installation using this mode?

If VM is able to talk to K8s api server then no issues, if not we won't be using this feature instead use the existing GRPC call to istiod.

src/identity/customca.rs
aryan16: Done

src/identity/manager.rs
liminw: Can you point me to the code that does it?

aryan16: https://github.com/istio/ztunnel/blob/master/src/identity/manager.rs#L233
https://github.com/istio/ztunnel/blob/master/src/tls/boring.rs#L165

Cargo.toml
aryan16: Updated, PTAL

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+        let signer_name = &self.signer;
+        let mut usage = Vec::new();
+        usage.push("client auth".to_string());
+
+        let obj_metadata = k8s_openapi::apimachinery::pkg::apis::meta::v1::ObjectMeta{
+            name: Some(csr_name),
+            ..Default::default()
+        };
+        let csr_spec = CertificateSigningRequestSpec{
+            expiration_seconds:Some(86400),
```
JimmyCYJ, line 66: Can we use a variable to hold 86400, and leave a comment that we will replace the hardcoded value with a value read from config?

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+            Err(_) => return Err(Error::EmptyResponse(id.clone()))?
+        };
+
+        let final_certs = check_signed_cert(&csr, &stored_csr_name).await;
+        match final_certs {
+            Ok(cert_chain) => {
+                let cert_chain_vec = cert_chain.0;
+                let final_cert_chain = vec![str::from_utf8(&cert_chain_vec).unwrap()];
+                let leaf = final_cert_chain[0].as_bytes();
+                let final_chain = if final_cert_chain.len() > 1 {
```
JimmyCYJ, line 95: Maybe ca_cert_bundle is better than final_chain?

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+fn generate_csr_name() -> String{
+    let mut rng = thread_rng();
+    let rand_number: u16 = rng.gen();
+    let rand_string = rand_number.to_string();
+    let base_string = "csr-ztunnel-";
+    format!("{}{}", base_string, rand_string)
+}
+
+async fn check_signed_cert(csr: &Api<CertificateSigningRequest>, csr_name: &String) -> Result<k8s_openapi::ByteString, anyhow::Error> {
+    let lp = ListParams::default().fields(&format!("{}{}", "metadata.name=", csr_name)).timeout(30);
```
JimmyCYJ, line 121: Is there any criteria about setting this timeout? What is the unit of this timeout, 30 seconds?

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+    let mut stream = csr.watch(&lp, "0").await?.boxed();
+    let res :std::option::Option<ByteString>;
+    while let Some(status) = stream.try_next().await? {
+        match status {
+            WatchEvent::Added(o) => {
+                println!("Added {}", o.metadata.name.unwrap_or_default());
```
JimmyCYJ, line 127: Does the metadata.name contain any information about which workload the CSR is created for?

src/identity/customca.rs
aryan16: Cache is handling that case already 
https://github.com/istio/ztunnel/blob/master/src/identity/manager.rs#L233
https://github.com/istio/ztunnel/blob/master/src/tls/boring.rs#L165

src/app.rs
```diff
@@ -130,11 +130,18 @@ pub async fn build_with_cert(
     })
 }
 
-pub async fn build(config: config::Config) -> anyhow::Result<Bound> {
+pub async fn build(config: config::Config) -> anyhow::Result<Bound, anyhow::Error> {
     if config.fake_ca {
         let cert_manager = identity::mock::new_secret_manager(Duration::from_secs(86400));
         build_with_cert(config, cert_manager).await
-    } else {
+    } else if config.custom_signer.is_some() {
```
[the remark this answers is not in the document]
aryan16: you mean the custom signer? No we are not validating the custom signer, do we need it?

src/identity/customca.rs
aryan16: sure, will do it. Thanks

src/identity/customca.rs
aryan16: nope its the CSR name which is completely random.

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+    pub fn new(signer: Option<std::string::String>, client: Client) -> Result<CustomSigner, Error> {
+        Ok(CustomSigner { signer, client})
+    }
+}
+
+#[async_trait]
+impl CertificateProvider for CustomSigner {
+    #[instrument(skip_all)]
+    async fn fetch_certificate(&self, id: &Identity) -> Result<tls::Certs, Error> {
+        let client = self.client.clone();
```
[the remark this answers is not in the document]
aryan16: When you call the clone() method on a kube::Client instance, Rust creates a new instance that is a copy of the original instance. However, the new instance shares the same connection to the Kubernetes API server as the original instance.

src/identity/customca.rs
aryan16: sure

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+            Err(_) => return Err(Error::EmptyResponse(id.clone()))?,
+        }
+    }
+}
+
+fn generate_csr_name() -> String{
+    let mut rng = thread_rng();
+    let rand_number: u16 = rng.gen();
+    let rand_string = rand_number.to_string();
+    let base_string = "csr-ztunnel-";
```
shankgan, line 116: would it make sense to include the service-account name in the csr request instead of keeping it completely random? We can update this in a future change if needed

src/config.rs
```diff
@@ -224,8 +230,10 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
         // TODO: full FindRootCAForXDS logic like in Istio
         xds_root_cert: RootCert::File("./var/run/secrets/istio/root-cert.pem".parse().unwrap()),
         ca_address,
+        custom_signer,
         // TODO: full FindRootCAForCA logic like in Istio
         ca_root_cert: RootCert::File("./var/run/secrets/istio/root-cert.pem".parse().unwrap()),
+        k8s_ca_crt: RootCert::File("./var/run/secrets/kubernetes.io/serviceaccount/ca.crt".parse().unwrap()),
```
shankgan, line 236: Is there a better place from where this can be picked up? Are we sure that the service account token corresponding to ztunnel will always be mounted here>

src/identity/customca.rs
aryan16: It's 30 sec for now, I am planning to make it to 60 sec. Any suggestions ?

src/config.rs
```diff
@@ -34,6 +34,7 @@ const XDS_ADDRESS: &str = "XDS_ADDRESS";
 const CA_ADDRESS: &str = "CA_ADDRESS";
 const TERMINATION_GRACE_PERIOD: &str = "TERMINATION_GRACE_PERIOD";
 const FAKE_CA: &str = "FAKE_CA";
+const CUSTOM_SIGNER: &str = "CUSTOM_SIGNER";
```
JimmyCYJ, line 37: As we are using a k8s client when custom signer is set, maybe rename as K8S_CUSTOM_SIGNER?

aryan16: sure, will update this one too.

src/config.rs
aryan16: yeah it's gonna be there whenever ztunnel will be used on K8s

src/identity/customca.rs
aryan16: sure, will add this Identity as well - https://github.com/istio/ztunnel/blob/master/src/identity/manager.rs#L35

src/identity/manager.rs
```diff
@@ -414,6 +415,19 @@ impl SecretManager<CaClient> {
     }
 }
 
+impl SecretManager<CustomSigner> {
+    pub async fn origin(cfg: crate::config::Config) -> Result<Self, anyhow::Error> {
```
howardjohn, line 419: what does "origin" mean?

src/identity/manager.rs
```diff
@@ -414,6 +415,19 @@ impl SecretManager<CaClient> {
...
+        let k8s_client = tls::create_k8s_client(cfg.k8s_ca_crt).await;
+        match k8s_client {
+            Ok(client) => {
+                let custom_signer = CustomSigner::new(Some("kubernetes.io/kube-apiserver-client".to_string()), client)?;
```
howardjohn, line 423: shouldn't this be cfg.signer? instead of apiserver-client

src/app.rs
```diff
@@ -130,11 +130,18 @@ pub async fn build_with_cert(
...
+        let cert_manager = identity::SecretManager::origin(config.clone()).await;
+        match cert_manager {
+            Ok(cert_mgr) => build_with_cert(config, cert_mgr).await,
+            Err(_) => Err(anyhow::Error::msg("Signer is unable to sign the certificate")),
```
howardjohn, line 141: Its not signing any certificate, its just building the client(?)

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+            san: id.to_string(),
+        }
+        .generate()?;
+        let cert: Vec<u8> = cs.csr;
+        let pkey = cs.pkey;
+        let csr_name = generate_csr_name();
+        let stored_csr_name = csr_name.clone();
+        let signer_name = &self.signer;
+        let mut usage = Vec::new();
+        usage.push("client auth".to_string());
```
howardjohn, line 59: can just do `vec!["client auth".to_string()]` and avoid mut

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+use crate::identity::Error;
+use crate::tls;
+use futures::{StreamExt, TryStreamExt};
+use k8s_openapi::{api::certificates::v1::{CertificateSigningRequest, CertificateSigningRequestSpec}, ByteString};
+use std::str;
+use rand::prelude::*;
+
+#[derive(Clone)]
+pub struct CustomSigner {
+    pub signer: Option<std::string::String>,
```
howardjohn, line 32:
```suggestion
    pub signer: String,
```

Not sure we need option since its required?

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+            metadata: obj_metadata,
+            spec: csr_spec,
+            ..Default::default()
+        };
+
+        // creating CSR
+        match csr.create(&pp, &csr_req).await {
+            Ok(o) => {
+                let name = o.metadata.name.unwrap_or_default();
+                assert_eq!(csr_req.metadata.name.unwrap_or_default(), name);
```
howardjohn, line 83: Do we need this assert? If we do, can we make it a debug_assert so it doesn't panic in prod

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+            }
+            Err(kube::Error::Api(ae)) => assert_eq!(ae.code, 409),
```
howardjohn, line 85:
```suggestion
            Err(kube::Error::Api(ae)) => assert_eq!(ae.code, 409),
```

This should go away, it can fail for any other reason as well. This would panic

src/identity/customca.rs
howardjohn: I don't think we need to generate a random string manually, you can set generateName in API Create request instead of Name and have k8s do it for you. Thati s how Pod works, for example

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+    let mut rng = thread_rng();
+    let rand_number: u16 = rng.gen();
+    let rand_string = rand_number.to_string();
+    let base_string = "csr-ztunnel-";
+    format!("{}{}", base_string, rand_string)
+}
+
+async fn check_signed_cert(csr: &Api<CertificateSigningRequest>, csr_name: &String) -> Result<k8s_openapi::ByteString, anyhow::Error> {
+    let lp = ListParams::default().fields(&format!("{}{}", "metadata.name=", csr_name)).timeout(30);
+    let mut stream = csr.watch(&lp, "0").await?.boxed();
```
howardjohn, line 122: just watch has a race condition; if it was finished before we start the watch it will break. Need a List+Watch

I would overall prefer a long lived watched on a label we set on each CSR but ok with list+watch approach like we do in chiron/utils.go in the short term. Note you need to put ResourceVersion as well.

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+    let res :std::option::Option<ByteString>;
+    while let Some(status) = stream.try_next().await? {
+        match status {
+            WatchEvent::Added(o) => {
+                println!("Added {}", o.metadata.name.unwrap_or_default());
+            }
+            WatchEvent::Modified(o) => {
+                let s = o.status.as_ref().expect("status exists on csr");
+                if s.certificate.is_some() {
```
howardjohn, line 131: can do `if let Some(cert) = s.certificate { ... }`. Then you don't need to unwrap later.

src/identity/customca.rs
```diff
@@ -0,0 +1,141 @@
...
+    let lp = ListParams::default().fields(&format!("{}{}", "metadata.name=", csr_name)).timeout(30);
+    let mut stream = csr.watch(&lp, "0").await?.boxed();
+    let res :std::option::Option<ByteString>;
+    while let Some(status) = stream.try_next().await? {
+        match status {
+            WatchEvent::Added(o) => {
+                println!("Added {}", o.metadata.name.unwrap_or_default());
+            }
+            WatchEvent::Modified(o) => {
+                let s = o.status.as_ref().expect("status exists on csr");
```
howardjohn, line 130: `expect` will panic if its not true; we should not let a bad API server response panic ztunnel

src/tls/boring.rs
```diff
@@ -238,6 +240,37 @@ pub fn grpc_connector(uri: String, root_cert: RootCert) -> Result<TlsGrpcChannel
...
+            conn.cert_store_mut()
+                .add_cert(x509::X509::from_pem(&b).map_err(Error::InvalidRootCert)?)
+                .map_err(Error::InvalidRootCert)?;
+        }
+        RootCert::Default => {} // Already configured to use system root certs
+    }
+    let mut http = hyper::client::HttpConnector::new();
+    http.enforce_http(false);
+    let https = hyper_boring::HttpsConnector::with_connector(http, conn)?;
+    let hyper_client = hyper::Client::builder().http2_only(true).build::<_,hyper::Body>(https);
```
howardjohn, line 264: Not sure we need http2 only for k8s connection

src/identity/manager.rs
aryan16: my bad, forgot to remove the testing code.

src/identity/customca.rs
aryan16: we need the resource name while watching the CSR, not sure if we let K8s to create the random name how will we watch the CSR ?

howardjohn: the Create call returns the resulting name

aryan16: got it, thanks

src/identity/manager.rs
aryan16: Associated function - https://doc.rust-lang.org/rust-by-example/fn/methods.html.

howardjohn: I know it is a function, I just do not get the name?

src/tls/boring.rs
aryan16: updated

src/identity/customca.rs
aryan16: updated, PTAL

src/identity/customca.rs
aryan16: Updated, PTAL

src/identity/manager.rs
aryan16: oh I saw the same name at many places in rust example codes as a replacement of `new`. Do you think I should update it?

costinm:
What's next ?  The DNS proxy ( which is REQUIRED if the underlying network is not secure ) ? Integration with the external SDS provider using UDS ?  

Can we please add the agent and just use it ? No need for ztunnel to duplicate all the integrations and do everything ( in a slightly different way ). 

A design doc on such changes with focus on security will also be greatly appreciated:
 - who approves the CSRs and on what criteria ? If Istiod, why not use the existing code that also generates the CSR
 - VMs - how does it change the install and security requirements ?  We didn't use k8s signing from VMs because 
 it required VMs to have K8S SA and access - which means higher escalation risks and permissions.  
 - what are the benefits compared to using the current method and letting Istiod generate the k8s sign request ? 

But even if we conclude direct integration with k8s from the proxy is best ( and I can be convinced it's a more
standard protocol than our gRPC call to Istiod ) - we should still do it in the agent, so both sidecars and ambient
are consistent and use whatever is best.

costinm:
On a related note - ztunnel should eventually match the integrations we have, including SDS and CSI-based providers ( in both dedicated mode and in future if we implement delegation ). 

I saw the FAKE provider - IMO it should be replaced (or we should add) with a mode that reads from /var/run/secrets... ( same default as istio-agent ), and fallback to SDS using the well-known path that the Envoy is using ( for Spire and all other integrations ). 
And once SDS is added - we can maybe remove the call to Istiod and keep ztunnel focused on proxying, not implementing multiple CA protocols, just XDS and files.

src/identity/manager.rs
```diff
@@ -414,6 +415,20 @@ impl SecretManager<CaClient> {
     }
 }
 
+impl SecretManager<CustomSigner> {
+    pub async fn origin(cfg: crate::config::Config) -> Result<Self, anyhow::Error> {
+        let k8s_client = tls::create_k8s_client(cfg.k8s_ca_crt).await;
+        match k8s_client {
```
costinm, line 421: This won't work on VMs. 

Pretty sure there is a way to initialize k8s client in rust that works with both 'in cluster' and kubeconfig. 

Same for the tokens - outside of K8S you need all the auth methods, the VM may use MDS or external commands to authenticate ( gcloud, etc ).

src/tls/boring.rs
```diff
@@ -238,6 +240,37 @@ pub fn grpc_connector(uri: String, root_cert: RootCert) -> Result<TlsGrpcChannel
     Ok(TlsGrpcChannel { uri, client: hyper })
 }
 
+// creating K8s client with boring
+pub async fn create_k8s_client(root_cert: RootCert) -> Result<kube::Client, anyhow::Error> {
+    let mut conn = ssl::SslConnector::builder(ssl::SslMethod::tls_client())?;
+    conn.set_verify(ssl::SslVerifyMode::PEER);
+    conn.set_alpn_protos(Alpn::H2.encode())?;
```
costinm, line 247: Is this the only SSL connection we create in ztunnel ? Any chance we could use the same code - or client library code ?

Issue #357, opened 2023-01-26
Need to actually support `arm64` builds in a way that is compatible with CI


bleggett:
- We aren't validating `fips`-less builds in CI
- We can't build `arm64` (or any other target Rust might otherwise support just fine) with FIPS
- FIPS is usually not relevant for contributors/local dev
- Contributors can't disable FIPS without local hacks.
- This leads to inevitable build entropy for supported build pathways not gated by CI, see https://github.com/istio/tools/pull/2344
- If `fips` is always going to be optional/have very specific and unique environmental requirements (which will always be the case - even upstream `boring` Rust lib has FIPS as an _opt-in_ flag, the inverse of what we have) - we should make sure our CI can build with and without it.

While we must build with and without `fips` for every CI build, the decision to make the `fips` feature a `default` feature complicates the ability to easily perform automated builds with and without `fips` - because `cargo` doesn't support negating features individually, they either must be optional, or they must be defaults - and the only way to disable defaults is to disable _all_ default features and selectively re-enable them, which sucks (both for forward compat and usability, see current solution), and makes doing CI validation of fips on/off impossible without locally editing files, see current best attempt https://github.com/istio/ztunnel/pull/139

Suggestion:

1. mimic upstream `boring` Rust lib and make `fips` a non-default feature
2. Change CI to build 2x ->
 - `BORING_BSSL_PATH="vendor/boringssl-fips/linux_x86_64" BORING_BSSL_INCLUDE_PATH="vendor/boringssl-fips/linux_x86_64/include/" cargo build --features fips` 
 - `cargo build`
 
 This would remove the current obstacles to simple automated builds of both `fips` and non-`fips` variants, and is more obvious/normative of an approach.
 
 
 EDIT: Changing this to `we should be able to build on more than one arch in a way that is actually automatable and doesn't create a lot of work for ourselves` - that's the core problem.

howardjohn:
non-fips was added exclusively to workaround a short term limitation (https://github.com/istio/ztunnel/issues/35) (note the issue was closed, should probably be opened as "support fips+arm64"). Do you have a use case for non-fips long term? on x86?

bleggett:
As someone who has worked at a place where FIPS compliance mattered - 

[...]

PR #388, opened 2023-02-09
Add vendored `linux/arm64` FIPS binaries

bleggett:
I don't think vendoring our own FIPS binaries is a great long term solution, but this should at least achieve parity with the current `x86` approach and let us build FIPS `ztunnel` binaries for the two most common architectures contributors are likely to have and that we will ultimately need to generate containerized artifacts for. Also updated README accordingly.

- I still think we probably should not be defining `BORING_BSSL` env vars in `.cargo/config.toml` - instead we should expect people to pass them as prefix args to `cargo build/run`.
- If we want to avoid that, we will need to add dynamic platform-based vendored path selection logic to `build.rs` in https://github.com/cloudflare/boring  - because of how their `build.rs` currently works, we cannot do that dynamic selection in our own `build.rs` - I've tried. Arguably that is something we should fix in upstream.
- We still actually need to do arm64 FIPS CI builds (https://github.com/istio/ztunnel/issues/357), and until the above are resolved we cannot.

Some of the hassle around letting https://github.com/cloudflare/boring auto-build `boringssl` itself should get a bit easier once the upstream PRs from @nmittler and @PiotrSikora are merged.


README.md
```diff
@@ -30,15 +33,45 @@ to ensure that compliance is granted only to the exact binary tested.
...
+We vendor FIPS boringssl binaries for 
+- `linux/x86_64`
+- `linux/arm64`
+
+To use these vendored libraries and build ztunnel for either of these OS/arch combos, for the moment you must manually edit
+[.cargo/config.toml](.cargo/config.toml) and change the values of BORING_BSSL_PATH and BORING_BSSL_INCLUDE_PATH under the `[env]` key to match the path to the vendored libraries for your platform, e.g:
+
+#### For linux/x86_64
+``` toml
+BORING_BSSL_PATH = { value = "vendor/boringssl-fips/linux_x86_64", force = true, relative = true }
```
howardjohn, line 48: Do you think its feasible to get the boring crate to allow some form of substition so we can just put `vendor/boringssl-fips/{platform}` or whatever?

README.md
```diff
@@ -30,15 +32,45 @@ to ensure that compliance is granted only to the exact binary tested.
...
+BORING_BSSL_INCLUDE_PATH = { value = "vendor/boringssl-fips/linux_x86_64/include/", force = true, relative = true }
+```
+
+#### For linux/arm64
+
+``` toml
+BORING_BSSL_PATH = { value = "vendor/boringssl-fips/linux_arm64", force = true, relative = true }
+BORING_BSSL_INCLUDE_PATH = { value = "vendor/boringssl-fips/linux_arm64/include/", force = true, relative = true }
```
howardjohn, line 56: nit: is the include/ arch specific or can we share it?

howardjohn (commented):
LGTM! does cross compile work?

bleggett:
> LGTM! does cross compile work?

Not tried it - just verified that a FIPS arm64 build works within `make shell`. If that works, CC ought to.

README.md
bleggett: Possibly/maybe - I think once the PRs @PiotrSikora and @nmittler already have outstanding in that repo get merged, that's something I'll look at if no one beats me to it.

README.md
bleggett: Yeah I thought about that, I think so, lemme give that a go.

howardjohn: I put a hold in case you want to do this but approved, feel free to remove

howardjohn (approved):

howardjohn:
cross compile failing in linking for me, but I may have done something wrong

howardjohn:
actually I guess you need to manually st the linker: CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc cargo build --target=aarch64-unknown-linux-gnu

README.md
bleggett: Yep - consolidated, ty for asking.

bleggett:
> actually I guess you need to manually st the linker: CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc cargo build --target=aarch64-unknown-linux-gnu

Generally speaking I find multiarch `buildx`  targets way easier to get working than CC - good to know.

howardjohn:
> > actually I guess you need to manually st the linker: CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc cargo build --target=aarch64-unknown-linux-gnu
> 
> Generally speaking I find multiarch `buildx` targets way easier to get working reliably/reproducibly than CC - good to know.

Good point, wasn't sure how slow it was.. In envoy we ended up using dedicated arm64 nodes because envoy takes forever to build. Here we have more flexibility, can cross compile, qemu, or use dedicated nodes - whatever is easy/fast

PR #392, opened 2023-02-13
make netns test-only dep

stevenctl:
This is preventing `cargo build` on OSX (using the instructions in the README) due to some unresolved imports from `nix` within `netns-rs`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -512,7 +512,7 @@ dependencies = [
  "autocfg",
  "cfg-if",
  "crossbeam-utils",
- "memoffset 0.7.1",
+ "memoffset",
  "scopeguard",
 ]
 
@@ -1243,15 +1243,6 @@ dependencies = [
 
 [[package]]
 name = "memoffset"
-version = "0.6.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5aa361d4faea93603064a027415f07bd8e1d5c88c9fbf68bf56a285428fd79ce"
-dependencies = [
- "autocfg",
-]
-
-[[package]]
-name = "memoffset"
 version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5de893c32cde5f383baa4c04c5d6dbdd735cfd4a794b0debdb2bb1b421da5ff4"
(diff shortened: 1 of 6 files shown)


howardjohn (dismissed):

stevenctl:
`[cfg(test)]` in `tests/` does nothing...

I think maybe the lib is built separately from the tests so we first built it with the test_helpers excluded :/

stevenctl:
> Cargo will build your library in its normal (non-test) configuration, then build each integration test/benchmark as a sepate binary crate which links to that library. So library code marked with #[cfg(test)] is not available in integration tests/benches.

https://users.rust-lang.org/t/what-are-the-rules-for-cfg-test/54122

Not sure the best way to deal with this. For now just gonna get `cargo build` to work on mac.

howardjohn:
I think we can put integ-only test helpers under test/ possibly (you can make it in a different file if you put it in a folder IIRC. Then we only need test_helpers for things used in unit tests

Or multi crates, main one imports wth cfg=test, integ tests import as normal?

stevenctl:
I believe multi-crates can only have 1 lib crate and N binary crates. 
Workspaces would work around this.. 

 `os = linux` won't work because cargo build attempts to build all the deps upfront. Could make it a cfg=linux specific dependency but the number of guards is getting large, and that's just for one lib. 
 
 I'm going to try CLion's remote development with our build-tools image and a couple of other approaches like devcontainer. 
 
 My goal was just to make sure my IDE functions well enough. I'll need linux to actually run anything anyway.

bleggett:
`My goal was just to make sure my IDE functions well enough. I'll need linux to actually run anything anyway.`

Yeah this is kinda why I'm OK (as a Mac user) with containerized Linux builds - as long as I can build on Linux/$arch I can make do with a container build image, and I prefer those anyway because they're a fixed buildenv.

Also I don't think Istio proper is ever gonna need `ztunnel` to build for macOS - not to say it can't but I think the goal is to support the *nixes and arches Istio ships containers for already as a baseline.


IMO 

- if `ztunnel` can't build for $rust-supported-target-arch, that's a bug/hygiene problem on our end - we shouldn't have arch specific code or assumptions in ztunnel. 
- if `ztunnel` can't build for $rust-supported-target-os then that's a bigger support ask, and case-by-case.

stevenctl:
Agreed. All the code navigation I need works fine even though cargo build doesn't (outside of docker). 
I do think we should still try to prune this dep from the main build though.

Issue #394, opened 2023-02-14
Publish `arm64` binaries


howardjohn:
We should publish these in CI. See https://github.com/istio/ztunnel/issues/357

PR #395, opened 2023-02-14
Publish arm64 binary

howardjohn:
This isn't ideal but we need to start publishing at least amd64, and my shortcut to do this (https://github.com/istio/istio/pull/43331) doesn't work without a lot of effort. For now we can just hack it.

CI job coming soon: https://github.com/istio/test-infra/pull/4570

Fixes https://github.com/istio/ztunnel/issues/394

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,5 @@
 [build]
 target-dir = "out/rust"
 [env]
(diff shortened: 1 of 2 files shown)


bleggett (approved):
Good enough for now, yeah

bleggett:
Whups - should have looked at the whole script, need this before the `cargo build --release`

bleggett:
Raised #399 to track the hack

PR #397, opened 2023-02-14
Fix version metric to report Istio version

howardjohn:
This aligns with istio/proxy impl; users care about Istio version, not ztunnel version

For https://github.com/istio/istio/issues/43249


bleggett (approved):

ymesika (approved):
lgtm

PR #401, opened 2023-02-16
check expected grpc error

### LINKED ISSUE #400: Seeing warnings in ztunnel logs that could cause unnecessary concerns
Happening every 30 mins:

2023-02-15T20:54:40.204428Z  WARN xds{id=1}: ztunnel::xds::client: XDS client error: gRPC error (Unknown error): error reading a body from connection: broken pipe, retrying

2023-02-15T21:24:40.813192Z  WARN xds{id=2}: ztunnel::xds::client: XDS client error: gRPC error (Unknown error): error reading a body from connection: broken pipe, retrying

@howardjohn mentioned this is ztunnel disconnects with istiod every 30 mins so istiod can rebalance and we hide this with Envoy.

hzxuzhonghu:
fix #400

### DIFF
--- a/src/xds/client.rs
+++ b/src/xds/client.rs
@@ -171,7 +171,6 @@ impl Config {
         }
     }
 }
-pub fn is_expected_grpc_error() {}
 pub struct AdsClient {
     config: Config,
     /// Stores all known workload resources. Map from type_url to name
--- a/src/xds/client.rs
+++ b/src/xds/client.rs
@@ -19,7 +19,6 @@ use std::sync::Arc;
 use std::time::Duration;
 use std::{fmt, mem};
 
-use hyper::http::status;
 use prost::{DecodeError, EncodeError};
 use prost_types::value::Kind;
 use prost_types::{Struct, Value};
--- a/src/xds/client.rs
+++ b/src/xds/client.rs
@@ -25,7 +25,6 @@ use prost_types::{Struct, Value};
 use thiserror::Error;
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
-use tonic::Code::DeadlineExceeded;
 use tracing::{debug, error, info, info_span, warn, Instrument};
 
 use crate::config::RootCert;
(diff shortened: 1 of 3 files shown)


bleggett:
This looks like it fixes https://github.com/istio/ztunnel/issues/400?

bleggett (commented):
Admittedly it's a small change so if you don't want to that's fine - but since this is sort of a behavioral quirk, describing/fencing the quirk with a unit test would be good.

bleggett (approved):

hzxuzhonghu:
>@hzxuzhonghu This looks like it fixes https://github.com/istio/ztunnel/issues/400?

Yes

ymesika (approved):
LGTM

src/xds/client.rs
```diff
@@ -266,6 +265,26 @@ impl AdsClient {
                 tokio::time::sleep(backoff).await;
                 backoff
             }
+            Err(e @ Error::GrpcStatus(_)) => {
+                let err_detail = e.to_string();
+                let status: tonic::Status = e.try_into().unwrap();
```
howardjohn, line 270: unwrap is not needed here, we can do:
```
            Err(ref e @ Error::GrpcStatus(ref status)) => {
                let err_detail = e.to_string();
```

and remove the TryInto impl above

src/xds/client.rs
```diff
@@ -266,6 +265,26 @@ impl AdsClient {
...
+                if status.code() == tonic::Code::Unknown
```
howardjohn, line 271: why Unknown? we don't have that in the Go version and it seems important

src/xds/client.rs
```diff
@@ -266,6 +265,26 @@ impl AdsClient {
...
+                        && status.message().contains("transport is closing"))
+                {
+                    debug!("XDS client terminated: {}, retrying", err_detail);
+                    self.metrics
+                        .increment(&ConnectionTerminationReason::Reconnect);
+                } else {
+                    warn!("XDS client error: {}, retrying", err_detail);
+                    self.metrics.increment(&ConnectionTerminationReason::Error);
+                }
+                // For gRPC errors, we connect immediately
```
howardjohn, line 284: Its not clear to me why we don't want backoff here but we do for Connection. Does envoy do the same?

hzxuzhonghu: This is just same as previous, i do not change it.  

@ramaraochavali Do you know envoy's behavior

ramaraochavali: Envoy there is a back off https://github.com/envoyproxy/envoy/blob/a4053922d7311e11c18e764ff5e5bfa8f71cd334/source/common/config/grpc_stream.h#L21 and there are efforts to make it configurable https://github.com/envoyproxy/envoy/pull/24701

src/xds/client.rs
hzxuzhonghu: https://github.com/istio/ztunnel/issues/400  unknown error here, not sure the error in  rust is same as go 

```
2023-02-15T20:54:40.204428Z WARN xds{id=1}: ztunnel::xds::client: XDS client error: gRPC error (Unknown error): error reading a body from connection: broken pipe, retrying```

src/xds/client.rs
hzxuzhonghu: Thanks， suggest open a new issue to tack that

hzxuzhonghu: https://github.com/istio/ztunnel/issues/413

src/xds/client.rs
hzxuzhonghu: good , i didn't know ref usage

howardjohn (approved):

PR #402, opened 2023-02-16
tips for faster dev on macs

stevenctl:
Maybe this should go in the wiki instead of README. This seems like the best workaround that lets me edit code natively and build/test in docker.


bleggett:
Yeah - the wiki is a better spot - we have Mac build instructions there: https://github.com/istio/istio/wiki/Preparing-for-Development-Mac


...but they are oudated now and don't cover ARM Mac quirks. I don't have edit perms - @linsun does.

stevenctl:
I do as well. Just going to throw them there. Almost nothing in the wiki has any info about ztunnel (yet).

Issue #334, opened 2023-01-12
CA: implement prioritization/rate limitting


howardjohn:
Depends on https://github.com/istio/ztunnel/pull/298

Currently we have 3 sources of CA calls:
1. On demand: we have a request incoming and need a cert (right now!). 
2. Background refresh: we have a cert and its near expiration, we need to refresh. By default this is 24hr lifetime, start refresh at 12hr
3. Prewarming. New workload (or ztunnel just started), we want to preload the cert to reduce latency on first call (avoid "cold starts")

In order of important, this likely looks like: On Demand >>>> Background refresh when _really close_ to expiration > Prewarming == Background refresh. Could be simplified to just "on demand is top priority".

Additionally, CA requests are expensive. 255 concurrent requests to 1 istiod would almost certainly overwhelm it (in CPU cost); other CAs may have different constraints. We are also not the only client. We need a sensible strategy that trades off not killing the CA with getting all the certs when we want.

At the very least, we should have a way to prioritize on demand requests. This could be something simple like adding some delay in prewarming, or have some priority queue.

bleggett:
Dumb q - do we (or do we intend to) support the SDS protocol so that ztunnel could play nice with a compatible replacement workload CA/SDS, like the rest of Istio does today?

howardjohn:
I started https://github.com/istio/ztunnel/issues/251 to define these integration interfaces. I hadn't put SDS in there but its plausible.

I will say in general the idea has been to have a few common integration points. For example, instead of 10 telemetry providers just use OTEL, which itself supports the kitchen sink of providers.

There is less of a standard around CAs compared to telemetry though.

keithmattix:
I've seen some semblance of this in the code; has this been implemented?

howardjohn:
yeah I think this is done. thanks!

PR #406, opened 2023-02-17
Clean up error handling in Secret Manager

qfel:
Clean up the implementation of #334 

Converts errors signaling background worker death to panics. This is generally preferable since SecretManager will not recover from them: all new certificate fetches will fail and no refreshes will be performed ever again.

They are also programming errors: it should not be possible to ever get into that state, save for bugs in SecretManager code (either due to early return in the worker or a panic). While we could try to restart the worker in such circumstances, that seems dangerous as a dead worker signals something is very wrong - better to propagate and panic.

### DIFF
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -355,36 +355,28 @@ impl<T: CertificateProvider> SecretManager<T> {
                 drop(certs);
                 // Notify the background worker to start refreshing the certificate.
                 match self.requests.send((id.clone(), pri)).await {
-                    // Note that this should not happen, the background worker is expected to be
-                    // running.
-                    //
-                    // TODO: Better error propagation.
-                    Err(_) => Err(Error::EmptyResponse(id.clone())),
+                    Err(e) => unreachable!("SecretManager worker died: {e}"),
                     Ok(()) => Ok(rx),
                 }
             }
         }
     }
 
-    async fn wait(
-        &self,
-        id: &Identity,
-        mut rx: watch::Receiver<CertState>,
-    ) -> Result<tls::Certs, Error> {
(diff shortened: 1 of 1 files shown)


hzxuzhonghu (changes requested):
It should return error instead of panic

qfel:
> It should return error instead of panic

vs

> Given we cannot recover from errors, this feels a bit like mutex poisoning where it is common practice to panic if it occurs. I am ok with this (despite normally not liking panics) for this case

So, which one?

src/identity/manager.rs
```diff
@@ -355,36 +355,28 @@ impl<T: CertificateProvider> SecretManager<T> {
                 drop(certs);
                 // Notify the background worker to start refreshing the certificate.
                 match self.requests.send((id.clone(), pri)).await {
-                    // Note that this should not happen, the background worker is expected to be
-                    // running.
-                    //
-                    // TODO: Better error propagation.
-                    Err(_) => Err(Error::EmptyResponse(id.clone())),
+                    Err(e) => unreachable!("SecretManager worker died: {e}"),
```
hzxuzhonghu, line 358: can you explain why self.requests.send cannot be recovered?

qfel: That means the worker has died, so this and *all* further calls will fail with this error. Basically the SecretManager instance is useless. You can try to recover from it by re-creating the SecretManager, but it's just silencing a programming error in a `try..catch(anything)` fashion. Rust has good support for resource ownership and makes it hard to maintain global state so it's less of a risk, but I have seen this kind of approach cause deadlocks in other languages.

hzxuzhonghu: IC, this is requests send channel.

hzxuzhonghu (approved):

Issue #404, opened 2023-02-16
Issues with L4 metrics cross node


linsun:
Using source (sleep) and destination (hw-v1 and v2), I am not seeing any metrics from hw-v1's ztunnel that has hw-v1 as source and sleep as destination. I think this missing traffic causes Kiali only shows 1 direction for sleep + hw-v1.

```
kubectl exec -it $ZTUNNEL -n istio-system -- curl localhost:15020/metrics
Defaulted container "istio-proxy" out of: istio-proxy, istio-init (init)
# HELP istio_connection_terminations The total number of completed connections to xds server.
# TYPE istio_connection_terminations counter
istio_connection_terminations_total{reason="Error"} 2
# HELP istio_build Istio component build info.
# TYPE istio_build gauge
istio_build{component="ztunnel",tag="7af8f5d"} 1
# HELP istio_tcp_connections_opened The total number of TCP connections opened.
# TYPE istio_tcp_connections_opened counter
istio_tcp_connections_opened_total{reporter="destination",source_workload="helloworld-v2-same-node",source_canonical_service="helloworld",source_canonical_revision="v2",source_workload_namespace="default",source_principal="spiffe://cluster.local/ns/default/sa/helloworld",source_app="helloworld",source_version="v2",source_cluster="Kubernetes",destination_service="unknown",destination_service_namespace="unknown",destination_service_name="unknown",destination_workload="sleep",destination_canonical_service="sleep",destination_canonical_revision="latest",destination_workload_namespace="default",destination_principal="spiffe://cluster.local/ns/default/sa/sleep",destination_app="sleep",destination_version="latest",destination_cluster="Kubernetes",request_protocol="tcp",response_flags="-",connection_security_policy="mutual_tls"} 1
[...]

PR #407, opened 2023-02-20
Switch between source and destination workloads

ymesika:
As discussed in #404

### DIFF
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -185,8 +185,8 @@ impl OutboundConnection {
             let inbound_connection_metrics = traffic::ConnectionOpen {
                 reporter: Reporter::destination,
                 derived_source: None,
-                destination: Some(req.source.clone()),
-                source: req.destination_workload.clone(),
+                source: Some(req.source.clone()),
+                destination: req.destination_workload.clone(),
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
lgtm, we have tests for these metrics though, they should be updated

src/proxy/outbound.rs
```diff
@@ -185,8 +185,8 @@ impl OutboundConnection {
             let inbound_connection_metrics = traffic::ConnectionOpen {
                 reporter: Reporter::destination,
                 derived_source: None,
-                destination: Some(req.source.clone()),
-                source: req.destination_workload.clone(),
+                source: Some(req.source.clone()),
```
howardjohn, line 188: actually this was intentionally backwards... let me look into it again

ymesika: I noticed your naming and figured you did it intentionally but see our discussion in the issue.
I think that when you pass it from outbound to inbound on same node the source should stay the source.. Only reporter should change.

ymesika:
> lgtm, we have tests for these metrics though, they should be updated

Do we? How come they didn't fail?

src/proxy/outbound.rs
howardjohn: Ah got it

PR #408, opened 2023-02-21
Add istio meta prefix metadata to proxy metadata

hanxiaop:
Env like `ISTIO_META_ISTIO_VERSION` is not reported, which lead to ztunnel has no version displayed in istiod.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -21,6 +21,7 @@ use std::{env, fs};
 
 use anyhow::anyhow;
 use bytes::Bytes;
+use serde_json::Value;
 use tokio::time;
 
 use crate::identity;
@@ -310,7 +311,7 @@ fn construct_proxy_config(mc_path: &str, pc_env: Option<&str>) -> anyhow::Result
 
     let mut pc = [mesh_config, proxy_config_env]
         .into_iter()
-        .flatten() 
+        .flatten()
         .fold(ProxyConfig::default(), |pc, v| pc.merge(v));
 
     // only include ISTIO_META_ prefixed fields in this map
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -388,4 +403,34 @@ pub mod tests {
...
     }
+
+    #[test]
+    fn test_istio_env_map() {
+        // Set up some test environment variables
+        env::set_var("ISTIO_META_FOO", "bar");
+        env::set_var("ISTIO_META_HELLO", "world");
+        env::set_var("NOT_ISTIO_VAR", "not_istio_value");
+
+        // Get all environment variables that start with "ISTIO_META_"
```
howardjohn, line 414: This test isn't actually testing anything, because non of our own code is being called - we just copy+pasted it into the test

src/config.rs
```diff
@@ -319,6 +321,19 @@ fn construct_proxy_config(mc_path: &str, pc_env: Option<&str>) -> anyhow::Result
         .filter_map(|(k, v)| Some((k.strip_prefix("")?.to_string(), v)))
         .collect();
 
+    let istio_env_vars: Vec<(String, String)> = env::vars()
+        .filter(|(key, _)| key.starts_with(ISTIO_META_PREFIX))
+        .map(|(key, val)| {
+            (
+                key.trim_start_matches(ISTIO_META_PREFIX).to_lowercase(),
```
howardjohn, line 328: Why lowercase?

Also I think L321 needs the stripping as well (see envoy logic: https://github.com/istio/istio/blob/b5df9a51c0290df8fd2692e7578bbef62a315d42/pkg/bootstrap/config.go#L563)

src/config.rs
hanxiaop: Added the extra test in the existing test `config_from_proxyconfig`

src/config.rs
hanxiaop: Fixed. It should not be lowercased. Just added these code logic https://github.com/istio/istio/blob/b5df9a51c0290df8fd2692e7578bbef62a315d42/pkg/bootstrap/config.go#L567-L576. However, the metadata struct defined here is `HashMap<String, String>`, which is not equal to `map[string]any`, should I revise it and add more logics or leave it like this and do it later when needed?

howardjohn (approved):

PR #410, opened 2023-02-21
ztunnel/cert: include intermediate certs in the chain

shashankram:
Currently, intermediate certs are not included in
cert chain passed to the TLS peer. This change
ensures intermediate certs are added to the cert
chain. The root cert is a part of the chain and
is not included as it should be available on the
peer to verify the cert chain.

### DIFF
--- a/src/tls/boring.rs
+++ b/src/tls/boring.rs
@@ -286,7 +286,13 @@ impl Certs {
         // key and certs
         conn.set_private_key(&self.key)?;
         conn.set_certificate(&self.cert.x509)?;
-        for chain_cert in self.chain.iter() {
+        for (i, chain_cert) in self.chain.iter().enumerate() {
+            // Only include intermediate certs in the chain.
+            // The last cert is the root cert which should already exist on the peer.
+            if i < (self.chain.len() - 1) {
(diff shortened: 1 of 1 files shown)


stevenctl (commented):
Reading through the openssl docs, I don't really understand what the difference is between add_cert and cert_store.add_cert. Can you help clarify? 

Can you also describe the bad behavior this bug fixes. E.g. "a client without some intermediate will successfully make a connection it shouldn't"

shashankram:
> Reading through the openssl docs, I don't really understand what the difference is between add_cert and cert_store.add_cert. Can you help clarify?
> 
> Can you also describe the bad behavior this bug fixes. E.g. "a client without some intermediate will successfully make a connection it shouldn't"

The change I made adds the intermediate certs to the chain. Currently, intermediate certs are not passed along in the chain presented to the peer, so the peer is unable to verify leaf certs signed using intermediate certs that are derived from the same root. Based on https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_get_cert_store.html, adding a cert to the "store" merely adds the certificate to a trust store used for peer cert verification, it does not pass that cert along in the chain.

stevenctl (approved):

PR #417, opened 2023-02-24
Add XDS keepalives, like we have in Go

howardjohn:
Attempt to fix https://github.com/istio/istio/issues/43592. I couldn't reproduce it unfortunately so don't have a great way to test


bleggett (approved):
I get that this is exploratory - in the long term I don't have any good insight into what these numbers should be in terms of HTTP/2 best practice or whether they should be configurable or not.

howardjohn:
> I get that this is exploratory - in the long term I don't have any good insight into what these numbers should be in terms of HTTP/2 best practice or whether they should be configurable or not.

fwiw we use the same in Istio today. But I also don't know much beyond that

bleggett (approved):

PR #418, opened 2023-02-24
auditing clone usage

ilrudie:
- some clones removed
- some clones were changing borrowed to owned, replaced with .to_owned() to make this clear

I have some comments that may be more relevant to the reviewer than they are to future developers. If reviewer feels they've served their purpose after they read through I am happy to remove.

working on: #10

### DIFF
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -200,7 +200,7 @@ impl OutboundConnection {
                 InboundConnect::DirectPath(stream),
                 origin_src,
                 req.destination,
-                self.pi.metrics.clone(),
+                self.pi.metrics.to_owned(), // self is a borrow so this clone is to return an owned
                 connection_metrics,
                 Some(inbound_connection_metrics),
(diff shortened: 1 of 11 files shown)


src/proxy/outbound.rs
```diff
@@ -372,7 +372,8 @@ impl OutboundConnection {
         }
         // For case source client and upstream server are on the same node
         if !us.workload.node.is_empty()
-            && self.pi.cfg.local_node == Some(us.workload.node.clone())
+            && !self.pi.cfg.local_node.is_none()
```
howardjohn, line 375: do we need this?

howardjohn (approved):

src/proxy/outbound.rs
ilrudie: No. Good catch. I was planning to unwrap local_node and wanted to ensure we didn't panic but it turns out that isn't necessary. Fix coming soon.

PR #419, opened 2023-02-24
remove DEST var check

jacob-delgado:


scripts/release.sh
```diff
@@ -36,7 +36,7 @@ SHA="$(git rev-parse --verify HEAD)"
 RELEASE_NAME="ztunnel-${SHA}-${ARCH}"
 ls -lh "${WD}/../out/rust/release/ztunnel"
 DEST="${DEST:-gs://istio-build/ztunnel}"
-if [[ "$CI" == "" && "$DEST" == "gs://istio-build/ztunnel" ]]; then
+if [[ "$CI" == "" ]]; then
   echo "Outside of CI, DEST must be explicitly set"
```
howardjohn, line 40: This just fails anytime CI is not set though?

The intent was to avoid anyone but CI pushing to `gs://istio-build/ztunnel` accidentally

PR #431, opened 2023-03-05
Set the baggage cluster ID from the env

ymesika:
Setting the cluster ID in the baggage header to the value of `CLUSTER_ID` env variable for the Waypoint to use as the source cluster on emitted metrics.

`CLUSTER_ID` already exists in ztunnel daemonset.

This is *not* setting the source cluster label on metrics generated by the source/destination ztunnel.


hzxuzhonghu:
LGTM

howardjohn (approved):

PR #436, opened 2023-03-07
rm meaningless ready task

hzxuzhonghu:


src/app.rs
```diff
@@ -42,7 +42,6 @@ pub async fn build_with_cert(
     let (drain_tx, drain_rx) = drain::channel();
 
     let ready = readiness::Ready::new();
-    let proxy_task = ready.register_task("proxy listeners");
```
hzxuzhonghu, line 45: L50 `workload manager` has registered

howardjohn: This is for the proxy listeners are bound which hapens at L75

howardjohn (commented):
Its not clear to me why we do not need these

src/app.rs
hzxuzhonghu: not sure i understand

hzxuzhonghu: I think there is only one task waiting for workloads discovery

howardjohn: This task isn't waiting for workload discovery though, its waiting until we bind the listeners to ports (15008, 15001, etc)

hzxuzhonghu: Not make much sense to me, if they bind failed, the process will exit.

howardjohn: but we don't want to mark as ready before then

hzxuzhonghu: Understand what you mean, will keep this

howardjohn (approved):

Issue #435, opened 2023-03-06
Cluster trust domain is hardcoded to `cluster.local`


bleggett:
See: https://github.com/istio/istio/issues/43673

cc @SirNexus

bleggett:
also @kdorosh

PR #437, opened 2023-03-07
Support trust domain

hzxuzhonghu:
Fix https://github.com/istio/ztunnel/issues/435

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -94,6 +94,8 @@ pub struct Workload {
     #[serde(default)]
     pub namespace: String,
     #[serde(default)]
+    pub trust_domain: String,
+    #[serde(default)]
     pub service_account: String,
 
     #[serde(default)]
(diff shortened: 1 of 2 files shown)


src/workload.rs
```diff
@@ -94,6 +94,8 @@ pub struct Workload {
     #[serde(default)]
     pub namespace: String,
     #[serde(default)]
+    pub trust_domain: String,
```
zhlsunshine, line 97: may need to display `trust_domain` in `fmt::Display for Workload` if add this field.

zhlsunshine:
IMO, I think that the `trust_domain` information should be added in `ProxyConfig` not in every `Workload`.

zhlsunshine:
IMO, I think that the `trust_domain` information should be added in or be fetched from `ProxyConfig`, not in every `Workload`.

howardjohn (approved):

howardjohn:
> IMO, I think that the `trust_domain` information should be added in or be fetched from `ProxyConfig`, not in every `Workload`.

Its suppose to be set if its *different* from the default trust domain (multi-trust domain meshes)

Issue #440, opened 2023-03-08
Support serverless environment


imroc:
## Background

Thanks to the great idea of separation of L4 and L7 and the current implementation of ztunnel in Rust, the resource occupation and performance loss are minimized, which also makes it possible to integrate mesh capabilities (ztunnel) into serverless environment, which can be dynamically enabled at any time when needed. It is very conducive to the large-scale implementation of service mesh.

But currently ztunnel cannot be directly integrated in serverless environment(one vm one pod). Below I will introduce the existing problems and how to solve them.

## Unable to prefetch cert

When ztunnel runs in a serverless environment, the following error will be reported:

![image](https://user-images.githubusercontent.com/7448852/223703453-aa84c09b-f073-47ec-99ee-7edabe7d0380.png)

That's because istiod uses NodeAuthorizer to verify whether the service account of the caller is trusted (istio-system/ztunnel) when ztunnel calls `CreateCertificate` to fetch cert, and in our serverless scenarios, ztunnel is not a daemonset, but a hidden container built into the pod , the servcie account used is the pod's own service account, so the verification will fail, resulting in failure to fetch cert:

![image](https://user-images.githubusercontent.com/7448852/223703933-d38deee9-29ed-4f6e-bdee-4cf4700ac668.png)

**Solution**: Let ztunnel support to customize the behavior, does not carry metadata in a serverless environment to avoid verification by NodeAuthorizer.

## Failed to transparently intercept traffic

In our serverless environment, ztunnel share net namespace with pod, so it's more like a sidecar, but only for L4 traffic. Just like envoy as a sidecar to intercept traffic transparently, needs to set `SO_MARK` for the packets sent to the local upstream([like this](https://github.com/istio/istio/blob/1.18.0-alpha.0/pilot/pkg/xds/filters/filters.go#L134)), so that the local upstream's return packets will be re-route to the loopback device by policy routing, and then be transparently proxied by ztunnel (avoid local upstream's return packets going out through `eth0`).

**Solution**: Let ztunnel support to set custom `SO_MARK` to local upstream.

imroc:
I have integrated the ztunnel into the serverless environment by modifying the ztunnel code, and successfully run ambient mode in serverless environment. Next, I will generalize my modification and submit PRs to fix this.

howardjohn:
Similar deployment models are "Istio as a sidecar" - useful when we cannot deploy daemonset, or on a VM (not always serverless).

[...]

PR #442, opened 2023-03-08
Support setting: DISABLE_IMPERSONATED_IDENTITY

imroc:
Let ztunnel be able to fetch certificates in serverless environment, part of #440

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -27,6 +27,8 @@ use crate::identity;
 
 const KUBERNETES_SERVICE_HOST: &str = "KUBERNETES_SERVICE_HOST";
 const NODE_NAME: &str = "NODE_NAME";
+const POD_NAME: &str = "POD_NAME";
+const POD_NAMESPACE: &str = "POD_NAMESPACE";
 const INSTANCE_IP: &str = "INSTANCE_IP";
 const CLUSTER_ID: &str = "CLUSTER_ID";
 const LOCAL_XDS_PATH: &str = "LOCAL_XDS_PATH";
@@ -38,6 +40,7 @@ const FAKE_CA: &str = "FAKE_CA";
 const ZTUNNEL_WORKER_THREADS: &str = "ZTUNNEL_WORKER_THREADS";
 const ENABLE_ORIG_SRC: &str = "ENABLE_ORIG_SRC";
 const PROXY_CONFIG: &str = "PROXY_CONFIG";
+const ENABLE_IMPERSONATED_IDENTITY: &str = "ENABLE_IMPERSONATED_IDENTITY";
 
 const DEFAULT_WORKER_THREADS: u16 = 2;
 const DEFAULT_ADMIN_PORT: u16 = 15000;
@@ -86,6 +89,10 @@ pub struct Config {
 
     /// The name of the node this ztunnel is running as.
     pub local_node: Option<String>,
+    /// The name of the pod this ztunnel is running as.
+    pub local_pod: Option<String>,
+    /// The namespace of the pod this ztunnel is running as.
+    pub local_pod_namespace: Option<String>,
     /// The local_ip we are running at.
     pub local_ip: Option<IpAddr>,
     /// The Cluster ID of the cluster that his ztunnel belongs to
@@ -122,6 +129,10 @@ pub struct Config {
 
     // CLI args passed to ztunnel at runtime
     pub proxy_args: String,
+
+    // If true, fetch certs for pod on the same node with impersonated identity.
+    // If false, only fetch cert for current pod (used in serverless environment).
+    pub enable_impersonated_identity: bool,
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -226,6 +237,8 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
         outbound_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15001),
 
         local_node: parse(NODE_NAME)?,
+        local_pod: parse(POD_NAME)?,
+        local_pod_namespace: parse(POD_NAMESPACE)?,
         local_ip: parse(INSTANCE_IP)?,
         cluster_id,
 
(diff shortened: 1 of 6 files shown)


imroc:
cc @howardjohn

src/identity/caclient.rs
```diff
@@ -54,14 +63,20 @@ impl CaClient {
...
-                fields: BTreeMap::from([(
-                    "ImpersonatedIdentity".into(),
-                    prost_types::Value {
-                        kind: Some(Kind::StringValue(id.to_string())),
-                    },
-                )]),
-            }),
+            metadata: {
+                if self.disable_impersonated_identity {
+                    None
```
stevenctl, line 68: IIRC there is code that checks if the `Result` matches the requested `id` and errors.

stevenctl: Yeah, down a few lines at `.verify_san(id)`

src/identity/caclient.rs
```diff
@@ -54,14 +63,20 @@ impl CaClient {
...
             validity_duration: 60 * 60 * 24, // 24 hours
-            metadata: Some(Struct {
-                fields: BTreeMap::from([(
-                    "ImpersonatedIdentity".into(),
-                    prost_types::Value {
-                        kind: Some(Kind::StringValue(id.to_string())),
-                    },
-                )]),
-            }),
+            metadata: {
```
stevenctl, line 66: I think the other issue of doing it at this level is that we end up caching this cert for a bunch of IDs if we try to impersonate

`https://github.com/istio/ztunnel/blob/367188d04ae8d19484cb74fa70067a03c5ffaaaf/src/identity/manager.rs#L341`

src/config.rs
```diff
@@ -252,6 +255,7 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
 
         enable_original_source: parse(ENABLE_ORIG_SRC)?,
         proxy_args: parse_args(),
+        disable_impersonated_identity: parse_default(DISABLE_IMPERSONATED_IDENTITY, false)?,
```
bleggett, line 258:
```suggestion
        disable_impersonated_identity: parse_default(DISABLE_IMPERSONATED_IDENTITY, false)?,
```

Can this just be `impersonate_identity`

src/identity/caclient.rs
imroc: In serverless environment, node is just a virtual concept, so we don't need to prefech certs for pods on the same "node", It is enough to just fetch cert for the current pod, I will optimize the cert prefetch logic.

src/config.rs
imroc: OK, will change to ENABLE_IMPERSONATE_IDENTITY, and default to true.

imroc:
commit resubmitted, PTAL @bleggett @stevenctl

src/config.rs
```diff
@@ -252,6 +265,7 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
 
         enable_original_source: parse(ENABLE_ORIG_SRC)?,
         proxy_args: parse_args(),
+        enable_impersonated_identity: parse_default(ENABLE_IMPERSONATED_IDENTITY, true)?,
```
bleggett, line 268:
```suggestion
        enable_impersonated_identity: parse_default(ENABLE_IMPERSONATED_IDENTITY, false)?,
```

Default on this should be `false` I think - 

We don't want to impersonate by default, and we only need to for some subset environments (serverless).

howardjohn: I think you have it backwards? Impersonation is the current behavior, ztunnel gets certs for spiffe://pod-on-my-node. With false it will get one for spiffe://ztunnel (but you would probably run ztunnel itself as a non-ztunnel SA). Basically like a sidecar

bleggett: Yeah - nevermind then, had it backwards.

bleggett (approved):

src/workload.rs
```diff
@@ -321,7 +321,27 @@ impl WorkloadManager {
         });
         let workloads: Arc<Mutex<WorkloadStore>> = Arc::new(Mutex::new(WorkloadStore {
             cert_tx: Some(tx),
-            local_node: config.clone().local_node,
+            local_node: {
```
stevenctl, line 324: Nit:

```rust
local_node: config.enable_impersonated_identity.then_some(config.local_node.clone()).flatten()

**Review by stevenctl — APPROVED** [review]

File: src/workload.rs [hunk 6]
```diff
@@ -614,7 +639,10 @@ impl WorkloadStore {
                 }
             }
         }
-        if Some(&w.node) == self.local_node.as_ref() {
+        if Some(&w.node) == self.local_node.as_ref()
+            || (Some(&w.name) == self.local_pod.as_ref()
+                && Some(&w.namespace) == self.local_namespace.as_ref())
```
hzxuzhonghu, line 644: Why do check workload name and namespace here?

hzxuzhonghu: cc @howardjohn

hzxuzhonghu: local_node `None` itself is very hack.

> fetch cert for current pod,

is this ztunnel itself?  Canot ztunnel fetch certs for those pods that run in same node (i mean in your case virtual kubelet)

imroc: Yes, ztunnel share pod's name and namespace.

imroc: > So in this case how could ztunnel be deployed, in only a single node or whole cluster as the daemonset

It's depends on the implementation. I think more fine-grained custom configurations may be needed to fine-grained control ztunnel behavior to be compatible with more serverless implementation scenarios.

hzxuzhonghu: virtual kubelet cannot represent serverless

hzxuzhonghu: We need a revist on the whole solution

imroc: How about `ENABLE_IMPERSONATED_IDENTITY` only controls whether fetch cert with impersonated identity, and another env controls only fetch cert for specified pod if not empty.

PR #443, opened 2023-03-08
getting correct xds_root_cert info for config with the same logic of FindRootCAForXDS in Istio

zhlsunshine:
getting correct `xds_root_cert` info for config with the same logic of `FindRootCAForXDS` in Istio

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -14,7 +14,7 @@
 
 use std::collections::HashMap;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::time::Duration;
 use std::{env, fs};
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -353,6 +373,45 @@ pub fn empty_to_none<A: AsRef<str>>(inp: Option<A>) -> Option<A> {
     inp
 }
 
+pub fn find_rootca_for_xds(citadel_ca_cert_path: String) -> Result<RootCert, Error> {
```
howardjohn, line 376: The existing istio implementation is legacy - we have to handle all sorts of old junk. I don't think we need most of this at all today. Probably just default to `./var/run/secrets/istio/root-cert.pem` and allow a single env to override.

BTW: ca_root_cert should have similar logic (but separate config)

zhlsunshine: Hi @howardjohn, okay, There should be 2 envs for xds and ca root cert, such as `XDS_CERT_PROVIDER` for xds root cert and  `CA_CERT_PROVIDER` for ca root cert. And both of them have `./var/run/secrets/istio/root-cert.pem` by default.

zhlsunshine:
Hi @howardjohn, re-submitted, and the change includes both xds and ca root cert.

src/config.rs
```diff
@@ -48,6 +48,11 @@ const DEFAULT_CLUSTER_ID: &str = "Kubernetes";
 
 const ISTIO_META_PREFIX: &str = "ISTIO_META_";
 
+/// Fetch the XDS/CA root cert file path based on below constants
+const XDS_ROOT_CERT_PROVIDER_ENV: &str = "XDS_CERT_PROVIDER";
```
howardjohn, line 52: I would name it XDS_ROOT_CA and CA_ROOT_CA to match the istio-agent

src/config.rs
```diff
@@ -195,6 +200,26 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
         Some(parse_default(CA_ADDRESS, default_istiod_address)?)
     });
 
+    let xds_root_cert_provider = parse_default(
+        XDS_ROOT_CERT_PROVIDER_ENV,
+        DEFAULT_ROOT_CERT_PROVIDER.to_string(),
+    )?;
+    let xds_root_cert = if Path::new(&xds_root_cert_provider).exists() {
```
howardjohn, line 207: can we handle SYSTEM special case as well?

src/config.rs
```diff
@@ -195,6 +200,26 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
...
+        RootCert::File(xds_root_cert_provider.parse().unwrap())
```
howardjohn, line 208: unwrap here isn't safe, can we propagate the error?

howardjohn: It was fine before since it was for a constant

howardjohn (approved):
LGTM, we can address the rest in followup

src/config.rs
zhlsunshine: Okay, will fix it later

src/config.rs
zhlsunshine: sure, it makes sense to match the isito-agent

src/config.rs
zhlsunshine: sure, make sure no panic here.

PR #444, opened 2023-03-08
Parse baggage on destination ztunnel

ymesika:
Implementing a TODO of parsing the baggage header on the destination ztunnel.

### DIFF
--- a/src/proxy/util.rs
+++ b/src/proxy/util.rs
@@ -14,6 +14,9 @@
 
 use std::io::{Error, ErrorKind};
 
+use hyper::http::HeaderValue;
+use tracing::warn;
+
 pub fn is_runtime_shutdown(e: &Error) -> bool {
     if e.kind() == ErrorKind::Other
         && e.to_string() == "A Tokio 1.x context was found, but it is being shutdown."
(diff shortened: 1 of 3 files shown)


src/proxy/util.rs
```diff
@@ -22,3 +25,89 @@ pub fn is_runtime_shutdown(e: &Error) -> bool {
...
+#[derive(Default)]
+pub struct Baggage {
+    pub cluster_id: Option<String>,
+    pub namespace: Option<String>,
+    pub workload_name: Option<String>,
+    pub service_name: Option<String>,
+    pub revision: Option<String>,
+}
+
+pub fn parse_baggage_header(h: Option<&HeaderValue>) -> Baggage {
```
howardjohn, line 38: can we add a fuzzer? see fuzz/ folder and https://github.com/rust-fuzz/cargo-fuzz, its pretty straightforward

howardjohn (approved):
This is a lot stricter than the full `baggage` spec but I think good for now to keep things simple

ymesika:
@howardjohn 
> full baggage spec

Do you mean the Otel baggage spec?

howardjohn:
Yeah, https://www.w3.org/TR/baggage/. I don't know we even want that complex parsing (unless there is a library we can use), just a note

kyessenov:
If ztunnel has WDS, why do we not look it up there instead?

howardjohn:
We do, baggage is fallback. If we aren't going to use baggage we should
just kill it entirely

src/proxy/util.rs
```diff
@@ -22,3 +25,89 @@ pub fn is_runtime_shutdown(e: &Error) -> bool {
...
+            assert_eq!(baggage.workload_name, Some("N1".to_string()));
+            assert_eq!(baggage.service_name, Some("N2".to_string()));
+            assert_eq!(baggage.revision, Some("V1".to_string()));
+        };
+    }
+
+    #[test]
+    fn baggage_parser_empty_values() {
+        let baggage_str = "k8s.cluster.name=,k8s.namespace.name=,k8s.deployment.name=,service.name=,service.version=";
+        if let Ok(header_value) = HeaderValue::from_str(baggage_str) {
```
kyessenov, line 94: I can't tell - do you properly handle multiline baggage header?

kyessenov:
I think you should send it to envoys until we can kill it there, and kill it in ztunnel.
It shouldn't be a workaround way to deal with split-brain WDS

src/proxy/util.rs
ymesika: Not sure too.
Are we expected to get multilined header? From the current Waypoint/Ztunnel code the header isn't multilined.
In what scenario will we get other component hbone connecting with baggage header?

src/proxy/util.rs
ymesika: Just to make sure I follow your line of thoughts about the need for fuzz - Is it because it exposes the ztunnel to attacks over the hbone port?

howardjohn: Any custom parsing of untrusted data should have fuzzing, this is one of the top sources of vulnerabilities and its ~trivial to prevent with a 5 line fuzz test

src/proxy/util.rs
kyessenov: HTTP allows you to break down any header into multilines, and it's more efficient in some cases to append, that concatenate with a comma.

ymesika: I see, so the commas are replaced with new lines in such cases?
I can add a test to cover that

PR #446, opened 2023-03-08
Cleanup and update deps

howardjohn:


hzxuzhonghu (approved):
Should we upgrade rust as well

PR #447, opened 2023-03-09
Added a fuzz target for baggage parsing

ymesika:
\+ `make clean`


ymesika:
@howardjohn Who/When are the fuzz tests being run? I can't see it being run as part of the CI and locally it ran for hours.

howardjohn (approved):
The fuzz tests run in OSS-fuzz, which is locked down to PSWG members. A bit unfortunate for folks writing tests, but good that they are private when they report vulnerabilities :slightly_smiling_face: .

If it fails PSWG will get notified and fix it or forward the failure to people.

PR #448, opened 2023-03-10
Prevent proxying outbound traffic with DirectLocal in serverless environment

imroc:
In serverless(`enable_impersonated_identity` is false), node is virtual, pods on the same "node" do not share the ztunnel, we can't let outbound traffic go through DirectLocal fast path.

### DIFF
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -374,7 +374,8 @@ impl OutboundConnection {
             return Err(Error::NoGatewayAddress(Box::new(us.workload.clone())));
         }
         // For case source client and upstream server are on the same node
-        if !us.workload.node.is_empty()
+        if self.pi.cfg.enable_impersonated_identity
+            && !us.workload.node.is_empty()
             && self.pi.cfg.local_node.as_ref() == Some(&us.workload.node) // looks weird but in Rust borrows can be compared and will behave the same as owned (https://doc.rust-lang.org/std/primitive.reference.html)
             && us.workload.protocol == Protocol::HBONE
         {


imroc:
Ref https://github.com/istio/ztunnel/issues/440#issuecomment-1463274803

cc @bleggett @stevenctl @howardjohn

hzxuzhonghu:
I am curious how can istio distinguish that whether a workload is on a virtual node, it has no such info. 

I think what you want is always disabling node fast path

imroc:
> I am curious how can istio distinguish that whether a workload is on a virtual node, it has no such info.
> 
> I think what you want is always disabling node fast path

istio does not need to know, it is configured on ztunnel, refer to the previous pr which have been merged: https://github.com/istio/ztunnel/pull/442

bleggett:
To @hzxuzhonghu 's point - 

Are we complicating things by having the concept of a "virtual" node?



Should ztunnel simply be modified to be ok with `local_node == null` and use that as the decisionmaking flag?

howardjohn:
IMO there is a valid mode of "sidecar" vs "node proxy" mode. Similar as done here but we should rename it.

Eventually we may need to know the node, as k8s service defines a way to "only send to same node", so we still need it.

These are not a new "virtual node" concept, this is how Istio works today with sidecars -- nothing too special.

bleggett:
> IMO there is a valid mode of "sidecar" vs "node proxy" mode. Similar as done here but we should rename it.
> 
> Eventually we may need to know the node, as k8s service defines a way to "only send to same node", so we still need it.
> 
> These are not a new "virtual node" concept, this is how Istio works today with sidecars -- nothing too special.

That's fine - I just don't want the `impersonated_identity` flag to become  a sort of hidden "sidecar  vs node-proxy" mode switch if we would be better served with a more explicit modeswitch.

howardjohn:
Yes I agree, I want a single mode switch that is "sidecar" vs "node" (better names are ok, too :-) )

imroc:
> I want a single mode switch that is "sidecar" vs "node" (better names are ok, too :-) )

I agree, How about adding a `PROXY_MODE` environment variable, the optional values are Node and Sidecar?

When `PROXY_MODE` is Sidecar:
1. Only fetch cert for the current pod.
2. Do not use impersonated identity.
3. Do not take the fast path.
4. Do not check recursion.

hzxuzhonghu:
>Only fetch cert for the current pod.


Istiod will not dispatch wl info of ztunnel, to ztunnel. so can you draft a design before.

imroc:
> Istiod will not dispatch wl info of ztunnel, to ztunnel. so can you draft a design before.

It's quite simple, just like the sidecar mode, ztunnel is not a workload, but a sidecar container in the pod, ztunnel can get pod info from env `POD_NAME` and `POD_NAMESPACE` using downward api. Does this need a design draft?

howardjohn:
I am not sure what "Do not check recursion." means but otherwise lgtm

imroc:
> I am not sure what "Do not check recursion." means but otherwise lgtm

@howardjohn ztunnel will check if the destination ip is the ztunnel ip, if it is, then returns Error::SelfCall(recursion error).  In sidecar mode, ztunnel shares ip with pod, so we should not check the recursion.

imroc:
Close in favor of #452 .

PR #449, opened 2023-03-10
enhance the xds and ca certificates finding

zhlsunshine:
enhance the xds and ca certificates finding:
1. change the environment var to match istio-agent
2. handle SYSTEM special case
3. propagate the error for RootCert::File generation


src/config.rs
```diff
@@ -211,22 +212,32 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
...
     let xds_root_cert = if Path::new(&xds_root_cert_provider).exists() {
-        RootCert::File(xds_root_cert_provider.parse().unwrap())
+        let xds_root_cert_match = match xds_root_cert_provider.parse::<PathBuf>() {
+            Ok(root_cert) => root_cert,
+            Err(_) => PathBuf::new(),
+        };
+        RootCert::File(xds_root_cert_match)
+    } else if xds_root_cert_provider.eq(&CERT_SYSTEM.to_string()) {
+        // handle SYSTEM special case for xds
+        RootCert::File(PathBuf::new())
```
howardjohn, line 225:
```suggestion
        RootCert::Default
```

src/config.rs
```diff
@@ -211,22 +212,32 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
...
-    )?;
+    let xds_root_cert_provider =
+        parse_default(XDS_ROOT_CA_ENV, DEFAULT_ROOT_CERT_PROVIDER.to_string())?;
     let xds_root_cert = if Path::new(&xds_root_cert_provider).exists() {
-        RootCert::File(xds_root_cert_provider.parse().unwrap())
+        let xds_root_cert_match = match xds_root_cert_provider.parse::<PathBuf>() {
+            Ok(root_cert) => root_cert,
+            Err(_) => PathBuf::new(),
+        };
+        RootCert::File(xds_root_cert_match)
```
howardjohn, line 222: We need to actually handle the error

```suggestion
        RootCert::File(xds_root_cert_provider.parse()?)
```

zhlsunshine: Hi @howardjohn, can not directly use `?` to handle such error, I encountered below error message:

```
expected enum `Infallible`, found enum `config::Error`
```

so I use `match` statement to handle the Err situation.

src/config.rs
zhlsunshine: okay, done!

src/config.rs
howardjohn: Its not handled, the error is ignored and we just return an invalid PathBuf::new()... we need to actually return an error.

` RootCert::File(xds_root_cert_provider.into())` will do it tohugh

zhlsunshine: done!

zhlsunshine:
Hi @howardjohn, could you please help to review or merge it? Thanks ^_^

src/config.rs
```diff
@@ -211,22 +212,24 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
...
-    let xds_root_cert_provider = parse_default(
-        XDS_ROOT_CERT_PROVIDER_ENV,
-        DEFAULT_ROOT_CERT_PROVIDER.to_string(),
-    )?;
+    let xds_root_cert_provider =
+        parse_default(XDS_ROOT_CA_ENV, DEFAULT_ROOT_CERT_PROVIDER.to_string())?;
     let xds_root_cert = if Path::new(&xds_root_cert_provider).exists() {
-        RootCert::File(xds_root_cert_provider.parse().unwrap())
+        RootCert::File(xds_root_cert_provider.into())
+    } else if xds_root_cert_provider.eq(&CERT_SYSTEM.to_string()) {
```
howardjohn, line 219: nit: technically this should probably be first in case they have a file named "system"

howardjohn: also I think you can jsut use `==` instead of `.eq` but not a big deal

howardjohn (approved):

Issue #424, opened 2023-02-28
throughput benchmarking is misleading / not working


kdorosh:
if you make this buffer much larger (10-1000x) https://github.com/istio/ztunnel/blob/dbbeb8f341e0b2651204094c0dace53718497f2b/src/proxy.rs#L163-L164 then `cargo bench --quick` will show impossible throughput results. HBONE will be much faster than direct and tcp proxy. Thus, I have very little confidence in the throughput test as it stands today

hzxuzhonghu:
I think it is reasonable, increasing the buffer size would improve throughput

kdorosh:
i think it is reasonable for an increased buffer size to improve throughput

i don't think it's reasonable that increasing that buffer size can result in throughput numbers for HBONE much higher than pure TCP proxy or direct connection to upstream (no proxy at all), which is what i saw when increasing this number 10-100x

jaellio:
@Stevenjin8

Stevenjin8:
Pretty sure this was silently fixed in #446. I can reproduce the issue before that, but not after.
Specifically, it seems like it was the update of the `h2` crate from `0.3.15` to `0.3.16`. Here are the [release notes](https://github.com/hyperium/h2/releases) for `h2@0.3.16`. I think

> Fix panic when trying to send data and connection window is available, but stream window is not.

might be what fixed it, but I'm just speculating.

Here are the results on commit cce6f79 (right before #446):

```
throughput/direct       time:   [2.9227 ms 2.9352 ms 2.9850 ms]
                        thrpt:  [3.2715 GiB/s 3.3271 GiB/s 3.3413 GiB/s]
throughput/tcp          time:   [2.8536 ms 2.8614 ms 2.8928 ms]
                        thrpt:  [3.3759 GiB/s 3.4129 GiB/s 3.4222 GiB/s]
Benchmarking throughput/hbone
2023-07-28T19:19:33.267542Z ERROR ztunnel::proxy::inbound: hbone server copy: io error: broken pipe dur=3.634443741s
throughput/hbone        time:   [74.989 µs 76.106 µs 76.386 µs]
                        thrpt:  [127.85 GiB/s 128.32 GiB/s 130.23 GiB/s]
```
and the results after running `cargo update -p h2 --precise 0.3.16`

```
throughput/direct       time:   [2.9268 ms 2.9770 ms 2.9896 ms]
[...]

PR #451, opened 2023-03-10
deps: update to hyper 1.0.0

howardjohn:
This moves us to Hyper 1.0. This is not 100% fully launched but its in rc3 and pretty close. AS we are alpha, I think its appropriate to bite the bullet now and move early. The main bummer is we need to fork `boring` for a bit since they won't move to 1.0 until post-rc.

This will unblock https://github.com/istio/ztunnel/pull/293

Benchmarks show some concerning data but I think its a symptom of https://github.com/istio/ztunnel/issues/424 rather than a real regression

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use bytes::Bytes;
 use std::borrow::Borrow;
 use std::collections::HashMap;
 use std::sync::Arc;
@@ -25,7 +26,9 @@ use drain::Watch;
 use gperftools::heap_profiler::HEAP_PROFILER;
 #[cfg(feature = "gperftools")]
 use gperftools::profiler::PROFILER;
-use hyper::{header::HeaderValue, header::CONTENT_TYPE, Body, Request, Response};
+use http_body_util::Full;
+use hyper::body::Incoming;
+use hyper::{header::HeaderValue, header::CONTENT_TYPE, Request, Response};
 use pprof::protos::Message;
 #[cfg(feature = "gperftools")]
 use tokio::fs::File;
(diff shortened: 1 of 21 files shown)


howardjohn:
fyi @stevenctl

src/admin.rs
```diff
@@ -81,17 +84,16 @@ pub struct CertsDump {
...
         workload_info: WorkloadInformation,
         shutdown_trigger: signal::ShutdownTrigger,
         drain_rx: Watch,
         cert_manager: Arc<SecretManager>,
-    ) -> hyper::Result<Self> {
+    ) -> anyhow::Result<Self> {
         Server::<State>::bind(
             "admin",
             config.admin_addr,
-            shutdown_trigger.clone(),
```
stevenctl, line 94: nit: remove `shutdown_trigger` arg

src/hyper_util.rs
```diff
@@ -71,100 +172,82 @@ pub fn plaintext_response(code: hyper::StatusCode, body: String) -> Response<Bod
...
+                                f(state, req)
+                            }),
+                        );
+                    match futures_util::future::select(Box::pin(drain.signaled()), serve).await {
+                        futures_util::future::Either::Left((_shutdown, mut server)) => {
+                            let drain = std::pin::Pin::new(&mut server);
+                            drain.graceful_shutdown();
+                            server.await
+                        }
+                        futures_util::future::Either::Right((server, _shutdown)) => server,
```
stevenctl, line 242: super nit: confusing to call this `server`, it's a `Connection`.

src/proxy/inbound.rs
```diff
@@ -78,67 +82,58 @@ impl Inbound {
...
+                    src_identity: socket
+                        .ssl()
+                        .peer_certificate()
+                        .and_then(|x| crate::tls::boring::extract_sans(&x).first().cloned()),
+                    src_ip: to_canonical(socket.get_ref().peer_addr().unwrap()).ip(),
+                    dst,
+                };
+                debug!(%conn, "accepted connection");
+                let enable_original_source = self.cfg.enable_original_source;
+                let server = crate::hyper_util::http2_server()
```
stevenctl, line 110: same here.. it's a "serverside connection" while this whole piece of machinery to accept connections feels like the server.

stevenctl:
LGTM would prefer another set of eyes.

stevenctl:
@bleggett

src/admin.rs
howardjohn: Why? we use it on L120 for quitquitquit

stevenctl (approved):

PR #452, opened 2023-03-11
Add PROXY_MODE setting, remove ENABLE_IMPERSONATED_IDENTITY

imroc:
1. Add `PROXY_MODE`, the proxy mode of ztunnel, `node` or `sidecar`, default to `node`.
2. Remove `ENABLE_IMPERSONATED_IDENTITY`, replaced by `PROXY_MODE`.

When set `PROXY_MODE` to `sidecar`:
1. Only fetch cert for the current pod (from env `POD_NAME` and `POD_NAMESPACE`).
2. Disable impersonated identity when fetching cert.
3. Disable fast path (`DirectLocal`).
4. Disable recursion check.

Refers https://github.com/istio/ztunnel/pull/448#issuecomment-1464672506

cc @howardjohn @bleggett @hzxuzhonghu

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -28,7 +28,7 @@ use tracing::{debug, error, info, instrument, trace};
 use xds::istio::security::Authorization as XdsAuthorization;
 use xds::istio::workload::Workload as XdsWorkload;
 
-use crate::config::ConfigSource;
+use crate::config::{ConfigSource, ProxyMode};
 use crate::identity::{Identity, SecretManager};
 use crate::metrics::Metrics;
 use crate::rbac::{Authorization, RbacScope};
@@ -320,27 +320,7 @@ impl WorkloadManager {
         });
         let workloads: Arc<Mutex<WorkloadStore>> = Arc::new(Mutex::new(WorkloadStore {
             cert_tx: Some(tx),
-            local_node: {
-                if config.enable_impersonated_identity {
-                    config.local_node.clone()
-                } else {
-                    None
-                }
-            },
-            local_pod: {
-                if config.enable_impersonated_identity {
-                    None
-                } else {
-                    config.local_pod.clone()
-                }
-            },
-            local_namespace: {
-                if config.enable_impersonated_identity {
-                    None
-                } else {
-                    config.local_pod_namespace.clone()
-                }
-            },
+            proxy_mode: config.proxy_mode.clone(),
             ..Default::default()
         }));
         let xds_workloads = workloads.clone();
@@ -592,13 +572,9 @@ pub struct WorkloadStore {
     #[serde(skip_serializing, default)]
     cert_tx: Option<mpsc::Sender<Identity>>,
 
-    // needed to determine whether or not to prefetch certs, will be None if ENABLE_IMPERSONATED_IDENTITY
-    // is set to false.
-    local_node: Option<String>,
-    // needed to determine whether or not to prefetch certs in serverless environment, will be None if
-    // ENABLE_IMPERSONATED_IDENTITY set to true.
-    local_pod: Option<String>,
-    local_namespace: Option<String>,
+    // needed to determine whether or not to prefetch certs
+    proxy_mode: ProxyMode,
+    local_node: Option<String>
 }
 
 impl WorkloadStore {
@@ -638,10 +614,8 @@ impl WorkloadStore {
                 }
             }
         }
-        if Some(&w.node) == self.local_node.as_ref()
-            || (Some(&w.name) == self.local_pod.as_ref()
-                && Some(&w.namespace) == self.local_namespace.as_ref())
-        {
+
+        if self.proxy_mode == ProxyMode::Node && Some(&w.node) == self.local_node.as_ref() {
             if let Some(tx) = self.cert_tx.as_mut() {
                 if let Err(e) = tx.try_send(widentity) {
                     info!("couldn't prefetch: {:?}", e)
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -574,7 +574,7 @@ pub struct WorkloadStore {
 
     // needed to determine whether or not to prefetch certs
     proxy_mode: ProxyMode,
-    local_node: Option<String>
+    local_node: Option<String>,
 }
 
 impl WorkloadStore {
--- a/src/config.rs
+++ b/src/config.rs
@@ -27,8 +27,7 @@ use crate::identity;
 
 const KUBERNETES_SERVICE_HOST: &str = "KUBERNETES_SERVICE_HOST";
 const NODE_NAME: &str = "NODE_NAME";
-const POD_NAME: &str = "POD_NAME";
-const POD_NAMESPACE: &str = "POD_NAMESPACE";
+const PROXY_MODE: &str = "PROXY_MODE";
 const INSTANCE_IP: &str = "INSTANCE_IP";
 const CLUSTER_ID: &str = "CLUSTER_ID";
 const LOCAL_XDS_PATH: &str = "LOCAL_XDS_PATH";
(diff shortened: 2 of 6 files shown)


src/workload.rs
```diff
@@ -639,10 +620,14 @@ impl WorkloadStore {
...
             }
         }
-        if Some(&w.node) == self.local_node.as_ref()
-            || (Some(&w.name) == self.local_pod.as_ref()
-                && Some(&w.namespace) == self.local_namespace.as_ref())
-        {
+        let should_fetch: bool = match self.proxy_mode {
+            ProxyMode::Sidecar => {
+                Some(&w.name) == self.local_pod.as_ref()
+                    && Some(&w.namespace) == self.local_namespace.as_ref()
```
hzxuzhonghu, line 626: Curious if in sidecar node, whose certificate would ztunnel use?

imroc: > Curious if in sidecar node, whose certificate would ztunnel use?

@hzxuzhonghu Of course it is in the pod where the ztunnel is located (ztunnel is not an independent pod, but a sidecar container in the pod), the pod name and namespace comes from env `POD_NAME` and `POD_NAMESPACE` using downward api, just like the sidecar mode.

hzxuzhonghu: Then how could ztunnel be injected to workload container? And also curious why not sidecar in this case.

imroc: > Then how could ztunnel be injected to workload container? And also curious why not sidecar mode in this case.

@hzxuzhonghu The core reason is that we want to integrate ztunnel in the serverless environment. In our serverless environment, each pod occupies a lightweight virtual machine exclusively, and ztunnel cannot be deployed by daemonset.
We want to integrate ztunnel into serverless pods in a sidecar-like way and enable it by default at scale.
The reason for not using sidecar directly is that in large-scale scenarios, resource overhead and performance loss 
 of envoy cannot be ignored, which is not conducive to large-scale use.

imroc: > Then how could ztunnel be injected to workload container? 

@hzxuzhonghu In our case, ztunnel is a hidden sidecar container, injected by the serverless agent in the lightweight virtual machine at runtime, user can not see the ztunnel container in pod.

imroc: I believe that other similar serverless environments will also use the ability of `PROXY_MODE`=sidecar to integrate with ambient mesh.

hzxuzhonghu: Can you explain how L25 equals ` Some(&w.name) == self.local_pod.as_ref()`

imroc: It's compares the pod name, just like the node name comparison when `PROXY_MODE`=node:

```rust
Some(&w.node) == self.local_node.as_ref()
```

hzxuzhonghu: As you say workload and ztunel are two pods And w.name actually is not pod name just pod IP.

imroc: No, I'm talking about ztunnel and workload are the same pod. 

And w.name actually the pod name not pod IP, check this out: https://github.com/istio/ztunnel/blob/ac83b1c51e0f7d57162fad858acc76b2bb0546b4/proto/workload.proto#L22

bleggett (approved):
The logic and goal of this seems reasonable to me and this PR looks fine - I will defer to @hzxuzhonghu on the need for a design doc. I don't think it would hurt to write a quick one-pager expanding on your PR description and bringing it up during the Ambient WG meeting just to make people aware.

imroc:
@bleggett Thanks, it has been fully tested in our environment, works fine and perfectly solves our problem, I think that I also have the responsibility  to write a simple design doc to explain to everyone. I will take the time to write it today.

src/workload.rs
hzxuzhonghu: I guess the desc is not right, from istiod , I remember it is pod IP address

hzxuzhonghu (changes requested):
Confirmed, for workload the name is actually ip not pod name.
```
// Lookup finds a given IP address.
func (a *AmbientIndex) Lookup(ip string) []*model.WorkloadInfo {
	a.mu.RLock()
	defer a.mu.RUnlock()
	// First look at pod...
	if p, f := a.byPod[ip]; f {
		return []*model.WorkloadInfo{p}
	}
	// Fallback to service. Note: these IP ranges should be non-overlapping
	return a.byService[ip]
}
```

imroc:
> Confirmed, for workload the name is actually ip not pod name.
> 
> ```
> // Lookup finds a given IP address.
> func (a *AmbientIndex) Lookup(ip string) []*model.WorkloadInfo {
> 	a.mu.RLock()
> 	defer a.mu.RUnlock()
> 	// First look at pod...
> 	if p, f := a.byPod[ip]; f {
> 		return []*model.WorkloadInfo{p}
> 	}
> 	// Fallback to service. Note: these IP ranges should be non-overlapping
> 	return a.byService[ip]
> }
> ```

@hzxuzhonghu `AmbientIndex.byPod` is a map of pod ip to workload: https://github.com/istio/istio/blob/485c30041d30c85f7b82c9a362d256c6d7913ff1/pilot/pkg/serviceregistry/kube/controller/ambientindex.go#L55-L56

But it not means that `wl.Name` is the pod ip.

The ip comes from `v1.Pod.Status.IP`, not `wl.name`:

https://github.com/istio/istio/blob/485c30041d30c85f7b82c9a362d256c6d7913ff1/pilot/pkg/serviceregistry/kube/controller/ambientindex.go#L746

And the ip in `wl` is actually `wl.Address`, not `wl.Name`

src/workload.rs
howardjohn: there is Resource.Name and the "name" field in workload proto they are distinct things

src/workload.rs
```diff
@@ -639,10 +620,14 @@ impl WorkloadStore {
                 }
             }
         }
-        if Some(&w.node) == self.local_node.as_ref()
-            || (Some(&w.name) == self.local_pod.as_ref()
-                && Some(&w.namespace) == self.local_namespace.as_ref())
-        {
+        let should_fetch: bool = match self.proxy_mode {
+            ProxyMode::Sidecar => {
```
howardjohn, line 624: For sidecar we can just always pre-getch our own identity, even before we get xds response. This is simpler and faster

imroc: Yes, I think so, let me optimize it in the next PR, ok?

howardjohn: I am a bit concerned with doing any pre-fetching in sidecar mode. can we change this to never do it for sidecars mode in the short term then?

hzxuzhonghu:
>And the ip in wl is actually wl.Address, not wl.Name

my mistake

src/workload.rs
imroc: > can we change this to never do it for sidecars mode in the short term then?

@howardjohn  OK, I changed to not pre-fetch cert in sidecar mode, and tested in the real environment, for the first request, there is not much performance impact.

howardjohn:
I skimmed it, looks good overall, will review closely and probably approve
tomorrow. Would be good to start adding tests for this as well

imroc:
Now the PR is ready, the main changes were merged into this commit:

```txt
Add PROXY_MODE setting, remove ENABLE_IMPERSONATED_IDENTITY

1. Add PROXY_MODE, the proxy mode of ztunnel, "node" or "sidecar", default to "node".
2. Remove ENABLE_IMPERSONATED_IDENTITY, replaced by PROXY_MODE.

When set PROXY_MODE to "sidecar":
1. Disable pre-fetching cert.
2. Disable impersonated identity when fetching cert.
3. Disable fast path (DirectLocal).
4. Disable recursion check.
```

And CI failed with some clippy blames, which is not related to this PR, a new commit is added to fix clippy errors:

```txt
Fix clippy blames.

1. Remove the unnecessary `let` binding, return the
   expression directly.
2. Remove the unnecessary conversion to the same
   type: `std::str::Lines<'_>`, delete `.into_iter()`.

```

imroc:
> Would be good to start adding tests for this as well.

OK, I'll take the time to add the tests.

howardjohn (approved):

hzxuzhonghu (approved):

PR #454, opened 2023-03-13
Remove redudent match in workload::byte_to_ip()

imroc:
Case `0` is not specical, already been handled properly in the following match:
```rust
n => Err(WorkloadError::ByteAddressParse(n)),
```


imroc:
Case `0` is not specical, already been handled properly in the following match:
```rust
n => Err(WorkloadError::ByteAddressParse(n)),
```

hzxuzhonghu (approved):

PR #455, opened 2023-03-13
Revert "Support setting: DISABLE_IMPERSONATED_IDENTITY"

hzxuzhonghu:
Reverts istio/ztunnel#442

In favor of new design https://github.com/istio/ztunnel/pull/452


imroc:
I think we don't need revert, new PR #452  is based on master, which has #443 merged

hzxuzhonghu:
Because i do not think we can achieve a consensus soon

PR #464, opened 2023-03-14
add state to secrets output

GregHanson:
add state to better capture any certs that may be in an unhealthy state

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -63,7 +63,7 @@ pub struct ConfigDump {
     certificates: Vec<CertsDump>,
 }
 
-#[derive(serde::Serialize, Debug, Clone)]
+#[derive(serde::Serialize, Debug, Clone, Default)]
 pub struct CertDump {
     // Not available via Envoy, but still useful.
     pem: String,
@@ -167,10 +167,7 @@ async fn dump_certs(cert_manager: &SecretManager) -> Vec<CertsDump> {
                 identity: id.to_string(),
                 state,
                 ca_cert: [CertDump {
-                    pem: "".to_string(),
-                    serial_number: "".to_string(),
-                    valid_from: "".to_string(),
-                    expiration_time: "".to_string(),
+                    ..Default::default()
                 }],
                 cert_chain: Vec::new(),
             };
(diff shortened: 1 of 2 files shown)


GregHanson:
feedback from https://github.com/istio/istio/pull/43907#issuecomment-1467256344

hanxiaop (commented):
lgtm

src/admin.rs
```diff
@@ -161,10 +162,23 @@ fn dump_cert(x509: &X509) -> CertDump {
...
-        .collect_certs(|id, certs| CertsDump {
-            identity: id.to_string(),
-            ca_cert: [dump_cert(certs.x509())],
-            cert_chain: certs.iter_chain().map(dump_cert).collect(),
+        .collect_certs(|id, certs, state| {
+            let mut dump = CertsDump {
+                identity: id.to_string(),
+                state,
+                ca_cert: [CertDump {
+                    pem: "".to_string(),
```
ymesika, line 170: Maybe add a `Default` annotation to the struct instead of setting empty strings?

ymesika (approved):
LGTM

nit comment

src/admin.rs
GregHanson: done

src/admin.rs
```diff
@@ -161,10 +162,20 @@ fn dump_cert(x509: &X509) -> CertDump {
...
     let mut dump = cert_manager
-        .collect_certs(|id, certs| CertsDump {
-            identity: id.to_string(),
-            ca_cert: [dump_cert(certs.x509())],
-            cert_chain: certs.iter_chain().map(dump_cert).collect(),
+        .collect_certs(|id, certs, state| {
+            let mut dump = CertsDump {
+                identity: id.to_string(),
+                state,
+                ca_cert: [CertDump {
```
howardjohn, line 169: nit: do we just want empty vec if there is none?

howardjohn (approved):

src/admin.rs
GregHanson: cc @qfel 

The `ca_cert` field is an array (with set size) and not a Vec, I know Piotr did this to be in line with Envoy API

PR #465, opened 2023-03-16
Rename proxy mode "Sidecar" to "Serverless"

imroc:
* Ztunnel is more of a VPC or network layer, not a full-blown sidecar.
* Also to avoid confusion with the old sidecars.

Check comments here: https://docs.google.com/document/d/135gWueRb5KW2vc0RerE2vGKBoCFxbqoLVBuTGJBcxO4/edit?disco=AAAArETr0pI


imroc:
cc @costinm

howardjohn (approved):

imroc:
Yep, that make sense, I have changed to "shared" and "dedicated", PTAL.

PR #467, opened 2023-03-16
socks5: change message from self-call to unknown dst

stevenctl:


src/proxy/outbound.rs
```diff
@@ -141,7 +141,7 @@ impl OutboundConnection {
         if block_passthrough && req.destination_workload.is_none() {
             // This is mostly used by socks5. For typical outbound calls, we need to allow calls to arbitrary
             // domains. But for socks5
```
howardjohn, line 143: nit: this was not complete sentence. Probably my fault

howardjohn (approved):

Issue #468, opened 2023-03-16
certificates field is empty in config_dump after IMPERSONATED_IDENTITY changes


GregHanson:
I noticed the `certificates` field is empty in ztunnel config_dump after https://github.com/istio/ztunnel/pull/452

```
$ kubectl exec -it -n istio-system ds/ztunnel -- curl localhost:15000/config_dump | jq .
...
...
      "ISTIO_VERSION": "1.18-dev"
    },
    "num_worker_threads": 2,
    "enable_original_source": null,
    "proxy_args": "proxy ztunnel"
  },
  "certificates": []
}
```

I would expect most of the certs to no longer be present after this change, but I would expect there to at least be one for ztunnel pod itself? https://github.com/istio/ztunnel/pull/442#discussion_r1131527224

GregHanson:
cc @imroc

imroc:
Nice catch, fixed in this PR https://github.com/istio/ztunnel/pull/469

PR #469, opened 2023-03-17
Fix: missing pre-fetched certificate in config_dump (#468)

imroc:
Re-add the `local_node.clone()` that was deleted by mistake.


imroc:
cc @GregHanson @howardjohn

PR #471, opened 2023-03-17
Add tests for Initializing and Unavailable cert states in config_dump

qfel:
I was going to add some comments on #464  but then realized it was merged.. so here's some of them.


GregHanson (approved):
LGTM, thanks for the follow-up

PR #472, opened 2023-03-17
Tweak config_dump format for certs

qfel:
- Expose errors to /config_dump instead of via tracing
 - Use empty ca_cert lists instead of a list with dummy object for unavailable certificates (note: can also not have the field altotheget for non-available certs, not sure which is preferred)

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -35,7 +35,7 @@ use tracing::error;
 
 use crate::config::Config;
 use crate::hyper_util::{empty_response, plaintext_response, Server};
-use crate::identity::{self, SecretManager};
+use crate::identity::SecretManager;
 use crate::tls::asn1_time_to_system_time;
 use crate::version::BuildInfo;
 use crate::workload::LocalConfig;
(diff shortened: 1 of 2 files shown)


src/admin.rs
```diff
@@ -72,12 +72,12 @@ pub struct CertDump {
     expiration_time: String,
 }
 
-#[derive(serde::Serialize, Debug, Clone)]
+#[derive(serde::Serialize, Debug, Clone, Default)]
 pub struct CertsDump {
     identity: String,
     state: String,
     // Make it an array to keep compatibility with Envoy's config_dump.
```
GregHanson, line 79: is this comment still relevant?

src/admin.rs
```diff
@@ -162,19 +162,21 @@ fn dump_cert(x509: &X509) -> CertDump {
...
             let mut dump = CertsDump {
                 identity: id.to_string(),
-                state,
-                ca_cert: [CertDump {
-                    ..Default::default()
-                }],
-                cert_chain: Vec::new(),
+                ..Default::default()
+            };
+            use crate::identity::CertState::*;
```
GregHanson, line 170: nit: what's the best practice in Rust for this?  most of the `use crate::` statements occur at the start of most of our files vs inline like above

qfel: No idea. I put it here since it's used only here in this file and locally made sense and made the code shorter. If I were to put it at the global scope I'd probably have to use CertState instead (since names like "Available" feel pretty amibuous).

src/admin.rs
qfel: Whoops, forgot to add the hold label. Will send out a separate PR.

PR #475, opened 2023-03-20
Add simple dashboard to ztunnel admin

hanxiaop:
For convenience. Used Envoy's style.
image

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -14,7 +14,6 @@
 
 use std::borrow::Borrow;
 use std::collections::HashMap;
-use std::fs;
 use std::sync::Arc;
 
 use std::{net::SocketAddr, time::Duration};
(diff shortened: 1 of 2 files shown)


src/admin.rs
```diff
@@ -129,12 +130,56 @@ impl Service {
...
+    for (index, (path, description)) in apis.iter().enumerate() {
+        api_rows.push_str(&format!(
+            "<tr class=\"{row_class}\"><td class=\"home-data\"><a href=\"{path}\">{path}</a></td><td class=\"home-data\">{description}</td></tr>\n",
+            row_class = if index % 2 == 1 { "gray" } else { "vert-space" },
+            path = path,
+            description = description
+        ));
+    }
+
+    let html = fs::read("src/assets/dashboard.html").unwrap();
```
howardjohn, line 170: We need to embed it, this file won't exist in the docker image (and shouldn't). There are rust macros to embed

howardjohn (commented):
Aren't most of these requiring POST anyways? I am not sure how much value this adds

hanxiaop:
> Aren't most of these requiring POST anyways? I am not sure how much value this adds

I want to add this because when I try to use commands like `istioctl dashboard envoy` or maybe a new command for ztunnel, it shows an empty page. Most of the time, I'm only interested in seeing the config dump, and I'm not sure if we need to display other APIs.

linsun:
This could be useful for dev and debugging imho

Issue #453, opened 2023-03-11
Ztunnel log format is not matching the Istio standard


hanxiaop:
The log format with colors is not readable directly when fetched via the bug-report.
https://github.com/istio/istio/pull/43878#discussion_r1132611318

MorrisLaw:
Hey @hanxiaop , I noticed this issue and was going to look into it. But it seems that you've already made the PR to fix it?

Should we close this out now? Or is there more work to be done?

hanxiaop:
> Hey @hanxiaop , I noticed this issue and was going to look into it. But it seems that you've already made the PR to fix it?
> 
> Should we close this out now? Or is there more work to be done?

@MorrisLaw I've only added the Ztunnel log to the bug report. However, if you look at the log format of other components and Ztunnel, you'll see a few differences. I was trying to address this earlier, but it appeared that it may require a lot of work. FYI, this https://github.com/istio/ztunnel/pull/476#discussion_r1147182614 might contain more details. If you're interested, feel free to unassign me and take on the work.

PR #476, opened 2023-03-20
Disable console log color

hanxiaop:
Part of https://github.com/istio/ztunnel/issues/453.
Since istio logs don't have colors, this PR will make ztunnel logs more convenient to be used in places like `bug-report`.


hanxiaop:
This unblocks https://github.com/istio/istio/pull/43878

hanxiaop:
> can we/should we auto-detect support? So local dev still has colors.
> 
> We may also want to adapt to Istio log format eventually

Yes. I think the color is still pretty useful, and I'm going to find a better way to handle the log.

src/telemetry.rs
```diff
@@ -37,8 +38,19 @@ pub fn setup_logging() {
     tracing_subscriber::registry().with(fmt_layer()).init();
 }
 
+fn is_local_dev() -> bool {
```
howardjohn, line 41: I was thinking something like https://github.com/dtolnay/isatty to detect if we should use color. Probably not this exact crate since its archive though, but detecting if we should use colors or not is fairly standard in linux programs

hanxiaop: Thanks for clarifying. I'll explore the crates.

hanxiaop: > We may also want to adapt to Istio log format eventually

By the way, IIUC, does this mean that we need to create a custom formatter to replace the current event formatter in https://github.com/istio/ztunnel/pull/476/files#diff-92d65ac8fde000b1d3ee1fda746b57e2586a6e7824e52256a9961016a10779deR53 under certain conditions?

hanxiaop: > I was thinking something like https://github.com/dtolnay/isatty to detect if we should use color. Probably not this exact crate since its archive though, but detecting if we should use colors or not is fairly standard in linux programs

Revised to `atty` to detect if we should use color.



> Yeah I think so, but not urgent. Might be a lot of work

Yes, I've looked into it, and it seems like there are many things that need to be dealt with, and some of the work may need to be finished upstream. I'll just leave the issue open and look back at it when I have time.

howardjohn (approved):

PR #486, opened 2023-03-30
Ambient xds evolution

kdorosh:
### Changes

Implementation for [ambient xds evolution](https://docs.google.com/document/d/1V5wkeBHbLSLMzAMbwFlFZNHdZPyUEspG4lHbnB0UaCg/edit#heading=h.qpcs4a7wyicb)

### Breaking Change

Note that this is a **breaking change** and breaks the xds contract between istiod and ztunnel. istiod used to send workloads, now it sends addresses which wrap either a workload or a service.

This break is perceived as ok since the istio changes are in flight / will be ready soon (not going to merge this and block devs without istio changes also validated in ci). We can only break this api contract because ambient is still in alpha.

### For Reviewers

In particular @howardjohn , the first commit here is the same as yours from https://github.com/istio/ztunnel/compare/master...howardjohn:ztunnel:ztunnel/service-xds?expand=1. The second commit are all my other changes.

Changes worth highlighting:
- I opted to use `staged_vips` in workload store rather than allowing all non-endpoints fields in `Service` struct to be optional
  - I thought this would make it easier for operator to see if any VIPs were stuck in `staged_vips`. It also makes the service struct simpler. This is not a strong feeling, I'm open to discuss doing it other ways
- Moved the authorization proto go package to avoid a conflict with workload.proto (both define an `Address` message)


src/proxy/inbound.rs
```diff
@@ -363,12 +363,26 @@ impl Inbound {
...
     ) -> (bool, bool) {
-        let has_waypoint = !upstream.waypoint_addresses.is_empty();
-        for i in upstream.waypoint_addresses.iter() {
-            if workloads.fetch_workload(i).await.map(|w| w.identity()) == conn.src_identity {
-                return (has_waypoint, true);
+        let has_waypoint = upstream.waypoint.address.is_some();
+        let wp_ip = match upstream.waypoint.address.as_ref() {
+            Some(addr) => match addr {
+                gatewayaddress::Address::IP(wp_ip) => wp_ip,
+                gatewayaddress::Address::Hostname(_) => return (has_waypoint, false),
```
howardjohn, line 370: can we add TODO to look this up from Service?

src/proxy/inbound.rs
```diff
@@ -363,12 +363,26 @@ impl Inbound {
...
+            },
+            None => return (has_waypoint, false),
+        };
+
+        if let Some(svc) = workloads.service_by_vip(*wp_ip).await {
```
howardjohn, line 375: nit: Having a `fetch_address` that has this logic would be nice. Then we don't need to handle service vs workload

src/proxy/outbound.rs
```diff
@@ -350,14 +350,28 @@ impl OutboundConnection {
...
-            let waypoint_address = us.workload.choose_waypoint_address().unwrap();
+        let waypoint_addr = &us.workload.waypoint;
+        if let Some(waypoint_address) = &waypoint_addr.address {
             // Even in this case, we are picking a single upstream pod and deciding if it has a remote proxy.
             // Typically this is all or nothing, but if not we should probably send to remote proxy if *any* upstream has one.
-            let waypoint_workload = match self.pi.workloads.fetch_workload(&waypoint_address).await
-            {
-                Some(wl) => wl,
-                None => return Err(Error::UnknownWaypoint(waypoint_address, downstream)),
+            let wp_ip_addr = match waypoint_address {
```
howardjohn, line 357: nit: would be nice to abstract this into something that is just `Workload -> Option<Workload>` (taking user workload -> waypoint`)

src/workload.rs
```diff
@@ -78,12 +82,32 @@ impl TryFrom<Option<xds::istio::workload::WorkloadStatus>> for HealthStatus {
...
+
 #[derive(Debug, Hash, Eq, PartialEq, Clone, serde::Serialize, serde::Deserialize)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct Workload {
     pub workload_ip: IpAddr,
     #[serde(default)]
-    pub waypoint_addresses: Vec<IpAddr>,
+    pub waypoint: GatewayAddress,
+    #[serde(default)]
+    pub network_gateway: GatewayAddress,
```
howardjohn, line 110:
```suggestion
    pub waypoint: Option<GatewayAddress>,
    #[serde(default)]
    pub network_gateway: Option<GatewayAddress>,
```

Then remove the Option<> inside of GatewayAddress.

We should either allow having an ip/host+port, or not, which is better represented with above

src/workload.rs
```diff
@@ -132,10 +156,11 @@ impl Workload {
         }
     }
     pub fn choose_waypoint_address(&self) -> Option<IpAddr> {
```
howardjohn, line 158: name is a bit misleading now since its not actually _choosing_ anymore.

src/workload.rs
```diff
@@ -191,20 +216,69 @@ pub fn byte_to_ip(b: &bytes::Bytes) -> Result<IpAddr, WorkloadError> {
...
+                address: match &w.address {
+                    Some(a) => match a {
+                        xds::istio::workload::gateway_address::Address::Ip(ip) => {
+                            Some(gatewayaddress::Address::IP(byte_to_ip(ip)?))
+                        }
+                        xds::istio::workload::gateway_address::Address::Hostname(hn) => {
+                            Some(gatewayaddress::Address::Hostname(hn.clone()))
+                        }
+                    },
+                    None => None,
```
howardjohn, line 245: This should be an error

src/workload.rs
```diff
@@ -191,20 +216,69 @@ pub fn byte_to_ip(b: &bytes::Bytes) -> Result<IpAddr, WorkloadError> {
...
+                },
+                port: w.port as u16,
+            },
+            None => GatewayAddress {
+                address: None,
+                port: 0,
+            },
+        };
+
+        let network_gw: GatewayAddress = match &resource.network_gateway {
```
howardjohn, line 255: nit: abstract this out? maybe a `TryFrom<XdsGatewayAddress>  for GatewayAddress`?

src/workload.rs
```diff
@@ -601,17 +741,25 @@ impl WorkloadStore {
...
+                    self.workload_to_vip.entry(wip).or_default().insert(vip);
+                } else {
+                    // Can happen due to ordering issues
+                    trace!("pod has VIP {vip}, but VIP not found");
+                    if let Some(ep_map) = self.staged_vips.get_mut(&vip) {
+                        ep_map.insert(wip, ep);
+                    } else {
+                        let mut ep_map = HashMap::new();
+                        ep_map.insert(wip, ep);
+                        self.staged_vips.insert(vip, ep_map);
```
howardjohn, line 761: I don't see anywhere where staged_vips is removed when we remove a workload. Can you add a testcase for this?

src/workload.rs
```diff
@@ -601,17 +741,25 @@ impl WorkloadStore {
...
+                    address: wip,
+                    port: pl.into(),
+                };
+                if let Some(svc) = self.vips.get_mut(&vip) {
+                    svc.endpoints.insert(ep.address, ep);
+                    self.workload_to_vip.entry(wip).or_default().insert(vip);
+                } else {
+                    // Can happen due to ordering issues
+                    trace!("pod has VIP {vip}, but VIP not found");
+                    if let Some(ep_map) = self.staged_vips.get_mut(&vip) {
```
howardjohn, line 756: nit: I think you can do something like `self.staged_vips.entry(vip).or_default().insert(ep)`

src/workload.rs
```diff
@@ -676,6 +824,19 @@ impl WorkloadStore {
         self.workloads.insert(wip, w);
     }
 
+    fn insert_svc(&mut self, svc: Service) {
+        // if svc already exists, just add new endpoints to existing svc
```
howardjohn, line 828: this never removes endpoints

howardjohn: Actually, looking closer maybe this is bug-free since its only called with Services that have endpoints from staged_vips, but it feels very fragile/confusing

howardjohn:
Overall looks pretty good, mostly just some nits

/test all

src/proxy/inbound.rs
```diff
@@ -363,12 +363,26 @@ impl Inbound {
...
-            if workloads.fetch_workload(i).await.map(|w| w.identity()) == conn.src_identity {
-                return (has_waypoint, true);
+        let has_waypoint = upstream.waypoint.address.is_some();
+        let wp_ip = match upstream.waypoint.address.as_ref() {
+            Some(addr) => match addr {
+                gatewayaddress::Address::IP(wp_ip) => wp_ip,
+                gatewayaddress::Address::Hostname(_) => return (has_waypoint, false),
+            },
+            None => return (has_waypoint, false),
+        };
```
ilrudie, line 373: **nit:** `has_waypoint` can probably just be hard coded to `true` instead of checking `is_some()`. The next line is matching on the same field and there is a `None` arm that handles the case where `has_waypoint` is false by returning immediately. That arm could just be `return (false, false)`

src/workload.rs
```diff
@@ -78,12 +82,32 @@ impl TryFrom<Option<xds::istio::workload::WorkloadStatus>> for HealthStatus {
     }
 }
 
+#[derive(Debug, Hash, Eq, PartialEq, Clone, Default, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct GatewayAddress {
+    #[serde(default)]
+    pub address: Option<gatewayaddress::Address>,
```
ilrudie, line 89: Expanding on what I think [John's comment](https://github.com/istio/ztunnel/pull/486/files#r1157494478) was. As far as I can tell a `GatewayAddress` is not really valid if the `address` field is `None`. All the "valid" states for `address` are a `Hostname` or an `IP` which are covered by the inner `enum Address` and the outer `enum Option` really only allows for expressing an invalid state. If there is some case where "None" would be used as a way to represent "localhost" for instance we should just add a more explicit default instead of relying on an implied meaning for `None`.

ilrudie (commented):
Adding one nit and (I hope correctly) expanding a little on one of John's comments.

proto/workload.proto
```diff
@@ -17,6 +17,55 @@ syntax = "proto3";
...
+  string namespace = 2;
+  // Hostname represents the FQDN of the service.
+  // For Kubernetes, this would be <name>.<namespace>.svc.<cluster domain>
+  string hostname = 3;
+  // Address represents the IPv4/IPv6 address for the service.
+  // This should be globally unique.
+  // This should not have a port number.
+  // TODO: Add network as discriminator
+  // TODO: Add dual stack support
+  bytes address = 4;
```
hzxuzhonghu, line 61: which ip address should this be when it is in multi clusters

proto/workload.proto
```diff
@@ -45,7 +95,12 @@ message Workload {
   string service_account = 7;
 
   // If present, the waypoint proxy for this workload.
-  repeated bytes waypoint_addresses = 8;
+  // All incoming requests must go through the waypoint.
+  GatewayAddress waypoint = 8;
+
+  // If present, East West network gateway this workload can be reached through.
+  // Requests from remote networks should traverse this gateway.
+  GatewayAddress network_gateway = 19;
```
hzxuzhonghu, line 103: how can present multi network gateways

src/workload.rs
hzxuzhonghu: is it to return the waypoint cluster ip?

hzxuzhonghu (commented):
I am concerned if this api is released, it can be very tricky to extend for multi networks case without breaking compatability

howardjohn:
The API in the Google doc has has some updates since this PR. specifically
around multi network. Please comment there if you think there are concerns

kdorosh:
I'll be working to address these PR comments early this week and then update this PR to reflect the latest state of the protos in [the design doc](https://docs.google.com/document/d/1V5wkeBHbLSLMzAMbwFlFZNHdZPyUEspG4lHbnB0UaCg/edit#heading=h.qpcs4a7wyicb), which I believe addresses most of the multi network concerns (as each service is multi network). I will keep that in mind as implementing and see if the design needs further refinement

src/proxy/inbound.rs
kdorosh: see https://github.com/istio/ztunnel/pull/486/commits/f319ce5041d37875353d2b5acf685200caeb535b

src/proxy/inbound.rs
kdorosh: see https://github.com/istio/ztunnel/pull/486/commits/f319ce5041d37875353d2b5acf685200caeb535b

src/proxy/outbound.rs
kdorosh: see b024929468b92d121376987ef3330847ecbc3891

src/workload.rs
kdorosh: see 9ec9a1260fce16f8abc343596431ccb2fdf9a39f

src/proxy/inbound.rs
kdorosh: this is going to change (no more `is_some` on address) once i implement https://github.com/istio/ztunnel/pull/486#discussion_r1157494478

proto/workload.proto
kdorosh: I believe per the design doc this will become `NetworkAddress` once I implement that (before this merges and I ask for more reviews)

proto/workload.proto
kdorosh: this will point to a service which itself can point to workloads in many networks

src/workload.rs
kdorosh: implemented in https://github.com/istio/ztunnel/pull/486/commits/6d7827f3110d69e6b1eb94d5975120c92fa3cf79

src/proxy/inbound.rs
kdorosh: implemented in https://github.com/istio/ztunnel/pull/486/commits/6d7827f3110d69e6b1eb94d5975120c92fa3cf79 and https://github.com/istio/ztunnel/pull/486/commits/38d0a5394347413e198ce5afd6e82ce1f07c1913

src/workload.rs
kdorosh: implemented in https://github.com/istio/ztunnel/pull/486/commits/0189043ac3e0d065052dacfbfffd1bd9223b43d8

src/workload.rs
kdorosh: implemented in https://github.com/istio/ztunnel/pull/486/commits/38d0a5394347413e198ce5afd6e82ce1f07c1913

src/proxy/inbound.rs
kdorosh: this function has been updated, closing this as dated

src/workload.rs
kdorosh: https://github.com/istio/ztunnel/pull/486/commits/a08e65d78de4d32530baba79c0608c404b6073d4 fixes the bug for removing endpoints and tests it

src/workload.rs
kdorosh: I think this is addressed by https://github.com/istio/ztunnel/pull/486/commits/9a0b67b8de69a9302619da99f300bdff0df16555 which moves this mutation to the shared implementation for insert_svc and https://github.com/istio/ztunnel/pull/486/commits/a08e65d78de4d32530baba79c0608c404b6073d4 which fixes the bug for removing endpoints and tests it

kdorosh:
will address merge conflicts monday. all other work on this PR is done. feel free to review the PR / direction again. @GregHanson is working on the control plane side of the changes to validate this against istio ci as well

proto/workload.proto
kdorosh: the api has been updated, please take a look again

proto/workload.proto
kdorosh: the api has been updated, please take a look again

src/proxy/outbound.rs
```diff
@@ -369,28 +379,27 @@ impl OutboundConnection {
...
-                destination: target,
-                destination_workload: Some(us.workload),
-                expected_identity: Some(waypoint_workload.identity()),
-                gateway: SocketAddr::from((waypoint_address, 15008)),
-                // Let the client remote know we are on the inbound path.
-                direction: Direction::Inbound,
-                request_type: RequestType::ToServerWaypoint,
-            });
+        match self.pi.workloads.find_waypoint(us.workload.clone()).await {
+            Ok(None) => {} // do nothing
```
ilrudie, line 383: **suggestion:** should we add some sort of debug log entry here so it's more straightforward to understand that we landed in this arm of the match statement?

src/config.rs
```diff
@@ -277,6 +280,7 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
         inbound_plaintext_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15006),
         outbound_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15001),
 
+        network: parse(NETWORK)?.unwrap_or_else(|| "defaultnw".to_string()),
```
ilrudie, line 283: **suggestion:** "defaultnw" as a hardcoded str is used quite a few times in different places. consider making it a `const`

src/workload.rs
```diff
@@ -277,6 +364,23 @@ impl xds::Handler<XdsWorkload> for Arc<Mutex<WorkloadStore>> {
     }
 }
 
+impl xds::Handler<XdsAddress> for Arc<Mutex<WorkloadStore>> {
+    fn handle(&self, updates: Vec<XdsUpdate<XdsAddress>>) -> Result<(), Vec<RejectedConfig>> {
+        let mut wli = self.lock().unwrap();
```
ilrudie, line 369: **suggestion:** consider using `expect` instead of `unwrap` so it might be easier to locate the source if `lock()` returns an error variant and this panics.

ilrudie (commented):
Just leaving a few non-blocking suggestions to consider

stevenctl:
LGTM

src/test_helpers/components.rs
```diff
@@ -224,6 +234,42 @@ impl<'a> TestWorkloadBuilder<'a> {
             .namespaces
             .child(&self.w.workload.node, &self.w.workload.name)?;
         self.w.workload.workload_ip = network_namespace.ip();
+
+        for (vip, ports) in &self.w.vips {
```
howardjohn, line 238: nit: Now that we have services we may want to replace the vip() method with one that adds a full Service instead of reverse engineering it here?

src/workload.rs
```diff
@@ -131,11 +167,14 @@ impl Workload {
             service_account: self.service_account.clone(),
         }
     }
-    pub fn choose_waypoint_address(&self) -> Option<IpAddr> {
-        self.waypoint_addresses
-            .iter()
-            .choose(&mut rand::thread_rng())
-            .copied()
+    pub fn waypoint_svc_ip_address(&self) -> Option<IpAddr> {
```
howardjohn, line 170: nit:

```rust
        self.waypoint.as_ref().and_then(|gw_address| {
            match &gw_address.destination {
                gatewayaddress::Destination::Hostname(_) => None, // TODO: support this
                gatewayaddress::Destination::Address(ip) => Some(ip.address),
            }
        })
```

is more idiomatic

src/workload.rs
```diff
@@ -240,7 +327,7 @@ impl TryFrom<&XdsWorkload> for Workload {
                 resource.status,
             ))?,
 
-            native_hbone: resource.native_hbone,
+            native_hbone: resource.native_tunnel,
```
howardjohn, line 330: nit: rename native_hbone in the internal representation as well

src/workload.rs
howardjohn: I think the panic will show the source file/line

src/workload.rs
```diff
@@ -512,59 +620,173 @@ impl WorkloadInformation {
...
+        hbone_port: u16,
+    ) -> Option<Upstream> {
+        self.fetch_address(network_addr(network, addr.ip())).await;
         let wi = self.info.lock().unwrap();
-        wi.find_upstream(addr, hbone_port)
+        wi.find_upstream(network, addr, hbone_port)
+    }
+
+    pub async fn find_waypoint(&self, wl: Workload) -> Result<Option<Upstream>, WaypointError> {
+        if let Some(gw_address) = &wl.waypoint {
```
howardjohn, line 635: nit:

```rust

let Some(gw_address) = &wl.waypoint else {
  return Ok(None)
}
```

can lead to less indent needed

src/workload.rs
```diff
@@ -512,59 +620,173 @@ impl WorkloadInformation {
...
+            // Even in this case, we are picking a single upstream pod and deciding if it has a remote proxy.
+            // Typically this is all or nothing, but if not we should probably send to remote proxy if *any* upstream has one.
+            let wp_nw_addr = match &gw_address.destination {
+                gatewayaddress::Destination::Address(ip) => ip,
+                gatewayaddress::Destination::Hostname(_) => {
+                    info!(%wl.name, "waypoint hostname lookup not supported yet");
+                    return Ok(None); // TODO: implement unsupported feature
```
howardjohn, line 642:
```suggestion
                    return Err(WaypointError::FindWaypointError);
```
?

src/workload.rs
```diff
@@ -512,59 +620,173 @@ impl WorkloadInformation {
...
+            debug!(%key, "sending demand request");
+            demand.demand(key.to_string()).await.recv().await;
+            debug!(%key, "on demand ready");
+        }
+    }
+
+    // keep private so that we can ensure that we always use fetch_address
+    fn find_address(&self, network_addr: NetworkAddress) -> Option<Address> {
+        // 1. handle workload ip, if workload not found fallback to service.
+        match self.find_workload(&network_addr) {
```
howardjohn, line 687: nit: don't use self.find_workload here. Th at means we lock,unlock 2x. We can just do the locking ourselves so we don't need 2x locking

src/workload.rs
ilrudie: You're right John. It will show file and line. I prefer the expect message myself. I also believe it's considered "good form" for rust to not have unwrap in production code. If you want unwrap functionality use expect along with a message about why it's thought to never panic. Sometimes we're smarter than the compiler is and for that there's expect; for everything else there pattern matching.

Anyway, don't block on this. It's not a hill anyone need die on.

ilrudie: `clippy::unwrap_used` could be added in CI at some point if we decide we want to remove all the unwraps from the codebase and prevent it from being introduced again

src/workload.rs
```diff
@@ -512,59 +620,173 @@ impl WorkloadInformation {
...
+        };
+        Ok(svc)
     }
 }
 
 /// A WorkloadStore encapsulates all information about workloads in the mesh
 #[derive(serde::Serialize, Default, Debug)]
 pub struct WorkloadStore {
-    workloads: HashMap<IpAddr, Workload>,
+    workloads: HashMap<NetworkAddress, Workload>,
```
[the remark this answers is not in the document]
howardjohn: We can have custom serialization. Probably good to keep our internal types clean and convert as needed for debug

src/test_helpers/components.rs
```diff
@@ -34,6 +34,8 @@ pub struct WorkloadManager {
     namespaces: netns::NamespaceManager,
     /// workloads that we have constructed
     workloads: Vec<LocalWorkload>,
+    /// services that we have constructed. VIP -> SVC
+    services: HashMap<IpAddr, Service>,
```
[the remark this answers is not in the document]
kdorosh: this is only test code which uses the default network. i specifically didn't update this here in an attempt to keep this PR smaller since it's already quite sizeable. open to adding more multi network stuff in follow up PRs?

I really just wanted the API correct and compiling here, so future PRs don't make single-network assumptions unconsciously. it was not my intent to get 100% multi network support here

src/workload.rs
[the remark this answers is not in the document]
kdorosh: I agree custom serialization and clean internal types is the way to go, implemented in https://github.com/istio/ztunnel/pull/486/commits/4d024f37cd2aa900965f377b748f0b5ca275c57c

examples/localhost.yaml
```diff
@@ -28,3 +30,29 @@ policies:
...
+- name: local
+  namespace: default
+  hostname: example.com
+  addresses:
+    - network: defaultnw
+      address: "127.10.0.1"
+  endpoints:
+    ? address: "127.0.0.1"
+      network: defaultnw
+    :
```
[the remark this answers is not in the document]
kdorosh: custom serialization implemented in https://github.com/istio/ztunnel/pull/486/commits/4d024f37cd2aa900965f377b748f0b5ca275c57c supports the `network/IP` syntax now

src/proxy/inbound_passthrough.rs
```diff
@@ -134,7 +140,11 @@ impl InboundPassthrough {
 
         // Find source info. We can lookup by XDS or from connection attributes
         let source_workload = if let Some(source_ip) = source_ip {
-            pi.workloads.fetch_workload(&source_ip).await
+            let network_addr_srcip = NetworkAddress {
+                network: pi.cfg.network.clone(), // inbound request must be on our network
```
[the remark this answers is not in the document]
kdorosh: added more detail in https://github.com/istio/ztunnel/pull/486/commits/21c681252f366afc80f444c75be044854a449442

src/proxy/inbound.rs
```diff
@@ -304,8 +311,12 @@ impl Inbound {
 
                 let baggage =
                     parse_baggage_header(req.headers().get_all(BAGGAGE_HEADER)).unwrap_or_default();
+                let src_network_addr = NetworkAddress {
+                    network: conn.dst_network.to_string(), // inbound request source network must match dst network
```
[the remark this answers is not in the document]
kdorosh: ah good catch, copy paste error but yes the idea was to punt on this since the solution isn't trivial. added more detail in https://github.com/istio/ztunnel/pull/486/commits/21c681252f366afc80f444c75be044854a449442 and opened a GH issue in this repo to track potential required updates to HBONE itself to carry the originating network

src/proxy/outbound.rs
hzxuzhonghu: Isn't it checking whether it has waypoint? If not, should drop

src/proxy/outbound.rs
```diff
@@ -369,28 +379,27 @@ impl OutboundConnection {
...
+                    // Always use HBONE here
+                    protocol: Protocol::HBONE,
+                    source: source_workload,
+                    // Use the original VIP, not translated
+                    destination: target,
+                    destination_workload: Some(us.workload),
+                    expected_identity: Some(waypoint_workload.identity()),
+                    gateway: wp_socket_addr,
+                    // Let the client remote know we are on the inbound path.
+                    direction: Direction::Inbound,
```
hzxuzhonghu, line 398: This looks very confusing to me, @howardjohn Can you elaborate it

src/workload.rs
kdorosh: addressed in https://github.com/istio/ztunnel/pull/486/commits/ecc06af22e028572215f01f205db579f13801728

src/workload.rs
kdorosh: addressed in a slightly different way in https://github.com/istio/ztunnel/pull/486/commits/ecc06af22e028572215f01f205db579f13801728. I opted to return an error here instead of None, so the proposed code here does not work anymore. But I think this is clearer for devs in the meantime until hostname is supported

proto/workload.proto
```diff
@@ -29,14 +97,14 @@ message Workload {
...
-  // TODO: Add network as discriminator
+  // TODO: support dual stack by making this repeated
   bytes address = 3;
   // Network represents the network this workload is on. This may be elided for the default network.
   // A (network,address) pair makeup a unique key for a workload *at a point in time*.
   string network = 4;
 
   // Protocol that should be used to connect to this workload.
-  Protocol protocol = 5;
+  TunnelProtocol tunnel_protocol = 5;
```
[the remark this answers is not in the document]
kdorosh: this is the latest per the design doc: 
Screenshot from 2023-05-08 10-23-07

what do you think this should say?

src/workload.rs
kdorosh: addressed in https://github.com/istio/ztunnel/pull/486/commits/7b747454057e6cb8ec04db2803bcb6ce8728a11e

src/workload.rs
kdorosh: addressed in https://github.com/istio/ztunnel/pull/486/commits/0d9ca6bfffa74f55f5af673d37c9592767df9087

src/workload.rs
kdorosh: addressed in https://github.com/istio/ztunnel/pull/486/commits/f22b23d2a70ec5cb89d78ccb81b3e4ff1cb2615f

src/workload.rs
```diff
@@ -512,59 +620,173 @@ impl WorkloadInformation {
...
+            };
+        }
+        Ok(None)
     }
 
     // Support workload and VIP
-    // It is to do on demand workload fetch if necessary, it handles both workload ip and clusterIP
-    async fn fetch_address(&self, addr: &SocketAddr) {
+    // It is to do on demand workload fetch if necessary, it handles both workload ip and services
+    pub async fn fetch_address(&self, network_addr: NetworkAddress) -> Option<Address> {
```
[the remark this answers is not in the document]
kdorosh: addressed in https://github.com/istio/ztunnel/pull/486/commits/e51aaa2aeae34b8620de65d6573072b6f1643477

src/rbac.rs
```diff
@@ -45,6 +45,7 @@ pub struct Authorization {
 pub struct Connection {
     pub src_identity: Option<Identity>,
     pub src_ip: IpAddr,
+    pub dst_network: String,
```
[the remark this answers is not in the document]
kdorosh: this was necessary for `assert_rbac()` to lookup the destination workload and enforce rbac on both inbound and outbound requests

src/proxy/outbound.rs
kdorosh: this is unchanged from before. `Direction::Inbound` is useful for logging, and the rest of this struct is unchanged in shape in an attempt to keep the PR size here as small as possible; although I agree we could probably refactor this into a more intuitive representation for an outgoing request

src/proxy/inbound_passthrough.rs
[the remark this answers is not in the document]
kdorosh: more detail also in https://github.com/istio/ztunnel/pull/486/commits/044c7bcf55f1f3b5bfb33ad6b1e10d4465cb9d8d

src/proxy/inbound.rs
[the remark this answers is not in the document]
kdorosh: I think we might be able to resolve this by implementing a `check_gateway` function similar to `check_waypoint` in `inbound.rs` and leaving the optional source workload there `None` if we receive from a gateway. looking into this now

src/config.rs
kdorosh: updated default network to `""` in https://github.com/istio/ztunnel/pull/486/commits/54631ccc0fa4a2fd54db23cb60e3f629fc30b350 cc @GregHanson

src/proxy/outbound.rs
```diff
@@ -722,8 +749,17 @@ mod tests {
             "127.0.0.1",
             "127.0.0.2:80",
             XdsWorkload {
+                network: "defaultnw".to_string(),
```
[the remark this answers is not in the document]
kdorosh: added some in https://github.com/istio/ztunnel/pull/486/commits/b45f85fb006e86900313fcdfc07c89fe6f2631e7

kdorosh (commented):
addressed all PR feedback and responded to anything that was unclear

kdorosh (commented):
addressed all PR feedback and responded to anything that was unclear

proto/workload.proto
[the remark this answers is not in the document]
hzxuzhonghu: TunnelProtocol

src/proxy/outbound.rs
hzxuzhonghu: SO I understand with it is now, if the waypoint has no workload found, it simply send the traffic to dest workload without through waypoint, right?

hzxuzhonghu: If so before a waypoint running up, the request to upstream may skip some auth policy

kdorosh: > if the waypoint has no workload found, it simply send the traffic to dest workload without through waypoint, right?

correct

> If so before a waypoint running up, the request to upstream may skip some auth policy

no; the contract for `find_waypoint(workload)` is that if the workload is configured with a waypoint service, then it either finds a backing workload for that service or returns an error. we only return `Ok(None)` if the workload itself does not have a waypoint at all

proto/workload.proto
```diff
@@ -17,7 +17,77 @@ syntax = "proto3";
...
+  // For Kubernetes, this is the Service name.
+  string name = 1;
+  // Namespace represents the namespace for the service.
+  string namespace = 2;
+  // Hostname represents the FQDN of the service.
+  // For Kubernetes, this would be <name>.<namespace>.svc.<cluster domain>
+  // TODO: support this field
+  string hostname = 3;
+  // Address represents the addresses the service can be reached at.
+  // There may be multiple addresses for a single service if it resides in multiple networks.
```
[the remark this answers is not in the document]
stevenctl: If it resides in two clusters for one network, it will get two cluster VIPs. 

https://github.com/istio/istio/blob/d1ec7e02a1a153869b176c53fa150834e0ef3f24/pilot/pkg/model/service.go#L80-L82

examples/localhost.yaml
```diff
@@ -28,3 +34,40 @@ policies:
     name: deny-9999
     namespace: default
     scope: Namespace
+services:
+- name: local
+  namespace: default
+  hostname: example.com
+  addresses:
+    - /127.10.0.1
+  endpoints:
```
howardjohn, line 43: seems like in our yaml we should either use `vips` or `endpoints`, needing to specify both seems tedious?

proto/workload.proto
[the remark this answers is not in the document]
howardjohn: I think this is referring to L108 `// Protocol`. But proto docs don't follow the Go pattern of starting the comment with the variable name so I think its fine personally

examples/localhost.yaml
kdorosh: not sure i understand the comment here, so ill elaborate on the intent for now:

so this is test config for our internal `Service` struct, which has both the service vips (in `addresses` and the endpoints behind the service `endpoints` which are basically the workloads)

this `Service` struct is constructed internally after we receive the xds workloads and services, in whatever order. if the xds service arrives we fill endpoints with all workloads it selects we currently know about

if the xds workload arrives first, we add its virtual IPs to `staged_vips` internal map so if/when the corresponding xds service arrives we can populate this internal `Service` struct with the right endpoints

howardjohn: This doesn't _have_ to be 1:1 with the XDS struct. Even in the workload we already have LocalWorkload which differs

kdorosh: ah I see what you mean, makes sense :+1: . I'll take a look at that update

kdorosh: addressed in https://github.com/istio/ztunnel/pull/486/commits/b5c28c2bf3503dcdf79b1f6a1e23a695e45681e5

src/proxy/inbound.rs
```diff
@@ -266,16 +271,28 @@ impl Inbound {
...
                         .body(Empty::new())
                         .unwrap());
                 };
-                let (has_waypoint, from_waypoint) =
-                    Self::check_waypoint(&workloads, &upstream, &conn).await;
+                let has_waypoint = upstream.waypoint.is_some();
+                let from_waypoint = Self::check_waypoint(&workloads, &upstream, &conn)
+                    .await
+                    .unwrap();
+                let from_gateway = Self::check_gateway(&workloads, &upstream, &conn)
```
howardjohn, line 289: this doesn't seem ideal since the identity of the gateway doesn't technically imply its from another network. But probably good enough

src/proxy/inbound.rs
```diff
@@ -266,16 +271,28 @@ impl Inbound {
...
                         .status(StatusCode::NOT_FOUND)
                         .body(Empty::new())
                         .unwrap());
                 };
-                let (has_waypoint, from_waypoint) =
-                    Self::check_waypoint(&workloads, &upstream, &conn).await;
+                let has_waypoint = upstream.waypoint.is_some();
+                let from_waypoint = Self::check_waypoint(&workloads, &upstream, &conn)
+                    .await
+                    .unwrap();
```
howardjohn, line 288: nit: if we are always going to unwrap these -- which looks like its because its unimplemented, may be simpler to do it in the check_gateway_address function. You can do `todo!()`

src/proxy/outbound.rs
howardjohn: The original code was copy+pasted from a prototype written in Go that had a different architecture :-) definitely can improve

src/proxy/inbound.rs
howardjohn: might be useful elsewhere, unless we plan to go awhile with things unimplemented; in that case, we should not unwrap. Basically, proper errors or `todo!()` is preferred. Not a blocker though

howardjohn (approved):

PR #491, opened 2023-04-04
wip: output certs for xds

stevenctl:


src/identity/manager.rs
```diff
@@ -318,12 +315,16 @@ impl Worker {
                             (CertState::Available(certs), refresh_at)
                         },
                     };
+
+                    // TODO log. also don't block on this before updating cert for user traffic
+                    let _ = self.output_cert(&state).await;
```
adiprerepa, line 320: why shouldn't we block? shouldn't the cert update before we are able to handle more traffic?

stevenctl: This code path is used both to fetch certs for user traffic, and the new code writes the certs to the filesystem to use for xDS/CA authentication.

costinm:
Added few comments on the bug - I don't think ztunnel should save or refresh certs, and the simplest option is to 
start using the MDS to get fresh JWTs.

This is not only for dedicated mode - ztunnel on a VM may serve multiple containers ( with some checks on 
Istiod side to verify the same-node property ). I think 2 changes are needed - one to allow ztunnel to get JWTs on demand
and from a MDS, and a separate one to allow ztunnel to authenticate with mTLS client certs managed by an external
agent.

hzxuzhonghu (commented):
Wondering why not expose secret dumping capability like sds dump

PR #495, opened 2023-04-07
Fix rbac rule matching

howardjohn:
Fixes https://github.com/istio/istio/issues/44256

Probably needs more tests but i am signing off for the weekend so snapshoting where I got

### DIFF
--- a/src/rbac.rs
+++ b/src/rbac.rs
@@ -153,6 +153,10 @@ impl Authorization {
                     trace!(matches = rule_match, "rule");
                 }
                 group_match &= rule_match;
+                if !group_match {
+                    // Short circuit
+                    break;
+                }
             }
             trace!(matches = group_match, "group");
             if group_match {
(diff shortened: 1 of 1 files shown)


peterj:
cc @ilrudie

src/rbac.rs
```diff
@@ -94,17 +94,18 @@ impl Authorization {
...
-                for mg in group.iter() {
+        for group in self.groups.iter() {
+            // Group typically has 1-3 elements (from,to,when)
+            // If ALL group matches, it is a match...
+            let mut group_match = true;
+            for rule in group.iter() {
+                // We can have multiple rules in a group, for example "Match ns=A,SA=B or ns=C"
+                // So we need ANY rule to match...
+                let mut rule_match = false;
+                for mg in rule.iter() {
```
ilrudie, line 105: **Suggestion:** In the proto each individual item in the matches of a  `message Rules` is of type `message Match`. Could `mg` be changed to something more descriptive or could a comment be added explaining what mg means? I do see why calling the variable `match` in this scope could be undesirable and potentially unclear as well.

src/rbac.rs
```diff
@@ -94,17 +94,18 @@ impl Authorization {
...
             return false;
         }
-        for rule in self.groups.iter() {
-            // If ANY rule matches, it is a match...
-            let mut rule_match = true;
-            for group in rule.iter() {
-                // We need ALL groups to match...
-                let mut group_match = true;
-                for mg in group.iter() {
+        for group in self.groups.iter() {
```
ilrudie, line 97: **Suggestion:** This naming is consistent with the internal API we're using but rubs a bit against the public API for an AuthorizationPolicy where what we're internally referring to as a group is called a Rule. It might be helpful to add a comment noting that naming comes from `proto/authorization.proto` and is inconsistent with `v1beta1 AuthorizationPolicy`.

Could alternatively or also be added to `impl TryFrom<&XdsRbac> for Authorization`.

ilrudie (commented):
Adding a couple of non-blocking suggestions based on my experience looking into the issue. Overall though it looks like the changes bring the naming of variables closer in line with the proto as well as correcting behavior raised by Peter.

src/rbac.rs
```diff
@@ -139,17 +140,22 @@ impl Authorization {
...
-                    trace!(matches = true, "empty group");
+                if rule.is_empty() {
+                    rule_match = true;
+                    trace!(matches = rule_match, "empty rule");
                 } else {
-                    trace!(matches = group_match, "group");
+                    trace!(matches = rule_match, "rule");
                 }
-                rule_match &= group_match;
+                group_match &= rule_match;
```
ilrudie, line 155: **suggestion:** Should this short circuit? If group match ever becomes false I think we can move on to assessing the next group because the current one is going to be false.

ilrudie: Tried a quick test of this locally. Here's the short circuit logic + revised test case to exercise short circuiting.
https://github.com/ilrudie/istio-ztunnel/commit/5ec391fdd0abd410050ef5bc0167465edb422e02

ilrudie: Happy to raise this as it's own PR to keep this one small and focused on the requested fix if that would be preferred.

howardjohn: I included it into this one, good call

src/rbac.rs
howardjohn: Yeah I think we should actually change `authorization.proto` names to be more like AuthorizationPolicy for clarity; we can make that breaking change now since we are pre-launch

ilrudie: Probably best to save breaking changes for after Kubecon EU. I'd be down to work on changes to the proto though.

ilrudie: Created #500 to track proposals, decision and potentially implementation

src/rbac.rs
```diff
@@ -139,17 +140,26 @@ impl Authorization {
...
-                if group.is_empty() {
-                    trace!(matches = true, "empty group");
+                if rule.is_empty() {
+                    rule_match = true;
+                    trace!(matches = rule_match, "empty rule");
                 } else {
-                    trace!(matches = group_match, "group");
+                    trace!(matches = rule_match, "rule");
+                }
+                group_match &= rule_match;
```
hzxuzhonghu, line 155: Why it is  still `AND`

ilrudie: For a "group" to match all elements (to/from/when in the authz policy) must match.

hzxuzhonghu: Right, i misread it. I thought group_match means all groups match

howardjohn:
@stevenctl can you take a look?

PR #498, opened 2023-04-11
Validate SAN trust domain during mTLS

### LINKED ISSUE #359: Verify inbound certificates properly
Currently, we only check the full SAN with a authz policy. otherwise we only check the root cert.

We also need to check the trust domain matches to get parity with Envoy's enforcement.

We should also double check (ie actually test) the root cert check actually happens, just in case - but it does appear to be there.

keithmattix:
Fixes #359

### DIFF
--- a/src/tls/boring.rs
+++ b/src/tls/boring.rs
@@ -316,11 +316,16 @@ impl Certs {
         ssl::SslVerifyMode::PEER | ssl::SslVerifyMode::FAIL_IF_NO_PEER_CERT
     }
 
-    pub fn mtls_acceptor(&self) -> Result<ssl::SslAcceptor, Error> {
+    pub fn mtls_acceptor(&self, dest_id: Option<&Identity>) -> Result<ssl::SslAcceptor, Error> {
         let _ctx = ssl::SslContext::builder(ssl::SslMethod::tls_server())?;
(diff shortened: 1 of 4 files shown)


src/tls/boring.rs
```diff
@@ -411,9 +419,31 @@ impl Verifier {
         cert.verify_san(identity)
     }
 
+    fn verifiy_san_trust_domain(&self, ctx: &mut X509StoreContextRef) -> Result<(), TlsError> {
+        let Self::SanTrustDomain(identity) = self else {
```
howardjohn, line 423: nit: I see this is copied from above but its kind of a weird pattern. If we are always getting SanTrustDomain's in this, we should just pass the inner identity probably and not take `self`?

howardjohn (approved):
A test for this would be nice. I think we can do it with a test like test_hbone_ip_mismatch

keithmattix:
Yeah, I plan to write a test, but I was running into a "not implemented" error locally and wanted to see if it was just my machine. Prow is reporting it too, so I think there's a mock I need to update somewhere

src/tls/boring.rs
hzxuzhonghu: And this match is redundant, guranteed in the caller

keithmattix: I didn't get a chance to refactor this before the PR auto-merged 😢

Issue #500, opened 2023-04-20
possible revisions to ztunnel authorization.proto naming


ilrudie:
In #495 a discussion was started around documenting vs changing the authorization.proto to align it a little better with the customer-facing APIs.

ilrudie:
closed by #501

PR #501, opened 2023-04-20
sample changes to proto

ilrudie:
Quick example of proposed changes for #500 so we have something tangible to look at.

TLDR:

- `groups` was changed to `rules` to align with the user-facing API and a rule now contains `clauses` which is my name for a generic representation of "to, from, when".
- `match` was left the same for now. I think that's pretty reasonable. Other idea was to call it `expression` but that's not really more descriptive.

I do realize this will conflict with #495 and will need some rework once that merges.

### DIFF
--- a/proto/authorization.proto
+++ b/proto/authorization.proto
@@ -45,13 +45,14 @@ message Rule {
 message Clause {
   // The logical behavior between the matches (if there are more than one)
   //  MatchBehavior match_behavior = 1;
+  // Matches are OR-ed
   // Match is a generic form of the authz policy's expressions contained in To, From and When.
   repeated Match matches = 2;
 }
 
 message Match {
-  // Values of specific type are ORed
-  // If multiple types are set, they are ANDed
+  // Values of specific type are OR-ed
+  // If multiple types are set, they are AND-ed
 
   repeated StringMatch namespaces = 1;
   repeated StringMatch not_namespaces = 2;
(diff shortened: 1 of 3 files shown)


src/rbac.rs
```diff
@@ -90,20 +90,20 @@ impl Authorization {
...
+        for rule in self.rules.iter() {
             // If ANY rule matches, it is a match...
             let mut rule_match = true;
-            for group in rule.iter() {
-                // We need ALL groups to match...
-                let mut group_match = true;
-                for mg in group.iter() {
+            for clause in rule.iter() {
+                // We need ALL clauses to match for a rule to match...
+                let mut cluase_match = true;
```
bleggett, line 102:
```suggestion
                let mut cluase_match = true;
```

sp

ilrudie: TY

proto/authorization.proto
```diff
@@ -29,22 +29,23 @@ message Authorization {
...
-  repeated Rules rules = 1;
+  repeated Clause clauses = 1;
 }
 
-message Rules {
+message Clause {
   // The logical behavior between the matches (if there are more than one)
   //  MatchBehavior match_behavior = 1;
+  // Match is a generic form of the authz policy's expressions contained in To, From and When.
   repeated Match matches = 2;
```
hzxuzhonghu, line 49: should also document the logic of match

ilrudie: Good catch. Thank you

howardjohn (approved):

ilrudie:
Will take a look at merge conflicts hopefully today. Do we want to give folks some time post-kubecon to review the names and comment if they dislike them?

PR #504, opened 2023-04-25
Initial MDS implementation

howardjohn:
* Store each connection in a shared map
* Capture special IP address, redirect to MDS handler
* MDS handler takes 4 tuple as input.
* Response is JSON containing identity

Also has an example Golang library implementation to add HTTP middleware that extracts the identity.

Istio tests: https://github.com/istio/istio/pull/44536 (blocked by this PR, of course)

### DIFF
--- a/go-metadata/istio.go
+++ b/go-metadata/istio.go
@@ -1,0 +1,132 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package istiometadata
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+// ConnectionMetadata returns the metadata about a connection
+type ConnectionMetadata struct {
+	Identity string `json:"identity"`
+}
+
+// metadataClient provides a default client for all metadata requests
+var metadataClient = http.Client{
+	Timeout: time.Second,
+}
+
+// metadataContextKey is the key the metadata handler will store connection metadata in
+var metadataContextKey = &contextKey{}
+
+type contextKey struct {
+	name string
+}
+
+func (k *contextKey) String() string { return "istio metadata context value" }
+
+func metadataServerURL() *url.URL {
+	u, err := url.Parse("http://169.254.169.111")
+	if err != nil {
+		panic(err.Error())
+	}
+	return u
+}
+
+func lookup(src, dst string) (*ConnectionMetadata, error) {
+	srcip, srcport, err := net.SplitHostPort(src)
+	if err != nil {
+		return nil, fmt.Errorf("invalid src: %v", err)
+	}
+	dstip, dstport, err := net.SplitHostPort(dst)
+	if err != nil {
+		return nil, fmt.Errorf("invalid dst: %v", err)
+	}
+
+	u := metadataServerURL()
+	params := url.Values{}
+	params.Add("srcip", srcip)
+	params.Add("srcport", srcport)
+	params.Add("dstip", dstip)
+	params.Add("dstport", dstport)
+	u.RawQuery = params.Encode()
+
+	resp, err := metadataClient.Get(u.String())
+	if err != nil {
+		return nil, fmt.Errorf("metadata lookup failed: %v", err)
+	}
+
+	mr := &ConnectionMetadata{}
+	if err := json.NewDecoder(resp.Body).Decode(mr); err != nil {
+		return nil, err
+	}
+	return mr, nil
+}
+
+// Handler is an HTTP middleware that looks up the metadata for each request, storing it in context.
+// It can later be looked up with ExtractFromRequest.
+func Handler(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Context().Value(metadataContextKey) != nil {
+			// We already have a value (connection re-use), skip
+			h.ServeHTTP(w, r)
+			return
+		}
+		local := r.Context().Value(http.LocalAddrContextKey)
+		cm, err := lookup(r.RemoteAddr, local.(*net.TCPAddr).String())
+		if err != nil {
+			// Nothing we can do
+			h.ServeHTTP(w, r)
+			return
+		}
+		// Serve underlying handler
+		ctx := context.WithValue(r.Context(), metadataContextKey, cm)
+		h.ServeHTTP(w, r.WithContext(ctx))
+	})
+}
+
+// ExtractFromRequest attempts to extract ConnectionMetadata from a request.
+// This requires Handler to be used.
+// If a connection is re-used between requests, only a single call will be made.
+func ExtractFromRequest(r *http.Request) *ConnectionMetadata {
+	v := r.Context().Value(metadataContextKey)
+	if v == nil {
+		return nil
+	}
+	return v.(*ConnectionMetadata)
+}
+
+// FetchFromRequest attempts to fetch ConnectionMetadata from a request.
+// This will make a call for each request, which may be inefficient.
+func FetchFromRequest(r *http.Request) (*ConnectionMetadata, error) {
+	local := r.Context().Value(http.LocalAddrContextKey)
+	return lookup(r.RemoteAddr, local.(*net.TCPAddr).String())
+}
+
+// FetchFromClientConnection attempts to fetch ConnectionMetadata from a client connection.
+func FetchFromClientConnection(c net.Conn) (*ConnectionMetadata, error) {
+	return lookup(c.LocalAddr().String(), c.RemoteAddr().String())
+}
+
+// FetchFromServerConnection attempts to fetch ConnectionMetadata from a client connection.
+func FetchFromServerConnection(c net.Conn) (*ConnectionMetadata, error) {
+	return lookup(c.RemoteAddr().String(), c.LocalAddr().String())
+}
--- a/go-metadata/istio.go
+++ b/go-metadata/istio.go
@@ -18,14 +18,18 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"time"
 )
 
 // ConnectionMetadata returns the metadata about a connection
 type ConnectionMetadata struct {
+	// Identity provides the identity of the peer.
+	// Example: spiffe://cluster.local/ns/a/sa/b.
 	Identity string `json:"identity"`
 }
 
@@ -43,8 +47,18 @@ type contextKey struct {
 
 func (k *contextKey) String() string { return "istio metadata context value" }
 
+const mdsHostEnv = "GCE_METADATA_HOST"
+
+// For now, we use a well-known IP which is intercepted.
+// TODO: consider creating a real service with node-affinity to drop redirection dependency.
+const defaultHost = "169.254.169.111"
+
 func metadataServerURL() *url.URL {
-	u, err := url.Parse("http://169.254.169.111")
+	host := os.Getenv(mdsHostEnv)
+	if host == "" {
+		host = defaultHost
+	}
+	u, err := url.Parse("http://" + host)
 	if err != nil {
 		panic(err.Error())
 	}
@@ -52,26 +66,22 @@ func metadataServerURL() *url.URL {
 }
 
 func lookup(src, dst string) (*ConnectionMetadata, error) {
-	srcip, srcport, err := net.SplitHostPort(src)
-	if err != nil {
-		return nil, fmt.Errorf("invalid src: %v", err)
-	}
-	dstip, dstport, err := net.SplitHostPort(dst)
-	if err != nil {
-		return nil, fmt.Errorf("invalid dst: %v", err)
-	}
-
 	u := metadataServerURL()
 	params := url.Values{}
-	params.Add("srcip", srcip)
-	params.Add("srcport", srcport)
-	params.Add("dstip", dstip)
-	params.Add("dstport", dstport)
+	params.Add("src", src)
+	params.Add("dst", dst)
 	u.RawQuery = params.Encode()
+	u.Path = "/connection"
 
 	resp, err := metadataClient.Get(u.String())
 	if err != nil {
 		return nil, fmt.Errorf("metadata lookup failed: %v", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		bdy, _ := io.ReadAll(resp.Body)
+		return nil, fmt.Errorf("metadata server returned %v: %v", resp.StatusCode, string(bdy))
 	}
 
 	mr := &ConnectionMetadata{}
@@ -86,7 +96,6 @@ func lookup(src, dst string) (*ConnectionMetadata, error) {
 func Handler(h http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.Context().Value(metadataContextKey) != nil {
-			// We already have a value (connection re-use), skip
 			h.ServeHTTP(w, r)
 			return
 		}
--- a/src/proxy/metadata.rs
+++ b/src/proxy/metadata.rs
@@ -1,0 +1,192 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::collections::HashMap;
+use std::convert::Infallible;
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+use anyhow::anyhow;
+use bytes::Bytes;
+use http_body_util::Full;
+use hyper::service::service_fn;
+use hyper::{Response, StatusCode};
+use serde::Serialize;
+use tokio::net::TcpStream;
+
+use tracing::{error, trace};
+
+use traffic::ConnectionOpen;
+
+use crate::identity::Identity;
+use crate::metrics::traffic::ConnectionClose;
+use crate::metrics::{traffic, MetricGuard, Metrics};
+
+use super::Error;
+
+#[derive(Clone, Copy, Hash, Debug, Eq, PartialEq)]
+pub struct ConnectionTuple {
+    pub src: SocketAddr,
+    pub dst: SocketAddr,
+}
+
+/// ConnectionTracker trackers connections. It stores a map of all active connections, and handles recording
+/// metrics for connection establishment and termination.
+#[derive(Clone)]
+pub struct ConnectionTracker {
+    conns: Arc<Mutex<HashMap<ConnectionTuple, Option<Identity>>>>,
+    metrics: Arc<Metrics>,
+}
+
+impl ConnectionTracker {
+    pub fn new(metrics: Arc<Metrics>) -> Self {
+        ConnectionTracker {
+            conns: Default::default(),
+            metrics,
+        }
+    }
+
+    /// fetch looks up a tuple and returns the connection metadata
+    pub fn fetch(&self, ctu: &ConnectionTuple) -> Option<ConnectionMetadata> {
+        let cm = self.conns.lock().unwrap();
+        cm.get(ctu)
+            .cloned()
+            .map(|identity| ConnectionMetadata { identity })
+    }
+
+    pub fn track<'a>(
+        &'a self,
+        ct: ConnectionTuple,
+        conn: &'a ConnectionOpen,
+    ) -> ConnectionGuard<'a> {
+        {
+            let mut c = self.conns.lock().unwrap();
+
+            let id = if let Some(ds) = &conn.derived_source {
+                ds.identity.clone()
+            } else {
+                conn.destination.as_ref().map(|w| w.identity())
+            };
+            trace!(tuple=?ct, id=?id, "tracking connection");
+            c.insert(ct, id);
+        }
+        let connection_close = self.metrics.increment_defer::<_, ConnectionClose>(conn);
+        ConnectionGuard::<'a> {
+            _metric: connection_close,
+            ct: self,
+            tup: ct,
+        }
+    }
+}
+
+pub struct ConnectionGuard<'a> {
+    /// _metric is just here to drop when we drop and record the connection close metric
+    _metric: MetricGuard<'a, ConnectionClose<'a>>,
+    ct: &'a ConnectionTracker,
+    tup: ConnectionTuple,
+}
+
+impl<'a> Drop for ConnectionGuard<'a> {
+    fn drop(&mut self) {
+        trace!(tuple=?self.tup, "forgetting connection");
+        let mut cm = self.ct.conns.lock().unwrap();
+        cm.remove(&self.tup);
+    }
+}
+
+#[derive(Clone, Debug, Serialize)]
+pub struct ConnectionMetadata {
+    #[serde(default)]
+    identity: Option<Identity>,
+}
+
+/// METADATA_SERVER_IP provides the well-known metadata server IP.
+/// This is captured by the redirection.
+pub const METADATA_SERVER_IP: IpAddr = IpAddr::V4(Ipv4Addr::new(169, 254, 169, 111));
+
+pub async fn handle_metadata_lookup(
+    ct: &ConnectionTracker,
+    stream: TcpStream,
+    remote_addr: SocketAddr,
+) -> Result<(), Error> {
+    trace!(remote=%remote_addr, "metadata lookup");
+    if let Err(e) = crate::hyper_util::http1_server()
+        .half_close(true)
+        .header_read_timeout(Duration::from_secs(2))
+        .max_buf_size(8 * 1024)
+        .serve_connection(
+            stream,
+            service_fn(|req: hyper::Request<hyper::body::Incoming>| {
+                let ct = ct.clone();
+                async move {
+                    let res: Result<_, Infallible> =
+                        serve_request(&ct, remote_addr, req).or_else(|e| {
+                            Ok(crate::hyper_util::plaintext_response(
+                                StatusCode::UNPROCESSABLE_ENTITY,
+                                e.to_string(),
+                            ))
+                        });
+                    res
+                }
+            }),
+        )
+        .await
+    {
+        error!("Error while serving HTTP connection: {}", e);
+    }
+    Ok(())
+}
+
+fn serve_request(
+    ct: &ConnectionTracker,
+    remote: SocketAddr,
+    req: hyper::Request<hyper::body::Incoming>,
+) -> anyhow::Result<Response<Full<Bytes>>> {
+    let query = req.uri().query().ok_or(anyhow!("missing query"))?;
+
+    let params = url::form_urlencoded::parse(query.as_bytes())
+        .into_owned()
+        .collect::<HashMap<String, String>>();
+    let src = SocketAddr::new(
+        params
+            .get("srcip")
+            .ok_or(anyhow!("missing srcip"))?
+            .parse()?,
+        params
+            .get("srcport")
+            .ok_or(anyhow!("missing srcport"))?
+            .parse()?,
+    );
+    let dst = SocketAddr::new(
+        params
+            .get("dstip")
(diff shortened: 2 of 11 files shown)


go-metadata/istio.go
```diff
@@ -0,0 +1,132 @@
...
+	Identity string `json:"identity"`
+}
+
+// metadataClient provides a default client for all metadata requests
+var metadataClient = http.Client{
+	Timeout: time.Second,
+}
+
+// metadataContextKey is the key the metadata handler will store connection metadata in
+var metadataContextKey = &contextKey{}
```
costinm, line 38: As API, not a big fan of using context as a hash map and forcing users to use this pattern. 

Better to have a clean interface - and let users wrap or save the info however they want. Most likely this will 
be integrated into another library ( otel, authz )

go-metadata/istio.go
```diff
@@ -0,0 +1,132 @@
...
+
+type contextKey struct {
+	name string
+}
+
+func (k *contextKey) String() string { return "istio metadata context value" }
+
+func metadataServerURL() *url.URL {
+	u, err := url.Parse("http://169.254.169.111")
```
costinm, line 47: For google there is a well-known env variable that may contain the MDS address.

I would use something like AMBIENT_MDS - the user may have multiple MDS servers ( platform, etc).
We should also consider using a Service (mds.istio-system.svc or ambientmds.istio-system.svc), with node affinity - it may be possible to run ztunnel without CNI ( using SOCKS for egress ), this would allow accessing ztunnel mds without magic.

go-metadata/istio.go
```diff
@@ -0,0 +1,132 @@
...
+
+	u := metadataServerURL()
+	params := url.Values{}
+	params.Add("srcip", srcip)
+	params.Add("srcport", srcport)
+	params.Add("dstip", dstip)
+	params.Add("dstport", dstport)
+	u.RawQuery = params.Encode()
+
+	resp, err := metadataClient.Get(u.String())
```
costinm, line 72: I  would add a path to the request - we may add more MDS queries later. 

Also not sure of query param encoding is best - but we can start with it. 
Maybe we shouldn't split the host port - and let Rust take care of that or just use the unsplit form as key, less
garbage and processing ?

go-metadata/istio.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return mr, nil
+}
+
+// Handler is an HTTP middleware that looks up the metadata for each request, storing it in context.
+// It can later be looked up with ExtractFromRequest.
+func Handler(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Context().Value(metadataContextKey) != nil {
+			// We already have a value (connection re-use), skip
```
costinm, line 89: Are you sure on connection reuse we get same context ? Don't remember.

go-metadata/istio.go
```diff
@@ -0,0 +1,132 @@
...
+		// Serve underlying handler
+		ctx := context.WithValue(r.Context(), metadataContextKey, cm)
+		h.ServeHTTP(w, r.WithContext(ctx))
+	})
+}
+
+// ExtractFromRequest attempts to extract ConnectionMetadata from a request.
+// This requires Handler to be used.
+// If a connection is re-used between requests, only a single call will be made.
+func ExtractFromRequest(r *http.Request) *ConnectionMetadata {
```
costinm, line 109: That's what I don't like about this pattern - user has to call ExtractFromRequest ( so has a dependency to this library anyways ), but the Handler is called in ALL requests. If user only needs peer info for some requests - it's more complicated to setup ( wrap only some request with the handler chain ), or not possible if the lookup is based on some other logic ( like checking XFF or other standard headers). 

Instead this method could just to the lookup on demand, and no wrapper needed.

src/proxy/metadata.rs
```diff
@@ -0,0 +1,192 @@
...
+            .get("dstip")
+            .ok_or(anyhow!("missing dstip"))?
+            .parse()?,
+        params
+            .get("dstport")
+            .ok_or(anyhow!("missing dstport"))?
+            .parse()?,
+    );
+    if remote.ip() != dst.ip() && remote.ip() != src.ip() {
+        anyhow::bail!("metadata server request must come from the src or dst address")
```
costinm, line 183: Not sure this is guaranteed - for example if the connection is IPv6, pretty sure it won't be the case ( dual stack and we use a MDS v4 address ). 

Can we keep a pointer to the dest and source pod - and check its IPs ? 

Or just add a TODO and bug - we can do this in a later phase.

src/proxy/outbound.rs
```diff
@@ -120,22 +121,31 @@ impl OutboundConnection {
...
     pub async fn proxy_to(
         &mut self,
         mut stream: TcpStream,
-        remote_addr: IpAddr,
+        remote_addr: SocketAddr,
         orig_dst_addr: SocketAddr,
         block_passthrough: bool,
     ) -> Result<(), Error> {
+        let remote_ip = remote_addr.ip();
+        if orig_dst_addr.ip() == METADATA_SERVER_IP {
```
costinm, line 135: That's ok - but may also be good to have MDS listen on a separate port ( 150xx) and let CNI handle the redirection ( or in whitebox mode use the Service ). Cleaner. 

But it's good to keep this code too.

go-metadata/istio.go
howardjohn: I generally agree but I think this is the only way to do it as an HTTP handler? unless do the magic to lookup the TCP connection for the request, and then add caching yourself which would just be duplicating this I think?

go-metadata/istio.go
howardjohn: Hmm actually probably not. Let me check

go-metadata/istio.go
costinm: Not sure - see my other comment bellow. In a handler user has access to peer and local IP - and can make the MDS call directly when needed.

costinm: I see what you mean - it's about avoiding to make this call on each request. 

How about we leave it out for now - users can probably use the same pattern if they want, better to focus this library on TCP ?

bleggett:
`Store each connection in a shared map` Have we thought about how this might conflict with eBPF-based + map-based conntracking?

Most eBPF CNIs will need to do something like this, and it will work almost exactly the same way (just in kernelspace).

go-metadata/istio.go
EItanya: I agree that I don't love implicit context storage, but this is what go-grpc uses all over the place, so in this instance probably better to follow the same pattern

go-metadata/istio.go
```diff
@@ -0,0 +1,132 @@
...
+	"fmt"
+	"net"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+// ConnectionMetadata returns the metadata about a connection
+type ConnectionMetadata struct {
+	Identity string `json:"identity"`
```
EItanya, line 29: Can we have commends on each field to better understand what they are? I understand there's only one right now, for instance, what identity is this exactly

src/proxy/metadata.rs
howardjohn: Ran into this myself with `localhost` testing -- since we don't use 127.0.0.1 to connect to MDS. I put a TODO. I think we can handle it once the workload API supports dual stack (currently doesn't)

linsun:
Sorry I'm missing how this would benefit users since we said we want to keep ztunnel small and lean.  MDS - is this google cloud metadata service or more generic?

howardjohn:
It has nothing to do with Google - this allows users applications to find the peer identity. Today they can do this with XFCC header, but that is HTTP based which we cannot do in ztunnel. This provides a generalized way.

Note this is needed for sandwiching as well

linsun:
> It has nothing to do with Google - this allows users applications to find the peer identity. Today they can do this with XFCC header, but that is HTTP based which we cannot do in ztunnel. This provides a generalized way.
> 
> Note this is needed for sandwiching as well

Thanks, btw, what does MDS stand for?

I thought we aren't using sandwich approach at all?

costinm:
On Thu, Apr 27, 2023 at 7:29 AM Eitan Yarmush ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In go-metadata/istio.go
> <https://github.com/istio/ztunnel/pull/504#discussion_r1179250379>:
>
> > +	"net/url"
> +	"time"
> +)
> +
> +// ConnectionMetadata returns the metadata about a connection
> +type ConnectionMetadata struct {
> +	Identity string `json:"identity"`
> +}
> +
> +// metadataClient provides a default client for all metadata requests
> +var metadataClient = http.Client{
> +	Timeout: time.Second,
> +}
> +
> +// metadataContextKey is the key the metadata handler will store connection metadata in
> +var metadataContextKey = &contextKey{}
>
> I agree that I don't love implicit context storage, but this is what
> go-grpc uses all over the place, so in this instance probably better to
> follow the same pattern
>

The pattern may be fine in some cases. That's not my concern - this library
should not be a framework, just expose the raw command.

It will likely be combined with checks for XFCC, ProxyProtocol, etc - there
are plenty of auth schemes an app may support, and
plenty of frameworks that set the identity in some way.

costinm:
MDS is 'metadata server' AFAIK, all cloud vendors have one. Not sure who
was first.

This has little to do with sandwich - any user application may need to know
the identity of the peer.
For example we may have authz to allow Alice to talk to Bob - but Bob still
wants to know who's talking to.

src/proxy/outbound.rs
```diff
@@ -121,22 +122,31 @@ impl OutboundConnection {
...
```
hzxuzhonghu, line 136: Would prefer this is not intercepted to ztunnel

howardjohn: I am not sure how we can opt out? since ztunnel will capture all traffic

costinm:
Interesting - will ztunnel break the platform MDS ? I think CNI needs to
probably exclude the range ( most platforms seem to be using
same /24).

But ztunnel MDS itself needs to be intercepted, how else can the pod
communicate with it ?

howardjohn:
Hmm we did fix platform MDS somehow and have a test for it. Let me see how
we did it

go-metadata/istio.go
```diff
@@ -0,0 +1,141 @@
...
+	"net/url"
+	"os"
+	"time"
+)
+
+// ConnectionMetadata returns the metadata about a connection
+type ConnectionMetadata struct {
+	// Identity provides the identity of the peer.
+	// Example: spiffe://cluster.local/ns/a/sa/b.
+	Identity string `json:"identity"`
```
louiscryan, line 33: Can we open a tracking issue to capture the other information about the identity that we would like to expose here

Over time I would expect to see us add either a field or a new API that provided details from the public key of the peer

I think the interesting question is what are we trying to prove to the caller about the veracity of the identity we just provided. Options include:

1. *How can the caller know that the MDS is reliably acting on its behalf*. Clearly theres a certain amount of infrastructural trust seeing as this is coming from the 'network' as perceived by the app. One option here if for the MDS to offer a "who am I" API. This is a bit like SDS but now its an implicit part of the infrastructure. Response to the call should either be the current public cert or some shorthand of it. Variant, can prove that the infra has the private key which is shielded from the app by signing some random bytes passed to the API which can then be verified with the public key. Likely want to deliver a shorthand in addition to x509 to shield users from parsing its gory details.

2. *How can the caller know that the infrastructure actually performed a secure handshake with the identified peer?* This is harder and of debateable value. If we provide proof that infrastructure already has the public and private keys as described in 1 above then the infra is already capable of impersonating the application for whatever purposes it sees fit. We could do something like a 'signed ping'
          
         Client generates random bytes and encrypts with public key of peer (opaque to local ztunnel)
         Client app calls 'secure ping' passing the encrypted bytes
         local ztunnel simply forwards encrypted bytes to the identified peer
         peer ztunnel decrypts random byes with private key and returns their signature
         local ztunnel forwars message signature to client
         Client compares message signature to locally computed hash

This proves that there is something between the client and the peer that has access to the private key associated with the identity. The client doesn't know who or what that is necessarily

linsun:
> MDS is 'metadata server' AFAIK, all cloud vendors have one. Not sure who was first. This has little to do with sandwich - any user application may need to know the identity of the peer. For example we may have authz to allow Alice to talk to Bob - but Bob still wants to know who's talking to.
> […](#)

Thanks, thinking loud here - could this be something opted in so it doesn't impact users don't use MDS.

howardjohn:
> > MDS is 'metadata server' AFAIK, all cloud vendors have one. Not sure who was first. This has little to do with sandwich - any user application may need to know the identity of the peer. For example we may have authz to allow Alice to talk to Bob - but Bob still wants to know who's talking to.
> > […](#)
> 
> Thanks, thinking loud here - could this be something opted in so it doesn't impact users don't use MDS.

This doesn't impact anyone if they don't use it, t hey have to explicitly call the metadata server for it to do anything

go-metadata/istio.go
bleggett:
```
Over time I would expect to see us add either a field or a new API that provided details from the public key of the peer
```

ideally one that follows XFCC semantics, and chains certs of all workloads in the pathway, including the ztunnel instance.

src/proxy/metadata.rs
```diff
@@ -0,0 +1,182 @@
...
+
+#[derive(Clone, Debug, Serialize)]
+pub struct ConnectionMetadata {
+    #[serde(default)]
+    identity: Option<Identity>,
+}
+
+/// METADATA_SERVER_IP provides the well-known metadata server IP.
+/// This is captured by the redirection.
+pub const METADATA_SERVER_IP: IpAddr = IpAddr::V4(Ipv4Addr::new(169, 254, 169, 111));
```
bleggett, line 116: Realistically, should we just be hardcoding this in OSS ztunnel?

src/proxy/outbound.rs
bleggett: `have MDS listen on a separate port ( 150xx) and let CNI handle the redirection`

+1 on this, I don't love the reliance on a hardcoded IP here.

src/proxy/metadata.rs
bleggett, line 116: Can we just hardcode a specific metadata server IP in OSS ztunnel like this?

If it's "well-known", to whom is it well-known?

howardjohn: I mean we _can_ - the IP is irrelevant other than the client and server need to be in agreement about it and it should not conflict with other IPs users want to use for other purposes.

> and would prefer a metadata port (like prometheus, xds, etc etc) so we don't need to rely on known IP keying in zt proper.

What is better about a special port instead of a special IP? Or did you want special IP + special port (which seems good?)

bleggett: I guess I'm not clear on why ztunnel needs a special IP at all?

If we can redirect any pod outbound on <whatever IP you want> to <whatever ztunnel you want> and <whatever ztunnel port we want> outside of ztunnel itself - why does ztunnel need to match a well-known IP?

howardjohn: Is it any simpler to have it in the CNI layer instead of in ztunnel though?

In general there is a tension of what goes at which layer. For example, we could skip. ztunnel entirely in cases where we know it won't do anything but pass through.

So far we have leaned towards keeping the CNI simpler so there is a better chance we can replace it with better approaches (and rust is easier than iptables/eBPF)

bleggett: `Is it any simpler to have it in the CNI layer instead of in ztunnel though?` 

Long term I think so, since the CNI (ours or someone else's) will be handling/redirecting the traffic based on intended dest anyway, and we already have the concept of directing packets to different well-known ztunnel listener ports and handling them differently as a result (for purposes of early exit/handling bypass in eBPF/iptables redirect logic, for instance)

`So far we have leaned towards keeping the CNI simpler so there is a better chance we can replace it with better approaches (and rust is easier than iptables/eBPF)`

Absolutely a fan of keeping CNI simpler, but sending captured traffic based on intended dest to a given ztunnel listener is what CNI is for - that's what it does best. It's how we do DNS redir for instance. CNI redir already does special-casing around certain dest ports (or whether src is host ip, etc etc) for purposes of data path optimization - it's already handling this stuff.

Feels logically consistent to maintain that pattern here, and it has the added benefit of making it very easy for the IP to be changed or redefined without ztunnel caring (or us needing to add a flag to ztunnel to let people define what they want).

costinm:
The reason for the hardcoded IP is pretty simple: in an app, how do you
make the http request ? You need a DNS name or IP.

Many MDS servers are also DNS servers - there is a plan to allow running a
DNS proxy in ztunnel. Also what hostname would you
use - and how would CNI know that some VIP is actually the MDS address ?

All MDS servers I know rely on a hardcoded 'well known' address in that
range. We can also use a special 150xx port, but if we
want ztunnel to be a resolver it'll need to listen on 53 ( yes, we can do
some complicated redirection to use a different port for DNS - but
why not use the simplest solution ). If we are using this pattern - I would
like to see a good reason why we should follow
a different pattern, and what would that be - can't think of a simpler or
cleaner solution.

Also this whole thing doesn't need a fancy iptables or eBPF or k8s - CNI
can simply add a route and add a second
address with the well-known IP on the ztunnel pod.

It will also work on VMs and non-K8S environments - even in cases where no
iptables are possible ( ztunnel can work without
any CNI or iptables, by setting SOCKS_PROXY env variable with most
applications ).

costinm:
On Fri, Apr 28, 2023 at 12:49 PM Louis Ryan ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In go-metadata/istio.go
> <https://github.com/istio/ztunnel/pull/504#discussion_r1180762108>:
>
> > +	"context"
> +	"encoding/json"
> +	"fmt"
> +	"io"
> +	"net"
> +	"net/http"
> +	"net/url"
> +	"os"
> +	"time"
> +)
> +
> +// ConnectionMetadata returns the metadata about a connection
> +type ConnectionMetadata struct {
> +	// Identity provides the identity of the peer.
> +	// Example: spiffe://cluster.local/ns/a/sa/b.
> +	Identity string `json:"identity"`
>
> Can we open a tracking issue to capture the other information about the
> identity that we would like to expose here
>
> Over time I would expect to see us add either a field or a new API that
> provided details from the public key of the peer
>

And probably some of the telemetry or labels and other info we got from the
PTR-DS. I agree in time we'll likely add more,
but the first step is to get identity in and validate.

I would also expose a lookup using only the peer IP - the 4-tuple is only
really needed if ztunnel can't impersonate the original
source IP.



> I think the interesting question is what are we trying to prove to the
> caller about the veracity of the identity we just provided. Options include:
>
>    1.
>
>    *How can the caller know that the MDS is reliably acting on its behalf*.
>    Clearly theres a certain amount of infrastructural trust seeing as this is
>    coming from the 'network' as perceived by the app. One option here if for
>    the MDS to offer a "who am I" API. This is a bit like SDS but now its an
>    implicit part of the infrastructure. Response to the call should either be
>    the current public cert or some shorthand of it. Variant, can prove that
>    the infra has the private key which is shielded from the app by signing
>    some random bytes passed to the API which can then be verified with the
>    public key. Likely want to deliver a shorthand in addition to x509 to
>    shield users from parsing its gory details.
>
>
I think this may be overkill and defeat the purpose. If the app is willing
to do all this - and link libraries and deal with certs - they are probably
better off
just using a proxyless library. With sidecars they don't get any guarantee
either, nor with most CNI.

It's also not the caller that needs to be assured - what is it going to do
? The network admin, who installs the CNI and ztunnel needs to be sure
all packets go to ztunnel, but that can be exposed using telemetry.




>
>    1.
>
>    *How can the caller know that the infrastructure actually performed a
>    secure handshake with the identified peer?* This is harder and of
>    debateable value. If we provide proof that infrastructure already has the
>    public and private keys as described in 1 above then the infra is already
>    capable of impersonating the application for whatever purposes it sees fit.
>    We could do something like a 'signed ping'
>
>      Client generates random bytes and encrypts with public key of peer (opaque to local ztunnel)
>      Client app calls 'secure ping' passing the encrypted bytes
>      local ztunnel simply forwards encrypted bytes to the identified peer
>      peer ztunnel decrypts random byes with private key and returns their signature
>      local ztunnel forwars message signature to client
>      Client compares message signature to locally computed hash
>
>
> This proves that there is something between the client and the peer that
> has access to the private key associated with the identity. The client
> doesn't know who or what that is necessarily
>

That seems for super-advanced users - it may be something for a specialized
ztunnel to support for specialized apps with libraries that can take
advantage of it,
but maybe not something to have in OSS and the simple/clean implementation
we have ? It is a cool idea.

bleggett:
> The reason for the hardcoded IP is pretty simple: in an app, how do you make the http request ? You need a DNS name or IP. Many MDS servers are also DNS servers - there is a plan to allow running a DNS proxy in ztunnel. Also what hostname would you use - and how would CNI know that some VIP is actually the MDS address ?

I'm suggesting if you have a need to map an arbitrary environmental IP of your choice to always get redirected to a listener in ztunnel, it's pretty simple to do that with a single iptables rule without ztunnel needing to care what that IP is, or needing to hardcode it in ztunnel. We could pick a fixed IP, anyone who wanted to do something different could change it, ztunnel wouldn't care either way.

> All MDS servers I know rely on a hardcoded 'well known' address in that range. 
 MDS clients do, I'm not sure the MDS server being added to ztunnel does.
 
> We can also use a special 150xx port, but if we want ztunnel to be a resolver it'll need to listen on 53 ( yes, we can do some complicated redirection to use a different port for DNS - but why not use the simplest solution ). 
 
ztunnel already listens on 15053 for this purpose, I believe. We have iptables rules that rewrite UDP on port 53 to 15053 within ztunnel's netns.
 
> If we are using this pattern - I would like to see a good reason why we should follow a different pattern, and what would that be - can't think of a simpler or cleaner solution. Also this whole thing doesn't need a fancy iptables or eBPF or k8s - CNI can simply add a route and add a second address with the well-known IP on the ztunnel pod. It will also work on VMs and non-K8S environments - even in cases where no iptables are possible ( ztunnel can work without any CNI or iptables, by setting SOCKS_PROXY env variable with most applications ).

I'm proposing we use the pattern already in use - ztunnel exposes well-known ports, and CNI redirects traffic, based on (variously) dest IP, dest port, and dest packet type to those ports - for this as well. A fixed IP within ztunnel isn't a requirement for this, and it feels weird to add it here when nothing else currently in ztunnel works like that.

costinm:
On Fri, Apr 28, 2023 at 6:18 PM Ben Leggett ***@***.***>
wrote:

> The reason for the hardcoded IP is pretty simple: in an app, how do you
> make the http request ? You need a DNS name or IP. Many MDS servers are
> also DNS servers - there is a plan to allow running a DNS proxy in ztunnel.
> Also what hostname would you use - and how would CNI know that some VIP is
> actually the MDS address ?
>
> I'm suggesting if you have a need to map an arbitrary environmental IP of
> your choice to always get redirected to a listener in ztunnel, it's pretty
> simple to do that with a single iptables rule without ztunnel needing to
> care what that IP is, or needing to hardcode it in ztunnel. We could pick a
> fixed IP, anyone who wanted to do something different could change it,
> ztunnel wouldn't care either way.
>
>
I see, you are right, that would work too.

However since ztunnel is intercepting all egress anyways - and istio-cni is
extremely complex and hopefully will become part of the native CNI,
I would rather keep it in ztunnel and simplify the requirement on the CNI
to 'just redirect all egress from the app to ztunnel'.


All MDS servers I know rely on a hardcoded 'well known' address in that
> range.
> MDS clients do, I'm not sure the MDS server being added to ztunnel does.
>
> We can also use a special 150xx port, but if we want ztunnel to be a
> resolver it'll need to listen on 53 ( yes, we can do some complicated
> redirection to use a different port for DNS - but why not use the simplest
> solution ).
>
> ztunnel already listens on 15053 for this purpose, I believe. We have
> iptables rules that rewrite UDP on port 53 to 15053 within ztunnel's netns.
>

Unless someone added DNS - I think agent does that. There is an option to
also listen on 53 and not require iptable rules. Just because
it is possible to use iptables - it doesn't mean we should ( and the rules
for 53 require raw and some very ugly stuff - UDP is far more
complicated )


> If we are using this pattern - I would like to see a good reason why we
> should follow a different pattern, and what would that be - can't think of
> a simpler or cleaner solution. Also this whole thing doesn't need a fancy
> iptables or eBPF or k8s - CNI can simply add a route and add a second
> address with the well-known IP on the ztunnel pod. It will also work on VMs
> and non-K8S environments - even in cases where no iptables are possible (
> ztunnel can work without any CNI or iptables, by setting SOCKS_PROXY env
> variable with most applications ).
>
> I'm proposing we use the pattern already in use - ztunnel exposes
> well-known ports, and CNI redirects traffic, based on (variously) dest IP,
> dest port, and dest packet type to those ports - for this as well. A fixed
> IP within ztunnel isn't a requirement for this, and it feels weird to add
> it here when nothing else currently in ztunnel works like that.
>

Another pattern already in use is CNI redirects all traffic - and ztunnel
routes it wherever it needs to go. I would rather reduce complexity in CNI.

I also mentioned we have the SOCKS mode that is independent of iptables (
works without any ) - with the current method it would work
just fine. And if we ever add a HTTP/1.1 CONNECT proxy ( for the few apps
that support HTTPS_PROXY and not SOCKS ) - it would work too.

go-metadata/istio.go
costinm: XFCC is a pretty horrible UX - you need a parser library, parse certs, etc. It may be useful for some super advanced users - but if they are willing to take all those dependencies they may be better off just using proxyless and having end to end mTLS.

All platforms I know ( not only google) have a basic, simple MDS server that returns metadata and tokens in 
a very simple way. 

I'm not saying it is an invalid use case to want fancy attestation - and it is clear that cloud vendors don't 
provide this right now, so it's not a bad idea if someone implemented it - but we should not mix it with
the very simple use case for this PR.

go-metadata/istio.go
costinm: There is a difference between a library and a framework. Yes, go-grpc and other frameworks follow this pattern - and it is not a bad pattern for a framework, where user can get peer identity in a consistent way from context.

My point is that for a library that interacts with ztunnel - it should not define those abstractions and implied requirements ( chained for ALL requests, etc). We will have other ways to get peer info - JWT, XFCC, etc - so this library would be used by frameworks together with other auth mechanisms, and in the end expose the context
data specific to the framework in use.

go-metadata/istio.go
```diff
@@ -0,0 +1,141 @@
...
+		panic(err.Error())
+	}
+	return u
+}
+
+func lookup(src, dst string) (*ConnectionMetadata, error) {
+	u := metadataServerURL()
+	params := url.Values{}
+	params.Add("src", src)
+	params.Add("dst", dst)
```
costinm, line 72: Getting connection metadata is very useful - as Louis mentioned we may extend this to return all the details
about the connection, including labels and more handshake info. 

And since we return it as json - we may as well consider an MDS endpoint that is also signing it.

But I think the use case for passing both src and dst is primarily for dealing with CNI implementations 
where the source IP can't be preserved. Calico for example seems to support that, and hopefully we can
do it for more CNIs.  

In the case of using ztunnel in a 'best practice' CNI - we may want to support metadata lookup using only the source IP ( which would return data from the PTR-DS/WDS ). 
If the ztunnel doesn't preserve the source IP - I think the lookup result should include the original source IP 
in addition to the peer identity. And maybe all the metadata lookup info too.

src/proxy/metadata.rs
costinm: I strongly disagree - routing traffic based on dst is what  ztunnel does.

CNI is responsible to redirect pod egress to ztunnel - it does not know where to send ( unless ztunnel and WDS/PTR-DS is integrated into the CNI ).  Knowing where to redirect an IP and how is based on Istiod configs.

costinm: And if we want at some point to have the 'redirect all egress from pod to daemon set' - shared by all
mesh implementations and all CNI providers - we need to keep the requirements on the CNI layer as
small as possible. 

We may add extra requirements - like exclude ports for proxyless - but getting into 'route IP address x in a different way' is too much.

src/proxy/outbound.rs
costinm: For the record - changed my mind after further thinking. See my other comments on why it's best for ztunnel to handle all redirection based on destination IP.

costinm (approved):

src/proxy/metadata.rs
bleggett: > I strongly disagree - routing traffic based on dst is what ztunnel does.

> CNI is responsible to redirect pod egress to ztunnel - it does not know where to send ( unless ztunnel and WDS/PTR-DS is integrated into the CNI ). Knowing where to redirect an IP and how is based on Istiod configs.

CNI knows what ztunnel listener to redirect traffic to, and ztunnel knows what dest to send traffic to - correct - using CNI to redirect MDS traffic to ztunnel listener would be the former, which is definitely what CNI is responsible for doing today.


> And if we want at some point to have the 'redirect all egress from pod to daemon set' - shared by all
mesh implementations and all CNI providers - we need to keep the requirements on the CNI layer as
small as possible.

Valid. I'm not sure it's a realistic or practical goal, given that no two CNIs are really the same and we want to keep ztunnel from being a kitchen-sink, but we can live in hope.

> We may add extra requirements - like exclude ports for proxyless - but getting into 'route IP address x in a different way' is too much.

It doesn't strike me as significantly different than `route IP port X in a different way` which CNI already does and will always do, but I'm not set on it.

src/proxy/outbound.rs
bleggett: This is also something we can potentially revisit later if ztunnel/CNI integ gets tighter/more involved, but I'm ok punting on it now and going with this impl.

bleggett:
Looking back at this:

1. It probably needs a new name, since it's unrelated to https://github.com/istio/istio/issues/43937
2. This is functionally identical to SPIFFE/SPIRE self-identity-lookup in a lot of ways, and if you were using SPIRE with Istio this would be a redundant mechanism. That's not necessarily an objection, but it concerns me a bit.

howardjohn:
Lets drop this for now to keep things minimal.

PR #513, opened 2023-05-04
Improve certificate prefetching

howardjohn:
Skip non-HBONE workloads, who should not need certs

Use Warmup priority. We implemented prioritization for certificate fetching to favor on-demand fetchers but forgot to start using it.

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -39,6 +39,7 @@ use xds::istio::workload::Service as XdsService;
 use xds::istio::workload::Workload as XdsWorkload;
 
 use crate::config::{ConfigSource, ProxyMode};
+use crate::identity::Priority::Warmup;
 use crate::identity::{Identity, SecretManager};
 use crate::metrics::Metrics;
 use crate::rbac::{Authorization, RbacScope};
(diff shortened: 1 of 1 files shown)


costinm (approved):
Looks great. The other issue I mentioned: ztunnel shouldn't get a cert for itself ( until we have a 'delegated' mode where ztunnel uses its own cert and doesn't need each workload
cert).

howardjohn:
We have traces but they need refinement and they are only going to stdout.

Metrics - agree.

ztunnel cert - I think ztunnel itself doesn't declare "hbone" protocol so
this might fix it. Let me check

src/workload.rs
```diff
@@ -596,6 +599,17 @@ impl WorkloadStore {
         self.remove(wip.to_string());
         let widentity = workload.identity();
         let status = workload.status;
+
+        // Determine if we should prefetch a certificate for this workload. Being "wrong" is not too bad;
+        // a missing cert will be fetched on-demand when we get a request, so will just result in some extra latency.
+        let should_fetch =
```
hzxuzhonghu, line 605: can we abstact this as a function

src/workload.rs
```diff
@@ -305,10 +306,12 @@ impl WorkloadManager {
         cert_manager: Arc<SecretManager>,
     ) -> anyhow::Result<WorkloadManager> {
         let (tx, mut rx) = mpsc::channel::<Identity>(256);
-        // todo ratelimit prefetching to a reasonable limit
         tokio::spawn(async move {
             while let Some(workload_identity) = rx.recv().await {
-                match cert_manager.fetch_certificate(&workload_identity).await {
```
hzxuzhonghu, line 311: nit: In case of misusing this function, maybe we should add some docs about when should use realtime fetching

src/workload.rs
```diff
@@ -305,10 +306,12 @@ impl WorkloadManager {
...
+                match cert_manager
+                    .fetch_certificate_pri(&workload_identity, Warmup)
+                    .await
+                {
                     Ok(_) => debug!("prefetched cert for {:?}", workload_identity.to_string()),
                     Err(e) => error!(
                         "unable to prefetch cert for {:?}, skipping, {:?}",
```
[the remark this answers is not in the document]
ilrudie: We do have some metrics for traffic and XDS in place. We could probably slide this into the XDS metrics structure that already exists.

On paper [OTLP](https://crates.io/crates/opentelemetry-otlp) support in Rust seems pretty good. hyper, tonic and prost are the defaults which are already used by ztunnel. I'm not sure how compatible our metrics today would be with the OpenTelemetry-Rust ecosystem though. We may need to rework the prom client metrics that already exist.

stevenctl (approved):

PR #526, opened 2023-05-19
config dump: cleanup json representation a bit

howardjohn:
* Use equivilent of "omitempty" liberally to hide unneeded fields
* Do not use tag types, these are much harder to process in other
  languages with less sophisticated type systems

### DIFF
--- a/src/workload.rs
+++ b/src/workload.rs
@@ -103,10 +103,10 @@ pub struct GatewayAddress {
 pub mod gatewayaddress {
     use super::{NamespacedHostname, NetworkAddress};
     #[derive(Debug, Hash, Eq, PartialEq, Clone, serde::Serialize, serde::Deserialize)]
-    #[serde(tag = "address", content = "content")]
+    #[serde(untagged)]
     pub enum Destination {
-        Hostname(NamespacedHostname),
         Address(NetworkAddress),
+        Hostname(NamespacedHostname),
     }
 }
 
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (commented):
It needs some tests

src/workload.rs
```diff
@@ -100,18 +100,18 @@ pub struct GatewayAddress {
 pub mod gatewayaddress {
     use super::{NamespacedHostname, NetworkAddress};
     #[derive(Debug, Hash, Eq, PartialEq, Clone, serde::Serialize, serde::Deserialize)]
-    #[serde(tag = "address", content = "content")]
+    #[serde(untagged)]
     pub enum Destination {
-        Hostname(NamespacedHostname),
         Address(NetworkAddress),
+        Hostname(NamespacedHostname),
```
kdorosh, line 106: at the very least I think we need a test here to ensure we can serialize / deserialize both types here

the reason I added tags in the first place was because in earlier versions of the code (when hostname was just a string) it would always serialize as hostname even if an address was provided, so i opted to be explicit about the type to avoid bugs

kdorosh: I think this test coverage can be provided by the linked admin.rs config dump test (serialize) and the local_client test that deserializes from yaml

src/workload.rs
```diff
@@ -100,18 +100,18 @@ pub struct GatewayAddress {
...
 }
 
 pub mod address {
     use crate::workload::{Service, Workload};
 
     #[derive(Debug, Eq, PartialEq, Clone, serde::Serialize, serde::Deserialize)]
-    #[serde(tag = "address", content = "content")]
+    #[serde(untagged)]
     pub enum Address {
         Workload(Box<Workload>),
```
kdorosh, line 116: if someone provides a service that just has fields from workload, now it will serialize as the wrong type? can we at least make some types required in proto so that cannot happen

kdorosh: im being pretty paranoid but this was a pain to debug the first time

hzxuzhonghu (approved):

PR #528, opened 2023-05-22
Fix build and test on macos

nmittler:
Addresses a couple of things:

* All of the `namespaced` tests actually require linux. Adding `cfg` blocks where necessary.
* The direct tests use multiple loopback addresses (e.g. 127.0.0.2). Macos requires manually configuring each one as an alias for the loopback device. Added a better error message to help here.


tests/namespaced.rs
```diff
@@ -12,697 +12,701 @@
...
-    () => {{
-        fn f() {}
-        fn type_name_of<T>(_: T) -> &'static str {
-            std::any::type_name::<T>()
-        }
-        let name = type_name_of(f);
-        &name[..name.len() - 3]
-    }};
-}
+#[cfg(all(test, target_os = "linux"))]
```
nmittler, line 15: FYI nothing actually changed in this file apart from adding a surrounding `mod namespaced` so I can easily enable only for linux.

Cargo.toml
```diff
@@ -5,7 +5,19 @@ edition = "2021"
...
-default = ["fips"]
+# We don't include the fips feature by default. The variables defined
+# in config.toml already override this setting by pointing to a pre-built
+# FIPS-compliant image of BoringSSL. For development, most folks don't care much
+# about FIPS and it's unlikely that they have a Clang 7 compiler handy.
+#
+# Additionally, since macos users can't use the pre-built image anyway, having
+# fips off by default reduces the number of steps required to build.
+#
+# To build ztunnel with a FIPS-compliant BoringSSL from scratch:
```
howardjohn, line 16: This is going to break the build which relies we want to be FIPS I think?

Cargo.toml
```diff
@@ -5,7 +5,19 @@ edition = "2021"
 rust-version = "1.65"
 
 [features]
-default = ["fips"]
+# We don't include the fips feature by default. The variables defined
+# in config.toml already override this setting by pointing to a pre-built
+# FIPS-compliant image of BoringSSL. For development, most folks don't care much
+# about FIPS and it's unlikely that they have a Clang 7 compiler handy.
```
howardjohn, line 11: You don't need it, we have precompiled them?

Cargo.toml
```diff
@@ -101,6 +115,8 @@ incremental = true
 criterion = { version = "0.4.0", features = ["async_tokio", "html_reports"] }
 diff = "0.1.13"
 matches = "0.1.9"
-netns-rs = "0.1.0"
 test-case = "3.0.0"
 #debug = true
+
+[target.'cfg(target_os = "linux")'.dev-dependencies]
```
howardjohn, line 121: nit: do we need this? since its already a core `dependency`?

howardjohn (commented):
Mostly LGTM, just worried about the real build using FIPS. Thanks for fixing this!!

Cargo.toml
nmittler: nope ... removed.

nmittler:
@howardjohn PTAL

howardjohn (approved):

PR #529, opened 2023-05-23
Update `Workload` primary xds key to UID

kdorosh:
### Changes

The second half of the scaffolding implementation for [ambient xds evolution](https://docs.google.com/document/d/1V5wkeBHbLSLMzAMbwFlFZNHdZPyUEspG4lHbnB0UaCg/edit#heading=h.qpcs4a7wyicb). The first half of the ztunnel changes merged in https://github.com/istio/ztunnel/pull/486

### Breaking Change

Note that this is a **breaking change** and breaks the xds contract between istiod and ztunnel. istiod used to send workloads keyed by `network/ip` (workloads only had one IP); now it sends workloads keyed by UID which allows the workload proto's shape to support multiple addresses (for eventual dual stack support) and no addresses (for "proxy" workloads in multi network).

This break is perceived as ok since the istio changes are in flight / will be ready soon. We can only break this api contract because ambient is still in alpha.

### For Reviewers

The intent here is to merge all the breaks to the API contract early while ambient is still in alpha (previously expiremental) just so we can add support for all the remaining features in due time. It is not my intent to add actual support for dual stack here, rather to enable us to as early as possible.


proto/workload.proto
```diff
@@ -74,8 +74,22 @@ message Service {
...
+  // cluster/group/version/kind/namespace/name/optional-index
+  //
+  // As an example, a ServiceEntry with two WorkloadEntries inlined could become
+  // two Workloads with the following UIDs:
+  // - cluster1/networking.istio.io/v1alpha3/ServiceEntry/default/external-svc/endpoint1
+  // - cluster1/networking.istio.io/v1alpha3/ServiceEntry/default/external-svc/endpoint2
+  //
+  // For VMs and other workloads other formats are also supported; for example,
+  // a single UID string: "0ae5c03d-5fb3-4eb9-9de8-2bd4b51606ba"
   string uid = 20;
```
howardjohn, line 93: UID not being a uuid feels a bit funky. I get it technical makes sense, just the 2 are typically used interchange-ably. Not saying we should change it necessarily

proto/workload.proto
```diff
@@ -74,8 +74,22 @@ message Service {
...
 }
 
+// Workload represents a workload - an endpoint (or collection behind a hostname).
+// The xds primary key is "uid" as defined on the workload below.
+// Secondary (alias) keys are the unique `network/IP` pairs that the workload can be reached at.
 message Workload {
   // UID represents a globally unique opaque identifier for this workload.
+  // For k8s resources, it is recommended to use the more readable format:
+  //
+  // cluster/group/version/kind/namespace/name/optional-index
```
howardjohn, line 84:
```suggestion
  // cluster/group/kind/namespace/name/section-name
```

WorkloadEntry v1alpha3 and WE v1beta1 are the same resource.

I would use "section name" to align with Gateway API naming (not that its important, as this is just internal doc)

proto/workload.proto
howardjohn: I guess maybe just `id` would make it feel 'right' to me

src/workload.rs
```diff
@@ -1248,42 +1251,72 @@ impl WorkloadStore {
...
-                Protocol::TCP => SocketAddr::from((us.workload.workload_ip, us.port)),
+                Protocol::TCP => {
+                    SocketAddr::from((self.choose_workload_ip(&us.workload)?, us.port))
+                }
             });
         }
+        Ok(())
+    }
+
+    // TODO: add more sophisticated routing logic, perhaps based on ipv4/ipv6 support underneath us.
```
howardjohn, line 1310: +1, randomly picking between ipv4 and ipv6 is .. weird. But no problem for now, we don't actually support dual stack yet

howardjohn (approved):

proto/workload.proto
kdorosh: I think uid is slightly better, albeit awkward, because it does _need_ to be universally unique

proto/workload.proto
```diff
@@ -74,8 +74,22 @@ message Service {
...
+// Secondary (alias) keys are the unique `network/IP` pairs that the workload can be reached at.
 message Workload {
   // UID represents a globally unique opaque identifier for this workload.
+  // For k8s resources, it is recommended to use the more readable format:
+  //
+  // cluster/group/kind/namespace/name/section-name
+  //
+  // As an example, a ServiceEntry with two WorkloadEntries inlined could become
+  // two Workloads with the following UIDs:
+  // - cluster1/networking.istio.io/v1alpha3/ServiceEntry/default/external-svc/endpoint1
```
hzxuzhonghu, line 88: How about pod?  It is like `cluster1/service` or `cluster1/pod`. Seems not very clear

kdorosh: the unit tests in this PR highlight the _recommended_ pod format, although if you wanted to use the pod k8s UID you could also use that

the recommended format for pod would be `cluster1//Pod/default/my-pod`

costinm:
At the very least, if we want to invent a new format - let's at least require that the UID is compatible with the SNI rules in TLS so it can be used in SNI routes ( that is normally hostname format - I don't understand what people have against the 30 year FQDN standard used in all the internet - but at least format the new istio-UID with the same chars and syntax)

howardjohn:
Hostname is not meeting the "unique" aspect.

Issue #101, opened 2022-11-09
Suggestion: restructure project


digitalsanctum:
Given that there's going to be things like benches and integration tests, I suggest we refactor using cargo workspaces [1]. 

An example of it in practice can be found in the tokio project [2]. 

Albeit this project is not that large yet, this seems like something that should be done sooner rather than later?

[1] https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html
[2] https://github.com/tokio-rs/tokio/blob/master/Cargo.toml

howardjohn:
I am not sure I see the benefit? Would we have a workspace with one Cargo.toml? Or multiple Cargo.toml? If so, what would be the split?

The one benefit I can clearly see is different deps for tests ,but doesn't dev-dependencies already solve that?

stevenctl:
If this repo ever produces a different binary, or we want to expose some of our code as a library I would see value in switching. 

Is there evidence of difficulties switching late? I'd rather do it as needed.

digitalsanctum:
@howardjohn The two main benefits are incremental compilation and organization. I'd propose just the conventional layout to start [1]. As @stevenctl alluded to, you could have an arbitrary number of libs and bins while sharing the same `Cargo.lock` and `target` directory.

Admittedly, it may be premature to adopt workspaces but I'd still propose the idiomatic project structure.

[1] https://doc.rust-lang.org/stable/cargo/guide/project-layout.html

howardjohn:
I think we are following the conventional layout now, and sounds like this isn't a clear need for multi-crates yet. I think its fair to consider this done for now, and if we need to split crates we can do that as needed. Thanks for the help

PR #533, opened 2023-05-24
[WIP] Restructure to support multiple crates

nmittler:
This moves the ztunnel code into a separate ztunnel crate directory. The top-level `Cargo.toml` just has a workspace consisting of the ztunnel crate.


nmittler:
FYI @howardjohn @keithmattix and anyone else that was interested in this from today's discussion.

CODEOWNERS
```diff
@@ -2,5 +2,5 @@
 /Makefile*       @istio/wg-test-and-release-maintainers
 /*.md            @istio/wg-test-and-release-maintainers
 /common/         @istio/wg-test-and-release-maintainers
-/common-protos/  @istio/wg-test-and-release-maintainers
+/proto/          @istio/wg-test-and-release-maintainers
```
nmittler, line 5: Just something I noticed ... I think this is correct?

Cargo.toml
```diff
@@ -1,91 +1,6 @@
-[package]
```
nmittler, line 1: an unfortunate diff ... this was just moved to the ztunnel directory.

ztunnel/build.rs
```diff
@@ -84,3 +88,11 @@ fn main() -> Result<(), anyhow::Error> {
     );
     Ok(())
 }
+
+fn workspace_dir() -> PathBuf {
+    let mut dir = std::env::current_dir().unwrap();
```
nmittler, line 93: This seems to work, although not sure if there is a better way to get the workspace dir.

bleggett:
Is moving ztunnel to a subfolder strictly necessary to support a secondary DNS crate?

bleggett (commented):
Is moving ztunnel to a subfolder strictly necessary to support a secondary DNS crate?

I'm fine with ztunnel being a module that is importable by $parent-wrapper but I'm not sure we need to churn-move the _current_ code to a subfolder to achieve that or support the DNS crate usecase (might be missing something tho).

We should be able to do 

```
<ztunnel-repo-root>/
             /dns (or /crates/dns, whatever)
```

and leave the existing ztunnel code where it is, no? That should be fully supported by `Cargo.toml`.

CODEOWNERS
howardjohn: No, common-protos is just not used here, should be removed if anything. protos/ should belong to ztunnel maintainers

nmittler:
@bleggett It would work. I guess my question is: what do we want? My feeling is that leaving everything as it is will result in a pretty lengthy and somewhat confusing list of top-level files/directories:

```
ztunnel/
  // Common stuff
  -  common/
  - docker/
  - proto/
  - scripts/
  - vendor/

  // Ztunnel-specific stuff
  - benches/
  - examples/
  - fuzz/
  - src/
  - tests/
  - build.rs
```

Putting all crates under subdirectories, the top-level is a bit cleaner and easier to understand:

```
  // Common stuff
  -  common/
  - docker/
  - proto/
  - scripts/
  - vendor/

  // Crates
  - crates/ztunnel/ (or just ztunnel)
```

NOTE: To simplify further, `proto` could be its own crate that generates the rust bindings. Something we could discuss in the future.

If we can get general agreement (maybe a tall order?) that having everything under subdirs is a better approach, it's probably better just to do it now.

WDYT?

CODEOWNERS
nmittler: done.

bleggett:
> @bleggett It would work. I guess my question is: what do we want? My feeling is that leaving everything as it is will result in a pretty lengthy and somewhat confusing list of top-level files/directories
>
> If we can get general agreement (maybe a tall order?) that having everything under subdirs is a better approach, it's probably better just to do it now. It's a one-time inconvenience.
> 
> WDYT?

My feeling is that since this is the `ztunnel` repo, it makes sense for `ztunnel` to be the top-level thing in the repo, and breaking that convention feels weird.

Crates-that-ztunnel-depends-on can go in subfolders, with the idea that they could (if we establish a need beyond `istio` proper) eventually be published separately and/or theoretically moved to their own repo.


If we get to the point where we have so many of those subcrates that it gets difficult/confusing to manage them in this repo, then they should probably be moved out, or un-crated (though honestly I expect to have very few in-tree subcrates, a DNS server is sort of a special case).

nmittler:
@linsun from the meeting today, my understanding was that we reached agreement that we should support multiple crates. Building on that, I think the question here is how that should be done. IMO It's not exactly a single vendor problem .. it's more of a general question about how we want to structure the code to accommodate multiple crates. There will soon be a DNS crate ... so 2. We might go so far as to break the XDS out into it's own crate ... in that case, 3 crates. So the question is, how do we want all this to look WRT the structure of the ztunnel codebase.

hzxuzhonghu:
As proposed ealier  https://github.com/istio/ztunnel/issues/101

linsun:
> @linsun from the meeting today, my understanding was that we reached agreement that we should support multiple crates. Building on that, I think the question here is how that should be done. IMO It's not exactly a single vendor problem .. it's more of a general question about how we want to structure the code to accommodate multiple crates. There will soon be a DNS crate ... so 2. We might go so far as to break the XDS out into it's own crate ... in that case, 3 crates. So the question is, how do we want all this to look WRT the structure of the ztunnel codebase.

From the ambient meeting, I think folks are interested in learning more and viewing the PR. I could be off here but I don't think folks agreed that we should support multiple crates, without understanding the benefits to ambient users.

bleggett:
> @linsun from the meeting today, my understanding was that we reached agreement that we should support multiple crates. 
> As proposed ealier #101
> There will soon be a DNS crate ... so 2. We might go so far as to break the XDS out into it's own crate ... in that case, 3 crates. So the question is, how do we want all this to look WRT the structure of the ztunnel codebase.

Yeah sorry - I'm for multiple crates (DNS needs to be a separate crate, or at least I agree I'd like to start with that viewpoint) - I don't think we need to restructure the whole repo to accomplish that, since it's pretty simple to incrementally break out subcrates into a subfolder.

Let's get A Subcrate (DNS) in place and see how we like the pattern.

bleggett:
> @louiscryan I seem to recall something to the effect of "we should support multiple crates" as being explicitly said (by you?) toward the end of the discussion. Please correct me if I'm misremembering, since much of the discussion here was based on that premise.

We can support multiple ztunnel sub-crates, we don't need to restructure the repo to do that.

nmittler:
@bleggett 

> We can support multiple ztunnel sub-crates, we don't need to restructure the repo to do that. Let's do the former for DNS, and not the latter.

There are actually 2 questions going on in this thread, and I want to back off on the discussion of the particulars of *how* we restructure for the moment. @linsun's comments seemed to be more fundamental questioning on whether or not ztunnel should even really support multiple crates. I want to get clarity on that first. Make sense?

nmittler:
Closing for now ... will open a DNS PR shortly so we can discuss this with a more concrete example.

Issue #487, opened 2023-03-30
DNS in Ambient Mesh


nmittler:
This is a top-level issue to track DNS as a feature for ambient mesh. In Istio, the control plane programmed pilot-agent, which acted as a DNS proxy for the container. Since ztunnel is not using pilot-agent and can also be multi-tenant, we need a path forward for DNS.

The first part of this work is to propose a design for ambient DNS. Should consider both single and multi-tenant scenarios. It should also discuss possible choices for technology (e.g. should we use an existing DNS solution? Should we build one from scratch?).

nmittler:
RFC is [here](https://docs.google.com/document/d/1EZDMZXgy4N-OjPC-FUvUdKwXewibxeowEWm1B20UeqE/edit#)

nmittler:
Re-opening until we have DNS added to our E2E testing of ztunnel.

howardjohn:
DNS is implemented and tested now, closing this out. thanks everyone

PR #536, opened 2023-05-29
Initial DNS proxy implementation

nmittler:
This copies much of the logic from Istio, however it also supports a multi-tenant proxy.

## New crate

The new `istio-dns` crate extends the Trust-DNS API to specifically support the DNS proxy use case. The existing Trust-DNS `Catalog` does not allow an `Authority` to conditionally indicate its answers as authoritative, which is something that a proxy needs to do. The code here was written with the intent of eventually upstreaming to Trust-DNS.

## Host aliases

Within ztunnel, a new `dns` package implements the `istio-dns` `Resolver`, and serves DNS directly from the `WorkloadStore`.

The logic for handling host aliases has been somewhat inverted from Istio. Istio pre-generated aliases for each client and added entries for all possible hosts to the lookup table.

However, in shared proxy mode we have to handle the problem that some host aliases are client-specific (e.g. just service-name without namespace). To account for this, we dynamically run the alias logic in reverse, trying to figure out the FQDN from the requested hostname. This means that no additional entries for aliases were necessary in the `WorkloadStore`.

## Forwarding

Ztunnel dns uses one of two types of DNS forwarders, depending on if shared or dedicated mode. When in shared mode, it needs to use the configuration for the client pod in order to forward to the appropriate upstream resolver.

Fixes https://github.com/istio/ztunnel/issues/487

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -34,6 +34,12 @@ use tokio::net::{TcpListener, TcpSocket, TcpStream};
 use tokio::time::timeout;
 use tracing::{error, trace, warn, Instrument};
 
+// Only public for testing. Needed by test_util.
+#[cfg(any(test, feature = "testing"))]
+pub mod dns;
+#[cfg(not(any(test, feature = "testing")))]
+mod dns;
+
 mod inbound;
 mod inbound_passthrough;
 mod outbound;
@@ -46,6 +52,7 @@ pub struct Proxy {
     inbound_passthrough: InboundPassthrough,
     outbound: Outbound,
     socks5: Socks5,
+    dns_proxy: Option<dns::DnsProxy>,
 }
 
 #[derive(Clone)]
@@ -67,7 +74,7 @@ impl Proxy {
         drain: Watch,
     ) -> Result<Proxy, Error> {
         let mut pi = ProxyInputs {
-            cfg,
+            cfg: cfg.clone(),
             state,
             cert_manager,
             metrics,
@@ -81,21 +88,28 @@ impl Proxy {
         let inbound_passthrough = InboundPassthrough::new(pi.clone()).await?;
         let outbound = Outbound::new(pi.clone(), drain.clone()).await?;
         let socks5 = Socks5::new(pi.clone(), drain).await?;
+        let dns_proxy = new_dns_proxy(pi.clone()).await?;
+
         Ok(Proxy {
             inbound,
             inbound_passthrough,
             outbound,
             socks5,
+            dns_proxy,
         })
     }
 
     pub async fn run(self) {
-        let tasks = vec![
+        let mut tasks = vec![
             tokio::spawn(self.inbound_passthrough.run().in_current_span()),
             tokio::spawn(self.inbound.run().in_current_span()),
             tokio::spawn(self.outbound.run().in_current_span()),
             tokio::spawn(self.socks5.run().in_current_span()),
         ];
+
+        if let Some(dns_proxy) = self.dns_proxy {
+            tasks.push(tokio::spawn(dns_proxy.run().in_current_span()));
+        }
 
         futures::future::join_all(tasks).await;
     }
@@ -105,7 +119,31 @@ impl Proxy {
             outbound: self.outbound.address(),
             inbound: self.inbound.address(),
             socks5: self.socks5.address(),
+            dns_proxy: self.dns_proxy.as_ref().map(|dns_proxy| dns_proxy.address()),
         }
+    }
+}
+
+async fn new_dns_proxy(pi: ProxyInputs) -> Result<Option<dns::DnsProxy>, Error> {
+    // Determine whether we should enable DNS from proxy metadata.
+    const DNS_CAPTURE_METADATA: &str = "ISTIO_META_DNS_CAPTURE";
+    let dns_enabled = pi
+        .cfg
+        .proxy_metadata
+        .get(DNS_CAPTURE_METADATA)
+        .map_or(false, |val| val.trim().to_lowercase() == "true");
+    if dns_enabled {
+        Ok(Some(
+            dns::DnsProxy::new(
+                pi.cfg.dns_proxy_addr,
+                pi.cfg.network,
+                pi.state.state, // TODO(nmittler): this is ugly.
+                dns::forwarder_for_mode(pi.cfg.proxy_mode)?,
+            )
+            .await?,
+        ))
+    } else {
+        Ok(None)
     }
 }
 
@@ -114,6 +152,7 @@ pub struct Addresses {
     pub outbound: SocketAddr,
     pub inbound: SocketAddr,
     pub socks5: SocketAddr,
+    pub dns_proxy: Option<SocketAddr>,
 }
 
 #[derive(thiserror::Error, Debug)]
--- a/src/proxy/dns.rs
+++ b/src/proxy/dns.rs
@@ -1,0 +1,1172 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub mod forwarder;
+pub mod name_util;
+pub mod proxy;
+pub mod resolver;
+
+use crate::config::ProxyMode;
+use crate::proxy::dns::name_util::trim_domain;
+use crate::proxy::dns::resolver::{Answer, Resolver};
+use crate::proxy::Error;
+use crate::state::service::Service;
+use crate::state::workload::{NetworkAddress, Workload};
+use crate::state::ProxyState;
+use itertools::Itertools;
+use log::warn;
+use once_cell::sync::Lazy;
+use rand::seq::SliceRandom;
+use rand::thread_rng;
+use std::collections::HashSet;
+use std::net::{IpAddr, SocketAddr};
+use std::ops::Deref;
+use std::str::FromStr;
+use std::sync::{Arc, RwLock};
+use std::time::Duration;
+use tokio::net::{TcpListener, UdpSocket};
+use trust_dns_proto::error::ProtoErrorKind;
+use trust_dns_proto::op::ResponseCode;
+use trust_dns_proto::rr::{Name, RData, Record, RecordType};
+use trust_dns_resolver::system_conf::read_system_conf;
+use trust_dns_server::authority::LookupError;
+use trust_dns_server::server::Request;
+use trust_dns_server::ServerFuture;
+
+const DEFAULT_TCP_REQUEST_TIMEOUT: u64 = 5;
+const DEFAULT_TTL_SECONDS: u32 = 30;
+
+static SVC_CLUSTER_LOCAL: Lazy<Name> = Lazy::new(|| Name::from_str("svc.cluster.local").unwrap());
+static CLUSTER_LOCAL: Lazy<Name> = Lazy::new(|| Name::from_str("cluster.local").unwrap());
+
+/// A proxy that serves known hostnames from ztunnel data structures. Unknown hosts are
+/// forwarded to an upstream resolver.
+pub(super) struct DnsProxy {
+    addr: SocketAddr,
+    server: ServerFuture<proxy::Proxy>,
+}
+
+impl DnsProxy {
+    /// Creates a new proxy.
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - The socket address on which to run the DNS proxy.
+    /// * `network` - The network of the current node.
+    /// * `state` - The state of this proxy.
+    /// * `forwarder` - The forwarder to use for requests not handled by this proxy.
+    pub(super) async fn new<S: AsRef<str>>(
+        addr: SocketAddr,
+        network: S,
+        state: Arc<RwLock<ProxyState>>,
+        forwarder: Arc<dyn Forwarder>,
+    ) -> Result<Self, Error> {
+        // Create the DNS server, backed by ztunnel data structures.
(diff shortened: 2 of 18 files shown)


crates/dns/src/handler.rs
```diff
@@ -0,0 +1,193 @@
...
+/// Creates an appropriate response [Edns], if one was available in the request.
+fn response_edns(request: &Request) -> Option<Edns> {
+    // check if it's edns
+    if let Some(req_edns) = request.edns() {
+        let mut resp_edns: Edns = Edns::new();
+        resp_edns.set_max_payload(req_edns.max_payload().max(512));
+        resp_edns.set_version(req_edns.version());
+
+        // DNSSEC is not currently supported by the proxy.
+        resp_edns.set_dnssec_ok(false);
```
howardjohn, line 187: Why not if we are just forwarding? I don't think we do this in the Go version

crates/dns/src/kube.rs
```diff
@@ -0,0 +1,251 @@
...
+
+    /// The name of the service.
+    pub name: String,
+
+    /// The namespace of the service.
+    pub namespace: String,
+}
+
+/// Generates the set of valid alternative hosts for the service FQDN
+/// (e.g. `s1.ns1.svc.cluster.local`), based on the location of the client.
```
howardjohn, line 45:
```suggestion
/// (e.g. `s1.ns1.svc.cluster.local`), based on the location of the client.
```

The example is not a FQDN in the context of DNS/the DNS library we use. FQDN means has a trailing `.`

crates/dns/src/kube.rs
```diff
@@ -0,0 +1,251 @@
...
+    // Add 'name.namespace.'.
+    if num_fqdn_labels > 2 {
+        let mut iter = service.fqdn.iter();
+        let name_label = iter.next().unwrap();
+        let ns_label = iter.next().unwrap();
+        out.push(Name::from_labels(vec![name_label, ns_label])?.into());
+    }
+
+    // TODO(nmittler): Do we need this?
+    // Do we need to generate entries for 'name.namespace.svc', 'name.namespace.svc.cluster', etc. ?
```
howardjohn, line 109: `.svc` - yes

`.svc.cluster` is not a valid domain

"If these are not that frequently used," is not relevant IMO. Correctness >>>> everything else

crates/dns/src/handler.rs
```diff
@@ -0,0 +1,193 @@
...
+
+/// Sends the response to the [ResponseHandler] and handles any errors.
+async fn send_response<'a, R: ResponseHandler>(
+    response: MessageResponse<
+        '_,
+        'a,
+        impl Iterator<Item = &'a Record> + Send + 'a,
+        impl Iterator<Item = &'a Record> + Send + 'a,
+        impl Iterator<Item = &'a Record> + Send + 'a,
+        impl Iterator<Item = &'a Record> + Send + 'a,
```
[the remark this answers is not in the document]
nmittler: TrustDNS :)

crates/dns/src/kube.rs
nmittler: Yeah ... I copied this comment over from the original code. I'm fine with just getting rid of it.

nmittler:
FYI @arvindsree

crates/dns/src/proxy.rs
```diff
@@ -0,0 +1,261 @@
...
+use tracing::{error, warn};
+use trust_dns_proto::op::{Edns, Header, MessageType, OpCode, ResponseCode};
+use trust_dns_proto::rr::Record;
+use trust_dns_resolver::error::ResolveErrorKind;
+use trust_dns_server::authority::{LookupError, MessageResponse, MessageResponseBuilder};
+use trust_dns_server::server::{Request, RequestHandler, ResponseHandler, ResponseInfo};
+
+/// A Trust-DNS [ResponseHandler] that proxies all DNS requests.
+///
+/// A proxy is fundamentally different than an `Authority` in TrustDNS, since the answers may
```
bleggett, line 26: Do we know if upstream would be amenable to this in theory?

it's hard to grok the cost of carrying this (more-verbose) workaround if we don't have a read on whether it can ever be upstreamed or not.

crates/dns/src/proxy.rs
```diff
@@ -0,0 +1,261 @@
...
+
+async fn send_lookup<R: ResponseHandler>(
+    request: &Request,
+    response_handle: R,
+    answer: Answer,
+) -> ResponseInfo {
+    let mut response_header = Header::response_from_request(request.header());
+
+    // We are the authority here, since we control DNS for known hostnames
+    response_header.set_authoritative(answer.is_authoritative());
```
bleggett, line 84: Does this mimic the current Go DNS logic?  Are we going to confuse anything if some responses are marked as authoritative and some not, from what the client (believes is) the same DNS server?

src/proxy/dns.rs
```diff
@@ -0,0 +1,532 @@
...
+    /// Creates a new proxy.
+    pub(super) async fn new(cfg: config::Config, pi: ProxyInputs) -> Result<Option<Self>, Error> {
+        // Create the DNS proxy only if configured to do so.
+        match cfg.enable_dns_proxy {
+            None | Some(false) => return Ok(None),
+            _ => Ok(Some(
+                Self::new_internal(cfg.dns_proxy_addr, cfg.proxy_mode, pi).await?,
+            )),
+        }
+    }
```
bleggett, line 60: Nit: I'd kinda prefer the cfg check to be moved out and live in `proxy.rs` rather than here (and then drop the `new`/`new_internal` stuff)

src/service.rs
bleggett, line 1: These are AFAICT good changes, but they're strictly unrelated to the DNS stuff and it would help shrink the DNS diff if they were moved into a sep. PR, FWIW.

src/workload.rs
bleggett, line 1: Ditto, this seems mostly unrelated (but good cleanup) - sep. PR would keep the reviewable diff down.

bleggett (commented):
Just minor comments after a cursory look, what's here seems good.

src/proxy/dns.rs
nmittler: Yeah I was going back and forth on this as well. I think my thought here was that I tried to keep as much DNS code as possible out of the other files. But it's a simple check ... I'm fine moving it.

crates/dns/src/proxy.rs
nmittler: The behavior should be the same as the Istio code: https://github.com/istio/istio/blob/5aa2900dca83190962aea2eac8712ca88ca63da3/pkg/dns/client/dns.go#L289

crates/dns/src/proxy.rs
nmittler: I'd like to land this before I approach the Trust-DNS folks. It definitely seems to be a use case they hadn't considered in it's current form, so I think it's worth at least having the discussion. Even if they don't accept it, it's not a lot of code and is stuff that needs to be done if we're going to use Trust-DNS anyway.

src/service.rs
nmittler: Opened #541

src/workload.rs
nmittler: Opened #541

bleggett:
- Note that port 53 capture is _technically_ implemented already in `istio-cni` iptables mode: https://github.com/istio/istio/blob/master/cni/pkg/ambient/net_linux.go#L461 and it should work if `ISTIO_META_DNS_CAPTURE` and related are enabled - but it hasn't tested with ambient so YMMV/twiddling may be required.

- Ultimately, as has been discussed before, we should IMO drop/deprecate/alias "DNS CAPTURE" as a separate configu line item and just support generic port redirection capture inclusions/exclusions for TCP/UDP instead - https://github.com/istio/istio/issues/43700 - that ultimately probably requires ztunnel to do UDP proxying tho which it doesn't yet.

crates/dns/src/proxy.rs
```diff
@@ -0,0 +1,261 @@
...
+        resp_edns.set_dnssec_ok(req_edns.dnssec_ok());
+
+        Some(resp_edns)
+    } else {
+        None
+    }
+}
+
+#[cfg(test)]
+#[cfg(any(unix, target_os = "windows"))]
```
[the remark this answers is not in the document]
kdorosh: at least for ci?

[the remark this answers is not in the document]
nmittler: This was copied from Trust-DNS. I suspect they might be accounting for cygwin, or something of the like?

nmittler:
@howardjohn @kdorosh @bleggett I believe this is ready to go.  The relevant Istio DNS tests were copied here and are passing. There are a couple of tests missing (due to current system limitations), which are marked as `TODO` with associated bugs.

Tests are currently unit/integration-level (DNS client/server using loopback) .. they do not use iptables/eBPF redirection. Happy to add it if folks can suggest how/where best to do so... otherwise, we can address in a follow-on PR.

bleggett:
> @howardjohn @kdorosh @bleggett I believe this is ready to go. The relevant Istio DNS tests were copied here and are passing. There are a couple of tests missing (due to current system limitations), which are marked as `TODO` with associated bugs.
> 
> Tests are currently unit/integration-level (DNS client/server using loopback) .. they do not use iptables/eBPF redirection. Happy to add it if folks can suggest how/where best to do so... otherwise, we can address in a follow-on PR.

I think a follow-on PR is fine - ideally we want to just re-run the existing`istio/istio` ambient integ test suite with DNS capture enabled, and that wouldn't live in this repo anyway.

bleggett (approved):

howardjohn:
`make test` doesn't actually test the crate as far as I can tell, and running the test locally fails

howardjohn:
Ideally the fix is not in `Makefile` either as then folks will miss it when they run `cargo test`

src/proxy.rs
```diff
@@ -67,7 +69,7 @@ impl Proxy {
         drain: Watch,
     ) -> Result<Proxy, Error> {
         let mut pi = ProxyInputs {
-            cfg,
+            cfg: cfg.clone(),
```
howardjohn, line 72:
```suggestion
            cfg,
```

src/proxy.rs
```diff
@@ -105,15 +114,43 @@ impl Proxy {
             outbound: self.outbound.address(),
             inbound: self.inbound.address(),
             socks5: self.socks5.address(),
+            dns_proxy: match &self.dns_proxy {
+                None => SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), 0),
```
howardjohn, line 118: nit: should we make it an Option?

src/proxy.rs
```diff
@@ -105,15 +114,43 @@ impl Proxy {
...
         }
     }
 }
 
+async fn new_dns_proxy(pi: ProxyInputs) -> Result<Option<dns::DnsProxy>, Error> {
+    // Determine whether we should enable DNS from proxy metadata.
+    const DNS_CAPTURE_METADATA: &str = "ISTIO_META_DNS_CAPTURE";
+    let dns_enabled = pi
+        .cfg
+        .proxy_metadata
```
howardjohn, line 130: cfg should probably have a `dns_proxy_enabled` (or perhaps `Option<SocketAddr>`) and we just use it directly?

Also I think the istio_meta_ is stripped (https://github.com/istio/ztunnel/blob/17620574ce78abf62b98c8dda2900e55bbae1d1d/src/config.rs#L403)

src/proxy/dns.rs
```diff
@@ -0,0 +1,1164 @@
...
+
+    /// Returns the address to which this DNS proxy is bound.
+    pub(super) fn address(&self) -> SocketAddr {
+        self.addr
+    }
+
+    /// Runs this DNS proxy to completion.
+    pub async fn run(self) {
+        // TODO(nmittler): Do we need to use drain?
+        if let Err(e) = self.server.block_until_done().await {
```
howardjohn, line 106: What marks it "done"? Do wen eed to close the listeners ourself?

howardjohn: Seems we get an error on sigterm:

```
WARN ztunnel::proxy::dns: DNS server shutdown error: Internal error in spawn: task 4 was cancelled
```

howardjohn (commented):
I mostly reviewed the impact on existing codepaths as this already has some LGTMs, I don't have a lot of time to thoroughly review the whole change, and I don't want to block it. Looks safe. main comment is the testing issue of the additional crate

src/proxy.rs
nmittler: It was an option previously, but the question came up as to why would it be an option and just not false by default.

nmittler:
@howardjohn 

> make test doesn't actually test the crate as far as I can tell, and running the test locally fails

Ah right. I think the "correct" way to support multiple crates would be to use a workspace. That has a number of advantages anyway, such as mananging one version for depdencies across crates.

Anyway, since the code in the crate was really something that should be upstreamed anyway .. I'll just get rid of the extra crate for now.

src/proxy.rs
nmittler: I was hoping to leverage the existing configuration for DNS capture for now, rather than introducing something new. I don't have a strong opinion here.

@bleggett thoughts?

Regarding using proxy metadata (since it seems I'm probably not using it correctly here), do we have other examples in ztunnel that use it? If we have constants defined somewhere, I can add the DNS one there.

howardjohn: I don't mean anything user facing, just we currently process all settings into `cfg` struct and now we are parsing them on-demand

src/proxy/dns.rs
nmittler: Yeah, probably worth upstreaming a fix a fix here. I'm guessing the behavior we want would be similar to our draining logic?

howardjohn: Maybe yeah. It probably matters less for DNS though since there is less persistent connections (none for UDP)

src/proxy.rs
nmittler: Ok sure .. just trying to figure out the desired end state here. Are you suggesting that I read the metadata and then set the flag on the config based on that?

src/proxy/dns.rs
nmittler: Raised an issue with Trust-DNS: https://github.com/bluejekyll/trust-dns/issues/1976. Let me know if the ask sounds right to you.

howardjohn (approved):

PR #537, opened 2023-06-01
make sure on-demend is requesting right type

GregHanson:
control plane only serves Address

control plane only has delta generators for type ADDRESS
https://github.com/istio/istio/blob/master/pilot/pkg/xds/delta.go#L431
->
https://github.com/istio/istio/blob/master/pilot/pkg/xds/workload.go#L43

delta.go does a specific map look up in its generator for type url - and there is only one for ADDRESS:
https://github.com/istio/istio/blob/master/pilot/pkg/xds/discovery.go#L561

otherwise a nil delta is returned

### DIFF
--- a/src/xds/client.rs
+++ b/src/xds/client.rs
@@ -231,7 +231,7 @@ impl Demander {
                 tx,
                 ResourceKey {
                     name,
-                    type_url: xds::WORKLOAD_TYPE.to_string(),
+                    type_url: xds::ADDRESS_TYPE.to_string(),
                 },
             ))
             .await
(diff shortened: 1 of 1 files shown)


GregHanson:
as far as I can tell, the on demand logic is never firing when things are working. I think I've only seen it during development and I had the control plane pushing a bad/incomplete WDS config

howardjohn (approved):

kdorosh (commented):
lgtm

howardjohn:
on-demand is off by default. But hoping to work on it more in the future

PR #541, opened 2023-06-08
Move service data to ServiceStore

nmittler:
This is a cleanup resulting from #536. This moves all service-related datastructures to a sub repository called `ServiceStore`. It has a simple API, which is used by the `WorkloadStore` and clarifies the contract between the two.

The separation also helps to generally simplify the `WorkloadStore`, which was becoming increasingly monolithic.


hzxuzhonghu (commented):
I would like the idea to separate them, but I donot think we can seperate service from workload store well, each store records map between svc->workload or reverse.

nmittler:
> I would like the idea to separate them, but I donot think we can seperate service from workload store well, each store records map between svc->workload or reverse.

This PR does it, it seems to work fine at least for the current implementation. Were you envisioning future work?

hzxuzhonghu:
Yes, the maps sometimes are bidirectional, I am thinking if in rust can we declare Store struct in one place and impement it in another

nmittler:
> Yes, the maps sometimes are bidirectional, I am thinking if in rust can we declare Store struct in one place and impement it in another

So do you have a concrete work item that needs this, or is it theoretical at this point?

ilrudie (commented):
I like the idea and the work here to break out service. I left a question about where service lives in our hierarchy but don't wish to block if folks like this and want to move it forward.

src/service.rs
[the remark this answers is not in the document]
nmittler: @ilrudie FYI, the DNS logic will be using services directly. I think making service a top-level construct makes sense and probably will happen eventually, although I'm not sure if it's worth doing here. I think this PR is a bit easier to consume for everyone and is a step in that direction. WDYT?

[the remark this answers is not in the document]
ilrudie: That makes sense to me. I hadn't see a reason to make it top level but you've got one so I'm OK with this. TY

src/service.rs
```diff
@@ -0,0 +1,327 @@
...
+    staged_vips: HashMap<NetworkAddress, HashMap<NetworkAddress, Endpoint>>,
+
+    /// Maintains a mapping of workload IP to VIP. This is used only to handle removal of service
+    /// endpoints when a workload is removed.
+    workload_to_vips: HashMap<NetworkAddress, HashSet<NetworkAddress>>,
+
+    /// Allows for lookup of services by network address, the service's xds secondary key.
+    /// this map stores the same services as `services_by_hostname` so we can mutate the services in both maps
+    /// at once when a service is updated or workload updates affect a service's endpoints.
+    by_vip: HashMap<NetworkAddress, Arc<RwLock<Service>>>,
```
howardjohn, line 97: next pr should remove the `Arc<RwLock` :slightly_smiling_face:

howardjohn (approved):

src/service.rs
nmittler: Yeah, I'm not a fan of that either ... any idea why this was done?

howardjohn: Something about feeling more confident in keeping the two maps in sync if they share the same object, but I think it can be done without

kdorosh: it was done for these three cases (which are covered by unit tests)
- https://github.com/istio/ztunnel/blob/b64f8255cc7baa64549384e083c0965114fbd177/src/workload.rs#L1443-L1445
- https://github.com/istio/ztunnel/blob/b64f8255cc7baa64549384e083c0965114fbd177/src/workload.rs#L1504-L1506
- https://github.com/istio/ztunnel/blob/b64f8255cc7baa64549384e083c0965114fbd177/src/workload.rs#L1531-L1533

either we need to perform this mutation on every copy of the stored data per index, or put the same data behind the index (i.e. by network/IP and ns/hostname). this opted to put the same data behind the index without pulling in new deps

perhaps a cleaner way of doing this would be something like https://docs.rs/double-map/latest/double_map/

src/service.rs
```diff
@@ -0,0 +1,327 @@
...
+
+    /// Allows for lookup of services by namespaced and hostname, the service's xds primary key.
+    /// this map stores the same services as `services_by_ip` so we can mutate the services in both maps
+    /// at once when a service is updated or workload updates affect a service's endpoints.
+    by_host: HashMap<String, Vec<Arc<RwLock<Service>>>>,
+}
+
+impl ServiceStore {
+    /// Returns the [Service] matching the given VIP.
+    pub fn get_by_vip(&self, vip: &NetworkAddress) -> Option<Arc<RwLock<Service>>> {
```
kdorosh, line 107: not a huge fan of returning the `Arc<RwLock`'s themselves, since this is the underlying data store. this feels like a leaky abstraction. id prefer we return a copy of the data that the caller can do whatever it pleases with (despite the performance cost)

this also can make for sneaky deadlocks (e.g. two different callers of this function trying to lock the same lock, and it being unclear that they are using the same one)

kdorosh (commented):
lgtm overall, my only comment is noted in https://github.com/istio/ztunnel/pull/541/files#r1224678218

kdorosh:
looks like my concern is being handled in https://github.com/istio/ztunnel/pull/545/files#r1224695814

PR #545, opened 2023-06-09
Refactoring WorkloadStore and ServiceStore

nmittler:
This addresses a few things:

* Creates a new `state` package. At the top level, contains `ProxyState` (container for the workloads and service stores), `DemandProxyState` (renamed from `WorkloadInformation`: wraps around `ProxyState` and adds an on-demand API), and `ProxyStateManager` (renamed from `WorkloadManager`)
* Moves `ServiceStore` out of `WorkloadStore`, making it a top-level store within `ProxyState`.
* Replaces `Mutex` with a `RwLock`. The `DemandProxyStore` now has a single `RwLock` for access to the entire `ProxyState`.
* Removes individual `RwLock` around `Workload` and `Service`. Instead, all call sites now just lock (read or write) the entire store.
* Extracted code that converts from XDS and updates the stores and put it into a new structure: `ProxyStateUpdater`. This implements all XDS `Handler` traits.


src/admin.rs
```diff
@@ -815,4 +811,29 @@ mod tests {
             resp_str.contains(r#"waypoint":{"destination":"defaultnw/127.0.0.10","port":15008}"#)
         );
     }
+
+    pub fn test_store(
```
ilrudie, line 815: The naming of this function suggests it may be a test in and of itself at first glance, consider a name which clarifies that it isn't a test

I think you moved it so it's only used in a local scope now as well so it can probably be private
```suggestion
   fn new_mock_store(
```

src/proxy/inbound.rs
```diff
@@ -345,7 +342,7 @@ impl Inbound {
                 };
                 let connection_metrics = traffic::ConnectionOpen {
                     reporter: Reporter::destination,
-                    source,
+                    source: source.map(|wl| wl),
```
ilrudie, line 345: **question:** what is this intended to do? It's iterating through source, grabbing each wl and not mutating the wl? Was it intended to clone or mutate them somehow?

src/proxy/outbound.rs
```diff
@@ -789,4 +786,21 @@ mod tests {
         gateway: &'a str,
         request_type: RequestType,
     }
+
+    pub fn test_store(xds_workloads: Vec<XdsWorkload>) -> anyhow::Result<WorkloadInformation> {
```
ilrudie, line 790: same as above. not sure it needs to be pub any more and I think the name could be less ambiguous about what it's doing.
```suggestion
    fn new_mock_store(xds_workloads: Vec<XdsWorkload>) -> anyhow::Result<WorkloadInformation> {
```

src/service.rs
```diff
@@ -188,59 +203,69 @@ impl ServiceStore {
...
+                self.insert(service);
             }
         }
     }
 
     /// Adds the given service.
     pub(super) fn insert(&mut self, mut service: Service) {
         // First mutate the service and add all missing endpoints
-        for addr in service.addresses.as_slice() {
+        for addr in &service.vips {
```
ilrudie, line 220:
```suggestion
        for vip in &service.vips {
```

src/service.rs
```diff
@@ -188,59 +203,69 @@ impl ServiceStore {
...
-                    service.endpoints.insert(wip.clone(), ep);
-                }
-            }
-
-            // If svc already exists, add old endpoints to new svc
-            if let Some(prev) = self.by_vip.get_mut(addr) {
-                let prev = prev.read().unwrap();
-                for (wip, ep) in prev.endpoints.iter() {
-                    service.endpoints.insert(wip.clone(), ep.clone());
+            if let Some(endpoints) = self.staged_vips.remove(addr) {
```
ilrudie, line 223:
```suggestion
            if let Some(endpoints) = self.staged_vips.remove(vip) {
```

src/proxy/outbound.rs
```diff
@@ -789,4 +786,21 @@ mod tests {
...
+            updater.insert_workload(w)?;
+        }
+
+        Ok(WorkloadInformation {
+            workloads,
+            services,
+            demand: None,
+        })
+    }
 }
```
kdorosh, line 806: can we dedup this with the version in `src/admin.rs`?

src/service.rs
```diff
@@ -94,18 +103,18 @@ pub struct ServiceStore {
...
-    by_host: HashMap<String, Vec<Arc<RwLock<Service>>>>,
+    by_host: HashMap<String, Vec<Arc<Service>>>,
 }
 
 impl ServiceStore {
     /// Returns the [Service] matching the given VIP.
-    pub fn get_by_vip(&self, vip: &NetworkAddress) -> Option<Arc<RwLock<Service>>> {
-        self.by_vip.get(vip).cloned()
+    pub fn get_by_vip(&self, vip: &NetworkAddress) -> Option<Service> {
+        self.by_vip.get(vip).map(|s| s.deref().clone())
```
kdorosh, line 117: ah fantastic this addresses my concern noted on the last PR

src/service.rs
```diff
@@ -188,59 +203,69 @@ impl ServiceStore {
...
-            if let Some(prev) = self.by_vip.get_mut(addr) {
-                let prev = prev.read().unwrap();
-                for (wip, ep) in prev.endpoints.iter() {
-                    service.endpoints.insert(wip.clone(), ep.clone());
+            if let Some(endpoints) = self.staged_vips.remove(addr) {
+                for (wip, ep) in endpoints {
+                    // Don't override if the new service has the endpoint.
+                    if !service.endpoints.contains_key(&wip) {
+                        service.endpoints.insert(wip.clone(), ep);
+                    }
```
kdorosh, line 228: this looks like a change in behavior.. was this a bug? is there a test to cover this case?

src/xds.rs
```diff
@@ -32,3 +47,265 @@ pub enum Error {
...
+        let mut workloads = self.workloads.write().unwrap();
+        let mut services = self.services.write().unwrap();
+        workloads.insert_workload(workload.clone())?;
+        while let Some(endpoint) = endpoints.pop() {
+            services.insert_endpoint(endpoint);
+        }
+        Ok(())
+    }
+
+    pub fn remove_workload(&self, xds_name: String) {
```
ilrudie, line 90: Could this be updated to take a borrow rather than an owned String?

src/workload.rs
```diff
@@ -117,10 +110,11 @@ pub mod address {
     #[derive(Debug, Eq, PartialEq, Clone, serde::Serialize, serde::Deserialize)]
     #[serde(untagged)]
     pub enum Address {
-        Workload(Box<Workload>),
-        Service(Box<Service>),
+        Workload(Workload),
+        Service(Service),
```
kdorosh, line 114: does this pass cargo clippy / cargo fmt? i remember this triggering warnings about objects of very different size within an enum

src/xds.rs
```diff
@@ -32,3 +47,265 @@ pub enum Error {
...
+        if parts.is_none() {
+            // we don't have namespace/hostname xds primary key for service
+            warn!(
+                "tried to remove service keyed by {} but it did not have the expected namespace/hostname format",
+                xds_name
+            );
+            return;
+        }
+
+        let (namespace, hostname) = parts.unwrap();
```
ilrudie, line 117: consider an `if let Some(parts)` for handling splits that return a None variant

src/workload.rs
```diff
@@ -900,64 +846,9 @@ pub struct WorkloadStore {
...
-                }
-            }
-        }
-        Ok(())
-    }
-
     // Determine if we should prefetch a certificate for this workload. Being "wrong" is not too bad;
     // a missing cert will be fetched on-demand when we get a request, so will just result in some extra latency.
-    fn should_prefetch_certificate(&self, w: &Workload) -> bool {
+    fn should_prefetch_certificate(&self, w: Arc<Workload>) -> bool {
```
kdorosh, line 851: not a huge fan of making this accept an Arc, i'd prefer to see the caller handle the lock

ilrudie (commented):
Looks like good work. Making a few suggestions to consider

src/proxy/inbound.rs
nmittler: Ah, this was the result of some intermediate changes. Removed the map.

src/workload.rs
```diff
@@ -516,67 +456,15 @@ impl WorkloadManager {
...
 /// WorkloadInformation wraps WorkloadStore, but is able to additionally request resources on-demand.
 /// It is designed to be cheap to clone.
 #[derive(serde::Serialize, Debug, Clone)]
 pub struct WorkloadInformation {
     #[serde(flatten)]
-    pub info: Arc<Mutex<WorkloadStore>>,
+    pub workloads: Arc<RwLock<WorkloadStore>>,
+
+    #[serde(flatten)]
+    pub services: Arc<RwLock<ServiceStore>>,
```
nmittler, line 467: FYI, I'm not loving this. Now there are 2 separate locks which need to be locked, which means if a call site needs both it has to lock in a consistent order. It would be better to have a single object that is lockable and contains all the various stores. I'm open to suggestions.

src/workload.rs
nmittler: Ah, originally I had switched the entire interface to returning `Arc<...>`, which would also fix this. Since then I went the other way and have everything returning copies, so we need the `Box` again. Fixed.

src/workload.rs
nmittler: This is only used within the struct, it's not a public API. The fact that it's an Arc means we don't need to copy.

nmittler: The latest code gets rid of the Arc.

src/state.rs
```diff
@@ -0,0 +1,415 @@
...
+pub struct DemandProxyState {
+    #[serde(flatten)]
+    pub state: Arc<RwLock<ProxyState>>,
+
+    /// If present, used to request on-demand updates for workloads.
+    #[serde(skip_serializing)]
+    demand: Option<Demander>,
+}
+
+impl DemandProxyState {
```
nmittler, line 82: Note to reviewers: The functions here are just a direct copy of those from `WorkloadInformation`.

src/state/service.rs
```diff
@@ -49,7 +50,13 @@ impl Service {
 #[derive(Debug, Eq, PartialEq, Clone, serde::Serialize, serde::Deserialize)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct Endpoint {
+    /// The service VIP for this endpoint.
+    pub vip: NetworkAddress,
```
nmittler, line 54: Added the vip to this structure to support the case of removing and then re-adding services. The VIP is needed to repopulate the waypoints to VIP map.

src/state/workload.rs
```diff
@@ -0,0 +1,1101 @@
+// Copyright Istio Authors
```
nmittler, line 1: Note to reviewers: This is just a move of `workload.rs`, minus the `WorkloadInformation` and `WorkloadManager`.

nmittler:
@ilrudie @kdorosh @howardjohn PTAL. I've done some further refactoring to move to a single RwLock. Updated the description to include the new changes.

src/service.rs
nmittler: The previous code relied on the fact that all inserted `Services` had no endpoints. This was because it was called with services derived from the raw XDS, which don't contain endpoints.

This is no longer the case, however, due to the fact that services are now immutable as they are stored as `Arc<Service>`. Now when a service is updated, we have to remove it and then add the updated one in its place.

Existing tests already cover this case.

src/state/service.rs
kdorosh: I like this change; I much prefer immutable services to the on-the-heap initial implementation I did. thank you

src/state/service.rs
```diff
@@ -94,18 +101,18 @@ pub struct ServiceStore {
...
     /// this map stores the same services as `services_by_hostname` so we can mutate the services in both maps
     /// at once when a service is updated or workload updates affect a service's endpoints.
-    by_vip: HashMap<NetworkAddress, Arc<RwLock<Service>>>,
+    by_vip: HashMap<NetworkAddress, Arc<Service>>,
 
     /// Allows for lookup of services by namespaced and hostname, the service's xds primary key.
     /// this map stores the same services as `services_by_ip` so we can mutate the services in both maps
     /// at once when a service is updated or workload updates affect a service's endpoints.
-    by_host: HashMap<String, Vec<Arc<RwLock<Service>>>>,
+    by_host: HashMap<String, Vec<Arc<Service>>>,
```
kdorosh, line 109: if service is immutable now, do we even need the `Arc` anymore? why can't we just put this on the stack?

kdorosh, line 109: can we make `by_host` take a `NamespacedHostname` rather than just a string? I'd prefer compilation errors to confusing runtime ones if we try to send a malformed key

src/state/service.rs
```diff
@@ -188,59 +201,69 @@ impl ServiceStore {
...
-            }
-
-            // If svc already exists, add old endpoints to new svc
-            if let Some(prev) = self.by_vip.get_mut(addr) {
-                let prev = prev.read().unwrap();
-                for (wip, ep) in prev.endpoints.iter() {
-                    service.endpoints.insert(wip.clone(), ep.clone());
+            if let Some(endpoints) = self.staged_vips.remove(vip) {
+                for (wip, ep) in endpoints {
+                    // Don't override if the new service has the endpoint.
```
kdorosh, line 223: can we add a debug log line (or perhaps warning) in the event this happens? I'd think this is a bug not a race if we hit this scenario

src/state/service.rs
nmittler: I think it's still useful to reduce the overall storage required.

nmittler: This is a change required by #536. The DNS proxy only has the requested hostname.

kdorosh (commented):
direction lgtm overall; love the move to make service immutable in the store. left a couple comments but nothing major

src/state/service.rs
kdorosh: I think that's reasonable :+1:

kdorosh: can we update the comment to correctly reflect that this is no longer keyed by namespaced hostname then?

src/cert_fetcher.rs
```diff
@@ -0,0 +1,101 @@
...
+            Some(&w.node) == self.local_node.as_ref() &&
+            // If it doesn't support HBONE it *probably* doesn't need a cert.
+            (w.native_tunnel || w.protocol == Protocol::HBONE)
+    }
+}
+
+impl CertFetcher for CertFetcherImpl {
+    fn prefetch_cert(&self, w: &Workload) {
+        if self.should_prefetch_certificate(w) {
+            let tx = self.tx.lock().unwrap();
```
howardjohn, line 95: `tx` does not need a mutex. Its like a channel in Go, it handles syncronization

src/proxy/inbound.rs
```diff
@@ -385,23 +378,23 @@ impl Inbound {
     }
 
     async fn check_waypoint(
-        workloads: &WorkloadInformation,
+        state: &DemandProxyState,
```
howardjohn, line 381: nit: I noticed in some places we use `DemandProxyState` and some `&DemandProxyState`.

Cloning should be cheap. All functions take `&self` so reference is sufficient. So I guess we just use `DemandProxyState` if we don'rt want to deal with lifetimes and `&DemandProxyState` otherwise?

I am not sure if there is a standard practice here

src/state/service.rs
```diff
@@ -92,20 +99,19 @@ pub struct ServiceStore {
...
     /// Allows for lookup of services by network address, the service's xds secondary key.
-    /// this map stores the same services as `services_by_hostname` so we can mutate the services in both maps
-    /// at once when a service is updated or workload updates affect a service's endpoints.
-    by_vip: HashMap<NetworkAddress, Arc<RwLock<Service>>>,
-
-    /// Allows for lookup of services by namespaced and hostname, the service's xds primary key.
-    /// this map stores the same services as `services_by_ip` so we can mutate the services in both maps
-    /// at once when a service is updated or workload updates affect a service's endpoints.
-    by_host: HashMap<String, Vec<Arc<RwLock<Service>>>>,
+    by_vip: HashMap<NetworkAddress, Arc<Service>>,
```
howardjohn, line 102: I am not sure we need the Arc? this seems to work:

```diff
diff --git a/src/cert_fetcher.rs b/src/cert_fetcher.rs
index 607d1ab..033b61e 100644
--- a/src/cert_fetcher.rs
+++ b/src/cert_fetcher.rs
@@ -17,7 +17,7 @@ use crate::config::ProxyMode;
 use crate::identity::Priority::Warmup;
 use crate::identity::{Identity, SecretManager};
 use crate::state::workload::{Protocol, Workload};
-use std::sync::{Arc, Mutex};
+use std::sync::Arc;
 use tokio::sync::mpsc;
 use tracing::{debug, error, info};
 
@@ -45,7 +45,7 @@ pub fn new(cfg: &config::Config, cert_manager: Arc<SecretManager>) -> Arc<dyn Ce
 struct CertFetcherImpl {
     proxy_mode: ProxyMode,
     local_node: Option<String>,
-    tx: Mutex<mpsc::Sender<Identity>>,
+    tx: mpsc::Sender<Identity>,
 }
 
 impl CertFetcherImpl {
@@ -72,7 +72,7 @@ impl CertFetcherImpl {
         Self {
             proxy_mode: cfg.proxy_mode.clone(),
             local_node: cfg.local_node.clone(),
-            tx: Mutex::new(tx),
+            tx,
         }
     }
 
@@ -92,8 +92,7 @@ impl CertFetcherImpl {
 impl CertFetcher for CertFetcherImpl {
     fn prefetch_cert(&self, w: &Workload) {
         if self.should_prefetch_certificate(w) {
-            let tx = self.tx.lock().unwrap();
-            if let Err(e) = tx.try_send(w.identity()) {
+            if let Err(e) = self.tx.try_send(w.identity()) {
                 info!("couldn't prefetch: {:?}", e)
             }
         }
diff --git a/src/state.rs b/src/state.rs
index 9588252..8ceca8b 100644
--- a/src/state.rs
+++ b/src/state.rs
@@ -72,7 +72,7 @@ pub struct ProxyState {
 #[derive(serde::Serialize, Debug, Clone)]
 pub struct DemandProxyState {
     #[serde(flatten)]
-    pub state: Arc<RwLock<ProxyState>>,
+    state: Arc<RwLock<ProxyState>>,
 
     /// If present, used to request on-demand updates for workloads.
     #[serde(skip_serializing)]
diff --git a/src/state/service.rs b/src/state/service.rs
index bea47c3..3b24c7e 100644
--- a/src/state/service.rs
+++ b/src/state/service.rs
@@ -99,13 +99,13 @@ pub struct ServiceStore {
     workload_to_vips: HashMap<NetworkAddress, HashSet<NetworkAddress>>,
 
     /// Allows for lookup of services by network address, the service's xds secondary key.
-    by_vip: HashMap<NetworkAddress, Arc<Service>>,
+    by_vip: HashMap<NetworkAddress, Service>,
 
     /// Allows for lookup of services by hostname, and then by namespace. XDS uses a combination
     /// of hostname and namespace as the primary key. In most cases, there will be a single
     /// service for a given hostname. However, `ServiceEntry` allows hostnames to be overridden
     /// on a per-namespace basis.
-    by_host: HashMap<String, Vec<Arc<Service>>>,
+    by_host: HashMap<String, Vec<Service>>,
 }
 
 impl ServiceStore {
@@ -231,9 +231,6 @@ impl ServiceStore {
         let vips = service.vips.clone();
         let hostname = service.hostname.clone();
 
-        // Create the Arc.
-        let service = Arc::new(service);
-
         // Map the vips to the service.
         for vip in &vips {
             self.by_vip.insert(vip.clone(), service.clone());
@@ -314,7 +311,7 @@ impl ServiceStore {
                 }
 
                 // Remove successful.
-                Some(prev.deref().clone())
+                Some(prev)
             }
         }
     }
```

Unless this is doing some expensive clone I am missing now

howardjohn: I guess in get_by_host

howardjohn:
I think all my comments except the lock around channel ended up being nits/bike shedding

howardjohn (approved):

src/admin.rs
```diff
@@ -12,42 +12,40 @@
...
-use crate::identity::SecretManager;
-use crate::tls::asn1_time_to_system_time;
-use crate::version::BuildInfo;
-use crate::workload::LocalConfig;
-use crate::workload::WorkloadInformation;
-use crate::{signal, telemetry};
 
 struct State {
-    workload_info: WorkloadInformation,
+    state: DemandProxyState,
```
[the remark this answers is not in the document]
nmittler: Yeah, exactly ... so there actually are 2 things. It's unfortunate that `DemandProxyState` has to exist, but something has to have the lock on the actual `ProxyState`.

howardjohn:
```mermaid
graph BT
    ProxyStateManager-->DemandProxyState-->ProxyState
    ProxyStateManager-->ProxyStateUpdater-->ProxyState
    XdsClient-->ProxyStateUpdater
    Admin-->ProxyStateManager
    Proxy-->ProxyStateManager
```

incase it helps anyone else

nmittler:
I've removed the hold, since @kdorosh also seemed to be ok with the changes. We can always address additional comments in a follow-up PR.

PR #546, opened 2023-06-12
Add cluster_id to AuthSource Token to support multicluster authentication

imroc:
Istiod uses `third-party-jwt` to authenticate ztunnel, [extractClusterID](https://github.com/istio/istio/blob/c62932476b70066d844b5fc1f8eac60aaa82e259/security/pkg/server/ca/authenticate/kubeauth/kube_jwt.go#L112) from ztunnel's request metadata, and then call the apiserver of the corresponding cluster according to the extracted `clusterid` for jwt token validation (see [ValidateK8sJwt](https://github.com/istio/istio/blob/c62932476b70066d844b5fc1f8eac60aaa82e259/security/pkg/server/ca/authenticate/kubeauth/kube_jwt.go#L153) ).

Currently, there is no `clusterid` in ztunnel's metadata, so will use the primary cluster's apiserver to validate ([see this](https://github.com/istio/istio/blob/c62932476b70066d844b5fc1f8eac60aaa82e259/security/pkg/server/ca/authenticate/kubeauth/kube_jwt.go#L173-L175)). In a multi-cluster environment, we need to add `clusterid` to the metadata of ztunnel. 

This PR adds `clusterid` to ztunnel's request metadata.

### DIFF
--- a/src/identity/auth.rs
+++ b/src/identity/auth.rs
@@ -21,6 +21,7 @@ use tonic::{Code, Request, Status};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum AuthSource {
+    // JWT authentication source which contains the token file path and the cluster id.
     Token(PathBuf, String),
 }
 
--- a/src/identity/auth.rs
+++ b/src/identity/auth.rs
@@ -21,13 +21,13 @@ use tonic::{Code, Request, Status};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum AuthSource {
-    Token(PathBuf),
+    Token(PathBuf, String),
 }
 
 impl AuthSource {
     pub fn load(&self) -> io::Result<Vec<u8>> {
         match self {
-            AuthSource::Token(path) => {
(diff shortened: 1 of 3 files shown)


imroc:
@howardjohn PTAL

src/identity/auth.rs
```diff
@@ -21,13 +21,13 @@ use tonic::{Code, Request, Status};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum AuthSource {
-    Token(PathBuf),
+    Token(PathBuf, String),
```
hzxuzhonghu, line 24: add comments

hzxuzhonghu (commented):
LGTM,

src/identity/auth.rs
imroc: Comments added: https://github.com/istio/ztunnel/pull/546/commits/86b8212ef864e298f86be6bd52d54a3b2cf75e23

hzxuzhonghu (approved):

PR #548, opened 2023-06-13
Add a help command for the ztunnel cli

film42:
This should probably be replaced by something like clap but even this is a small quality of life improvement. Feel free to close if you're not interested.


howardjohn (approved):

hzxuzhonghu (commented):
/ok-to-test

PR #550, opened 2023-06-13
Hostname lookup

ilrudie:
- adds code to handle the `gatewayaddress::Destination::Hostname` enum variant when looking up an address

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -23,7 +23,7 @@ use crate::proxy::inbound::InboundConnect::{DirectPath, Hbone};
 use crate::proxy::{ProxyInputs, TraceParent, BAGGAGE_HEADER, TRACEPARENT_HEADER};
 use crate::rbac::Connection;
 use crate::socket::to_canonical;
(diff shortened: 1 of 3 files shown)


src/proxy/inbound.rs
```diff
@@ -383,45 +379,40 @@ impl Inbound {
...
-                false
-            }
-            None => false,
-        };
-        Ok(from_gateway)
+                None => false,
+            };
+            return from_gateway;
+        }
+        false // this occurs if gateway_address was None
```
hzxuzhonghu, line 415: nit: return early

src/proxy/inbound.rs
```diff
@@ -477,3 +468,178 @@ impl crate::tls::CertProvider for InboundCertProvider {
...
+        },
+    };
+    use std::{
+        collections::HashMap,
+        net::{Ipv4Addr, SocketAddrV4},
+        sync::RwLock,
+    };
+
+    #[tokio::test]
+    async fn check_gateway() {
```
hzxuzhonghu, line 490: test_check_gateway

PR #556, opened 2023-06-20
implement service entry SANs matching

GregHanson:


GregHanson:
dependent on https://github.com/kdorosh/istio/pull/1 and also https://github.com/istio/istio/pull/45471

src/tls/boring.rs
```diff
@@ -386,7 +383,7 @@ enum Verifier {
     None,
 
     // Allows exactly one identity, making sure at least one of the presented certs matches that identity
-    San(Identity),
+    San(Vec<Identity>),
```
hzxuzhonghu, line 386: update comment as well

hzxuzhonghu (commented):
lgtm

src/state.rs
```diff
@@ -40,6 +40,7 @@ pub mod workload;
 pub struct Upstream {
     pub workload: Workload,
     pub port: u16,
+    pub sans: Vec<String>,
```
ymesika, line 43: If this is always empty vector (didn't see any place where san is being added to it) why do you need it?

ymesika (commented):
Left a comment but overall lgtm.

src/state.rs
GregHanson: thanks - this snuck in during that latest merge with latest

ymesika (approved):
lgtm

howardjohn (approved):

kdorosh (commented):
how did we test that this works? i don't see any tests added covering the new behavior AFAICT

GregHanson:
follow-up PR to add some tests here:
https://github.com/istio/ztunnel/pull/601

PR #559, opened 2023-06-20
Add RBAC to benches

Stevenjin8:


benches/throughput.rs
```diff
@@ -20,6 +20,7 @@ use std::time::Duration;
 use criterion::{
     criterion_group, criterion_main, BenchmarkId, Criterion, SamplingMode, Throughput,
 };
+use ipnet::IpNet;
```
MorrisLaw, line 23: The initial [failure ](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_ztunnel/559/test_ztunnel/1671242834755194880#1:build-log.txt%3A582-585)you have now is because this isn't being used.

benches/throughput.rs
```diff
@@ -106,104 +172,115 @@ fn initialize_environment(mode: Mode) -> (Arc<Mutex<TestEnv>>, Runtime) {
...
+                b.to_async(&rt).iter(|| async {
+                    tcp::run_client(&mut env.lock().await.hbone, *size, Mode::ReadWrite).await
+                })
+            });
+        }
+    }
+}
+
+pub fn throughput(c: &mut Criterion) {
+    for policies in [vec![], create_test_policies()] {
```
keithmattix, line 216: Same as above; I'd make separate bench_with_input test cases for RBAC

scripts/bench-performance.sh
```diff
@@ -0,0 +1,24 @@
+#! /bin/bash
+#set -eo
+
+echo $PULL_BASE_SHA
+if [[ ! -v PULL_BASE_SHA ]]; then
+    echo "PULL_BASE_SHA is not set"
+    exit 1
+fi 
+if [[ ! -v PULL_PULL_SHA ]]; then
```
keithmattix, line 9: I'd use `PULL_HEAD_SHA` instead of `PULL_PULL_SHA`. The former is more common terminology

benches/throughput.rs
```diff
@@ -50,32 +51,95 @@ struct TestEnv {
...
         env::set_var("RUST_LOG", "error")
     }
     helpers::initialize_telemetry();
+    // NOTE: Runtime to run stuff.
     let rt = tokio::runtime::Builder::new_multi_thread()
         .enable_all()
         .build()
         .unwrap();
-    // Global setup: spin up an echo server and ztunnel instance
+    // Global setup: spin up an echo server and tunnel instance
```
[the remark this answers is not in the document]
Stevenjin8: My bad. I don't remember changing this. Maybe it was from a merge? regardless, I will look at the diff in the future.

scripts/bench-performance.sh
Stevenjin8: I deleted this file because I think it makes more sense to work with this in a separate PR (#561). I also think the terminology is weird but it's the value I got when I followed [these instructions](https://github.com/istio/istio/wiki/Working-with-Prow#writing-tests) to run my tests locally.

keithmattix: Fair; consider this one non blocking. It's up to you!

Stevenjin8: I feel like the way to go is to put the SHA's as arguments, but I'm not sure if that will work well with prow. Will investigate

keithmattix:
@Stevenjin8 your CLA check is failing; can you sign the CLA (as an individual) and make sure your commits are signed (`git commit -s`).

benches/throughput.rs
```diff
@@ -50,32 +51,95 @@ struct TestEnv {
...
+            rules: rules.clone(),
+        });
+    }
+
+    return policies;
+}
+
+fn initialize_environment(
+    mode: Mode,
+    policies: Vec<Authorization>,
```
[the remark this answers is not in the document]
Stevenjin8: What would `None` represent then? No policies? I would find that confusing because there are then two ways of representing no policies `None` and `vec![]`.

keithmattix (approved):

Stevenjin8:
Oops

hzxuzhonghu (approved):

PR #561, opened 2023-06-20
bench script

Stevenjin8:
Adds a script to compare performance across two different tests. The end goal is to run this as a prow command. Would it be better to have the commit SHAs be argument of the scripts? I'm unsure as to how that would pan out in the prow command.

Relevant PR: https://github.com/istio/test-infra/pull/4777

Update: Using environment variables seems like the way to go.

### DIFF
--- a/scripts/benchtest.sh
+++ b/scripts/benchtest.sh
@@ -1,0 +1,12 @@
+#! /bin/bash
+set -eux
+
+REPORT_PATH="out/rust/criterion/"
+
+git checkout $PULL_BASE_SHA
+cargo bench --profile dev -- --quick # FIXME: dev for faster compilation, quick for faster runs
+
+git checkout $PULL_PULL_SHA
+cargo bench --profile dev -- --quick # FIXME: dev for faster compilation, quick for faster runs
+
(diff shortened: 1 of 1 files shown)


howardjohn (commented):
Should we use save-baseline and baseline to compare the two?

https://github.com/istio/ztunnel/blob/719985743665be2b1b5539d6ab155427f870803e/benches/README.md#L12-L14

howardjohn:
You want `/test bench_ztunnel`. Or `/test ?` to get a list of all commands

scripts/benchtest.sh
```diff
@@ -0,0 +1,12 @@
+#! /bin/bash
+set -eux
+
+REPORT_PATH="out/rust/criterion/"
+
+git checkout "$PULL_BASE_SHA"
+cargo bench -- --save-baseline master
+
+git checkout "$PULL_PULL_SHA"
+cargo bench -- --baseline-lenient master
```
Stevenjin8, line 10: `--baseline` will error when new benchamrks are added.

howardjohn:
Looks good. The actual results are a bit concerning though, looks like we have huge noise

howardjohn (approved):

PR #563, opened 2023-06-21
Add proxy meta cluster id config to ztunnel

hanxiaop:


howardjohn (changes requested):
Can you give context? what is this fixing? Current setup seems fine

hanxiaop:
> Can you give context? what is this fixing? Current setup seems fine

Can you check https://github.com/istio/istio/pull/45576#discussion_r1238017620 and https://github.com/istio/istio/pull/45576#discussion_r1238018119? Thanks

src/config.rs
```diff
@@ -362,6 +362,11 @@ impl ProxyConfig {
 }
 
 fn construct_proxy_config(mc_path: &str, pc_env: Option<&str>) -> anyhow::Result<ProxyConfig> {
+    // set cluster id to proxy metadata so that it can be used in the node metadata
+    let cluster_id = parse_default(CLUSTER_ID, DEFAULT_CLUSTER_ID.to_string())?;
+    let var_name = format!("{}{}", ISTIO_META_PREFIX, CLUSTER_ID);
+    env::set_var(var_name, cluster_id);
```
hzxuzhonghu, line 368: Why donot we parse `ISTIO_META_CLUSTERID` directly, same as sidecar.

Sorry if this was discussed somewhere before

hanxiaop: I actually made some changes to set `ISTIO_META_CLUSTER_ID` in https://github.com/istio/istio/pull/45576/files. I thought directly setting the environment variable here might be simpler than replacing the container with a new environment. However, I'm not sure if it's a great approach. It would be helpful to gather more opinions from @howardjohn and @GregHanson.

The reason why the cluster ID metadata is needed is discussed in this comment: https://github.com/istio/istio/pull/45576/files#r1238017620.

howardjohn: I am not sure I understand. To see it in Istiod we set `CLUSTER_ID`. We already set that as an env var. We already parse that into cluster_id. We should just explicitly add it in `fn node(&self) -> Node `

hanxiaop: Thanks, fixed. That was what I was trying to do.

hzxuzhonghu: @howardjohn I think we should update the `CLUSTER_ID` to `ISTIO_META_CLUSTER_ID` to be consistent

hzxuzhonghu: We may need other env like sidecar in future, ISTIO_META is to explicitly mark it will be sent to pilot via node metadata

hanxiaop: Not sure about the detail of legacy part, but I think since we have full control over ztunnel, we can avoid those `ISTIO_META_` prefix values. Also, we have compatibility in parsing `ISTIO_META_` prefix values, and if we really need those specific values, we can still add them in the old way.

hzxuzhonghu: I donot think `ISTIO_META_` is legacy, this is the convention to tell sidecar which env to bring to pilot.
We can deploy ztunnel on VM,  then we will diverge with sidecar by not respecting the naming standard

hzxuzhonghu: I've seen `NETWORK` env also without `ISTIO_META_`. As ztunnel is a daemonset, so i do not even know how we can customize each ztunnel. From my perspective, now we can have the same env values per cluster.  We lose the capability to support multi network per cluster.

hanxiaop: >  would it be better to bring this up in Ambient WG to gather more input?

This sounds good. I'm not quite sure about the correct approach in Ambient yet.

ilrudie: I've added this to the Ambient WG agenda. It won't be this week due to the holiday but hopefully we can get a quick direction on the 29th if consensus isn't reached here first.

bleggett: There are two things that are generally true:

- It is good to prefix env vars with a product-specific prefix - what the prefix is doesn't matter much.
- If the same env var means the same thing it should have a consistent name.

so I don't see the problem with ISTIO_META_CLUSTER_ID (other than the fact that what it does and why you'd want to change it is not documented here: https://istio.io/latest/docs/reference/commands/pilot-agent/)

hanxiaop: Thanks for the explaination. IIUC this is what originally I was trying to do. I'll revert and check.

hanxiaop: I reopened https://github.com/istio/istio/pull/45576, as this will set proxy meta of ztunnel with CLUSTER_ID

hzxuzhonghu (approved):
Thank you

hanxiaop:
@howardjohn PTAL, thanks

howardjohn (approved):

PR #565, opened 2023-06-22
Add tests for change_log_level

MorrisLaw:
This PR is a follow-up for comments left on [PR #557](https://github.com/istio/ztunnel/pull/557).

The following changes were made:
- changed level for `"Parsed level: {:?}"` message to debug instead of info. cc/ @keithmattix 
- changed status code for other error in `change_log_level` to also be BAD_REQUEST. cc/ @howardjohn 
- add unit tests for `change_log_level`. cc/ @GregHanson

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -440,13 +440,15 @@ async fn handle_gprof_heap(_req: Request<Incoming>) -> Response<Full<Bytes>> {
 
 #[cfg(test)]
 mod tests {
+    use super::change_log_level;
     use super::dump_certs;
     use super::handle_config_dump;
     use super::ConfigDump;
+    use crate::admin::HELP_STRING;
     use crate::config::construct_config;
     use crate::config::ProxyConfig;
     use crate::identity;
-    use crate::test_helpers::new_proxy_state;
+    use crate::test_helpers::{new_proxy_state, helpers};
     use crate::xds::istio::security::string_match::MatchType as XdsMatchType;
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
thanks!

src/admin.rs
```diff
@@ -361,11 +361,11 @@ fn change_log_level(reset: bool, level: &str) -> Response<Full<Bytes>> {
     match tracing::level_filters::LevelFilter::from_str(level) {
         Ok(level_filter) => {
             // Valid level, continue processing
-            tracing::info!("Parsed level: {:?}", level_filter);
+            tracing::debug!("Parsed level: {:?}", level_filter);
```
GregHanson, line 364: nit: this API isn't going to be called often (outside of debugging), could probably leave at `info` level

GregHanson (commented):
lgtm - thanks for the tests!

src/admin.rs
MorrisLaw: cc/ @keithmattix since he suggested the change initially. Any strong opinions against leaving as info?

keithmattix: No strong opinions; we can always lower the level later. Feel free to change it back to info

nmittler:
@MorrisLaw these tests are failing on CI: https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_ztunnel/536/test_ztunnel/1671993444202975232/build-log.txt.

Can you rollback? It's blocking my change. Thanks!

PR #567, opened 2023-06-23
Revert "Add tests for change_log_level"

MorrisLaw:
Reverts istio/ztunnel#565 because of a failure [noticed in CI after merge](https://github.com/istio/ztunnel/pull/565#issuecomment-1603356939).

PR #569, opened 2023-06-23
Add unit test for change_log_level

MorrisLaw:
This is a redo (and improvement) of the PR I merged yesterday and then subsequently had to [revert ](https://github.com/istio/ztunnel/pull/567) because of race conditions caused by accessing the same [global log value](https://github.com/istio/ztunnel/blob/63aafbf44be4df0f9fd3aaab67ed824ecee534ad/src/telemetry.rs#L24) at the same time before the correct assert could be called, I believe that's what led to the previous[ error noticed in CI](https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_ztunnel/536/test_ztunnel/1671993444202975232/build-log.txt). This new test combines all of the tests into one function to ensure that it's done in order. I validated this with the old unit tests and after with the new unit test by running `cargo test admin::tests::test_change_log_level`.

This PR also separates the repetitive logic of getting a `String` value from `Response<Full<Bytes>>` into it's own test_helper function called `get_response_str`. It also has the added effect of making the unit test more readable imo. This seemed like the best spot to place it but I'm open to suggestions if this is not ideal.

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -849,14 +849,14 @@ mod tests {
 
         let resp = change_log_level(true, "trace");
         let resp_str = get_response_str(resp).await;
-        assert!(resp_str.contains("current log level is trace"));
+        assert!(resp_str.contains("current log level is trace\n"));
 
         let resp = change_log_level(true, "info");
         let resp_str = get_response_str(resp).await;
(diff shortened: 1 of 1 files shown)


howardjohn (approved):
nice :+1:

PR #572, opened 2023-06-26
Update API for headless services

nmittler:
Preferring service keys rather than VIPs for referencing services from workload.

Also, change requirements to allow both addresses and hostnames for addressing a headless service.


proto/workload.proto
```diff
@@ -152,11 +157,11 @@ message Workload {
...
-  // Virtual IPs defines a set of virtual IP addresses the workload can be reached at.
-  // Typically these represent Service ClusterIPs.
-  // The key is a NetworkAddress string of the format network/IP, or just an IP if network
-  // of the service desired matches the network of the workload.
-  map<string, PortList> virtual_ips = 15;
+  // Deprecated. Use `services` to map this workload to service resources.
+  map<string, PortList> virtual_ips = 15 [deprecated = true];
+
+  // The services for which this workload is an endpoint.
+  repeated NamespacedHostname services = 22;
```
howardjohn, line 164: I think we need `map<NamespacedHostname, PortList>`. The port part is still important

howardjohn (commented):
looks good

proto/workload.proto
nmittler: Doesn't the service have the ports?

proto/workload.proto
```diff
@@ -152,11 +157,12 @@ message Workload {
...
-  // Typically these represent Service ClusterIPs.
-  // The key is a NetworkAddress string of the format network/IP, or just an IP if network
-  // of the service desired matches the network of the workload.
-  map<string, PortList> virtual_ips = 15;
+  // Deprecated. Use `services` to map this workload to service resources.
+  map<string, PortList> virtual_ips = 15 [deprecated = true];
+
+  // The services for which this workload is an endpoint.
+  // The key is the NamespacedHostname string of the format namespace/hostname.
+  map<string, PortList> services = 22;
```
hzxuzhonghu, line 165: With this, does that mean we will fetch two resource before a request can be redirected? Currently we only need fetch once

proto/workload.proto
```diff
@@ -58,12 +58,13 @@ message Service {
...
   string namespace = 2;
   // Hostname represents the FQDN of the service.
-  // For Kubernetes, this would be <name>.<namespace>.svc.<cluster domain>
+  // For Kubernetes, this would be <name>.<namespace>.svc.<cluster domain>.
   // TODO: support this field
   string hostname = 3;
   // Address represents the addresses the service can be reached at.
   // There may be multiple addresses for a single service if it resides in multiple networks,
   // multiple clusters, and/or if it's dual stack (TODO: support dual stack).
+  // For a headless kubernetes service, this list will be empty.
```
hzxuzhonghu, line 67: How about cross network

proto/workload.proto
nmittler: Yes, that's right. But to be fair, it's no different than if we were processing the raw k8s resources.

proto/workload.proto
hzxuzhonghu: I mean if a headless service instance is in another network, do not we need to specify the network gateway address?

proto/workload.proto
hzxuzhonghu: I am not sure as two xds requests are both though network

proto/workload.proto
nmittler: Still not sure I see what that has to do with Service. To access any workload that resides on another network, you just use the gateway address in the workload.

proto/workload.proto
nmittler: @howardjohn any thoughts here?

proto/workload.proto
hzxuzhonghu: Oh IC. headeless service access will require workload, then when ztunnel would fetch headless service?

proto/workload.proto
howardjohn: Not sure what you mean? So will try to answer but may be off

In standard XDS we load all Service and workloads at startup, just like Envoy. Envoy has like 8 resources so not sure why 2 is an issue

In "on demand", we would lookup a single Address. The control plane will return the service+workloads associated with that address. Its technically 1 type (Address), and definitely 1 request.

proto/workload.proto
nmittler: So on demand is not currently supported for DNS. I raised https://github.com/istio/ztunnel/issues/577

proto/workload.proto
hzxuzhonghu: With this change, it will fetch two resource on flying. That means a request latency will be client->ztunnel->ztunnel>server round trip plus fetch worklod time + fetch service time. 

>The control plane will return the service+workloads associated with that address. Its technically 1 type (Address), and definitely 1 request.

I do not understand about this, message proto of one of type `Workload` and `Service`, how could we return both with one request

hzxuzhonghu: >In standard XDS we load all Service and workloads at startup, just like Envoy

Where does ztunnel do standard xds request for full Address? If we do this, isn't ztunnel store all the services workloads of the mesh, how can it be resource saving

howardjohn: > I do not understand about this, message proto of one of type Workload and Service, how could we return both with one request

Ztunnel does not request Workload and Service, it requests `Address` which is a oneof of these types

> Where does ztunnel do standard xds request for full Address? If we do this, isn't ztunnel store all the services workloads of the mesh, how can it be resource saving

Ztunnel on-demand is optional and not enabled by default. It can scale to 100k pod cluster with 50mb RAM due to more efficient protocol/implementation compared to Envoy, _without_ on-demand. On-demand is only needed when we want to hyper-scale beyond that.

With on-demand, service+workload is already fetched on request. This PR doesn't change it really. Note there are some control plane optimizations as well -- if we request a Service we will return all workloads associated with it, we always return node-local workloads, etc.

PR #583, opened 2023-07-03
Remove unused vip field

nmittler:
This field was removed by https://github.com/istio/istio/pull/45764 and https://github.com/istio/ztunnel/pull/572.

The control plane is no longer sending it and the field has been removed from the proto.


hzxuzhonghu (approved):
lgtm

Issue #582, opened 2023-07-03
store all `IpAddr` as Ipv6 in ztunnel


kdorosh:
cc @nmittler @howardjohn for thoughts on the proposed solution

## Problem

We can get lookup errors in ztunnel if ipv4 addresses are stored in an ipv6 address space. For example, as part of testing https://github.com/istio/istio/pull/45621#discussion_r1245546181 I noticed that the lookups were failing in ztunnel [here](https://github.com/istio/ztunnel/blob/bbe8b3a7f55470fd5804d1f28c6177f29cadc02f/src/proxy/dns.rs#L126-L127) because the `client_addr` in trust_dns had the ipv4 client address stored as an ipv6 space.

For example, a lookup would be made for `::ffff:10.244.2.32` and this would fail to match even though `10.244.2.32` was in the workload store.

One idea I'm having is to use https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#method.to_ipv4_mapped to put everything in an ipv6 space in ztunnel. Thoughts?

kdorosh:
alternatively i may want to spend more time digging into the trust_dns library to understand why my ipv4 requests are being presented as ipv6-formatted `IpAddr` to the workload store

nmittler:
I'm not sure I see how this is related to Trust-DNS. For a given request, it just returns the appropriate records based on the requested record type (v4/v6).

I suspect the issue is somewhere in the ambient logic (either ztunnel, istiod, or both). IIUC, the client IP of the DNS request is IPv6, but the ztunnel [WorkloadStore](https://github.com/istio/ztunnel/blob/bbe8b3a7f55470fd5804d1f28c6177f29cadc02f/src/state/workload.rs#L487) doesn't have a mapping for that IP? Is my understanding correct?

bleggett:
> I suspect the issue is somewhere in the ambient logic (either ztunnel, istiod, or both). IIUC, the client IP of the DNS request is IPv6, but the ztunnel [WorkloadStore](https://github.com/istio/ztunnel/blob/bbe8b3a7f55470fd5804d1f28c6177f29cadc02f/src/state/workload.rs#L487) doesn't have a mapping for that IP? Is my understanding correct?

Yeah - it's in the WE as IPv4, and it enters lookup as an IPv6-wrapped IPv4 address, so lookup fails.

[...]

PR #585, opened 2023-07-05
DNS proxy should handle IPv4-in-IPv6 like TCP proxy does.

bleggett:
Fixes: https://github.com/istio/ztunnel/issues/582

Notes:

- Duplicating [`to_canonical`](https://github.com/istio/ztunnel/blob/bbe8b3a7f55470fd5804d1f28c6177f29cadc02f/src/socket.rs#L59) within the DNS proxy crate seemed simpler - it's trivial and as noted should be ultimately replaced with `stdlib` in both places and removed from the code eventually anyway. But open to better ideas that aren't `add a 'utils' crate`, which IMO isn't necessary.

- The implicit assumption here is that if client uses an IPv4-in-6 address, they want A records back, and not AAAA records. While the below causes the DNS proxy to behave like the TCP proxy already does, it's not 100% clear to me that this is the correct behavior for a DNS server. There's an argument to be made that a 4-in-6 client address means you should return both A and AAAA records.

### DIFF
--- a/src/proxy/dns.rs
+++ b/src/proxy/dns.rs
@@ -30,7 +30,7 @@ use once_cell::sync::Lazy;
 use rand::seq::SliceRandom;
 use rand::thread_rng;
 use std::collections::HashSet;
-use std::net::{IpAddr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 use std::ops::Deref;
 use std::str::FromStr;
 use std::sync::{Arc, RwLock};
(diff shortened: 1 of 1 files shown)


src/proxy/dns.rs
```diff
@@ -908,6 +916,38 @@ mod tests {
...
+            forwarder,
+        };
+
+        let ip4n6_client_ip = ip("::ffff:202:202");
+        let req = req(n("www.google.com"), ip4n6_client_ip, RecordType::A);
+        match store.lookup(&req).await {
+            Ok(_) => {}
+            Err(e) => {
+                if let Some(resp_code) = e.as_response_code() {
+                    assert_ne!(ResponseCode::ServFail, *resp_code);
```
kdorosh, line 944: perhaps we want a stronger assertion here on a valid response code

kdorosh (commented):
looks reasonable to me

src/proxy/dns.rs
bleggett: > perhaps we want a stronger assertion here on a valid response code

Yeah I was trying to minimally mock but I agree it can be better.

nmittler (approved):

kdorosh:
@stevenctl can you take a look at this?

kdorosh (approved):

stevenctl (approved):

PR #589, opened 2023-07-07
DNS support for headless services

nmittler:
Additionally, this adds support for DNS resolution for workload hostnames.


nmittler:
This partially resolves https://github.com/istio/istio/issues/45895, however istiod has to also be updated to properly set pod hostname.

howardjohn (approved):

PR #592, opened 2023-07-09
Make all components optional

### LINKED ISSUE #581: Make all proxy components optional
With a DNS proxy now embedded in ztunnel, it is desirable to be able to just run ztunnel in "dns mode", without any of the other components running.

Fundamentally, the proxy is pretty simple today ... it's basically just a catalog of `components`. Suggest formalizing a basic component API and have them passed into the proxy, to be started/stopped as a single unit.

nmittler:
The DNS proxy is already optional. This PR also makes the HBONE proxy optional, which allows running ztunnel as a stand-alone DNS proxy.

Moved the DNS proxy to a top-level package.

This also cleans up the metrics/stats and readiness servers to make the code more consistent.

Fixes #581


nmittler:
Added a hold on this until https://github.com/istio/ztunnel/pull/597 lands

howardjohn (dismissed):

nmittler:
@howardjohn mind taking another pass? I refactored the startup app flow a bit more and separated the dns metrics out of the global metrics struct.

src/app.rs
```diff
@@ -12,78 +12,133 @@
...
+    } else {
+        None
+    };
+
+    // Create and start the readiness server.
+    let readiness_server = readiness::Server::new(config.clone(), drain_rx.clone(), ready.clone())
+        .await
+        .context("readiness server starts")?;
+    let readiness_address = readiness_server.address();
+    readiness_server.spawn();
```
howardjohn, line 62: I think we need this in the data plane task pool, as discussed on slack

src/metrics.rs
```diff
@@ -83,57 +66,70 @@ impl Metrics {
...
+        if let Some(deferred_fn) = mem::take(&mut self.deferred_fn) {
+            (deferred_fn)(self.param);
         } else {
-            error!("defer record failed, event is gone");
+            error!("defer deferred record failed, event is gone");
         }
     }
 }
 
+pub trait DeferRecorder {
```
howardjohn, line 121: meant for another PR?

src/app.rs
nmittler: My takeaway from our discussion is that this one was on the edge of control/data plane and didn't need to be in the data plane task pool?

src/metrics.rs
nmittler: I had to extract this from `Metrics` because I separated out the DNS metrics. Now both structures implement this.

nmittler:
@howardjohn this is ready to go ... PTAL.

I've refactored things a bit further. Metrics are now completely split out into the pieces that are needed by each component. I've also reworked the `app`'s data plane work pool and use a channel to add tasks to the data plane pool. We now eagarly start the readiness server (in the dataplane pool), so that we can properly serve readiness for all the other components until they're actually ready to serve.

howardjohn (approved):

PR #597, opened 2023-07-11
Adding metrics for DNS.

### LINKED ISSUE #594: Add metrics to DNS proxy
Need to add metrics to the DNS proxy in ztunnel. Similar to what was done in Istio: https://github.com/istio/istio/blob/master/pkg/dns/client/monitoring.go.

nmittler:
Fixes #594.

### DIFF
--- a/src/metrics/dns.rs
+++ b/src/metrics/dns.rs
@@ -1,0 +1,232 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::metrics::{DefaultedUnknown, Recorder};
+use crate::state::workload::address::Address;
+use crate::state::workload::Workload;
+use prometheus_client::encoding::EncodeLabelSet;
+use prometheus_client::metrics::counter::Counter;
+use prometheus_client::metrics::family::Family;
+use prometheus_client::metrics::histogram::Histogram;
+use prometheus_client::registry::Registry;
+use std::time::Duration;
+use trust_dns_server::server::Request;
+
+pub struct Metrics {
+    pub requests: Family<DnsLabels, Counter>,
+    pub forwarded_requests: Family<DnsLabels, Counter>,
+    pub forwarded_failures: Family<DnsLabels, Counter>,
+    pub forwarded_duration: Family<DnsLabels, Histogram>,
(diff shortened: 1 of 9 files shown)


src/proxy/dns.rs
```diff
@@ -385,30 +392,86 @@ impl DnsStore {
...
         // Find the service for the requested host.
         let requested_name = Name::from(request.query().name().clone());
         let Some(service_match) = self.find_server(&client, &requested_name) else {
             // Unknown host. Forward to the upstream resolver.
-            return self.forwarder.forward(&client, request).await;
+            return self.forward(&client, request).await;
         };
 
+        // Increment counter for all requests.
+        self.metrics.increment(&DnsRequest {
```
ymesika, line 469: Looks like it's being incremented with similar labels either in `forward` or here.
Wondering if it should be placed once at line 455?

nmittler: Ah good catch. This call site should have the destination set. Updated.

nmittler:
@ymesika 

> It seems like the tests were updated to include the metrics registry but not checking whether eventually the expected metrics were added.

I'm actually not sure where metrics are tested today. Got any links handy?

ymesika:
> @ymesika
> 
> > It seems like the tests were updated to include the metrics registry but not checking whether eventually the expected metrics were added.
> 
> I'm actually not sure where metrics are tested today. Got any links handy?

Here for example: https://github.com/istio/ztunnel/blob/master/tests/direct.rs#L274

src/metrics/dns.rs
```diff
@@ -0,0 +1,278 @@
...
+            requests,
+            forwarded_requests,
+            forwarded_failures,
+            forwarded_duration,
+        }
+    }
+}
+
+#[derive(Clone, Hash, Debug, PartialEq, Eq, EncodeLabelSet)]
+pub struct DnsLabels {
```
howardjohn, line 76: This seems like a LOT of labels. I am not sure this much is appropriate for Prometheus. Source IP is unbounded cardinality for example

howardjohn: I am a bit torn. I see the value but also the cost. I will think about it a bit more (and welcome other opinions here)

src/metrics/dns.rs
```diff
@@ -0,0 +1,278 @@
...
+    request_protocol: String,
+
+    // Source workload.
+    source_workload: DefaultedUnknown<String>,
+    source_canonical_service: DefaultedUnknown<String>,
+    source_canonical_revision: DefaultedUnknown<String>,
+    source_workload_namespace: DefaultedUnknown<String>,
+    source_principal: DefaultedUnknown<Identity>,
+    source_app: DefaultedUnknown<String>,
+    source_version: DefaultedUnknown<String>,
```
howardjohn, line 89: Largely these are legacy I think, in favor of source_canonical_service/source_canonical_revision. I wonder if we drop them for this new metric

howardjohn (commented):
I only had a chance to skim, looks reasonable. I will actually review later today

nmittler:
@howardjohn I've removed a bunch of the metrics labels. PTAL

howardjohn (approved):

PR #599, opened 2023-07-11
cname mapping for dns

kdorosh:
remediation for https://github.com/istio/istio/pull/45900#discussion_r1260089767 and optimizes DNS query resolution the same way as the istio DNS sidecar

the true fix still needs to be worked on as per https://github.com/istio/ztunnel/issues/600. but this optimization is still nice to have

### DIFF
--- a/src/proxy/dns.rs
+++ b/src/proxy/dns.rs
@@ -414,7 +414,7 @@ impl Resolver for DnsStore {
             // search process down to just two DNS queries. This will eliminate unnecessary upstream DNS queries from the
             // agent, reduce load on DNS servers and improve overall latency. This idea was borrowed and adapted from
             // the autopath plugin in coredns. This idea is also borrowed from the sidecar istio-dns solution:
-            // https://github.com/istio/istio/blob/master/pkg/dns/client/dns.go#L580-L594
+            // https://github.com/istio/istio/blob/f052a3ff61c95c427f626fe4b952751fd900e803/pkg/dns/client/dns.go#L580-L594
             let namespaced_domain = append_name(as_name(&client.namespace), &self.svc_domain);
             if str::ends_with(requested_name.to_string().as_str(), namespaced_domain.to_string().as_str()) {
                 let mut records = Vec::new();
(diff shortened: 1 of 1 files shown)


kdorosh:
/test test

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: failed to bind to address 127.0.0.1:39290: Address already in use (os error 98)', benches/throughput.rs:128:14
```

bleggett (approved):
LGTM since it replicates logic from existing istio-dns.

Assuming it's difficult to unit test this given that the result will look the same.

src/proxy/dns.rs
```diff
@@ -405,6 +405,23 @@ impl Resolver for DnsStore {
...
+            // variant of the host+ the first search domain in resolv.conf as the first query
+            // is likely to be host.ns.svc.cluster.local (e.g., www.google.com.ns1.svc.cluster.local) due to
+            // the list of search namespaces in resolv.conf (unless the app explicitly does www.google.com. which is unlikely).
+            // We will resolve www.google.com.ns1.svc.cluster.local with a CNAME record pointing to www.google.com.
+            // which will cause the client's resolver to automatically resolve www.google.com. , and short circuit the lengthy
+            // search process down to just two DNS queries. This will eliminate unnecessary upstream DNS queries from the
+            // agent, reduce load on DNS servers and improve overall latency. This idea was borrowed and adapted from
+            // the autopath plugin in coredns. This idea is also borrowed from the sidecar istio-dns solution:
+            // https://github.com/istio/istio/blob/f052a3ff61c95c427f626fe4b952751fd900e803/pkg/dns/client/dns.go#L580-L594
+            let namespaced_domain = append_name(as_name(&client.namespace), &self.svc_domain);
```
nmittler, line 418: I'm not sure this is correct. The problem Istio is trying to solve is for entries that are actually contained within our cache. This is for cache misses, where we forward to the upstream resolver.

For the forwarding case, I would expect we should pass exactly the hostname passed to us.

kdorosh:
closing this PR per https://github.com/istio/ztunnel/pull/599#discussion_r1261418869 and in favor of https://github.com/istio/ztunnel/pull/603

PR #601, opened 2023-07-12
add tests for service subjectAltNames

GregHanson:
some followup tests for https://github.com/istio/ztunnel/pull/556

### DIFF
--- a/examples/localhost.yaml
+++ b/examples/localhost.yaml
@@ -44,6 +44,8 @@ services:
     - /127.10.0.1
   ports:
     80: 8080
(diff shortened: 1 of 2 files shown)


hzxuzhonghu (approved):

kdorosh (commented):
I'd slightly prefer a direct.rs test that makes requests to non standard identities (one that fails and then one that we configure to explicitly allow) than this, which isn't really a integ test. But this is better than nothing for sure

howardjohn (approved):

PR #603, opened 2023-07-12
Fix for DNS system forwarder

### LINKED ISSUE #600: DNS forwarder not propagating nxdomain properly
i think https://github.com/istio/ztunnel/blob/6e04f923b4b00b3a14b31ee10df9da0d6b9ffea9/src/proxy/dns/forwarder.rs#L40

per https://github.com/istio/ztunnel/blob/6e04f923b4b00b3a14b31ee10df9da0d6b9ffea9/src/proxy/dns.rs#L472-L473

should be returning NXDOMAIN from coredns on invalid queries (e.g. someservice.serviceentry.ns.svc.cluster.local), rather than NOERROR with no results. coredns returns NXDOMAIN when queried directly, so if our proxy does not propagate that then this will be problematic down the road

some context with the integration test to easily repro https://github.com/istio/istio/pull/45900#discussion_r1260089767

nmittler:
Currently, if a DNS request is forwarded to the upstream resolver and fails with NXDomain, the client will see a successful response with no records.

This is due to the fact that the DNS system forwarder returns an error of type ResolveError, which was not being handled properly by the Handler.

This fixes the logic and adds tests for the system forwarder.

Fixes #600

### DIFF
--- a/src/proxy/dns.rs
+++ b/src/proxy/dns.rs
@@ -42,6 +42,7 @@ use tokio::net::{TcpListener, UdpSocket};
 use trust_dns_proto::error::ProtoErrorKind;
 use trust_dns_proto::op::ResponseCode;
 use trust_dns_proto::rr::{Name, RData, Record, RecordType};
+use trust_dns_resolver::config::ResolverConfig;
 use trust_dns_resolver::system_conf::read_system_conf;
 use trust_dns_server::authority::LookupError;
 use trust_dns_server::server::Request;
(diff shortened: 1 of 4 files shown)


kdorosh:
resolves https://github.com/istio/ztunnel/issues/600 ?

kdorosh (approved):
changes lgtm, also confirmed it fixes the istio integration test error https://github.com/istio/istio/pull/45900#discussion_r1261402070

bleggett (approved):
LGTM as a more robust fix than https://github.com/istio/ztunnel/pull/599

howardjohn (approved):

Issue #608, opened 2023-07-13
support on-demand hostname for workloads


kdorosh:
edit: the idea for the implementation here is inspired by https://github.com/envoyproxy/envoy/issues/20562

---

old, slightly confusing original context:

~it's not completely implemented, but what we do have is wrong.~ what we have is per-pod hostnames for statefulsets: https://github.com/istio/ztunnel/blob/9583efce62c8d9d4c29c987a6479d13803ea8149/src/dns/server.rs#L324-L331

we still need workload entry hostname support, which functions a bit differently.

the original intent behind the workload `hostname` field was to have a place to represent the [`WorkloadEntry.address`](https://istio.io/latest/docs/reference/config/networking/workload-entry/#WorkloadEntry) field, for example:

```yaml
apiVersion: networking.istio.io/v1alpha3
kind: WorkloadEntry
metadata:
  name: details-svc
spec:
  # use of the service account indicates that the workload has a
  # sidecar proxy bootstrapped with this service account. Pods with
  # sidecars will automatically communicate with the workload using
  # istio mutual TLS.
  serviceAccount: details-legacy
  address: vm1.vpc01.corp.net # this would be the hostname on the workload we resolve via async polling
  labels:
    app: details-legacy
    instance-id: vm1
```

The new proposal after headless services support in ztunnel is to add a new `async_hostname` field for the WE hostname behavior.

The address field in the WorkloadEntry above translates to an [envoy DNS cluster](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/service_discovery), and the workload hostname field in ztunnel was originally intended for this. Note that this means the DNS "cluster" in ztunnel here should use the system forwarder (any address field that's only in istio's DNS cache should return nxdomain to avoid potential looping DNS queries).

it's important we have some way for ztunnel to do DNS resolution as a client for workloads for geographic DNS resolution, e.g. https://github.com/istio/istio/blob/7bf52db38c91c89a4d56d6a94f402fd9ddaf6465/pilot/pkg/serviceregistry/kube/conversion.go#L151-L155

also note that the ztunnel DNS "cluster" implementation should likely mirror envoy's implementation as closely as possible to keep semantics the same for users. this also means we don't just want this lookup available for DNS proxy, but also need this to work during the request path (e.g. a service entry `se.istio.io` resolves to a workload with hostname `aws.elb.com` which resolves to final dest) by pre-resolving the provided hostnames asynchronously

cc @nmittler

[...]

PR #612, opened 2023-07-18
Implement on-demand hostname for Workloads

kdorosh:
### Summary

implementation here is inspired by https://github.com/envoyproxy/envoy/issues/20562

in short, implements on-demand DNS resolution when a workload doesn't have any ips but has a hostname. If a request goes to a workload with only hostnames, we do a blocking on-demand DNS lookup and use the result. The result is cached for any future requests until the shortest DNS record ttl is expired, then we invalidate all records and refresh DNS (to ensure load balancing does not favor records with longer ttls at the end of lifetime).

### Future Work

https://github.com/istio/ztunnel/pull/612/commits/326b887217c1904b95f04847933a03fd66392c34 details a lot of code I wrote but decided to omit as a premature optimization. This code was intended to allow smart early refresh for on-demand DNS iff we received more than one request to a hostname during the DNS ttl. This was built with the assumption that earlier request patterns would continue, so we could limit blocking DNS queries if there was a consistent load to only the very first request.

That said, this implementation only blocks on the first request per DNS ttl and is much simpler, and probably fine for most cases. We can add more complex refresh behavior later if there is interest.

cc @nmittler for likely having an interest in this PR

resolves https://github.com/istio/ztunnel/issues/608


proto/workload.proto
```diff
@@ -103,9 +101,20 @@ message Workload {
...
+  // The hostname for the workload to be resolved asynchronously by the ztunnel.
+  //
+  // If the resolved DNS query has several endpoints, any subsequent requests will
+  // be load balanced between these endpoints (i.e. envoy STRICT_DNS cluster)
+  // TODO: add support for another load-balancing mode where each connection sends
+  // to the first endpoint resolved by its own DNS query for the lifetime of the
+  // connection in the connection pool (i.e. envoy LOGICAL_DNS cluster)
+  //
+  // Each workload must have either an address or async_hostname; not both.
+  string async_hostname = 23;
```
nmittler, line 116: I definitely do not think we should go forward with this. What is the purpose of this again? It's not clear how this is different than `hostname`. I know fundamentally one is supposed to come from WLE and one from pod, but nothing about this naming makes that clear.

Could we instead make `hostname` and array?

@howardjohn wdyt?

howardjohn: Am I oversimplifying this or do we just need to do:

```
if workload.ip != null {
 use ip 
} else if workload.hostname != null {
 use hostname
}
```

?
Not sure why we need a new field for that.

A headless pod will have a hostname + IP, so we don't need to use the hostname on the request path (i mean, we could --it would just resolve to the IP though so its moot)

kdorosh: I think that is over-simplifying. If we want to keep service entry + workload entry semantics the same as in classic istio then we need async DNS resolution for SE/WE hostnames. Today's behavior:

https://github.com/istio/istio/blob/30b23bab0017d3be258d79230ec05871afcad144/pilot/pkg/serviceregistry/serviceentry/conversion.go#L330 for service entry resolution

https://github.com/istio/istio/blob/30b23bab0017d3be258d79230ec05871afcad144/pilot/pkg/networking/core/v1alpha3/cluster.go#L671-L674 for how that converts to envoy clusters

These envoy clusters from the envoy docs https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/upstream/service_discovery#arch-overview-service-discovery-types-strict-dns

> When using strict DNS service discovery, Envoy will continuously and asynchronously resolve the specified DNS targets. Each returned IP address in the DNS result will be considered an explicit host in the upstream cluster. This means that if the query returns three IP addresses, Envoy will assume the cluster has three hosts, and all three should be load balanced to. If a host is removed from the result Envoy assumes it no longer exists and will drain traffic from any existing connection pools. Consequently, if a successful DNS resolution returns 0 hosts, Envoy will assume that the cluster does not have any hosts. ***Note that Envoy never synchronously resolves DNS in the forwarding path. At the expense of eventual consistency, there is never a worry of blocking on a long running DNS query.***

SE has the resolution field (not currently supported) and WE addresses can be hostnames. Both need to be looked at more closely, I'll look more tomorrow. Parity with current features is intent of this PR though

howardjohn:
I don't think Envoy has the right behavior here, it is why we get constant
complaints about DOSing coredns. Should only query when we actually send a
request ("on demand")

keithmattix:
+1 to @howardjohn  - it would be great to just skip to the on-demand DNS feature we've been asking Envoy for for years

bleggett:
> +1 to @howardjohn - it would be great to just skip to the on-demand DNS feature we've been asking Envoy for for years

Are there extant issues/tickets with context out there for this request?



> I don't think Envoy has the right behavior here, it is why we get constant complaints about DOSing coredns. Should only query when we actually send a request ("on demand")

Are those. complaints just cases of misconfigured [respect_dns_ttl](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-respect-dns-ttl) and  [dns_refresh_rate](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/cluster/v3/cluster.proto#envoy-v3-api-field-config-cluster-v3-cluster-dns-refresh-rate) properties?

howardjohn:
No, 30s ttl is common. Query every 30s * every SE * every proxy == DDOS.

https://github.com/envoyproxy/envoy/issues/20562

kdorosh:
Makes sense to me, thanks for the additional context. I'll update the implementation to the best-of-both-worlds approach. I think I should be able to reuse the hostname field that already exists as well. :eyes: on impl in coming days

src/dns/server.rs
```diff
@@ -656,14 +664,33 @@ struct SystemForwarder {
 }
 
 impl SystemForwarder {
-    fn new() -> Result<Self, Error> {
+    fn new(custom_name_servers: Vec<SocketAddr>) -> Result<Self, Error> {
```
howardjohn, line 667: Do the changes here mean we hit our own internal DNS system for lookups? If so, interesting... we don't in Envoy. Part of me likes it (in memory lookup is obviously cheaper than a real one...). But worried we may end up with loops or something.

Will revist this comment as I read more

src/proxy/dns.rs
```diff
@@ -0,0 +1,271 @@
...
+// This can be important in niche cases, like geographic dns resolution, as documented in
+// https://github.com/istio/istio/blob/7bf52db38c91c89a4d56d6a94f402fd9ddaf6465/pilot/pkg/serviceregistry/kube/conversion.go#L151-L155
+pub(super) struct PollingDns {
+    pi: ProxyInputs,
+    drain: Watch,
+    // workload UID to task
+    tasks: HashMap<String, TaskContext>,
+}
+
+// TODO(kdorosh) DRY the copied code following here and generally clean up any latest placeholder test values
```
howardjohn, line 59: Is set_id a test value?

src/state.rs
```diff
@@ -72,6 +73,46 @@ pub struct ProxyState {
...
+        };
+        Ok(**ip)
+    }
+
+    // TODO: ipv6 support
+    pub fn set_ips_for_workload(&mut self, workload_uid: String, ips: Vec<Ipv4Addr>) {
+        let set = HashSet::from_iter(ips.iter().map(|x| IpAddr::V4(*x)));
+        self.by_workload_uid.insert(workload_uid.to_owned(), set);
+        trace!(
+            "set_ips_for_workload({}), by_workload_uid len: {}",
```
howardjohn, line 111: nit: Might be more appropriate to use things like https://docs.rs/tracing/latest/tracing/attr.instrument.html for this.

Basically we are abusing a string to trace a function call + add k/v attributes, both of which the library can handle natively/better.

NBD especially for trace, more a comment for awareness

src/state.rs
```diff
@@ -174,6 +215,43 @@ impl ProxyState {
...
+        // TODO: add more sophisticated routing logic, perhaps based on ipv4/ipv6 support underneath us.
+        // if/when we support that, this function may need to move to get access to the necessary metadata.
+        // Randomly pick an IP
+        // TODO: do this more efficiently, and not just randomly
+        let Some(ip) = workload.workload_ips.choose(&mut rand::thread_rng()) else {
+
+            if !workload.hostname.is_empty() {
+                let ip = self
+                    .resolved_dns
+                    .load_balance_for_workload(workload.clone().uid)?;
```
howardjohn, line 246: Can you remind me our current status on "dns Services"? is it a TODO, pending, we killed it? I forgot entirely :-)

I absolutely don't mind starting with one, just want to make sure I understand the end state

src/proxy/inbound.rs
```diff
@@ -402,8 +402,12 @@ impl Inbound {
             let from_gateway = match state.fetch_destination(&gateway_address.destination).await {
                 Some(address::Address::Workload(wl)) => Some(wl.identity()) == conn.src_identity,
                 Some(address::Address::Service(svc)) => {
-                    for (ip, _ep) in svc.endpoints.iter() {
-                        if state.fetch_workload(ip).await.map(|w| w.identity()) == conn.src_identity
+                    for (_ep_uid, ep) in svc.endpoints.iter() {
+                        if state
+                            .fetch_workload_by_uid(&ep.workload_uid)
```
howardjohn, line 407: Interesting, so we fetch by UID because some don't have IP. Makes sense - maybe could use a comment inthe code, it wasn't obvious to me at first why.

Also I think this will never actually require a lookup _in current Istiod implementation_, as we push all Workloads when we push a Service. Good to have like this, though (and we had it before your PR as well, of course)

src/proxy/outbound.rs
```diff
@@ -410,6 +416,14 @@ impl OutboundConnection {
...
             return Err(Error::NoGatewayAddress(Box::new(us.workload.clone())));
         }
+
+        let workload_ip = self
+            .pi
+            .state
+            .state
+            .read()
+            .unwrap()
+            .load_balance(&us.workload)?;
```
howardjohn, line 426: This usage is not great - may be nicer to have load_balance on DemandProxyState?

I think we could probably get to a point where `state` is not `pub` in `DemandProxyState`

src/state.rs
```diff
@@ -72,6 +73,46 @@ pub struct ProxyState {
...
+            "load_balance_for_workload({}) by_workload_uid len: {}",
+            &workload_uid,
+            self.by_workload_uid.len()
+        );
+        // TODO: add support for different async dns modes: e.g. logical DNS vs strict DNS.
+        // this is effectively envoy strict DNS; returning a pre-resolved IP address at random.
+        let Some(ipset) = self.by_workload_uid.get(&workload_uid) else {
+            return Err(Error::NoResolvedAddresses(workload_uid));
+        };
+        let ips_vec = ipset.iter().collect::<Vec<_>>();
```
howardjohn, line 99: I think you can choose() without collecting

src/proxy/dns.rs
```diff
@@ -0,0 +1,271 @@
...
+                    .read()
+                    .unwrap()
+                    .workloads
+                    .get_async_dns_workloads();
+
+                // kill tasks that no longer need to be running
+                let current_workload_uids =
+                    dns_workloads.iter().map(|w| w.uid.clone()).collect_vec();
+                let current_workload_uid_set: HashSet<String> =
+                    HashSet::from_iter(current_workload_uids.iter().cloned());
```
howardjohn, line 177: no need to make collect into current_workload_uids if we are then just going to re-collect it into a Set. We can go straight from iter -> Set

src/proxy/dns.rs
```diff
@@ -0,0 +1,271 @@
...
+        let accept = async move {
+            loop {
+                let dns_workloads = self
+                    .pi
+                    .state
+                    .state
+                    .read()
+                    .unwrap()
+                    .workloads
+                    .get_async_dns_workloads();
```
howardjohn, line 171: this seems like we are going to resolve for all workloads? Is that right? I thought we were only resolving for workloads that have at least one request recently?

howardjohn (commented):
I only had a chance to skim it,  looks like the right direction. mostly reviewed the non-dns specific code

src/dns/server.rs
kdorosh: This PR is still very WIP. This was for configuring a non system default dns name server so the nip.io style hosts resolve to known IPs in the direct test

kdorosh:
Yeah I still haven't had a chance to add the on demand thing so right now it still polls everything, but I'm working on that next and then cleaning up

src/proxy/outbound.rs
kdorosh: yeah I think it will have to move, since `load_balance()` will need to be async for on-demand DNS queries. and the rust async stuff wasn't loving async functions on the proxy state (understandably.. we don't want to lock our store and then await a DNS query so no one else can read in the meantime).

anyways this will require a bit of refactoring to get to compile and be reasonable, and that's my current focus

src/proxy/dns.rs
kdorosh: yeah that's not right, this is still WIP. im currently reworking the async await part for on-demand and then updating the dns polling to be correct and only poll once per DNS ttl iff we received a request in last DNS ttl lifetime

src/state.rs
kdorosh: im not sure exactly what you mean? what is a "dns service"?

src/proxy/dns.rs
kdorosh: yes; this code was copied from test code and still needs to be make production-ready and live in the right place. first im getting the on-demand skeleton flushed out, ci passing, and then cleaning up

kdorosh:
/test

```
thread 'test_tcp_request' panicked at 'called `Result::unwrap()` on an `Err` value: failed to bind to address 127.0.0.1:45366: Address already in use (os error 98)', /home/prow/go/src/istio.io/ztunnel/src/test_helpers/app.rs:73:10
```

kdorosh:
/test test

```
thread 'test_tcp_request' panicked at 'called `Result::unwrap()` on an `Err` value: failed to bind to address 127.0.0.1:45366: Address already in use (os error 98)', /home/prow/go/src/istio.io/ztunnel/src/test_helpers/app.rs:73:10
```

src/config.rs
```diff
@@ -161,6 +161,9 @@ pub struct Config {
 
     // CLI args passed to ztunnel at runtime
     pub proxy_args: String,
+
+    // custom DNS nameservers; primarily for testing
+    pub dns_nameservers: Vec<SocketAddr>,
```
nmittler, line 166: Should this be an option? If set, even if it's empty, I think the expectation would be that it overrides the system settings.

src/dns/server.rs
```diff
@@ -656,14 +664,33 @@ struct SystemForwarder {
...
         // Get the resolver config from /etc/resolv.conf.
         let (cfg, opts) = read_system_conf()?;
 
         // Extract the parts.
         let domain = cfg.domain().cloned();
         let search_domains = cfg.search().to_vec();
-        let name_servers = cfg.name_servers().to_vec();
+        let mut name_servers = cfg.name_servers().to_vec();
+
+        for s in &custom_name_servers {
```
nmittler, line 676: Let's not merge. Instead, make the nameservers list an Option and if set, just overwrite the configuration we pass into the resolver.

src/dns/server.rs
```diff
@@ -633,16 +636,21 @@ pub trait Forwarder: Send + Sync {
         client: Option<&Workload>,
         request: &Request,
     ) -> Result<Answer, LookupError>;
+
+    fn resolver(&self) -> Arc<dyn Resolver>;
```
nmittler, line 640: We shouldn't expose this. The point of this interface is to hide those details.

nmittler, line 640: We shouldn't expose this. This interface is meant to hide those details.

kdorosh:
no need to take a look yet, this PR is still very WIP @nmittler @howardjohn

nmittler:
@kdorosh just backing up for a second...

IIUC the problem we're trying to solve is that the request is made to an IP, but internally we need to map that IP to a hostname, as defined in WE ... correct? If so, then this is not really related to the ztunnel DNS code per se ... it's really a function of the proxy. The proxy will just need to map the IP to a hostname, look it up, and then proceed with the request.

Might this be something that hyper could do for us? I don't think the logic will need to reach back to our DNS code, just use the local resolver.

... or am I missing something?

kdorosh:
@nmittler feel free to ping me again on slack when you're out of your meeting, we can chat in slack

> The proxy will just need to map the IP to a hostname, look it up, and then proceed with the request.

that's what we do, we look up the request (on-demand on the datapath using the system resolver) and then cache the response for the DNS ttl for any future requests per 1. in https://github.com/envoyproxy/envoy/issues/20562

perhaps there's some more code reuse I can leverage from stuff you wrote earlier? I'm definitely open to that. I wanted to write better tests against what I have now (lots of copy paste) and then refactor and DRY

nmittler:
@kdorosh just to clarify, I was thinking there might be a way to leverage hyper services for this (e.g. https://tikv.github.io/doc/hyper/client/connect/dns/index.html).

kdorosh:
fantastic, thank you. I will look into those to replace the test DNS client code I copied from some other parts of the codebase (for now). still focused on writing more complete and better tests before i spend too much time cleaning up the current state (which, for the most part, works)

nmittler:
SG .. also, to give credit where it's due .... @howardjohn suggested hyper-trust-dns in an offline chat :)

howardjohn:
I think hyper can plugin any function of Name -> Addr (or similar) btw,
doesn't have to be that one

kdorosh:
/test test

os port binding flake again

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: failed to bind to address 127.0.0.1:60161: Address already in use (os error 98)', benches/throughput.rs:125:14
```

src/dns/server.rs
kdorosh: addressed, this code was removed

src/proxy/dns.rs
kdorosh: addressed

src/state.rs
kdorosh: addressed

src/state.rs
kdorosh: bump on this

src/state.rs
kdorosh: simplified and removed this code

src/proxy/dns.rs
kdorosh: addressed, this code was removed

src/dns/server.rs
kdorosh: this code was removed

kdorosh:
ok now its ready for review @howardjohn @nmittler ; also updated the PR description and responded to your comments. thanks

proto/workload.proto
kdorosh: comment here added the context i was missing https://github.com/istio/ztunnel/pull/612#issuecomment-1642303664. opted to implement on demand rather than copy envoy's impl

src/state.rs
```diff
@@ -259,6 +268,227 @@ impl DemandProxyState {
...
+            let resolver_result =
+                TokioAsyncResolver::new(dns_resolver_cfg, dns_resolver_opts, TokioHandle);
+            if resolver_result.is_err() {
+                warn!(
+                    "system dns async resolution: error creating resolver for workload {} is: {:?}",
+                    &workload_uid, resolver_result
+                );
+                return;
+            }
+            let r = resolver_result.unwrap();
```
ilrudie, line 395: **question:**  Do you know what the cost for creating a new resolver per call is? Assuming the resolver could be reused is it worth considering initializing the resolver when DemandProxyState is initialized and storing the resolver for use when resolve_on_demand_dns is called to avoid that cost during the resolution?

kdorosh: i have not benchmarked it. happy to explore optimizations here if others have recommendations but the tokio resolver needs to be in the same runtime as it's currently executing in (since each call to `resolve_on_demand_dns` invokes a `tokio::spawn` we need to be careful we aren't trying to create new runtimes within a runtime, or sharing memory unsafely across runtime environments)

kdorosh:
@stevenctl would be a good first reviewer before john or nathan can get to it (if they desire / have the bandwidth)

src/dns/server.rs
```diff
@@ -365,7 +365,10 @@ impl Store {
                     service
                         .endpoints
                         .iter()
-                        .filter_map(|(addr, _)| {
+                        .filter_map(|(_, ep)| {
+                            let Some(addr) = ep.address.clone() else {
```
stevenctl, line 369: super-nit: 

```suggestion
                            let Some(addr) = &ep.address else {
```

src/state.rs
ilrudie: Probably worth getting some idea how expensive it is before we spend any time optimizing. No need to delay the PR over this.

src/dns/server.rs
kdorosh: addressed

src/state.rs
```diff
@@ -259,6 +267,227 @@ impl DemandProxyState {
...
+            .by_hostname
+            .get(hostname)
+            .filter(|rdns| {
+                rdns.initial_query.is_some()
+                    && rdns.initial_query.unwrap().elapsed() < rdns.dns_refresh_rate
+            })
+            .cloned()
+    }
+
+    pub async fn set_gateway_address(
```
stevenctl, line 470: Why did this become a method if there is no reference to `self`?

stevenctl: Seems like maybe it would be a method of `Upstream`.

src/state.rs
```diff
@@ -299,12 +535,21 @@ impl DemandProxyState {
...
-            Some(upstream) => {
+            Some(mut upstream) => {
                 debug!(%wl.name, "found waypoint upstream");
-                Ok(Some(upstream))
+                match self
+                    .set_gateway_address(&mut upstream, workload_ip, gw_address.port)
+                    .await
+                {
+                    Ok(_) => Ok(Some(upstream)),
+                    Err(e) => {
```
stevenctl, line 548: Seems like set_gateway_address has no error paths. Could change the signature and remove branch.

kdorosh: `.waypoint_svc_ip_address()?` can return an error in

```rust
pub async fn set_gateway_address(us: &mut Upstream, workload_ip: IpAddr, hbone_port: u16) {
    if us.workload.gateway_address.is_none() {
        us.workload.gateway_address = Some(match us.workload.protocol {
            Protocol::HBONE => {
                let ip = us
                    .workload
                    .waypoint_svc_ip_address()?
                    .unwrap_or(workload_ip);
                SocketAddr::from((ip, hbone_port))
            }
            Protocol::TCP => SocketAddr::from((workload_ip, us.port)),
        });
    }
}
```

src/state.rs
kdorosh: great call, got lost in refactor and cleanup. addressed in https://github.com/istio/ztunnel/pull/612/commits/61189f87f24444cfc266f86f35e7c8e5df29c252

kdorosh: i left it where it originally lived (check the new PR diff) but im happy to move it

src/proxy/outbound.rs
```diff
@@ -381,14 +382,48 @@ impl OutboundConnection {
...
+            .state
+            .load_balance(
+                &mutable_us.workload,
+                &source_workload,
+                &self.pi.cfg.dns_resolver_config,
+                &self.pi.cfg.dns_resolver_opts,
+                self.pi.metrics.clone(),
+            )
+            .await?;
+        let us = match set_gateway_address(&mut mutable_us, workload_ip, self.pi.hbone_port) {
```
stevenctl, line 397: Do we need to call this here, or does the call within fetch_waypoint do enough for us?

stevenctl (commented):
Need to look over the DNS lookup stuff still, lgtm so far

src/proxy/outbound.rs
kdorosh: this originally kept the original behavior (in case of waypoint, we called `set_gateway_address` twice. that said, we only really need to do it once. so i updated it in https://github.com/istio/ztunnel/pull/612/commits/630671d02e0cab2ed676da33becb613f38be4737

howardjohn:
will take a look tomorrow

src/proxy/outbound.rs
```diff
@@ -381,35 +382,72 @@ impl OutboundConnection {
...
-        let us = us.unwrap();
+        let mut mutable_us = us.unwrap();
+        let workload_ip = self
+            .pi
+            .state
+            .load_balance(
+                &mutable_us.workload,
+                &source_workload,
+                &self.pi.cfg.dns_resolver_config,
+                &self.pi.cfg.dns_resolver_opts,
```
hzxuzhonghu, line 393: nit: can we make these inner fields of state rather than passing

src/state.rs
```diff
@@ -259,6 +267,206 @@ impl DemandProxyState {
...
+                src_workload,
+                dns_resolver_config,
+                dns_resolver_opts,
+                metrics,
+            )
+            .await?;
+        Ok(ip)
+    }
+
+    async fn load_balance_for_workload(
```
hzxuzhonghu, line 305: should make it more readable, it is resolving hostname

src/state.rs
```diff
@@ -259,6 +267,206 @@ impl DemandProxyState {
...
+                    .inc();
+                r
+            }
+            None => {
+                metrics
+                    .as_ref()
+                    .on_demand_dns_cache_misses
+                    .get_or_create(&labels)
+                    .inc();
+                let jh = tokio::spawn(Self::resolve_on_demand_dns(
```
hzxuzhonghu, line 334: We can call Self::resolve_on_demand_dns.await

src/state.rs
kdorosh: addressed in https://github.com/istio/ztunnel/pull/612/commits/c8615127acf2d41d00812688db412f1d69d514fc

src/proxy/outbound.rs
kdorosh: addressed in https://github.com/istio/ztunnel/pull/612/commits/c8615127acf2d41d00812688db412f1d69d514fc

kdorosh:
/test test

os port reuse flake

```
thread 'test_quit_lifecycle' panicked at 'called `Result::unwrap()` on an `Err` value: failed to bind to address 127.0.0.1:57248: Address already in use (os error 98)', tests/direct.rs:166:56
```

src/state.rs
```diff
@@ -259,6 +283,188 @@ impl DemandProxyState {
...
+        let Some(ip) = rdns.ips.iter().choose(&mut rand::thread_rng()) else {
+            return Err(Error::EmptyResolvedAddresses(workload_uid));
+        };
+        Ok(*ip)
+    }
+
+    fn resolve_on_demand_dns(mut state: DemandProxyState, workload: &Workload) -> JoinHandle<()> {
+        let workload_uid = workload.uid.to_owned();
+        let hostname = workload.hostname.to_owned();
+        tokio::spawn(async move {
```
howardjohn, line 379: why spawn?

kdorosh: no good reason, was copied from other async code and i missed that during cleanup. made simpler in https://github.com/istio/ztunnel/pull/612/commits/454372bbe8873e4bfaf0a697d21394e41c476350

src/state.rs
```diff
@@ -259,6 +282,175 @@ impl DemandProxyState {
...
+        let Some(ip) = rdns.ips.iter().choose(&mut rand::thread_rng()) else {
+            return Err(Error::EmptyResolvedAddresses(workload_uid));
+        };
+        Ok(*ip)
+    }
+
+    async fn resolve_on_demand_dns(mut state: DemandProxyState, workload: &Workload) {
+        let workload_uid = workload.uid.to_owned();
+        let hostname = workload.hostname.to_owned();
+        async move {
```
ilrudie, line 364: I may be missing something but it seems like we don't need a new async block since the function is async now.

kdorosh: good call, thanks https://github.com/istio/ztunnel/pull/612/commits/54307a9281c4a54a26ca43aaad1d1be3e7c34282

src/proxy/outbound.rs
```diff
@@ -381,35 +382,68 @@ impl OutboundConnection {
...
                     request_type: RequestType::ToServerWaypoint,
-                    upstream_sans: us.sans.clone(),
+                    upstream_sans: mutable_us.sans,
                 });
             }
             // we expected the workload to have a waypoint, but could not find one
             Err(e) => return Err(Error::UnknownWaypoint(e.to_string())),
         }
+
+        let us = match set_gateway_address(&mut mutable_us, workload_ip, self.pi.hbone_port) {
```
stevenctl, line 435: I think all calls to `fetch_upstream` (either above, or inside of `fetch_waypoint` end up calling `load_balance`. 

If I didn't mix myself up, I think the calls to `load_balance` and `set_gateway` could be inside `fetch_upstream`. Maybe awkward though.

src/state.rs
```diff
@@ -259,6 +282,173 @@ impl DemandProxyState {
...
+            hostname: hostname.to_owned(),
+            ips,
+            initial_query: Some(now),
+            dns_refresh_rate,
+        };
+        state.set_ips_for_hostname(hostname, rdns);
+    }
+
+    pub fn set_ips_for_hostname(&mut self, hostname: String, rdns: ResolvedDns) {
+        self.state
```
stevenctl, line 430: Does this mean that several requests to a given hostname occurring before we hit `set_ips_for_hostname` will all resolve in parallel and get the same response? 

Can probably be optimized later. If I have this right, it would good to add a TODO.

stevenctl (approved):

src/state.rs
```diff
@@ -357,34 +560,26 @@ impl DemandProxyState {
     }
 }
 
-fn set_gateway_address(us: &mut Upstream, hbone_port: u16) -> anyhow::Result<()> {
+pub fn set_gateway_address(
```
EItanya, line 563: maybe a nit, but why does this take a `mut Upstream` and not exist on `impl Upstream`?

src/config.rs
```diff
@@ -265,6 +272,9 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
         RootCert::Static(Bytes::from(ca_root_cert_provider))
     };
 
+    use trust_dns_resolver::system_conf::read_system_conf;
+    let (cfg, opts) = read_system_conf().unwrap();
```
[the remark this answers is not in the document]
kdorosh: addressed in https://github.com/istio/ztunnel/pull/622

src/proxy/metrics.rs
```diff
@@ -28,6 +28,10 @@ pub struct Metrics {
     pub connection_close: Family<CommonTrafficLabels, Counter>,
     pub received_bytes: Family<CommonTrafficLabels, Counter>,
     pub sent_bytes: Family<CommonTrafficLabels, Counter>,
+
+    // on-demand DNS is not a part of DNS proxy, but part of ztunnel proxy itself
+    pub on_demand_dns_cache_hits: Family<OnDemandDnsLabels, Counter>,
```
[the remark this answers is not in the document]
kdorosh: addressed in https://github.com/istio/ztunnel/pull/622

PR #615, opened 2023-07-21
Set NODELAY when accepting H2 connections

Stevenjin8:
Longest I've spent on a single line of code.

This change improves latency performance by 10-20x. We should set NODELAY because we are already setting it on one end of the H2 connection. Also, this [Will I need TCP_NODELAY for my HTTP/2 connections?](https://http2.github.io/faq/#will-i-need-tcp_nodelay-for-my-http2-connections).

Not sure if this is the correct place to do this. Feels more of a `hyper` thing than a TLS thing. I want to put it in here

https://github.com/istio/ztunnel/blob/be1e6d8d333b5dd26cf123babd1239eb5366f862/src/hyper_util.rs#L48

But, `conn` is of type `SslStream<TcpStream>` and I can't find a way to get the `TcpStream` out. I'm also happy to add a `set_nodelay` method to `SslStream` upstream.

### DIFF
--- a/src/hyper_util.rs
+++ b/src/hyper_util.rs
@@ -52,6 +52,7 @@ pub fn tls_server<T: CertProvider + Clone + 'static>(
                     None
                 }
                 Ok(s) => {
+                    s.get_ref().set_nodelay(true).unwrap();
                     debug!("TLS handshake succeeded");
                     Some(s)
                 }
--- a/src/hyper_util.rs
+++ b/src/hyper_util.rs
@@ -52,11 +52,13 @@ pub fn tls_server<T: CertProvider + Clone + 'static>(
                     None
                 }
                 Ok(s) => {
-                    s.get_ref().set_nodelay(true).unwrap();
                     debug!("TLS handshake succeeded");
                     Some(s)
                 }
             }
+        }).map(|conn| {
+            conn.get_ref().set_nodelay(true).unwrap();
+            conn
         })
 }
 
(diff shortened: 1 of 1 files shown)


src/tls/boring.rs
```diff
@@ -569,6 +569,7 @@ where
 
     fn accept(&self, conn: TcpStream) -> Self::AcceptFuture {
         let mut acceptor = self.acceptor.clone();
+        conn.set_nodelay(true).unwrap();
```
howardjohn, line 572: you can do `SslStream.get_ref` to get the inner TcpStream. Then you can call set_nodelay.

So I think this is probably not the best place.

Maybe in http_util::tls_server add

```
        .map(|conn| {
            conn.get_ref().set_nodelay()
        })
```

?

Stevenjin8: > you can do `SslStream.get_ref` to get the inner TcpStream. Then you can call set_nodelay.
> 
> So I think this is probably not the best place.
> 
> Maybe in http_util::tls_server add
> 
> ```
>         .map(|conn| {
>             conn.get_ref().set_nodelay()
>         })
> ```
> 
> ?

@howardjohn NGL I'm a bit surprised this works. I thought that setting NODELAY after the TLS handshake would cause handshakes to be really slow, but that doesn't seem to be the case (at least locally).

howardjohn: oh I see the problem, its after the TLS handshake. I didn't think of that. I guess current way is fine then.. I don't mind either way I guess

howardjohn (approved):

PR #620, opened 2023-07-25
Async ssl config provider for xDS and CA connections

EItanya:
Switch the implementation of TLS in the xDS and CA clients to be an async SSL config provider. Currently the only implementation of this config provider is a file based one (`FileClientCertProviderImpl`). It was done this way to reduce the scope of this PR to just the interface. 

The reason for this change is to allow for more flexible certificate based auth mechanisms in the future. For example, JWT authentication will not work by default on AWS because their Identity document is not a JWT, therefore using something like SPIRE to attest the identity, and then connect via mTLS is a very attractive option.

The actual mechanism which will most likely be implemented as a follow up will be `SDS`. There is no "agreed" upon standard API for certificates, but this is the closest one, and mirrors the behavior of the pilot-agent.


howardjohn:
Why does XDS connection to Istiod have anything to do with externally managed certificates?

EItanya:
Depending on your system, and where the zTunnel is running you may use an external system to retrieve the certs to connect to istiod, potentially even the same system that you use to get the workload certs.

Right now there's no option for an mTLS connection with `istiod` only the authz route with the token which essentially requires running the agent in non-kubernetes environments. This just enables authenticating with `istiod` using the certificate mechanism instead.

In this case `externally managed certificates` just means certs which aren't loaded from the filesystem, or are potentially dynamic. An outcome from this could be a file watcher on a directory to provide the cert + key.

stevenctl:
If I have a VM running zTunnel (let's say I don't want L7 mesh) that got a cert from something besides SDS, I should be able to mount it. 

If I have a VM that fetched a cert using a token, future auth with istiod should use mTLS. You can nearly make zTunnel on a VM work with our existing [bootstrap](https://istio.io/latest/docs/setup/install/virtual-machine/#create-files-to-transfer-to-the-virtual-machine) besides the token expiring and not being able to talk to istiod later on.

EItanya:
`SDS` is a perfect example that could implement this interface. The answer to this could also be that we prefer to leave this out of the zTunnel altogether. But to me this seems like something simple enough that it shouldn't require a separate component mounting the certs.

bleggett:
```
Right now there's no option for an mTLS connection with istiod only the authz route with the token which essentially requires running the agent in non-kubernetes environments. This just enables authenticating with istiod using the certificate mechanism instead.
```

Supporting mTLS-based auth with Istiod in addition to token-based auth strikes me as pretty reasonable, as a baseline.

EItanya:
> Supporting mTLS-based auth with Istiod in addition to token-based auth strikes me as pretty reasonable, as a baseline.

100%, and that's what I was trying to get across. I think the extra step I took is that I would like to see that mechanism be async to support mechanisms other than files.

EItanya:
@howardjohn curious to get your updated thoughts. Essentially I created a client cert interface in the `tls` package abstract the getting of the client cert builder. Therefore now the xds, and soon the ca client can just take in that interface, and an optional auth interface. I already implemented it on the xds side.

src/identity/auth.rs
```diff
@@ -23,51 +23,46 @@ use tonic::{Code, Request, Status};
...
-        match self {
-            AuthSource::Token(_, cluster_id) => {
+                request.metadata_mut().insert("authorization", token);
                 if !cluster_id.is_empty() {
                     let id = AsciiMetadataValue::try_from(cluster_id.as_bytes().to_vec())
                         .map_err(|e| Status::new(Code::Unauthenticated, e.to_string()))?;
                     request.metadata_mut().insert("clusterid", id);
                 }
             }
+            // When no authentication is required, do nothing.
```
shashankram, line 64: Is this specifically referring to non-token based auth? If yes, would reword the comment to "When no token-based authentication ....`

shashankram (commented):
Looks reasonable

kdorosh (commented):
looks reasonable to me; would like to see some more updates to the PR description to provide more context on use cases (e.g. AWS) and how this helps the community

src/xds/client.rs
```diff
@@ -383,11 +381,20 @@ impl AdsClient {
...
-        let mut response_stream = client
-            .delta_aggregated_resources(tonic::Request::new(outbound))
-            .await
-            .map_err(Error::Connection)?
-            .into_inner();
+        let tls_grpc_channel = tls::grpc_connector(
+            self.config.address.clone(),
+            self.config.tls_builder.fetch_cert().await.unwrap(),
+        )
+        .unwrap();
```
howardjohn, line 388:
```suggestion
        ?;
```

howardjohn (approved):

src/xds/client.rs
EItanya: Fixed

PR #622, opened 2023-07-26
Update ondemand dns metric name

kdorosh:
Align better with prometheus best practices for naming https://prometheus.io/docs/practices/naming/

use `_total` and `..._cache_misses_total` rather than `..._cache_hits_total` and `..._cache_misses_total`

### DIFF
--- a/tests/direct.rs
+++ b/tests/direct.rs
@@ -225,7 +225,7 @@ async fn test_vip_request() {
 
 fn on_demand_dns_assertions(metrics: ParsedMetrics) {
     for metric in &[
-        ("istio_on_demand_dns_cache_hits_total"),
+        ("istio_on_demand_dns_total"),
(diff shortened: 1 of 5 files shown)


tests/direct.rs
```diff
@@ -236,8 +236,15 @@ fn on_demand_dns_assertions(metrics: ParsedMetrics) {
             "expected metric {metric} to have len(1)"
         );
         let value = m.unwrap()[0].value.clone();
+        let expected = match *metric {
+            "istio_on_demand_dns_total" => prometheus_parse::Value::Untyped(2.0),
+            "istio_on_demand_dns_cache_misses_total" => prometheus_parse::Value::Untyped(1.0),
```
howardjohn, line 241: we could also do a label cache=hit or cache=miss?

howardjohn (approved):

PR #627, opened 2023-07-29
enhance wl xds logging

ilrudie:
Found the logging of "handling delete {}" for both update and remove to be difficult to keep track of. PR looks to clarify xds handling with better log messages.

- lower log level for wl deletes to debug
- add debug log for wl inserts

### DIFF
--- a/src/xds.rs
+++ b/src/xds.rs
@@ -76,6 +76,8 @@ impl ProxyStateUpdater {
     }
 
     pub fn insert_workload(&self, w: XdsWorkload) -> anyhow::Result<()> {
+        debug!("handling insert {}", w.uid);
+
         // Convert the workload.
         let workload = Workload::try_from(&w)?;
 
@@ -104,7 +106,7 @@ impl ProxyStateUpdater {
     }
 
     pub fn remove(&self, xds_name: &String) {
-        info!("handling delete {}", xds_name);
+        debug!("handling delete {}", xds_name);
         let mut state = self.state.write().unwrap();
 
         // remove workload by UID; if xds_name is a service then this will no-op


src/xds.rs
```diff
@@ -76,11 +93,13 @@ impl ProxyStateUpdater {
     }
 
     pub fn insert_workload(&self, w: XdsWorkload) -> anyhow::Result<()> {
+        info!("handling insert {}", &w.uid);
```
hzxuzhonghu, line 96: is this too noisy

hzxuzhonghu: And with insert log, it will be clear whether a WL is updated or delete

ilrudie: PR doesn't change the number of logs being generated at the info level. Prior to this PR for every insert we would log "handling delete {}" and with this PR we'll log "handing insert {}" instead.

If you change to debug log level then you will see additional logs because you'll also be seeing the delete logged for every insert. I thought being able to see the delete may be handy for debugging but would be too noisy normally.

hzxuzhonghu: So can we tune both of then to debug or trace

src/xds.rs
```diff
@@ -54,6 +55,22 @@ pub enum Error {
     OnDemandSend(),
 }
 
+/// DeleteReason enumerates the possible reasons that ztunnel is removing a workload from proxy state
+pub enum DeleteReason {
+    XdsRemove,
+    /// XdsPreClean means we're inserting a workload which involves deleting the old version of the workload to ensure a clean insert
+    XdsPreClean,
+}
```
hzxuzhonghu, line 63: About this, i would suggest not making it complex

hzxuzhonghu: @howardjohn

howardjohn: Yeah I mostly agree this seems a bit excessive just for debug logs

ilrudie: Are we OK with every update logging about the update and the delete so long as it's at debug?

hzxuzhonghu: Yes, sounds good. Workload is a fast-changing resource

hzxuzhonghu (approved):

PR #630, opened 2023-08-01
metrics: report Service from inbound side as well

howardjohn:
This mirrors Envoy implementation and is required for correct
dashboards/kiali/etc.

Basically we try to guess the Service since we cannot really know; same
as Envoy.

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -410,10 +410,10 @@ pub async fn relay(
     }
 }
 
-// pick_upstream_service selects an upstream service for inbound metrics.
+// guess_inbound_service selects an upstream service for inbound metrics.
 // There may be many services for a single workload. We find the the first one with an applicable port
 // as a best guess.
-pub fn pick_upstream_service(
+pub fn guess_inbound_service(
     conn: &Connection,
     upstream_service: Vec<Service>,
     dest: &Workload,
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -419,7 +419,6 @@ pub fn pick_upstream_service(
     dest: &Workload,
 ) -> Option<ServiceDescription> {
     let dport = conn.dst.port();
-    error!("upstream service... {upstream_service:?}");
     let netaddr = network_addr(&dest.network, conn.dst.ip());
     let euid = endpoint_uid(&dest.uid, Some(&netaddr));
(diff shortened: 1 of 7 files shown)


src/proxy.rs
```diff
@@ -407,6 +410,37 @@ pub async fn relay(
...
+    let euid = endpoint_uid(&dest.uid, Some(&netaddr));
+    upstream_service
+        .iter()
+        .find(|s| {
+            for (sport, tport) in s.ports.iter() {
+                if tport == &dport {
+                    // TargetPort directly matches
+                    return true;
+                }
+                if s.endpoints.get(&euid).and_then(|e| e.port.get(sport)) == Some(&dport) {
```
hzxuzhonghu, line 433: can you clarify when endpoint port is different from service port

src/proxy.rs
```diff
@@ -407,6 +410,37 @@ pub async fn relay(
...
+// pick_upstream_service selects an upstream service for inbound metrics.
+// There may be many services for a single workload. We find the the first one with an applicable port
+// as a best guess.
+pub fn pick_upstream_service(
+    conn: &Connection,
+    upstream_service: Vec<Service>,
+    dest: &Workload,
+) -> Option<ServiceDescription> {
+    let dport = conn.dst.port();
+    error!("upstream service... {upstream_service:?}");
```
ymesika, line 422: Debugging leftover?

ymesika (approved):
LG except the comment I left.

src/proxy.rs
howardjohn: yes added a comment. tl;dr Named port in k8s

src/proxy.rs
```diff
@@ -407,6 +410,38 @@ pub async fn relay(
     }
 }
 
+// pick_upstream_service selects an upstream service for inbound metrics.
+// There may be many services for a single workload. We find the the first one with an applicable port
+// as a best guess.
+pub fn pick_upstream_service(
```
hzxuzhonghu, line 416: One last suggestion is better to rename this function, this is only for inbound usage. We should prevent outbound usage by mistake

howardjohn: done

hzxuzhonghu (approved):

PR #631, opened 2023-08-02
HBONE connections benches

### LINKED ISSUE #623: Benching HBONE handshake performance
Right now, we have benchmarks for time taken to establish a new TCP connection over an existing HBONE connection because of connection pooling

Stevenjin8:
Benchmarks how long it takes to establish a new HBONE connection.
Do this by creating `n` workloads giving us `O(n^2)` ways to create new HBONE connections. The actual limiting factor here is not the number of workload source/destination pairs, but the number of possible open file descriptions.
This happens to me when I change the duration of the test from 5 to 10 seconds.

I confirmed this is creating new HBONE connections by running `tcpdump` on `lo` and seeing the number of TLS handshakes was more or less the same as the number of test iterations.

resolves #623


Stevenjin8:
I have no idea how to fix the CI error

```
error: non-binding `let` on a future
   --> benches/throughput.rs:465:9
    |
465 | /         let _ = tokio::spawn(async move {
466 | |             let _ = tokio::join!(app.wait_termination(), echo.run());
467 | |         });
    | |___________^
    |
    = help: consider awaiting the future or dropping explicitly with `std::mem::drop`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#le
```

The other benches do the by doing `let t = tokio::spawn(...)` and then returning `t`. However, `t` gets eventually bound to `_` so it never gets awaited

benches/throughput.rs
```diff
@@ -374,12 +380,139 @@ pub fn metrics(c: &mut Criterion) {
...
+/// Thus, we identify hosts by a u8 in the form 127.0.1.x
+fn hbone_connection_ip(x: u8) -> IpAddr {
+    IpAddr::V4(Ipv4Addr::new(127, 0, 1, x))
+}
+
+fn hbone_connection_config() -> ztunnel::config::ConfigSource {
+    let mut workloads: Vec<LocalWorkload> = Vec::with_capacity(MAX_HBONE_WORKLOADS as usize);
+    // We can't create one work load with many IPs because ztunnel could connect to any one causing
+    // inconsistent behavior.
+    for i in 1..MAX_HBONE_WORKLOADS + 1 {
```
MorrisLaw, line 408: Why do we loop to `MAX_HBONE_WORKLOADS` + 1?

benches/throughput.rs
```diff
@@ -132,16 +139,16 @@ fn initialize_environment(
             let _ = tokio::join!(app.wait_termination(), echo.run());
         });
         let mut hbone = ta
-            .socks5_connect(helpers::with_ip(
-                echo_addr,
-                TEST_WORKLOAD_HBONE.parse().unwrap(),
-            ))
+            .socks5_connect(
```
MorrisLaw, line 142: This section is repeated often.

e.g.

```
 let mut s =
    e.ta.socks5_connect(helpers::with_ip(
        e.echo_addr,
        TEST_WORKLOAD_TCP.parse().unwrap(),
    ))
    .await;
```

Wdyt about abstracting that logic into a helper function, maybe something called `newTCPStream` or similar. Have it maybe take in the `TEST_WORKLOAD_<SOURCE/HBONE/TCP>` string value as an argument. Would help to maintain that DRY principle while also making it convenient for the next person to extend this test with more benches?

MorrisLaw:
> 

`let t` won't work for your test? Is that why you did `let _`?

Stevenjin8:
> > 
> 
> `let t` won't work for your test? Is that why you did `let _`?

`let t` doesn't work because it ends up being an unused variable. But the warning comes from the fact that I never await the result of the spawn, which is the same for the other benches.

benches/throughput.rs
jaellio: I believe the upper bound is exclusive.

Stevenjin8: I originally had `0..MAX_HBONE_WORKLOADS`, but that means that a workload would get assigned the ip 127.0.1.0, which isn't a valid host address. So I just shifted everything by 1.

tests/direct.rs
```diff
@@ -197,7 +201,10 @@ async fn run_requests_test(
         let dst = SocketAddr::from_str(target)
             .unwrap_or_else(|_| helpers::with_ip(echo_addr, target.parse().unwrap()));
         for _ in 0..num_queries {
-            let mut stream = app.socks5_connect(dst).await;
+            let mut stream = app.socks5_connect(dst,
+
```
Stevenjin8, line 205: format

benches/throughput.rs
```diff
@@ -286,24 +293,23 @@ pub fn connections(c: &mut Criterion) {
...
+                    helpers::with_ip(e.echo_addr, TEST_WORKLOAD_TCP.parse().unwrap()),
+                    TEST_WORKLOAD_SOURCE.parse().unwrap(),
+                )
                 .await;
             tcp::run_client(&mut s, 1, Mode::ReadWrite).await
         })
     });
-    // TODO(https://github.com/istio/ztunnel/issues/15): when we have pooling, split this into "new hbone connection"
-    // and "new connection on existing HBONE connection"
+    // This tests connection time over an existing HBONE connection.
```
jaellio, line 304: Could you clarify this comment? Thanks!

benches/throughput.rs
```diff
@@ -374,12 +380,139 @@ pub fn metrics(c: &mut Criterion) {
...
+/// Reserve IPs in 127.0.1.0/24 for these HBONE connection tests.
+/// Thus, we identify hosts by a u8 in the form 127.0.1.x
+fn hbone_connection_ip(x: u8) -> IpAddr {
+    IpAddr::V4(Ipv4Addr::new(127, 0, 1, x))
+}
+
+fn hbone_connection_config() -> ztunnel::config::ConfigSource {
+    let mut workloads: Vec<LocalWorkload> = Vec::with_capacity(MAX_HBONE_WORKLOADS as usize);
+    // We can't create one work load with many IPs because ztunnel could connect to any one causing
+    // inconsistent behavior.
```
jaellio, line 407: You could also clarify in the comment what approach your *are* taking. 

`Instead, we create many workloads with unique IPs to ensure a new HBONE connection..."

benches/throughput.rs
```diff
@@ -374,12 +380,139 @@ pub fn metrics(c: &mut Criterion) {
...
+        });
+        (echo_addr, ta)
+    });
+
+    let ta: Arc<Mutex<TestApp>> = Arc::new(Mutex::new(ta));
+    // Host addresses can't end with 0.
+    let addresses = Arc::new(Mutex::new((1u8, 2u8)));
+
+    let mut c = c.benchmark_group("hbone_connection");
+    // WARNING increasing the measurement time could lead to running out of IP pairs or having to
```
jaellio, line 476:
```suggestion
    // WARNING increasing the measurement time could lead to running out of IP pairs or having too
```

benches/throughput.rs
Stevenjin8: I'll think about this. The whole echo server setup is clever but also confusing. You can access the echo server through any ip address on `lo`, but if you want to use ztunnel/socks5, the address you pick to access the echo server needs to be a registered workload.

Point is, I feel like the way its abstracted right now is already confusing, so I didn't want to abstract it any further. I do think there is probably a better way to do this. My first idea it to make `TEST_WORKLOAD_*` actual IP address instead of a string that can be parsed into an ip address. Thoughts?

benches/throughput.rs
```diff
@@ -374,12 +380,139 @@ pub fn metrics(c: &mut Criterion) {
     });
 }
 
+/// Iterate through possible IP pairs restricted to 0 < ip_pair.0 < ip_pair.1 <= MAX_HBONE_WORKLOADS.
+fn next_ip_pair(ip_pair: (u8, u8)) -> (u8, u8) {
```
jaellio, line 384: Note this is only n*(n-1)/2 combos. If we wanted more new HBONE connections from the same number of IPs this could be optimized.

benches/throughput.rs
MorrisLaw: Ahh ok that makes sense

benches/throughput.rs
MorrisLaw: The abstraction I refer to is to really help reduce the repetitive and confusing code in the tests. Using a helper function should actually make it more understandable.

benches/throughput.rs
```diff
@@ -374,12 +380,138 @@ pub fn metrics(c: &mut Criterion) {
...
+            .await
+            .unwrap();
+        let ta = TestApp::from((&app, cert_manager));
+        ta.ready().await;
+
+        let echo = tcp::TestServer::new(Mode::ReadWrite, 0).await;
+        let echo_addr = echo.address();
+        let _ = tokio::spawn(async move {
+            let _ = tokio::join!(app.wait_termination(), echo.run());
+        });
```
keithmattix, line 467: Adding a `.await` here should fix the CI

Stevenjin8: TODO: Drop works, but hoping that someone has a better solution than just orphaning the tasks.

howardjohn (approved):

PR #632, opened 2023-08-02
xds: refactor client to be type-agnostic

howardjohn:
This is just a minor cleanup; when I first wrote this I didn't know
enough about rust to make this properly generic. Now instead of
hardcoding type-awareness into the client it can handle arbitrary types.

No behavior change is expected.

The AdsClient has part of it split into a new `State` type to help with
borrow-checking: we cannot give mutable access to AdsClient and call a
handler otherwise.

### DIFF
--- a/src/xds/client.rs
+++ b/src/xds/client.rs
@@ -27,13 +27,11 @@ use tracing::{debug, error, info, info_span, warn, Instrument};
 
 use crate::config::RootCert;
 use crate::metrics::IncrementRecorder;
-use crate::xds::istio::security::Authorization;
-use crate::xds::istio::workload::Address;
 use crate::xds::metrics::{ConnectionTerminationReason, Metrics};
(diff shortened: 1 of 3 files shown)


src/xds/client.rs
```diff
@@ -112,71 +98,177 @@ pub fn handle_single_resource<T: prost::Message, F: FnMut(XdsUpdate<T>) -> anyho
...
     proxy_metadata: HashMap<String, String>,
-
-    address_handler: Box<dyn Handler<Address>>,
-    authorization_handler: Box<dyn Handler<Authorization>>,
+    handlers: HashMap<String, Box<dyn RawHandler>>,
     initial_watches: Vec<String>,
     on_demand: bool,
 }
 
+pub struct State {
```
hzxuzhonghu, line 162: nit: confusing with ProxyState

hzxuzhonghu (approved):

PR #649, opened 2023-08-10
inbound passthrough always denies when there is a waypoint

stevenctl:
Until we make a proper decision around whether to hairpin, we should not sometimes do it if the IP is in-cluster and the identity happens to exist on the node.

https://github.com/istio/istio/issues/44530

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -170,6 +170,9 @@ pub enum Error {
     #[error("unknown source: {0}")]
     UnknownSource(IpAddr),
 
+    #[error("unauthenticated source: {0}")]
+    UnauthenticatedSource(IpAddr),
+
     #[error("unknown waypoint: {0}")]
     UnknownWaypoint(String),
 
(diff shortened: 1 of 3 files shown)


keithmattix:
I agree with the change, but won't this break istio/istio tests?

howardjohn (approved):

stevenctl:
```rust
diff --git a/tests/namespaced.rs b/tests/namespaced.rs
index b90d106..a7a045e 100644
--- a/tests/namespaced.rs
+++ b/tests/namespaced.rs
@@ -421,7 +421,7 @@ mod namespaced {
             .register()?;
         let zt = manager.deploy_ztunnel(DEFAULT_NODE)?;
 
-        run_tcp_to_hbone_client(client, manager.resolver(), "server")?;
+        run_tcp_to_hbone_client(client, manager.resolver(), "server").unwrap();
 
         let metrics = [
             (CONNECTIONS_OPENED, 1),
@@ -598,7 +598,7 @@ mod namespaced {
         client
             .run(move || async move {
                 info!("Running client to {srv}");
-                let mut stream = TcpStream::connect(srv).await.unwrap();
+                let mut stream = TcpStream::connect(srv).await?;
                 hbone_read_write_stream(&mut stream).await;
                 Ok(())
             })?
```


I would thik this should make it fail, no? I wanted to make sure it would fail an dthen change it to `expect_err`. I can't figure out why this doesn't bubble up the error that is currently panicking CI inside the `.run` lambda.

Issue #638, opened 2023-08-04
Improve ztunnel dashboard UI


hanxiaop:
Ref: https://github.com/istio/istio/pull/45695#issuecomment-1662312782

1. ztunnel `/config_dump` does not return a pretty printed json object
2. ztunnel `/logging` is unusable since it expects a POST and a level, this should be similar to envoy logging API.
3. ztunnel currently doesn't have a `/stats` API on 15000, but it does have one on :`15020/metrics`

hanxiaop:
> Doesn't logging in envoy use a POST as well? Modifications _should_ require a POST? Or are we just talking about viewing the current level?

We are talking about viewing the current level, currently it's not viewable directly like envoy.

> For stats, we don't need to copy envoy IMO.

+1, I think `15020/metrics` is good.

Stevenjin8:
I can do the pretty print

Stevenjin8:
@hanxiaop

> ztunnel /logging is unusable since it expects a POST and a level, this should be similar to envoy logging API.

I don't think this is completely true. I can get the logging level of ztunnel through an empty post:

```
root@ztunnel-4v22b:/# curl -d "" localhost:15000/logging
current log level is info
```

Though, it is much less descriptive and formatted differently than envoy:

```
istio-proxy@my-pod:/$ curl -d "" localhost:15000/logging
active loggers:
  admin: warning
  alternate_protocols_cache: warning
...
```

Neither ztunnel nor enovy support GET requests.

So I wouldn't say that " /logging is unusable since it expects a POST and a level," but the ztunnel api is definitely different than the envoy API.

hanxiaop:
@Stevenjin8 Thanks for testing this. We are talking about the `/logging` API in the dashboard. Now, you can try to see the ztunnel dashboard using a command like `istioctl d envoy ztunnel-bg4b7.istio-system`. Once you click on the `/logging` API, it shows
```
Invalid HTTP method
 
usage: POST /logging						(To list current level)
usage: POST /logging?level=<level>				(To change global levels)
usage: POST /logging?level={mod1}:{level1},{mod2}:{level2}	(To change specific mods' logging level)

hint: loglevel:	error|warn|info|debug|trace|off
hint: mod_name:	the module name, i.e. ztunnel::proxy
```

I'm not sure, but I think we may need to modify the dashboard page, and perhaps the API a little, to make it work?

Stevenjin8:
[...]

PR #654, opened 2023-08-15
Add GET method for admin logging endpoint

Stevenjin8:
This way we can properly access the endpoint with hyperlinks. #638


GregHanson:
left some similar feedback [here](https://github.com/istio/ztunnel/issues/638#issuecomment-1683961476).  Not sure adding GET support to the API is the way to go.  I think the dashboard ultimately just needs work on the html side IMO.  That said I think this one can be addressed again post-beta

howardjohn:
I don't think we should spend any time improving the UI of this.

PR #655, opened 2023-08-16
Update Hyper to 1.0.0-rc4

howardjohn:
This is 1 step closer to 1.0.0....

The main change is they made their own Read/Write types, so we need to
wrap a bunch of things in `TokioIo`. NBD.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,9 +55,9 @@ checksum = "34fde25430d87a9388dadbe6e34d7f72a462c8b43ac8d309b42b0a8505d7e2a5"
 
 [[package]]
 name = "anyhow"
-version = "1.0.72"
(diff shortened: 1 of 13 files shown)


howardjohn:
note to self: need to sort key usage

hzxuzhonghu:
what's the improvement

howardjohn:
Nothing but we need to update eventually since this will become the 1.0.0.
Better adopt to the breaking change sooner than later.

Also this bumps the boringssl library which Nate said fixed OSX build
although I couldn't verify myself.

ericvn:
And hoping this fixes the failure in https://github.com/istio/ztunnel/pull/656.

stevenctl (approved):

PR #658, opened 2023-08-18
Don't special-case DNS resolution based on the client

kdorosh:
Don't special-case DNS resolution based on the client. Instead, acknowledge that ServiceEntry is a global API in ambient and treat it consistently for all clients.

If users really want a specific service they can always use the VIP.

Also, with the old namespace preference we have an issue where we cannot reach workloads behind the hostname that reside outside the client namespace since the ServiceEntry selector only works within its own namespace

cc @nmittler

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -19,8 +19,8 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
-use itertools::Itertools;
 use once_cell::sync::Lazy;
+use rand::prelude::IteratorRandom;
 use rand::seq::SliceRandom;
 use rand::thread_rng;
 use tokio::net::{TcpListener, UdpSocket};
@@ -310,11 +310,12 @@ impl Store {
                     // We found a match. We always return `Some` result, even if there
                     // are zero records returned.
 
-                    // Get the service matching the client namespace. If no match exists, just
-                    // return the first service.
+                    // Pick any service with the host. We want to round-robin
+                    // the resolved VIP behind any given hostname if there are
+                    // conflicts across namespaces.
                     let service = services
                         .iter()
-                        .find_or_first(|service| service.namespace == client.namespace)
+                        .choose(&mut rand::thread_rng())
                         .cloned()
                         // Should never be empty, since we delete the Vec when it's empty.
                         .unwrap();
@@ -706,6 +707,7 @@ impl Forwarder for SystemForwarder {
 
 #[cfg(test)]
 mod tests {
+    use itertools::Itertools;
     use std::cmp::Ordering;
     use std::collections::HashMap;
     use std::net::{SocketAddrV4, SocketAddrV6};
--- a/src/state/service.rs
+++ b/src/state/service.rs
@@ -118,8 +118,8 @@ pub struct ServiceStore {
 
     /// Allows for lookup of services by hostname, and then by namespace. XDS uses a combination
     /// of hostname and namespace as the primary key. In most cases, there will be a single
-    /// service for a given hostname. However, `ServiceEntry` allows hostnames to be overridden
-    /// on a per-namespace basis.
+    /// service for a given hostname. However, `ServiceEntry` is global (the VIPs and hostnames
+    /// are valid for any mesh pod) so if there are several we will return all services.
     by_host: HashMap<String, Vec<Arc<Service>>>,
 }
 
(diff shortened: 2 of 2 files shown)


src/dns/server.rs
```diff
@@ -310,11 +310,12 @@ impl Store {
...
                     // are zero records returned.
 
-                    // Get the service matching the client namespace. If no match exists, just
-                    // return the first service.
+                    // Pick any service with the host. We want to round-robin
+                    // the resolved VIP behind any given hostname if there are
+                    // conflicts across namespaces.
                     let service = services
                         .iter()
-                        .find_or_first(|service| service.namespace == client.namespace)
```
howardjohn, line 317: > Also, with the old namespace preference we have an issue where we cannot reach workloads behind the hostname that reside outside the client namespace since the ServiceEntry selector only works within its own namespace

Isn't it "find_or_first"? so it can go cross namespace?

kdorosh: If there is a service entry in our namespace with the host we requested, we will always stay within our namespace. I think we should load balance across all workloads, just like we would if the ServiceEntry lived outside our namespace.

I think we should have simple behavior and not special case

howardjohn: >If there is a service entry in our namespace with the host we requested, we will always stay within our namespace

So you would rather 50% of the time we go to namespace 1 and namespace 2? That doesn't seem great and isn't consistent with how sidecars work either -- they are "undefined behavior" when you have multiple, but it consistently goes to one or the other

kdorosh: Yes, I would prefer to load balance. I don't think sidecar's "undefined behavior" should be our role model or example here.

1. If we didn't want to load balance, that's fine. We can use the VIP
2. If we don't load-balance, now two different clients get very subtly different behavior because of the namespace they live in. This also includes things like eventually supporting DNS PTR
3. as precedent: kube service FQDNs are also global
4. I don't think the current "undefined behavior" for sidecars is good for multi-tenancy reasons. As I understand it today, If I define the same hostname + svc port on different service entries then istiod freezes until one of the ServiceEntries is "fixed":

```
        "outbound|80||serviceentry.istio.io": {
            "proxy": "istio-ingressgateway-9cd748746-xl8rk.istio-system",
            "message": "Duplicate cluster outbound|80||serviceentry.istio.io found while pushing CDS"
        }
    }
}
```

I don't like that users can break or stick existing configuration or using that as a precedent.

Instead I think we should make ServiceEntry global, like k8s services. If we want to prevent users from duplicating hostnames across ServiceEntries without sticking configuration, instead having a defined and well understood behavior for which ones get ignored (the older?), I'm open to that as well.

howardjohn: In all cases I know of Istio behavior in the event of conflicts is to consistently pick one (where one is typically the oldest, if there is not a "most relevant"). Randomly flipping between the two doesn't seem to ever be the right behavior for a user.

> If we want to prevent users from duplicating hostnames across ServiceEntries

There is no way we can actually _prevent_ this, which is why its a conflict and we need to pick one.

yuval-k: i think the question is what the semantics should be?
the way i see it, as service entry:
- It defines a global hostname
- It selects backend workloads

What does it mean to define a service entry in multiple namespaces then?
I would think of it as adding more backends to the global hostname.

With this in mind, I don't see why if the origin namespace of a request to a service entry should impact the backend selection. 
i.e. the backend should be selected the the pool of all backends in a consistent manner, regardless of where the traffic originated form.

WDYT? Do you see the semantics differently?

howardjohn: The way it was designed was each workload sees one service. We cannot merge across two different namespaces, the behavior is wonky. User A may use a SE for google.com to overwrite to their own local `google-fake` and another for the real google. One may set port 443 to HTTP to use TLS origination and the other HTTPS since they do ti from the app, etc.

So we pick one in Istio for that reason. We don't have a good way to pick one, though, so prefer same namespace then oldest. This allows them to always 'win' the conflict on their own, and if not we use the old state.

I am sure you have a customer use case, so what is the motivation on your end? The current logic behaves more like Istio sidecars so I am a bit surprised its not ok?

ilrudie: Does `.iter()` make any guarantees about ordering? Is `find_or_first(...)` going to produce the true oldest or just whichever happened to be configured first when this ztunnel started up?

nmittler:
@kdorosh perhaps it would be better to write a short doc for this first to explain the problem you're trying to solve? This feels a bit like designing-by-PR

src/dns/server.rs
howardjohn: iter on Vec is ordered, yes. I don't know that we create the Vec in a cohesive ordered fashion, though.

So I am not sure the current implementation meets my expectation either, just to be clear (although it might)

kdorosh:
closing in favor of a design doc to demonstrate the tradeoffs for the different possible semantics of ServiceEntry for ambient

kdorosh:
design doc [here](https://docs.google.com/document/d/1uDWoWxHyMCE4oUc-nTJPfVoQyikZP-UMp-_BrAA-PQE/edit)

discussion in slack about next steps [here](https://istio.slack.com/archives/C049TCZMPCP/p1692812392839189)

kdorosh:
closing since i don't actually want to merge this PR. I want namespaced service entries

PR #662, opened 2023-08-22
Add tcp keepalives for xDS client connection

keithmattix:
Fixes istio/istio#46624


howardjohn (approved):
I don't think this actually fixes the issue. The issue is more that RBAC is applied on new connections only, and envoy (waypoint) pools HTTP connections so it isn't recreating connections when the RBAC changes.

That being said I think this is a good change

keithmattix:
Ah I missed that nuance in your previous comment; I'll reopen the issue

PR #664, opened 2023-08-23
add connection denied metrics for RBAC decisions in zTunnel

EItanya:
Add `istio_tcp_connection_denied_total` metrics for `rbac` decisions made by the `zTunnel`.

This PR adds metrics using the current `CommonTrafficLabels` when `rbac` decisions are made.
Also it adds tests to `namespaced.rs` to exercise the RBAC as well as the metrics.

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -33,7 +33,7 @@ use super::Error;
 use crate::baggage::parse_baggage_header;
 use crate::config::Config;
 use crate::identity::SecretManager;
-use crate::metrics::Recorder;
+use crate::metrics::{IncrementRecorder, Recorder};
 use crate::proxy;
 use crate::proxy::inbound::InboundConnect::{DirectPath, Hbone};
 use crate::proxy::metrics::{ConnectionOpen, Metrics, Reporter};
@@ -290,13 +290,8 @@ impl Inbound {
                 }
                 if from_waypoint {
                     debug!("request from waypoint, skipping policy");
-                } else if !state.assert_rbac(&conn).await {
-                    info!(%conn, "RBAC rejected");
-                    return Ok(Response::builder()
-                        .status(StatusCode::UNAUTHORIZED)
-                        .body(Empty::new())
-                        .unwrap());
                 }
+
                 if has_waypoint && !from_waypoint {
                     info!(%conn, "bypassed waypoint");
                     return Ok(Response::builder()
@@ -347,6 +342,18 @@ impl Inbound {
                     connection_security_policy: metrics::SecurityPolicy::mutual_tls,
                     destination_service: ds,
                 };
+
+                if !state.assert_rbac(&conn).await {
+                    info!(%conn, "RBAC rejected");
+                    let denied_metrics: metrics::ConnectionDenied =
+                        metrics::ConnectionDenied::from(&connection_metrics);
+                    metrics.as_ref().increment(&denied_metrics);
+                    return Ok(Response::builder()
+                        .status(StatusCode::UNAUTHORIZED)
+                        .body(Empty::new())
+                        .unwrap());
+                }
+
                 let status_code = match Self::handle_inbound(
                     Hbone(req),
                     enable_original_source.then_some(source_ip),
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -292,13 +292,6 @@ impl Inbound {
                     debug!("request from waypoint, skipping policy");
                 }
 
-                if has_waypoint && !from_waypoint {
-                    info!(%conn, "bypassed waypoint");
-                    return Ok(Response::builder()
-                        .status(StatusCode::UNAUTHORIZED)
-                        .body(Empty::new())
-                        .unwrap());
-                }
                 let source_ip = if from_waypoint {
                     // If the request is from our waypoint, trust the Forwarded header.
                     // For other request types, we can only trust the source from the connection.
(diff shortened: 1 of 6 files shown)


src/proxy/inbound.rs
```diff
@@ -347,6 +342,18 @@ impl Inbound {
                     connection_security_policy: metrics::SecurityPolicy::mutual_tls,
                     destination_service: ds,
                 };
+
+                if !state.assert_rbac(&conn).await {
```
kdorosh, line 346: the PR title (and lack of description) indicate this PR just adds metrics, but this seems like a change in behavior from before, since it's no longer in the else block

kdorosh: if it's from a waypoint then why do we need to assert rbac now? before we did not, that seems correct

EItanya: That's a fair point, here's the issue: we need source info for metrics, and that is computed afterwards. I'm open to suggestions/alternatives here.

src/proxy/metrics.rs
```diff
@@ -27,6 +27,7 @@ use crate::state::workload::Workload;
 pub struct Metrics {
     pub connection_opens: Family<CommonTrafficLabels, Counter>,
     pub connection_close: Family<CommonTrafficLabels, Counter>,
+    pub connection_deny: Family<CommonTrafficLabels, Counter>,
```
howardjohn, line 30: will take a closer look in a bit (and the other PR), but high level comment -- we have the "Core Istio metrics" and new metrics. The former is bound by backwards compat.

I wonder if for the new ones we should:
1. mark all of them experimental. Maybe in the metric description we just put `[EXPERIMENTAL]`, I think that is how k8s does it.
2. Use a reduced set of labels. We have like 5 labels that are ~the same, mostly for backwards compat. We don't really need all of them IMO.

WDYT?

src/proxy/metrics.rs
```diff
@@ -108,6 +109,8 @@ pub struct ConnectionClose<'a>(&'a ConnectionOpen);
 
 pub struct BytesTransferred<'a>(&'a ConnectionOpen);
 
+pub struct ConnectionDenied<'a>(&'a ConnectionOpen);
```
howardjohn, line 112: We report denied in Envoy as well -- but as a response flag. Should we do the same here?

src/proxy/metrics.rs
EItanya: I'd definitely be open to that, do you have a list that you'd like to get rid of?

src/proxy/metrics.rs
EItanya: Could you explain a little bit more what you mean by that. In the context of a connection, those are mostly used for debugging/logging purposes. There's no real way to return them to the user.

howardjohn: I mean we already have a way to report "connection denied by RBAC" in the existing metrics scheme, as we have a response_flag label that indicates why it was closed. Seems like if that works for envoy it should work for us, too.

EItanya: Ok, I think I see what you're saying now. Currently that's a `CommonTrafficLabel`, but I'm not sure `connection denied by RBAC` really makes sense on any of the existing metrics` ConnectionOpen`. Are you saying put the reason on `connection_closed` instead?

I do think there could be value in having a separate metric specifically for RBAC, especially to implement `AUDIT` in the future. Having context data as multiple extra labels could get confusing.

howardjohn: I don't know if we put it on open or closed in Istio envoy TCP, we can check. For HTTP its more clear since there is just "Requests" and we put code=403,flag=<something>.

EItanya: So I looked at the code, and the TCP metrics are all a bit async because the reportHelper is called on a timer.
However, it's called with endStream [here](https://github.com/istio/proxy/blob/ec5563aa0ff7333af61d51b4d5837f218c19b70b/source/extensions/filters/http/istio_stats/istio_stats.cc#L790C5-L790C17), and then the response code is added to the `tags_` [a few lines down](https://github.com/istio/proxy/blob/ec5563aa0ff7333af61d51b4d5837f218c19b70b/source/extensions/filters/http/istio_stats/istio_stats.cc#L798). Basically it doesn't seem like it's meant to show up for TCP metrics. Which in my opinion makes sense, it's much more of an HTTP concept. 

More reason I suppose to do something like you [mentioned above](https://github.com/istio/ztunnel/pull/664#discussion_r1304515927)

kdorosh: I agree we should start with the same format as envoy for these metrics; users should not have to scrape two different formats

EItanya: The problem with doing the same format as `envoy` for this particular thing is that the denies are actually captured and returned by the RBAC filter. So the stats will be `RBAC` stats, not `istio` stats. You can make some inference based on the response_code, but that could also come from upstream so it's not as accurate.

howardjohn: I think the confusion here is that I was wrong :slightly_smiling_face:.

In metrics, we have "response flag". In ztunnel it is always `-`.

In logs there is also `RESPONSE_CODE_DETAILS`. _That_ is where rbac is reported (also, envoy docs lie and say its only HTTP -- its not true).

To me an ideal outcome would be to define a new response flag (perhaps `RBACDenied (RD)`) in envoy. Implement it for envoy RBAC, which is then reporting in metrics and logs. Report the same flag in ztunnel.

WDYT?

Technically we can have ztunnel specific response flags as well but not ideal

howardjohn:
#956

PR #672, opened 2023-08-30
optimize the function load_balance_for_hostname to avoid starting multiple DNS tasks

zhlsunshine:
optimize the function `load_balance_for_hostname` to avoid starting multiple background on-demand DNS tasks.


src/state.rs
```diff
@@ -341,9 +358,42 @@ impl DemandProxyState {
...
+                        cached_resolve_dns
+                            .complete_resolved_dns
+                            .store(true, Ordering::Relaxed);
+                    } else {
+                        while !cached_resolve_dns
+                            .complete_resolved_dns
+                            .load(Ordering::Relaxed)
+                        {
+                            tokio::time::sleep(Duration::from_millis(100)).await;
+                        }
```
hzxuzhonghu, line 393: not an expert on dns, can you confirm what does core-dns do with such case

zhlsunshine: Hi @hzxuzhonghu,  there should be many threads calling `load_balance_for_hostname` at the same time, and `resolve_on_demand_dns`, which will do DNS resolve task including using `lookup_ip` to get ip address of hostname, would be called many times if high concurrency requests are sent for specified hostname which need to be resolved.
The optimization is to make sure that the first request for hostname to resolve the DNS once and other same hostname request will wait here.
I will add some comments in the code. Thanks!

howardjohn: Yeah but sleeping 100ms here is not appropriate

howardjohn: This is a fairly common pattern https://pkg.go.dev/golang.org/x/sync/singleflight

zhlsunshine: Greate usage for golang to use channel mechanism to implement singleflight pattern and `singleflight` has different context in different language. 😀
I will use the rust way to eliminate the sleep logic, thanks @howardjohn!

zhlsunshine: Hi @howardjohn & @hzxuzhonghu, after some tries, I find out that `AtomicBool` is the easier way to implement the `singleflight` than MPSC channel in this case. Because I have to consider the async function in `load_balance_for_hostname` and async function calling `resolve_on_demand_dns`.
And I eliminate the sleep logic, then add the timeout for in other concurrency requests. Is this okay for now? Thanks!

src/state.rs
```diff
@@ -456,6 +526,54 @@ impl DemandProxyState {
...
+            let timeout_duration = Duration::from_secs(10);
+            // Start the timer
+            let start_time = Instant::now();
+            if !cached_atom_bool
+                .complete_resolved_dns
+                .load(Ordering::Relaxed)
+            {
+                let elapsed_time = start_time.elapsed();
+                // Check if the timer is timed out or not
+                if elapsed_time >= timeout_duration {
```
hzxuzhonghu, line 568: dead loop until >10s or has complete_resolved_dns is true

zhlsunshine: Hi @hzxuzhonghu, yes, this dead loop will break in 2 cases:

1. timeout after 10s
2. complete_resolved_dns is true

I’m not sure that 10 seconds is appropriate or not, any suggestion?

hzxuzhonghu: The critical point is cpu wasting polling on a value, here it should be waiting for a signal/sem

zhlsunshine: Hi @hzxuzhonghu, get your points now. I change the code by using `use tokio::sync::Notify;` to implement the notifications for other requests. thanks for your advice!
I also have some details to discuss with you @howardjohn & @hzxuzhonghu, it may be not suitable to use MPSC channel in this use case after some experiments. As you know, MPSC stands for "Multiple Producers and Single Consumer", but in this case it's more like "Single Producers and Multiple Consumer", so `tokio::sync::Notify` should be the more proper. Thanks!

zhlsunshine:
Hi @howardjohn & @hzxuzhonghu, is there any comment for this PR? thanks!

hzxuzhonghu:
Will take a look

src/state.rs
```diff
@@ -341,9 +370,56 @@ impl DemandProxyState {
                     .on_demand_dns_cache_misses
                     .get_or_create(&labels)
                     .inc();
-                // TODO: optimize so that if multiple requests to the same hostname come in at the same time,
+                // optimize so that if multiple requests to the same hostname come in at the same time,
```
hzxuzhonghu, line 373: remove

src/state.rs
```diff
@@ -341,9 +370,56 @@ impl DemandProxyState {
...
                 // we don't start more than one background on-demand DNS task
-                Self::resolve_on_demand_dns(self.to_owned(), workload).await;
+                if state
+                    .get_cached_resolve_dns_for_hostname(&hostname)
```
hzxuzhonghu, line 376: we should directly check state.get_ips_for_hostname(&hostname).is_none here

and then could remove the the resolved hostnames from `crd_hashmap` to save memory

src/state.rs
```diff
@@ -341,9 +370,56 @@ impl DemandProxyState {
...
+                    let cached_resolve_dns = Arc::new(SingleFlight::new());
+                    let cached_resolve_dns_clone = cached_resolve_dns.clone();
+                    state.set_cached_resolve_dns_for_hostname(
+                        hostname.to_owned(),
+                        cached_resolve_dns_clone,
+                    );
+
+                    // Check if this is the first request of doing DNS task for hostname
+                    if cached_resolve_dns
+                        .should_resolved_dns
```
hzxuzhonghu, line 388: I donot think we need this field.

hzxuzhonghu (commented):
I donot fully understand the idea.

I think we just need `CachedResolvedDNSMap` which stores the hash map of hostname to SingleFlight

The single flight  only need to store the `Notify`

src/state.rs
hzxuzhonghu: crd_hashmap has a potential mem leak

src/state.rs
zhlsunshine: okay, I will remove these comments

src/state.rs
zhlsunshine: Hi @hzxuzhonghu, the crd hashmap need to store the information for hostname which need to be resolved, so what's kind of case would leak the mem?

src/state.rs
zhlsunshine: Hi @hzxuzhonghu, this filed has a confusion name, it may be like this: first_resolved_dns

zhlsunshine:
> I donot fully understand the idea.
> 
> I think we just need `CachedResolvedDNSMap` which stores the hash map of hostname to SingleFlight
> 
> The single flight only need to store the `Notify`

Hi @hzxuzhonghu, I think I can give more explains here for these 2 AtomicBools:
Although state is Arc<RwLock<ProxyState>>, all the reading and writing operations themselves for ProxyState are atomic just in the code block, such as read `CachedResolvedDNSMap` via `get_cached_resolve_dns_for_hostname` and write `CachedResolvedDNSMap` via `set_cached_resolve_dns_for_hostname`, so we need these 2 AtomicBools to make sure that only the first arrive request to write the hashmap, and others need to wait for the first completed.

zhlsunshine:
Hi @howardjohn & @hzxuzhonghu & @nmittler, is there anything that I can do to make it approved? Thanks!

src/state.rs
```diff
@@ -109,6 +114,30 @@ pub struct ResolvedDns {
...
+/// A CachedResolvedDNSMap contains do dns task for hostname via calling resolve_on_demand_dns
+#[derive(serde::Serialize, Default, Debug, Clone)]
+pub struct CachedResolvedDNSMap {
+    crd_hashmap: HashMap<String, Arc<SingleFlight>>,
+}
+
+#[derive(serde::Serialize, Default, Debug)]
+pub struct SingleFlight {
+    should_resolved_dns: AtomicBool,
+    complete_resolved_dns: AtomicBool,
```
hzxuzhonghu, line 126: I donot feel comfortable with the above fields

zhlsunshine: Hi @hzxuzhonghu, I have changed them:
1. change `should_resolved_dns` to `is_resolved_dns_running`
2. change `complete_resolved_dns` to `is_resolved_dns_completed`

zhlsunshine:
> In addition, we can have a test for it

Hi @hzxuzhonghu, sure, I added a concurrency requests test named `build_concurrency_requests_unknown_dest` for this change!

zhlsunshine:
Hi @hzxuzhonghu, I have simplified the code for the DNS resolving optimization by removing 2 atomic bools variables: `is_resolved_dns_running` and `is_resolved_dns_completed`.  Please help to review it, thanks!

zhlsunshine:
Hi @hzxuzhonghu & @howardjohn, I have removed 2 AtomicBool fields `is_resolved_dns_running` and `is_resolved_dns_completed` for `SingleFlight` and use another fields `calling_index` to record the request index, then trigger the DNS resolving task only when `calling_index` is 1 (the first request arriving). Could you please help to review it? Thanks!

src/state.rs
```diff
@@ -341,9 +371,36 @@ impl DemandProxyState {
                     .on_demand_dns_cache_misses
                     .get_or_create(&labels)
                     .inc();
-                // TODO: optimize so that if multiple requests to the same hostname come in at the same time,
+                // optimize so that if multiple requests to the same hostname come in at the same time,
```
hzxuzhonghu, line 374: Is this comment still valid?

src/state.rs
```diff
@@ -456,6 +513,34 @@ impl DemandProxyState {
...
+        cached_atom_bool: SingleFlight,
+    ) {
+        let mut binding = self.state.write().unwrap();
+        let sf_map_element = binding
+            .resolved_dns
+            .crd_hashmap
+            .entry(hostname)
+            .or_insert(cached_atom_bool);
+        // increment 1 for every requests in concurrency
+        sf_map_element.calling_index += 1;
```
hzxuzhonghu, line 528: Can we get rid of this index as we talked before? 

We may need to return the cached SingleFilight in this function

zhlsunshine: Hi @hzxuzhonghu, according to our discussion before, we need to simplify the SingleFlight implementation, therefore, I removed the AtomicBool logic and keep the `Notify`. And as you know, all requests are coming in concurrency and are unorder, we finally have to know the specified request routine which really triggers the DNS resolving task, so the index is necessary.

zhlsunshine: BTW, I also think to use channel similar with golang to implement this, but it's more complicated than directly use i32 index, so this should be the most simplified method.

src/state.rs
```diff
@@ -95,6 +96,9 @@ pub struct ResolvedDnsStore {
     // in a future with support for per-pod DNS resolv.conf settings we may need
     // to change this to a map from source workload uid to resolved IP addresses.
     by_hostname: HashMap<String, ResolvedDns>,
+
+    // crd_hashmap is a map which store IP addresses for unkonwn domain after DNS resolving tasks
+    crd_hashmap: HashMap<String, SingleFlight>,
```
hzxuzhonghu, line 101: nit: rename

btw, when will we cleanup the map

src/state.rs
hzxuzhonghu: Understand this, and i can see the above op is not atomic, so a flag like this is needed

src/state.rs
zhlsunshine: Yes, I removed the `TODO` to show the optimization purpose. Do you think it's necessary? If no, I can remove this comment.

src/state.rs
```diff
@@ -456,6 +513,34 @@ impl DemandProxyState {
             .cloned()
     }
 
+    pub fn set_cached_resolve_dns_for_hostname(
+        &mut self,
+        hostname: String,
+        cached_atom_bool: SingleFlight,
```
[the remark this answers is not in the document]
zhlsunshine: okay, done!

src/state.rs
zhlsunshine: okay, done!
Yeah, that's a good question, I think it should be sync with the hashmap: `by_hostname`.

zhlsunshine: `crd_hashmap`  ---> `sf_by_hostname`

src/state.rs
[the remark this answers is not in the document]
zhlsunshine: `cached_atom_bool` ---> `cached_by_hostname`

zhlsunshine:
Hi @hzxuzhonghu, new submit for answering your question: `when will we cleanup the map?` 
I add `initial_query` and `dns_refresh_rate` for this map which is similar with the hashmap of `by_hostname`. But the refresh rate is 1 hour by default. Please let me know if you have better suggestion. Thanks.! ^_^

src/state.rs
hzxuzhonghu: IC

src/state.rs
```diff
@@ -95,6 +96,9 @@ pub struct ResolvedDnsStore {
     // in a future with support for per-pod DNS resolv.conf settings we may need
     // to change this to a map from source workload uid to resolved IP addresses.
     by_hostname: HashMap<String, ResolvedDns>,
+
+    // sf_by_hostname is a map which store IP addresses for unkonwn domain after DNS resolving tasks
```
hzxuzhonghu, line 100: nit: the comment is not right

hzxuzhonghu (commented):
Need to think about this, tbh i donot think this is a good way

src/state.rs
zhlsunshine: Nice catch! It does not store the IP addresses,  done!

zhlsunshine:
> Need to think about this, tbh i donot think this is a good way

Hi @hzxuzhonghu, there is no standard for the refresh rate,  am I right? I also notice that the expire time is `60` seconds by default for Hashmap of `by_hostname`, should I set by following that?

zhlsunshine:
Hi @hzxuzhonghu, as discussing together, basically the ip addresses would be stored in `by_hostname` once the calling of  `resolve_on_demand_dns` is completed, so it's unnecessary to refresh rate for the elemnts in `sf_by_hostname`.

src/state.rs
```diff
@@ -341,9 +372,39 @@ impl DemandProxyState {
                     .on_demand_dns_cache_misses
                     .get_or_create(&labels)
                     .inc();
-                // TODO: optimize so that if multiple requests to the same hostname come in at the same time,
+                // optimize so that if multiple requests to the same hostname come in at the same time,
                 // we don't start more than one background on-demand DNS task
-                Self::resolve_on_demand_dns(self.to_owned(), workload).await;
+                if state
+                    .get_cached_resolve_dns_for_hostname(&hostname)
+                    .is_none()
```
hzxuzhonghu, line 379: what about else branch? seems you donot handle it

src/state.rs
```diff
@@ -341,9 +372,39 @@ impl DemandProxyState {
...
+                            // Check if this is the first request of doing DNS task for hostname
+                            if sf_map_element.calling_index == 1 {
+                                // doing the dns tasks
+                                Self::resolve_on_demand_dns(self.to_owned(), workload).await;
+                                // need to remove the element in sf_by_hostname because the DNS resolving task
+                                // is completed and the IP addresses already are stored in Hashmap of by_hostname
+                                state.remove_cached_resolve_dns_for_hostname(&hostname);
+                                sf_map_element.notify_waiters();
+                            } else {
+                                sf_map_element.wait_for_notifying();
```
hzxuzhonghu, line 399: I am a little concerned a concurrent call for waiting may be after another call notify, in this case one thread may stuck forever

src/state.rs
```diff
@@ -456,6 +517,43 @@ impl DemandProxyState {
...
+        cached_by_hostname: SingleFlight,
+    ) {
+        let mut binding = self.state.write().unwrap();
+        let sf_map_element = binding
+            .resolved_dns
+            .sf_by_hostname
+            .entry(hostname)
+            .or_insert(cached_by_hostname);
+        // increment 1 for every requests in concurrency
+        sf_map_element.calling_index += 1;
```
hzxuzhonghu, line 532: sf_map_element is cloned, this modify does not affect the original element

src/state.rs
```diff
@@ -109,6 +114,32 @@ pub struct ResolvedDns {
...
+impl SingleFlight {
+    pub fn new() -> Self {
+        SingleFlight {
+            calling_index: 0,
+            wait_for_notification: Arc::new(RwLock::new(Notify::new())),
+        }
+    }
+
+    pub fn wait_for_notifying(&self) {
+        self.wait_for_notification.read().unwrap().notified();
```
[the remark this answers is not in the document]
zhlsunshine: `self.wait_for_notification.read().unwrap().notified()` should suspend the current thread, so except the first request thread would handle the DNS resolving task, the others would get into this and be suspended here.

src/state.rs
zhlsunshine: > what about else branch? seems you donot handle it

Hi @hzxuzhonghu, okay, there should be waiting for the DNS resolving task because I added the remove element logic.

src/state.rs
zhlsunshine: Hi @hzxuzhonghu, this modification does affect the original element because `or_insert` will return a mutating reference of this element in Hashmap.

zhlsunshine: However, I also should add `*` for `sf_map_element`, so it should be like this:
`(*sf_map_element).calling_index += 1;`. Thanks!

src/state.rs
zhlsunshine: Yeah,  I checked it `tokio::sync::Notify` just support the 1 notify vs 1 waiter, I will change it into `tokio::sync::mpsc`.

src/state.rs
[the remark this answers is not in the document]
zhlsunshine: Changed!

zhlsunshine:
Hi @hzxuzhonghu, I have changed the notification mechanism from `tokio::sync::Notify` to `Mutex and Condvar` which can support multiple waiters and single notifier. And the basic logic is not changed.

zhlsunshine:
Hi @hzxuzhonghu any update? Thanks!

src/state.rs
```diff
@@ -341,9 +388,49 @@ impl DemandProxyState {
...
+                            // Check if this is the first request of doing DNS task for hostname
+                            if sf_map_element.calling_index == 1 {
+                                // doing the dns tasks
+                                Self::resolve_on_demand_dns(self.to_owned(), workload).await;
+                                // need to remove the element in sf_by_hostname because the DNS resolving task
+                                // is completed and the IP addresses already are stored in Hashmap of by_hostname
+                                state.remove_cached_resolve_dns_for_hostname(&hostname);
+                                sf_map_element.notify_waiters();
+                            } else {
+                                sf_map_element.wait_for_notifying();
```
hzxuzhonghu, line 415: @zhlsunshine need to confirm whether it can be waken up after a previous notify, otherwise there is a race this thread can be blocking forever

zhlsunshine: Hi @hzxuzhonghu, yes, I confirm that it can be waken up after a previous notify. This is the common usage of A Condition Variable recorded in official sites of Rust [here](https://doc.rust-lang.org/stable/std/sync/struct.Condvar.html#:~:text=Condition%20variables%20are%20typically%20associated,the%20current%20thread%20of%20execution.)

hzxuzhonghu: I think quite the opposite, the example actually does notify after wait

```
#![allow(unused)]
fn main() {
    use std::sync::{Arc, Condvar, Mutex};
    use std::{thread, time};

    let pair = Arc::new((Mutex::new(false), Condvar::new()));
    let pair2 = Arc::clone(&pair);

    // Inside of our lock, spawn a new thread, and then wait for it to start.
    thread::spawn(move || {
        println!("in thread");

        let (lock, cvar) = &*pair2;
        let mut started = lock.lock().unwrap();
        *started = true;
        // We notify the condvar that the value has changed.
        println!("notify");
        cvar.notify_one();
    });

    // Wait for the thread to start up.
    thread::yield_now();
    let ten_millis = time::Duration::from_millis(10);
    thread::sleep(ten_millis);
    let (lock, cvar) = &*pair;
    let mut started = lock.lock().unwrap();
    println!("waiting");
    started = cvar.wait(started).unwrap();
    println!("waiting return");
}

```

Trey this example, you can see it never end

zhlsunshine: Hi @hzxuzhonghu, I have tried above code in my environment, so you care a case which make sure the notifying operation occur before the waiting operation. Yeah, it would be suspended in main thread.  For addressing this, I list 2 solution about it:

1. In our case, the DNS Resolving tasks can not be ahead of the waiting requests, so this can reduce the chance of suspending
2. Using `wait_timeout` instead of `wait`, and the timeout can be 100 milliseconds

Is this okay?

zhlsunshine: Hi @hzxuzhonghu, I have tried above code in my environment, so you create a case which make sure the notifying operation occur before the waiting operation. Yeah, it would be suspended in main thread.  For addressing this, I list 2 solution about it:

1. In our case, the DNS Resolving tasks can not be ahead of the waiting requests, so this can reduce the chance of suspending
2. Using `wait_timeout` instead of `wait`, and the timeout can be 100 milliseconds

Is this okay?

zhlsunshine: Hi @hzxuzhonghu, I have tried above code in my environment, yeah, it would be suspended in main thread. So you create a case which make the notifying operation occurred before the waiting operation. For addressing this, I list 2 solution about it:
1. In our case, the DNS Resolving tasks can not be ahead of the waiting requests, so this can reduce the chance of suspending
2. Using `wait_timeout` instead of `wait`, and the timeout can be 100 milliseconds

Is this okay?

zhlsunshine: Hi @hzxuzhonghu, I use above code you provided and change them based on the solution, the final code is below:

```
#![allow(unused)]
fn main() {
    use std::sync::{Arc, Condvar, Mutex};
    use std::{thread, time};

    let pair = Arc::new((Mutex::new(false), Condvar::new()));
    let pair2 = Arc::clone(&pair);

    // Inside of our lock, spawn a new thread, and then wait for it to start.
    thread::spawn(move || {
        println!("in thread");

        let (lock, cvar) = &*pair2;
        let mut started = lock.lock().unwrap();
        *started = true;
        // We notify the condvar that the value has changed.
        println!("notify");
        cvar.notify_all();
    });

    // Wait for the thread to start up.
    thread::yield_now();
    let ten_millis = time::Duration::from_millis(100);
    thread::sleep(ten_millis);
    let (lock, cvar) = &*pair;
    let mut started = lock.lock().unwrap();
    println!("waiting");
    // Use wait_timeout with a timeout of 100 milliseconds
    let (new_started, _timeout_result) = cvar.wait_timeout(started, ten_millis).unwrap();
    // Check the condition after waiting
    if *new_started {
        println!("Thread started.");
    } else {
        println!("Timeout: Thread did not start.");
    }
    println!("waiting return");
}
```

The output looks like below:
image

hzxuzhonghu: I am not sure this is a good way, and also whether 100ms is the right timeout

zhlsunshine: Hi @hzxuzhonghu, basically, the DNS resolving task can last from microseconds to milliseconds which is depended on many facts, and I think 100 milliseconds should be a conservative value for this.

src/state.rs
```diff
@@ -109,6 +113,54 @@ pub struct ResolvedDns {
...
+        }
+    }
+
+    pub fn wait_for_notifying(&self) {
+        let timeout_millis = std::time::Duration::from_millis(100);
+        let lock = self.mutex.lock().unwrap();
+
+        // Must define _unused to avoid immediately dropping of the lock and use wait_timeout with a timeout of 100 milliseconds
+        let (_is_started, _timeout_result) =
+            self.condvar.wait_timeout(lock, timeout_millis).unwrap();
```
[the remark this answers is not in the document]
hzxuzhonghu: is there a similar condvar in tokio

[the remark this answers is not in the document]
zhlsunshine: Hi @nmittler [here ](https://github.com/istio/ztunnel/pull/672#discussion_r1355979606) is the exact place to record why we need to use `wait_timeout`. Hope this can be helpful.

src/state.rs
```diff
@@ -109,6 +113,54 @@ pub struct ResolvedDns {
...
+    }
+
+    pub fn notify_waiters(&self) {
+        println!("Notifier notifying...");
```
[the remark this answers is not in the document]
zhlsunshine: okay, done!

src/state.rs
```diff
@@ -109,6 +113,54 @@ pub struct ResolvedDns {
     dns_refresh_rate: std::time::Duration,
 }
 
+#[derive(serde::Serialize, Default, Debug)]
+pub struct SingleFlight {
```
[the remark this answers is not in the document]
zhlsunshine: sure, add the comments for this struct to describe its function.

src/state.rs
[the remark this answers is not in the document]
nmittler: @zhlsunshine I'm not sure that thread addresses my general concern of blocking a tokio thread for this. Using a timeout still means that a single invocation is blocking the thread for up to 100 milliseconds, which is a very long time. If multiple requests come in at the same time, then we could effectively take down the tokio runtime. I suspect we should be using tokio synchronization primitives for this.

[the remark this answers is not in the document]
howardjohn: Yeah this is broken; it will block the entire thread. We only have 2 threads. Async code shouldn't block

howardjohn (changes requested):

src/state.rs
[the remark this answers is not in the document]
zhlsunshine: Thanks @nmittler & @howardjohn for your comments and information sharing!  Yeah, it's fairly coarse by using `std::sync::Mutex`, and `tokio::sync::Notify` may result in spurious wakeups for abnormal cases especially for unexpected pending notification. How about to use `tokio::sync::Notify`, and remove the `Notify` instance from `sf_by_hostname` once current DNS resolving task completed. Moreover, the waiters will register the waiting list before the notifier because of the DNS resolving task.  Please check the later PR.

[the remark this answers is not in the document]
zhlsunshine: Thanks @nmittler & @howardjohn for your comments and information sharing!  Yeah, it's fairly coarse by using `std::sync::Mutex`, and `tokio::sync::Notify` may result in spurious wakeups for abnormal cases especially for unexpected pending notification. How about to use `tokio::sync::Notify`, and remove the `Notify` instance from `sf_by_hostname` once current DNS resolving task completed. Moreover, the waiters will register the waiting list before the notifier because of the DNS resolving task.  Please help to review the later PR.

zhlsunshine:
Hi @nmittler, @howardjohn and @hzxuzhonghu,  PTAL this new commits, thanks!

src/state.rs
```diff
@@ -109,6 +114,47 @@ pub struct ResolvedDns {
...
+    }
+}
+
+impl SingleFlight {
+    pub fn new() -> Self {
+        SingleFlight {
+            // is_first_req_in is set to true because
+            // there always is the first request
+            is_first_req_in: true,
+            wait_for_notification: Arc::new(RwLock::new(Notify::new())),
```
nmittler, line 145: I don't think we need a lock for this.

See the example usage: https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html#examples-5.

nmittler (commented):
Even though you switched to tokio structures, I still don't think this approach will work. You're still performing the wait within the `ProxyState`, which is itself locked by a mutex. This will still cause thread blocking because we're using a regular `sync.Mutex` (which I don't think we should change).

Instead, perhaps `find_desination` should return a future? Then the caller can unlock the `ProxyState` mutex before waiting on the result.

src/state.rs
zhlsunshine: Yeah, it's unnecessary to use this lock. I removed it.

zhlsunshine:
> Even though you switched to tokio structures, I still don't think this approach will work. You're still performing the wait within the `ProxyState`, which is itself locked by a mutex (see the call site). This will still cause thread blocking because we're using a regular `sync.Mutex` (which I don't think we should change).
> 
> Instead, perhaps `find_desination` should return a future? Then the caller can unlock the `ProxyState` mutex before waiting on the result.

Hi @nmittler, I understand the `locked` is for both `get_cached_resolve_dns_for_hostname` and `set_cached_resolve_dns_for_hostname`. If yes, this should be necessary if we want to implement the multiple waiters and single notifier, am I correct. And these 2 `ProxyState` lock operations are similar to `set_ips_for_hostname` and `get_ips_for_hostname`, all these locks are operations based on specified hostname. Should they be reasonable?

zhlsunshine:
> Even though you switched to tokio structures, I still don't think this approach will work. You're still performing the wait within the `ProxyState`, which is itself locked by a mutex (see the call site). This will still cause thread blocking because we're using a regular `sync.Mutex` (which I don't think we should change).
> 
> Instead, perhaps `find_desination` should return a future? Then the caller can unlock the `ProxyState` mutex before waiting on the result.

Hi @nmittler,  I changed the `wait_for_notifying` to an async function, and call it with `await`, I believe that `wait_for_notifying` would yield to executor once it is blocked, therefore, the whold thread would not be blocked, is it okay?

src/state.rs
```diff
@@ -341,9 +387,51 @@ impl DemandProxyState {
                     .on_demand_dns_cache_misses
                     .get_or_create(&labels)
                     .inc();
-                // TODO: optimize so that if multiple requests to the same hostname come in at the same time,
+                // optimize so that if multiple requests to the same hostname come in at the same time,
                 // we don't start more than one background on-demand DNS task
-                Self::resolve_on_demand_dns(self.to_owned(), workload).await;
+                if state
```
nmittler, line 392: I think we need to simply all of this ... it's getting pretty complex. It might be cleaner to separate out the DNS resolution logic into a separate, dedicated structure. For example:

```rust
struct DnsResolver {
  state: Arc<RwLock<ProxyState>>,
  notifyMap: Mutex<HashMap<string,Arc<Notify>>
}

impl DnsResolver {
  async fn resolve_host(&self, host) Option<ResolvedDns>{
    // If we don't have the addresses (i.e. get_ips_for_hostname)
    // then:
    //     Get or create a `Notify` for the host in `notifyMap`.
    //     If created:
    //         resolve DNS and notify waiters
    //     else:
    //         wait for notify
    // Return the addresses (i.e. get_ips_for_hostname)
  }
}

```

Then the code for `loadbalance_for_hostname` would be reduced to something like:


```rust
let rdns = match self.dnsResolver.resolve_host(&workload.hostname)
...
```

src/state.rs
```diff
@@ -95,6 +96,10 @@ pub struct ResolvedDnsStore {
     // in a future with support for per-pod DNS resolv.conf settings we may need
     // to change this to a map from source workload uid to resolved IP addresses.
     by_hostname: HashMap<String, ResolvedDns>,
+
+    // sf_by_hostname is a map which store the hostname and its SingleFlight properties to provides
+    // a duplicate function call suppression mechanism for DNS resolving tasks.
+    sf_by_hostname: HashMap<String, SingleFlight>,
```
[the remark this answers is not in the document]
zhlsunshine: Hi @nmittler, The reason why I put `sf_by_hostname` in `state` which is an `Arc<RwLock<ProxyState>>` field of `DemandProxyState` is that `sf_by_hostname` hashmap should be similar to `by_hostname` hashmap in `ResolvedDnsStore` to re-use the RwLock to read or write data in safe way in concurrency. So I think it could not be a member of `DemandProxyState`.

src/state.rs
```diff
@@ -109,6 +114,47 @@ pub struct ResolvedDns {
     dns_refresh_rate: std::time::Duration,
 }
 
+/// A SingleFlight can make sure that only the first request to really handle DNS resolving task
+/// for specified hostname and other requests in concurrency need to wait for the task completion.
+#[derive(serde::Serialize, Default, Debug)]
+pub struct SingleFlight {
```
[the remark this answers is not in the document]
zhlsunshine: Hi @nmittler, Yeah, it's a good idea, I make the whole struct of `ResolvedDnsStore` as non-pub struct. Done!

src/state.rs
zhlsunshine: Hi @nmittler, thanks for your suggestion and I also followed your idea to change the code to make it simple. However, I find that the function `resolve_on_demand_dns` which really do the DNS resolving task must be done inside the `DemandProxyState` because it refers to many data owned by `DemandProxyState`, such as insert the resolved ip addresses into HashMap of `by_hostname` which is belonged to `DemandProxyState`.

Any idea?

nmittler: Can you just pass in a function reference? Basically, `DnsResolver` just needs some function that actually resolves DNS.

src/state.rs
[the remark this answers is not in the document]
nmittler: I understand that there will have to be a lock, but it's probably fine if we have another lock for this. That's why I suggested a separate structure that just handles resolution.

Having this here is polluting the simplicity of ProxyState. SingleFlight and the whole resolution process is a feature of DemandProxyState.

src/state.rs
nmittler: I'm open to other opinions on this @howardjohn @hzxuzhonghu

zhlsunshine: > Can you just pass in a function reference? Basically, `DnsResolver` just needs some function that actually resolves DNS.

Okay, l will try it out, I understand your mean is that move all checking logic into `DnsResolver`, then call `resolve_on_demand_dns` of `DemandProxyState`,  am I right?

zhlsunshine: > Can you just pass in a function reference? Basically, `DnsResolver` just needs some function that actually resolves DNS.

Okay, @nmittler, l will try it out, I understand your mean is that move all checking logic into `DnsResolver`, then call `resolve_on_demand_dns` of `DemandProxyState`,  am I right?

zhlsunshine: > Can you just pass in a function reference? Basically, `DnsResolver` just needs some function that actually resolves DNS.

Okay, @nmittler, I understand your mean is that move all checking logic into `DnsResolver`, then call `resolve_on_demand_dns` of `DemandProxyState`,  am I right?

zhlsunshine: > I think we need to simply all of this ... it's getting pretty complex. It might be cleaner to separate out the DNS resolution logic into a separate, dedicated structure. For example:
> 
> ```rust
> struct DnsResolver {
>   state: Arc<RwLock<ProxyState>>,
>   notifyMap: Mutex<HashMap<string,Arc<Notify>>
> }
> 
> impl DnsResolver {
>   async fn resolve_host(&self, host) Option<ResolvedDns>{
>     // If we don't have the addresses (i.e. get_ips_for_hostname)
>     // then:
>     //     Get or create a `Notify` for the host in `notifyMap`.
>     //     If created:
>     //         resolve DNS and notify waiters
>     //     else:
>     //         wait for notify
>     // Return the addresses (i.e. get_ips_for_hostname)
>   }
> }
> ```
> 
> Then the code for `loadbalance_for_hostname` would be reduced to something like:
> 
> ```rust
> let rdns = match self.dnsResolver.resolve_host(&workload.hostname)
> ...
> ```

Hi @nmittler, I added more description in `load_balance_for_hostname`, so do you mean I need to move the logic of `Err` branch into another function?

zhlsunshine:
Hi @howardjohn & @hzxuzhonghu, what's your opinions about this PR? Thanks!

src/state.rs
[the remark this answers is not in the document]
zhlsunshine: Hi @nmittler, thanks for more description here, I'm working on this and submit another commit later

src/state.rs
zhlsunshine: Hi @nmittler, I made a big change based on your command, all DNS resolving relates are now in a struct `DnsResolver`. Please help to review again, thanks Mittler!

zhlsunshine:
Hi @howardjohn, @hzxuzhonghu and @nmittler, there is a change for this PR based on @nmittler 's suggestion, creating a new struct named `DnsResolver` to holds all DNS resolving relates to simplify the `ProxyState`, and `DnsResolver` is part feature of `DemandProxyState`. Could you please help to review it again? Thanks in advance!

src/state.rs
zhlsunshine: Hi @nmittler, I made a change based on your command, all DNS resolving relates are now in a struct `DnsResolver`. Please help to review again, thanks Mittler!

src/state.rs
```diff
@@ -71,34 +72,35 @@ impl fmt::Display for Upstream {
...
-    #[serde(flatten)]
-    pub resolved_dns: ResolvedDnsStore,
+    #[serde(skip_serializing)]
+    dns_resolver_opts: ResolverOpts,
 }
 
 /// A ResolvedDnsStore encapsulates all resolved DNS information for workloads in the mesh
-#[derive(serde::Serialize, Default, Debug)]
-pub struct ResolvedDnsStore {
+#[derive(serde::Serialize, Default, Debug, Clone)]
```
nmittler, line 89: why Clone?

src/state.rs
```diff
@@ -109,6 +111,244 @@ pub struct ResolvedDns {
...
+        self.wait_for_notification.notified().await;
+    }
+
+    fn notify_waiters(&self) {
+        self.wait_for_notification.notify_waiters();
+    }
+}
+
+impl DnsResolver {
+    pub fn new(dns_resolver_cfg: ResolverConfig, dns_resolver_opts: ResolverOpts) -> Self {
```
nmittler, line 156: remove `pub` on all methods? The struct is not public.

src/state.rs
```diff
@@ -109,6 +111,244 @@ pub struct ResolvedDns {
     dns_refresh_rate: std::time::Duration,
 }
 
+/// A SingleFlight can make sure that only the first request to really handle DNS resolving task
+/// for specified hostname and other requests in concurrency need to wait for the task completion.
+#[derive(serde::Serialize, Default, Debug)]
+struct SingleFlight {
+    // is_first_req_in can record requests number and it can be used to
+    // check the request number, such as the first request.
+    is_first_req_in: bool,
```
nmittler, line 120: I think I mentioned in a previous review, but I don't think you need this flag since we have a lock around this structure. Instead just lock, check if a single flight already exists, if so use it, if not add a new one. Does that work?

src/state.rs
```diff
@@ -333,20 +570,17 @@ impl DemandProxyState {
         let workload_uid = workload.uid.to_owned();
         let hostname = workload.hostname.to_owned();
         metrics.as_ref().on_demand_dns.get_or_create(&labels).inc();
-        let rdns = match state.get_ips_for_hostname(&workload.hostname) {
+        let rdns = match state.dns_resolver.get_ips_for_hostname(&hostname) {
```
nmittler, line 573: Can this entire method (load_balance_for_hostname) just be moved to `DnsResolver`? All of this should be internal details of it, I think.

src/state.rs
zhlsunshine: Hi @nmittler, as the definition of `resolved_dns: Arc<RwLock<ResolvedDnsStore>>`, so `ResolvedDnsStore` must declare the default clone to support concurrency safe accessing.

src/state.rs
zhlsunshine: okay, remove `pub`, done!

src/state.rs
zhlsunshine: Hi @nmittler, as you know `DnsResolver` is part of `DemandProxyState`, and it needs to modify itself on its 2 HashMaps, so it's mutable reference. If moving `load_balance_for_hostname` into `DnsResolver`, it would impact that the whole `DemandProxyState` need to be a mutable reference as well. So hope it can be as it is now. Does it make sense?

src/state.rs
zhlsunshine: Hi @nmittler, this flag should be here because `checking the index of the request` and `doing the dns resolving task` are not at the same time. When multiple requests get into `resolve_host` in concurrency, the flag must be set to distinguish the first request or other request.

src/state.rs
nmittler: I was actually concerned when I saw that [DemandProxyState.resolve_on_demand_dns](https://github.com/istio/ztunnel/blob/1b20547a865b70f90492ffb18cbe3fb86e3beea8/src/state.rs#L368) accepting a non-reference `self` just so that it can switch to a mutable reference. Not sure when that was introduced, but it seems a bit hacky and should probably be fixed if possible. Perhaps the issue is that we have some methods that needlessly require `&mut self`(e.g. if you lock an object that you write to, you don't need mut)?

@howardjohn I may be off here, since my rust is a little rusty at this point :) ... thoughts?

zhlsunshine:
Hi @nmittler, as we discussed, it becomes more simplicity.

src/state.rs
zhlsunshine: done after the discussion.

src/state.rs
```diff
@@ -109,6 +113,178 @@ pub struct ResolvedDns {
...
+    }
+
+    async fn resolve_host(
+        &mut self,
+        hostname: &String,
+        workload: &Workload,
+    ) -> Option<ResolvedDns> {
+        // optimize so that if multiple requests to the same hostname come in at the same time,
+        // we don't start more than one background on-demand DNS task
+        let is_the_first_req = self.get_singleflight_by_hostname(hostname.to_owned());
```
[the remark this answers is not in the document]
zhlsunshine: Hi @nmittler, there are 2 mutable borrows occurs in both `self.get_or_create_notify` (first mutable borrow) and `self.resolve_on_demand_dns` (second mutable borrow), so I have to use this `let element = self.get_cached_resolve_dns_for_hostname(hostname);` to avoid this problem. 
However, I can change `get_singleflight_by_hostname` into `get_or_create_notify`.  Does it make sense?

src/state.rs
```diff
@@ -71,34 +72,37 @@ impl fmt::Display for Upstream {
...
-#[derive(serde::Serialize, Default, Debug)]
-pub struct ResolvedDnsStore {
+#[derive(serde::Serialize, Default, Debug, Clone)]
+struct ResolvedDnsStore {
     // by_hostname is a map from hostname to resolved IP addresses for now.
     //
     // in a future with support for per-pod DNS resolv.conf settings we may need
     // to change this to a map from source workload uid to resolved IP addresses.
+    #[serde(skip_serializing)]
     by_hostname: HashMap<String, ResolvedDns>,
```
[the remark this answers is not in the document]
zhlsunshine: I can change it, how about `resolved_hostname`?

nmittler:
@zhlsunshine I created https://github.com/istio/ztunnel/pull/722, which re-works this a bit based on some things we discussed. PTAL.

zhlsunshine:
> @zhlsunshine could we step back for a moment and discuss the motivation for this PR? Are you seeing a real issue with DNS performance, specifically due to concurrent requests for the same host? I just want to make sure we're solving a real problem here.

Hi @nmittler, frankly, I'm focusing on some validations of ambient mesh in CSP environment in recent, what I'm sure about is that the original code has the problem: there may be multiple DNS resolving for the same hostname in concurrent requests.  
BTW, I have no more comment for your rework [PR#722](https://github.com/istio/ztunnel/pull/722).

zhlsunshine:
> @zhlsunshine could we step back for a moment and discuss the motivation for this PR? Are you seeing a real issue with DNS performance, specifically due to concurrent requests for the same host? I just want to make sure we're solving a real problem here.

Hi @nmittler, frankly, I'm focusing on some validations of ambient mesh in CSP environment in recent, what I'm sure about is that the original code has the problem: there may be multiple DNS resolving for the same hostname in concurrent requests.  
BTW, I have no more comment for your rework [PR#722](https://github.com/istio/ztunnel/pull/722). And I also insists on that you should be a co-author of this PR thanks to your help and comments. ^_^

nmittler:
@zhlsunshine just to be clear, you haven't actually seen performance problems in a deployed system? My concern is fundamentally about the complexity of the solution. I'd like to see that the problem is real before we optimize for it.

@howardjohn @hzxuzhonghu any thoughts?

howardjohn:
I agree with Nate

zhlsunshine:
> @zhlsunshine just to be clear, you haven't actually seen performance problems in a deployed system? My concern is fundamentally about the complexity of the solution. I'd like to see that the problem is real before we optimize for it.
> 
> @howardjohn @hzxuzhonghu any thoughts?

Hi @nmittler, yes, I haven't. But multiple requests with the same hostname given in concurrency and whose resolved IP addresses are not in the DNS resolved cache, then this situation leads to the same number of times of DNS resolving operations for the same hostname, it's really waste, so this optimization should be valid even we do not test the performance for it.

nmittler:
@zhlsunshine 

> Hi @nmittler, yes, I haven't. But multiple requests with the same hostname given in concurrency and whose resolved IP addresses are not in the DNS resolved cache, then this situation leads to the same number of times of DNS resolving operations for the same hostname, it's really waste, so this optimization should be valid even we do not test the performance for it.

I understand what you're trying to optimize for. What I'm suggesting is that it may not be a huge issue in practice. I'd rather not add this type of complexity until we actually have a customer that is being impacted.

howardjohn:
One thing to note is the behavior currently is equal to without DNS proxy.
So it's not like this is fixing a regression, purely a performance benefit
, and the intent of the proxy is not performance.

zhlsunshine:
Hi @howardjohn, besides the performance change, we also did some code refactor for the DNS resolver to make it more clear, does it make sense?

howardjohn:
https://github.com/istio/ztunnel/pull/1098

PR #680, opened 2023-09-14
fix : make ambient work with intermidiary certs

yuval-k:
make the test certs logic resemble more what the control plane does, and add a test that the certs loaded correctly.

This solves two problems:
- If the cert-chain has multiple certs. using `stack_from_pem` instead of `from_pem` solves that.
- Sometimes the user might add the root cert to the chain as well. While technically a user error , we should be forgiving, to prevent user frustration. So I added `dedup_by` that removes to adjacent certs that are the same.  

Fixes [istio/istio#46995](https://github.com/istio/istio/issues/46995).


src/tls/boring.rs
```diff
@@ -65,7 +66,17 @@ pub fn cert_from(key: &[u8], cert: &[u8], chain: Vec<&[u8]>) -> Certs {
...
     let chain = chain
         .into_iter()
-        .map(|pem| ZtunnelCert::new(x509::X509::from_pem(pem).unwrap()))
+        .flat_map(|pem| x509::X509::stack_from_pem(pem).unwrap())
+        .dedup_by(|x, y| {
+            match (
+                x.digest(MessageDigest::sha256()),
+                y.digest(MessageDigest::sha256()),
+            ) {
+                (Ok(x), Ok(y)) => x.as_ref() == y.as_ref(),
```
hzxuzhonghu, line 75: do we  need to dedup, when can this happen

yuval-k: I mentioned that in the PR description
> Sometimes the user might add the root cert to the chain as well. While technically a user error , we should be forgiving, to prevent user frustration. So I added dedup_by that removes to adjacent certs that are the same.

yuval-k: note that if you don't dedupe them you get this error:
```
2023-09-14T15:17:26.046203Z  WARN proxy{uid=9be91780-225e-4ae1-a553-4cd9a0b8617b}:outbound{id=849484871dbcc962fd9f4ca41950145e}: ztunnel::proxy::outbound: failed dur=591.669µs err=tls error: invalid operation: ErrorStack([Error { code: 184549481, library: "X.509 certificate routines", function: "OPENSSL_internal", reason: "CERT_ALREADY_IN_HASH_TABLE", file: "../crypto/x509/x509_lu.c", line: 356 }])
```

stevenctl (approved):

PR #681, opened 2023-09-15
Add support for HTTP/1.1, negotiated by ALPN

howardjohn:


yuval-k:
Hi @howardjohn, What's the context/background for this change? I may have missed it, but haven't seen anything mentioned in the weekly meeting notes.

I'm asking as given that both envoy (and other modern proxies/servers) support h2, I'm not sure if it is worth the extra complexity to add h1 support.
This means that we have 2 protocols with different buffering and connection semantics to tune and optimize now and down the road.

howardjohn:
> Hi @howardjohn, What's the context/background for this change? I may have missed it, but haven't seen anything mentioned in the weekly meeting notes.
> 
> I'm asking as given that envoy (and other modern proxies/servers) support h2, I'm not sure if it is worth the extra complexity to add h1 support. This means that we have 2 protocols with different buffering and connection semantics to tune and optimize now and down the road.

The motivations were compatibility and performance. However, our possible compatibility use cases haven't come to fruition yet and I haven't had time to test/prove performance benefits at all. Lets keep this on the backburner for now unless its value becomes more apparent

PR #682, opened 2023-09-16
Add port for hbone single TLS

keithmattix:
ztunnel version of istio/istio#46945

### DIFF
--- a/proto/workload.proto
+++ b/proto/workload.proto
@@ -220,7 +220,7 @@ message GatewayAddress {
   // port to reach the gateway at for mTLS HBONE connections
   uint32 hbone_mtls_port = 3;
   // port to reach the gateway at for single tls HBONE connections
-  // used for traffic originating outside the mesh reaching a waypoint-enabled destination
+  // used for sending unauthenticated traffic originating outside the mesh to a waypoint-enabled destination
   optional uint32 hbone_single_tls_port = 4;
 }
 
--- a/proto/workload.proto
+++ b/proto/workload.proto
@@ -217,8 +217,11 @@ message GatewayAddress {
     NamespacedHostname hostname = 1;
     NetworkAddress address = 2;
   }
-  // port to reach the gateway at
-  uint32 port = 3;
+  // port to reach the gateway at for mTLS HBONE connections
+  uint32 hbone_mtls_port = 3;
+  // port to reach the gateway at for single tls HBONE connections
+  // used for traffic originating outside the mesh reaching a waypoint-enabled destination
+  optional uint32 hbone_single_tls_port = 4;
 }
 
 // NetworkAddress represents an address bound to a specific network.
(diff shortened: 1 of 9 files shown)


proto/workload.proto
```diff
@@ -217,8 +217,11 @@ message GatewayAddress {
     NamespacedHostname hostname = 1;
     NetworkAddress address = 2;
   }
-  // port to reach the gateway at
-  uint32 port = 3;
+  // port to reach the gateway at for mTLS HBONE connections
+  uint32 hbone_mtls_port = 3;
+  // port to reach the gateway at for single tls HBONE connections
+  // used for traffic originating outside the mesh reaching a waypoint-enabled destination
```
bleggett, line 223:
```suggestion
  // used for sending unauthenticated traffic originating outside the mesh to a waypoint-enabled destination
```

src/admin.rs
```diff
@@ -675,14 +675,16 @@ mod tests {
                     network: "defaultnw".to_string(),
                     address: [127, 0, 0, 10].to_vec(),
                 })),
-                port: 15008,
+                hbone_mtls_port: 15008,
+                hbone_single_tls_port: Some(15006),
```
bleggett, line 679:
```suggestion
                hbone_single_tls_port: Some(15006),
```

Is it smart to recontextualize this (existing) port we use for plaintext?

I get that we should treat single-tls as untrusted, but should we dual-purpose the port already used for no-tls?

Should we just pick a new port?

src/proxy/inbound.rs
```diff
@@ -656,7 +657,8 @@ mod test {
                 namespace: "gatewayns".to_string(),
                 hostname: "gateway".to_string(),
             }),
-            port: 15006,
+            hbone_mtls_port: 15006,
+            hbone_single_tls_port: Some(15006),
```
bleggett, line 661: Maybe dumb q - why are these the same here?

src/admin.rs
keithmattix: My understanding is what the waypoint didn't have a plaintext port

src/proxy/inbound.rs
keithmattix: Lol good catch; typo

src/admin.rs
bleggett: it doesn't, but ztunnel does, and already uses 15006 for this purpose, so it's a bit confusing to repurpose that same port for this single-tls hairpin functionality, versus picking a new dedicated port for this.

(there might be an argument here that I'm missing tho)

keithmattix: Hmm not a bad idea. 15009?

bleggett: > Hmm not a bad idea. 15009?

Or [15003 to give us an "empty slot" on either side for future use](https://github.com/istio/ztunnel/blob/master/ARCHITECTURE.md), but yeah whatever you prefer/others have opinions on.

bleggett (approved):
LGTM cc @istio/wg-networking-maintainers-ztunnel

stevenctl (approved):

PR #684, opened 2023-09-19
remove use of new_socket_addr() in dns server tests

ilrudie:
Removes `new_socket_addr()` which creates a race condition that can cause test flakes.

`new_socket_addr()` binds to a free port, captures the resulting address and returns it but then drops the resulting socket. If the socket isn't cleaned up prior to `Server::new(...)` attempting to bind (or another concurrent test has bound to that port) it can cause a panic resulting in an error like:

```
thread 'test_tcp_request' panicked at 'called `Result::unwrap()` on an `Err` value: failed to bind to address 127.0.0.1:45366: Address already in use (os error 98)', /home/prow/go/src/istio.io/ztunnel/src/test_helpers/app.rs:73:10
```

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -716,7 +716,7 @@ mod tests {
     use crate::metrics;
     use crate::test_helpers::dns::{
         a, aaaa, cname, ip, ipv4, ipv6, n, new_message, new_tcp_client, new_udp_client,
-        send_request, server_request, socket_addr,
+        send_request, server_request,
     };
     use crate::test_helpers::helpers::subscribe;
     use crate::test_helpers::{new_proxy_state, test_default_workload};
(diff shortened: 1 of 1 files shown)


src/dns/server.rs
```diff
@@ -1193,12 +1201,20 @@ mod tests {
...
         let domain = "cluster.local".to_string();
-        let addr = new_socket_addr().await;
         let state = state();
         let forwarder = Arc::new(SystemForwarder::new().unwrap());
-        let server = Server::new(domain, addr, NW1, state, forwarder, test_metrics())
-            .await
-            .unwrap();
+        let server = Server::new(
+            domain,
+            SocketAddr::from_str("127.0.0.1:0")
```
howardjohn, line 1208: nit: I think you can do something like `SocketAddr::from(([127,0,0,1], 0))` or similar but I forget the syntax

howardjohn (approved):

src/dns/server.rs
ilrudie: Great point. It's marginally less readable but because it's from simpler types it can be compile-time checked. I like that more but I missed the comment before it auto-merged. I'll do a quick follow up.

PR #689, opened 2023-09-24
minor optimize fetch upstream on demand

wulianglongrd:
In the best case, we can avoid calling `self.state.read().unwrap().find_address` twice in the `fetch_address` function.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -490,7 +490,7 @@ impl DemandProxyState {
     // only support workload
     pub async fn fetch_workload_by_uid(&self, uid: &str) -> Option<Workload> {
         // Wait for it on-demand, *if* needed
-        debug!(%uid, "fetch workload");
+        debug!(%uid, "fetch workload by uid");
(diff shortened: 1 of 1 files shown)


src/state.rs
```diff
@@ -490,7 +490,7 @@ impl DemandProxyState {
     // only support workload
     pub async fn fetch_workload_by_uid(&self, uid: &str) -> Option<Workload> {
         // Wait for it on-demand, *if* needed
-        debug!(%uid, "fetch workload");
+        debug!(%uid, "fetch workload by uid");
```
wulianglongrd, line 493: This is to avoid confusion with the `fetch_workload` function. Currently, both functions print the same log.

howardjohn (changes requested):
fetch_on_demand should never be called with the lock. This is making a network call that may take an unbounded amount of time.

If we have other places with the lock held we should fix them (possibly assert in fetch_on_demand it is not locked)

PR #690, opened 2023-09-25
xds: pass ISTIO_METAJSON_* env vars in node matadata

shashankram:
ISTIO_METAJSON_* env vars are parsed and passed along in the xDS node metadata. This can include labels and annotations. The env var JSON values are unmarshalled to the proto Values/Structs.

Testing done:
Set the `ISTIO_METAJSON_LABELS='{"app":"foo","version":"v1"}'` env var and verified it gets relayed to Istiod's `NodeMetadata.Labels`

### DIFF
--- a/src/xds/client.rs
+++ b/src/xds/client.rs
@@ -20,6 +20,7 @@ use std::{fmt, mem};
 use prost::{DecodeError, EncodeError};
 use prost_types::value::Kind;
 use prost_types::{Struct, Value};
+use serde_json;
 use thiserror::Error;
 use tokio::sync::mpsc;
 use tokio::sync::oneshot;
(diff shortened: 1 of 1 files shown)


src/xds/client.rs
```diff
@@ -447,6 +479,19 @@ impl AdsClient {
             .fields
             .append(&mut Self::build_struct(self.config.proxy_metadata.clone()).fields);
 
+        // Lookup ISTIO_METAJSON_* environment variables and add them to the node metadata
+        for (key, val) in std::env::vars().filter(|(key, _)| key.starts_with(ISTIO_METAJSON_PREFIX))
```
stevenctl, line 483: nit: the other `env` reading is done in `config.rs`. Should this go there for consistency and be passed in on another field `metadata_json`?

howardjohn:
What is the motivation?

stevenctl (approved):

shashankram:
> What is the motivation?

Attributes like labels and annotations can be relayed via the node metadata by using the `ISTIO_METAJSON_*` env var. This is used by Istiod, for e.g. in the workload autoregistration controller. Would make the UX consistent with sidecars when these env vars are set.

src/xds/client.rs
shashankram: Couldn't add it to ProxyConfig because of prost types not implementing some traits. That, and it made sense to put it in `node()` because that's where we are reading node metadata other than proxy_metadata, such as name, namespace, instance IPs, etc.

PR #716, opened 2023-10-24
bump dependency

hzxuzhonghu:


howardjohn (approved):

hzxuzhonghu:
@hanxiaop Why does it need to be backported

hanxiaop:
> @hanxiaop Why does it need to be backported

There are some errors when updating common files of 1.20 ztunnel repo. See https://github.com/istio/ztunnel/pull/731#issuecomment-1818150804. Or should we only fix the usage?

hzxuzhonghu:
We can fix the lint.

hanxiaop:
> We can fix the lint.

Sounds good, I'll revise

PR #722, opened 2023-11-03
Avoid redundant DNS requests

nmittler:
This modifies the approach described in https://github.com/istio/ztunnel/pull/672. Attempting to clarify some things we had discussed offline.


src/state.rs
```diff
@@ -591,6 +443,183 @@ impl DemandProxyState {
...
 }
 
+/// A Dns Resolver is responsible for the DNS resolving task for given hostnames
+#[derive(Default, Debug, Clone)]
+struct DnsResolver {
+    // Map of resolved hostnames.
+    resolved: Arc<RwLock<HashMap<String, ResolvedDns>>>,
+
+    // Map of in-progress resolution requests.
+    in_progress: Arc<Mutex<HashMap<String, Arc<Notify>>>>,
```
zhlsunshine, line 453: I'm thinking that: is it possible that putting both `resolved` and `in_progress` in a struct such as `ResolvedDnsStore`? Then we just need to define `resolved_dns: Arc<RwLock<ResolvedDnsStore>>` as a field in `DnsResolver`.

nmittler: I kept them separate initially, because of how they're used. The `resolved` map is the fast-path and shouldn't need a write lock for the majority cases (once some amount of resolution has been done). Accessing `in_progress` needs a write lock every time it's accessed, because it's a "get or create" operation.  That said, you could benchmark it with one lock vs 2 and see if it makes much of a difference in practice.

src/state.rs
```diff
@@ -591,6 +443,183 @@ impl DemandProxyState {
...
+            // Wait for the in-progress resolution to complete.
+            n.notified().await;
+        }
+
+        // At this point, resolution has completed. Just serve from local
+        // cache.
+        self._find_resolved_host(&hostname)
+    }
+
+    fn _get_or_create_notify(&self, hostname: &String) -> (Arc<Notify>, bool) {
```
zhlsunshine, line 530: I think rust has explicit property for function, such as `pub` key word, so I prefer to remove `_` for this private function.

src/state.rs
```diff
@@ -591,6 +443,183 @@ impl DemandProxyState {
...
+            TokioHandle,
+        );
+        if resolver_result.is_err() {
+            warn!(
+                "system dns async resolution: error creating resolver for workload {} is: {:?}",
+                &workload_uid, resolver_result
+            );
+            return;
+        }
+        let r = resolver_result.unwrap();
```
ymesika, line 571: Is this safe?

nmittler:
Closing this, since it was just meant for discussion.

PR #726, opened 2023-11-16
dependendencies: bump to hyper 1.0.0

howardjohn:
Two forks left: 
* https://github.com/cloudflare/boring/pull/194
* https://github.com/davidpdrsn/tower-hyper-http-body-compat/pull/7 OR https://github.com/hyperium/tonic/issues/1307

This trades forks on: boring, hyper-util
For forks on: boring, tower-hyper-http-body-compat

So technically its a 1:1 tradeoff, but this is more future proof and http-body-compat is much smaller than hyper-util. IMO this is good to go for now

Fixes https://github.com/istio/ztunnel/issues/805

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,5 @@
 [build]
 target-dir = "out/rust"
 [env]
-BORING_BSSL_PATH = { value = "vendor/boringssl-fips/linux_x86_64", force = true, relative = true }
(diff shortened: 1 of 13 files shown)


ilrudie:
Hyper boring was superseded by https://github.com/cloudflare/boring/pull/194 so the closed PR shown right now isn't a blocker.

stevenctl (approved):

stevenctl:
What's the reasoning for/against `x.y` vs `x.y.z` versions of deps? Just want to rely on lockfile for pinning now?

Issue #729, opened 2023-11-16
floating forget_certificate function


GregHanson:
`forget_certificate()` was added to SecretManager in https://github.com/istio/ztunnel/pull/421, but as far as I can tell, the method was never called outside of tests.

It's mentioned a few times in comments in src/identity/manager.rs but that's it.  Was it logic around it added and then removed in different follow-ups?  Some pending design work?

GregHanson:
cc @qfel

GregHanson:
related https://github.com/istio/ztunnel/issues/351

GregHanson:
for the ProxyState, on workload add/updates, there is always a delete followed by an insert:
https://github.com/istio/ztunnel/blob/master/src/xds.rs#L104C1-L122C52

should forget_certficates be called be handled the same?

Issue #351, opened 2023-01-23
certificates: stop trying to refresh when identity is no longer on the node


howardjohn:
Currently I think we wil try forever to refresh these identities. Once they are no longer on the node, we don't need them and it will fail when we do try to refresh.

This should probably be done after https://github.com/istio/ztunnel/issues/334  to avoid overlapping work

qfel:
Is this
   1. Stop refreshing on specific error code from the cert service
   2. Add a method to SecretManager to allow explicit prune of specific cert (so that it will stop refreshing)
or both?

I can take a look since I mostly wrapped up #334 (at least the "implement" portion, since no code relies on the warmup functionality yet).

howardjohn:
I think (2). 1 may be transient - we may think we have pod X on our node but the CA doesn't yet (stale cache), so we would want to keep retrying.

The issue I am thinking is

0: istiod says "pod x is on your node". We fetch X identity
1: istiod says "pod x is removed"
2: we continue to refresh X forever

keithmattix:
Has this been implemented?

linsun:
@qfel @howardjohn is this done? I saw https://github.com/istio/ztunnel/pull/421 is merged

can we close this?

howardjohn:
That adds the `forget_certificate` but we don't use it iiuc

linsun:
@adiprerepa  are you working on this, or should we unassign?

linsun:
cc @ilrudie while i think this is nice to have for beta, probably not a blocker.

ilrudie:
Looks like it's been assigned. @adiprerepa, are you actively working on this?

adiprerepa:
@ilrudie I was, no longer though. feel free to unassign.

ilrudie:
I've opened #818 with an initial implementation of forgetting certs

ilrudie:
Yeah, Its still an problem afaik. I didn't love the impl in 818 and John had also expressed some concerns. I decided it needed a full rework so closed the old issue

PR #730, opened 2023-11-17
clear SecretManager cache when workload is updated/removed

GregHanson:
part of https://github.com/istio/ztunnel/issues/729

### DIFF
--- a/src/xds.rs
+++ b/src/xds.rs
@@ -150,6 +150,9 @@ impl ProxyStateUpdater {
                     .remove_endpoint(&prev.uid, &endpoint_uid(&prev.uid, None));
             }
 
+            // clear the corresponding identity from SecretManager
+            self.cert_fetcher.clear_cert(&prev.identity());
+
             // We removed a workload, no reason to attempt to remove a service with the same name
             return;
         }
--- a/src/cert_fetcher.rs
+++ b/src/cert_fetcher.rs
@@ -15,7 +15,7 @@
 use crate::config;
 use crate::config::ProxyMode;
 use crate::identity::Priority::Warmup;
-use crate::identity::{Identity, SecretManager};
+use crate::identity::{Identity, Request, SecretManager};
 use crate::state::workload::{Protocol, Workload};
 use std::sync::Arc;
 use tokio::sync::mpsc;
@@ -24,6 +24,7 @@ use tracing::{debug, error, info};
 /// Responsible for pre-fetching certs for workloads.
 pub trait CertFetcher: Send + Sync {
     fn prefetch_cert(&self, w: &Workload);
+    fn clear_cert(&self, id: &Identity);
 }
 
 /// A no-op implementation of [CertFetcher].
@@ -31,6 +32,7 @@ pub struct NoCertFetcher();
 
 impl CertFetcher for NoCertFetcher {
     fn prefetch_cert(&self, _: &Workload) {}
+    fn clear_cert(&self, _: &Identity) {}
 }
 
 /// Constructs an appropriate [CertFetcher] for the proxy config.
@@ -45,26 +47,35 @@ pub fn new(cfg: &config::Config, cert_manager: Arc<SecretManager>) -> Arc<dyn Ce
 struct CertFetcherImpl {
     proxy_mode: ProxyMode,
     local_node: Option<String>,
-    tx: mpsc::Sender<Identity>,
+    tx: mpsc::Sender<Request>,
 }
 
 impl CertFetcherImpl {
     fn new(cfg: &config::Config, cert_manager: Arc<SecretManager>) -> Self {
-        let (tx, mut rx) = mpsc::channel::<Identity>(256);
+        let (tx, mut rx) = mpsc::channel::<Request>(256);
 
         // Spawn a task for handling the pre-fetch requests asynchronously.
         tokio::spawn(async move {
-            while let Some(workload_identity) = rx.recv().await {
-                match cert_manager
-                    .fetch_certificate_pri(&workload_identity, Warmup)
-                    .await
-                {
-                    Ok(_) => debug!("prefetched cert for {:?}", workload_identity.to_string()),
-                    Err(e) => error!(
-                        "unable to prefetch cert for {:?}, skipping, {:?}",
-                        workload_identity.to_string(),
-                        e
-                    ),
+            while let Some(req_type) = rx.recv().await {
+                match req_type {
+                    Request::Fetch(workload_identity, pri) => {
+                        match cert_manager
+                            .fetch_certificate_pri(&workload_identity, pri)
+                            .await
+                        {
+                            Ok(_) => {
+                                debug!("prefetched cert for {:?}", workload_identity.to_string())
+                            }
+                            Err(e) => error!(
+                                "unable to prefetch cert for {:?}, skipping, {:?}",
+                                workload_identity.to_string(),
+                                e
+                            ),
+                        }
+                    }
+                    Request::Forget(workload_identity) => {
+                        cert_manager.forget_certificate(&workload_identity).await;
+                    }
                 }
             }
         });
@@ -92,9 +103,15 @@ impl CertFetcherImpl {
 impl CertFetcher for CertFetcherImpl {
     fn prefetch_cert(&self, w: &Workload) {
         if self.should_prefetch_certificate(w) {
-            if let Err(e) = self.tx.try_send(w.identity()) {
+            if let Err(e) = self.tx.try_send(Request::Fetch(w.identity(), Warmup)) {
                 info!("couldn't prefetch: {:?}", e)
             }
+        }
+    }
+
+    fn clear_cert(&self, id: &Identity) {
+        if let Err(e) = self.tx.try_send(Request::Forget(id.clone())) {
+            info!("couldn't clear identity: {:?}", e)
         }
     }
 }
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -365,7 +365,7 @@ async fn maybe_sleep_until(till: Option<Instant>) -> bool {
     }
 }
 
-enum Request {
+pub enum Request {
     Fetch(Identity, Priority),
     Forget(Identity),
 }


howardjohn (commented):
I am not sure we want to trigger it here, the cert is shared by multiple pods

keithmattix:
> I am not sure we want to trigger it here, the cert is shared by multiple pods

Not only that, but if users are using the default service account all the pods in their namespace, the cert could be shared by multiple workloads.

src/xds.rs
```diff
@@ -150,6 +150,9 @@ impl ProxyStateUpdater {
                     .remove_endpoint(&prev.uid, &endpoint_uid(&prev.uid, None));
             }
 
+            // clear the corresponding identity from SecretManager
+            self.cert_fetcher.clear_cert(&prev.identity());
```
GregHanson, line 154: ok, calling the clear in the remove() function sounds like a no-go, since multiple pods/workloads can use the same identity/service account.  I think it is still safe in the  `insert_workload()` since an update would result in a delete immediately followed by a prefect_cert.

GregHanson: in addition to delete, I am also thinking of a update as well.  In the event of a root cert rotation, there is a full push, where all the identities on the node need to be updated which is why I am thinking of the `insert` case as well

GregHanson: example flow on a Full Push:
```
2023-11-17T19:48:58.633342Z  INFO xds{id=2}: ztunnel::xds::client: received response type_url="type.googleapis.com/istio.workload.Address" size=15
2023-11-17T19:48:58.634024Z DEBUG xds{id=2}: ztunnel::xds: handling insert Kubernetes//Pod/istio-system/istiod-5fc86c58d7-6zhjk
2023-11-17T19:48:58.634266Z DEBUG xds{id=2}: ztunnel::xds: handling delete Kubernetes//Pod/istio-system/istiod-5fc86c58d7-6zhjk
2023-11-17T19:48:58.634654Z DEBUG xds{id=2}: ztunnel::xds: handling insert Kubernetes//Pod/kube-system/coredns-5d78c9869d-lfhbw
2023-11-17T19:48:58.634861Z DEBUG xds{id=2}: ztunnel::xds: handling delete Kubernetes//Pod/kube-system/coredns-5d78c9869d-lfhbw
2023-11-17T19:48:58.635178Z DEBUG xds{id=2}: ztunnel::xds: handling insert Kubernetes//Pod/local-path-storage/local-path-provisioner-6bc4bddd6b-zktwb
2023-11-17T19:48:58.635393Z DEBUG xds{id=2}: ztunnel::xds: handling delete Kubernetes//Pod/local-path-storage/local-path-provisioner-6bc4bddd6b-zktwb
2023-11-17T19:48:58.635580Z DEBUG xds{id=2}: ztunnel::xds: handling insert Kubernetes//Pod/metallb-system/controller-574698dfb9-8bpgc
2023-11-17T19:48:58.635667Z DEBUG xds{id=2}: ztunnel::xds: handling delete Kubernetes//Pod/metallb-system/controller-574698dfb9-8bpgc
2023-11-17T19:48:58.635774Z DEBUG xds{id=2}: ztunnel::xds: handling insert Kubernetes//Pod/istio-system/ztunnel-pjz4l
2023-11-17T19:48:58.635835Z DEBUG xds{id=2}: ztunnel::xds: handling delete Kubernetes//Pod/istio-system/ztunnel-pjz4l
2023-11-17T19:48:58.635938Z DEBUG xds{id=2}: ztunnel::xds: handling insert Kubernetes//Pod/istio-system/ztunnel-24rvb
2023-11-17T19:48:58.636008Z DEBUG xds{id=2}: ztunnel::xds: handling delete Kubernetes//Pod/istio-system/ztunnel-24rvb
2023-11-17T19:48:58.636110Z DEBUG xds{id=2}: ztunnel::xds: handling insert Kubernetes//Pod/kube-system/coredns-5d78c9869d-nhbj8
2023-11-17T19:48:58.636263Z DEBUG xds{id=2}: ztunnel::xds: handling delete Kubernetes//Pod/kube-system/coredns-5d78c9869d-nhbj8
2023-11-17T19:48:58.637799Z DEBUG xds{id=2}: ztunnel::xds: handling insert Kubernetes//Pod/kube-system/metrics-server-699dbf548-97d54
2023-11-17T19:48:58.637915Z DEBUG xds{id=2}: ztunnel::xds: handling delete Kubernetes//Pod/kube-system/metrics-server-699dbf548-97d54
2023-11-17T19:48:58.638154Z DEBUG xds{id=2}: ztunnel::xds: handling insert Kubernetes//Pod/istio-system/ztunnel-bzsz5
2023-11-17T19:48:58.638243Z DEBUG xds{id=2}: ztunnel::xds: handling delete Kubernetes//Pod/istio-system/ztunnel-bzsz5
2023-11-17T19:48:58.638353Z DEBUG xds{id=2}: ztunnel::xds: handling insert Kubernetes//Pod/default/sleep-9454cc476-xmbq7
2023-11-17T19:48:58.638423Z DEBUG xds{id=2}: ztunnel::xds: handling delete Kubernetes//Pod/default/sleep-9454cc476-xmbq7
2023-11-17T19:48:58.639160Z DEBUG xds{id=2}: ztunnel::xds::client: sending response type_url="type.googleapis.com/istio.workload.Address" nonce="208bbcbe-7874-46a9-b9ff-c63767938a97" type=Ack
2023-11-17T19:48:58.639407Z DEBUG ztunnel::xds::client: sending request type_url="type.googleapis.com/istio.security.Authorization"
2023-11-17T19:48:58.639609Z DEBUG ztunnel::xds::client: sending request type_url="type.googleapis.com/istio.workload.Address"
2023-11-17T19:48:58.639763Z DEBUG ztunnel::identity::manager: gihanson cleared identity "spiffe://cluster.local/ns/istio-system/sa/istiod"
2023-11-17T19:48:58.639834Z DEBUG ztunnel::identity::manager: gihanson cleared identity "spiffe://cluster.local/ns/kube-system/sa/coredns"
2023-11-17T19:48:58.639885Z DEBUG ztunnel::identity::manager: gihanson cleared identity "spiffe://cluster.local/ns/local-path-storage/sa/local-path-provisioner-service-account"
2023-11-17T19:48:58.639936Z DEBUG ztunnel::identity::manager: gihanson cleared identity "spiffe://cluster.local/ns/metallb-system/sa/controller"
2023-11-17T19:48:58.639985Z DEBUG ztunnel::identity::manager: gihanson cleared identity "spiffe://cluster.local/ns/istio-system/sa/ztunnel"
2023-11-17T19:48:58.640030Z DEBUG ztunnel::identity::manager: gihanson cleared identity "spiffe://cluster.local/ns/istio-system/sa/ztunnel"
2023-11-17T19:48:58.640071Z DEBUG ztunnel::identity::manager: gihanson cleared identity "spiffe://cluster.local/ns/kube-system/sa/coredns"
2023-11-17T19:48:58.640116Z DEBUG ztunnel::identity::manager: gihanson cleared identity "spiffe://cluster.local/ns/kube-system/sa/metrics-server"
2023-11-17T19:48:58.640148Z DEBUG ztunnel::identity::manager: gihanson cleared identity "spiffe://cluster.local/ns/istio-system/sa/ztunnel"
2023-11-17T19:48:58.640178Z DEBUG ztunnel::identity::manager: gihanson cleared identity "spiffe://cluster.local/ns/default/sa/sleep"
2023-11-17T19:48:58.640323Z DEBUG ztunnel::identity::manager: gihanson new identity "spiffe://cluster.local/ns/default/sa/sleep"
```

GregHanson:
the logic could be moved to WorkloadStore, potentially adding another index `by_identity` which could keep track of workloads by their identity, and know to clear the identity from SecretManager when all workloads for an identity are removed.  @howardjohn @keithmattix thoughts?

keithmattix:
Is there a bug caused by not clearing the cache? Do we anticipate a perf bump from clearing the certs? Or is the motivation from a security aspect of ztunnel not holding any more certs than it needs to? The last option is the only one that's compelling to me so far tbh

GregHanson:
there's a few outlined in https://github.com/istio/ztunnel/issues/351#issuecomment-1434804006

but there is also the case where a root cert rotation is performed.  All workloads get pushed via WDS as a full push update - but the SecretManager is never notified so the cache is stale until ztunnel pods are restarted or the certs expire

howardjohn:
Clearing a cert has a very very high cost. It can easily take >1s of CPU time, and some CAs have a per-cert cost (I have seen 50c a cert). We need to be very cautious about when we request one/delete one

GregHanson:
> We need to be very cautious about when we request one/delete one

I'm not sure how to improve this on a FullPush scenario in the event of something like a root cert rotation - unless we modify the WDS API to signal a cert/identity change?

howardjohn:
A full push should not trigger certificate rotation, they are completely
unrelated concepts.

GregHanson:
So then you suggest a brand new object to the WDS api?

hzxuzhonghu:
Since clearing a secret maybe a costy op, i think maybe we can do a lazy clear.  We can do it when rotating, that could be an appropriate point

bleggett:
Ideally we need a way to actually instruct ztunnel to proactively do cache invalidation for its certs.

If we had SDS this would be relatively straightforward, but since we don't we will probably have to modify WDS.

linsun:
> Since clearing a secret maybe a costy op, i think maybe we can do a lazy clear. We can do it when rotating, that could be an appropriate point

Are you thinking only clearing and the secret whenever the workload is serve as a source or target of a traffic flow?

hzxuzhonghu:
@linsun Nope, i am thinking about lazy clearing it when we need to refresh it, so we can reduce the lock time, concern from @howardjohn https://github.com/istio/ztunnel/pull/730#issuecomment-1817118784

GregHanson:
I think @hzxuzhonghu's [proposed](https://github.com/istio/ztunnel/pull/730#issuecomment-1820138618) method should handle the identity removal as needed.

For cert invalidation, on a root cert rotation for example, I agree with Ben's proposal for modifying the WDS API.  @ilrudie proposed adding the certificate serial number as a field to the Workload API, which could be used for signaling the secret cache in ztunnel that the existing certs for that identity need to be invalidated

linsun:
> I think @hzxuzhonghu's [proposed](https://github.com/istio/ztunnel/pull/730#issuecomment-1820138618) method should handle the identity removal as needed.
> 
> For cert invalidation, on a root cert rotation for example, I agree with Ben's proposal for modifying the WDS API. @ilrudie proposed adding the certificate serial number as a field to the Workload API, which could be used for signaling the secret cache in ztunnel that the existing certs for that identity need to be invalidated

Root cert rotation is different from intermediate cert got updated. Which scenario are you after? I thought it is for intermediate cert.
Q for identity removal here, do remove identity + add the new identity when ztunnel refresh the cert as part of the 24 hour validity duration?

linsun:
> @ilrudie proposed adding the certificate serial number as a field to the Workload API, which could be used for signaling the secret cache in ztunnel that the existing certs for that identity need to be invalidated

Any concerns with this approach? I think @GregHanson is working on POC this.  Would certificate serial number be strong enough here? cc @istio/wg-security-maintainers @hzxuzhonghu @howardjohn @keithmattix

howardjohn:
The xDS control plane only knows the serial number in demo setups. We cannot assume XDS and CA are the same or have visibility into each other.

linsun:
> The xDS control plane only knows the serial number in demo setups. We cannot assume XDS and CA are the same or have visibility into each other.

I don't agree that is only for demo setups!  If users were following our docs to plugin intermediate CA certs into Istiod, Istiod know the CA cert serial number. I believe in recent istiod, there is no restart needed for istiod to pick up the updated intermediate CA certs.

I do agree with your statement that XDS and CA may not be the same where istiod may not be the CA where cert manager or spire is the CA. In this case, I think it is reasonable to assume the updated intermediate CA certs by picked up by the spire agent or wait till ztunnel refreshes the cert as part of the 24 hour validity duration.

keithmattix:
> by picked up by the spire agent

SPIRE is not the only way for users to integrate Istio with an external PKI. cert-manager (through istio-csr) and Kubernetes CSR are both supported options where istiod never sees the minted leaf certs. Generally, I think we should be wary of any solution that assumes that the thing generating the WDS payload is the thing providing the certs.

linsun:
> > by picked up by the spire agent
> 
> SPIRE is not the only way for users to integrate Istio with an external PKI. cert-manager (through istio-csr) and Kubernetes CSR are both supported options where istiod never sees the minted leaf certs. Generally, I think we should be wary of any solution that assumes that the thing generating the WDS payload is the thing providing the certs.

Agreed spire is not the only way, i was giving it as an example. For public cloud, I agree, users probably won't be looking into [plugin intermediate CA certs](https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/) into Istiod, but I can see it being used for on prem. Istiod as a CA or plugin external CA both are supported scenarios of Istio.

howardjohn:
Not a PKI expert here but IMO...

* Root cert rotation should **not** actually trigger new certificates to be minted. This is both expensive and unsafe. Correct way to rotate a root is to distribute the new roots _and then_ start moving new certificates over, then removing the old one. You cannot just atomically swap the old and new, nor just atomically move everything over to 'new' even keeping 'old'.
* Eventually you do need to mint certificates with 'new'. Is 12hr short enough for this, or do we need something between "instant" and "12hr"? 12hr seems fine to me... plus you can always force it w/ restart
* Our current root cert distribution with istiod is broken/wrong, as its coupled to signing

GregHanson:
> Root cert rotation should not actually trigger new certificates to be minted.

@howardjohn same opinion when it's a change in an intermediate cert update?

howardjohn:
> > Root cert rotation should not actually trigger new certificates to be minted.
> 
> @howardjohn same opinion when it's a change due to an intermediate cert update?

Yes, if you immediately get a new cert, a workload will then send request to with the cert. The server may not have the new intermediate yet and not trust it?

linsun:
@GregHanson can you pls update the github issue with the behaviors of sidecars and ztunnel on intermediate cert changes?  I think that will help everyone understand and decide if this is acceptable.

GregHanson:
|  |  ztunnel | sidecar |
| --- | --- | --- |
| intermediate cert changes picked up immediately? | NO | NO |
| intermediate cert changes picked up at 12 hour refresh interval | YES | YES |
| pick up root cert changes immediately? | NO | YES (feature flag & PCDS)

GregHanson:
also not a PKI expert, but my thoughts are if there is an error after root cert or intermediate cert change - do you really want to wait up to 12 hours to find out about it?  I guess right now the only way to sanity check is as John said, to perform selective restarts of pods to force the reload.

bleggett:
> Not a PKI expert here but IMO...
> 
> * Root cert rotation should **not** actually trigger new certificates to be minted. This is both expensive and unsafe. Correct way to rotate a root is to distribute the new roots _and then_ start moving new certificates over, then removing the old one. You cannot just atomically swap the old and new, nor just atomically move everything over to 'new' even keeping 'old'.

You always need to distribute a replacement root before you do anything, but things downstream of the root CA have no control over how rapidly the replaced root cert will be invalidated - they may be told how rapidly but they don't have control over it, and that's by design.

Istiod doesn't really have control over how or when root certs or rotated (beyond the reasonable assumption that it should be able to obtain a new one before the old one is revoked), and really should only care about its intermediate certs. Istio is at most an intermediate CA in terms of responsibility.

> * Eventually you do need to mint certificates with 'new'. Is 12hr short enough for this, or do we need something between "instant" and "12hr"? 12hr seems fine to me... plus you can always force it w/ restart
> * Our current root cert distribution with istiod is broken/wrong, as its coupled to signing

The time duration istio has to perform this is (time you get replacement cert from your CA) + (time until the old cert is revoked, which the CA can tell you). I don't really know as Istio itself needs to have a duration configuration for this.

hzxuzhonghu:
I suggest separating root cert update with clearing a cert while a pod deleted. Otherwise it is too diverged

GregHanson:
closing for now since the implementation will require further design work

PR #734, opened 2023-11-20
Update 1.20 common-files

hanxiaop:
Cherry-pick lint fix from https://github.com/istio/ztunnel/pull/716 and fix https://github.com/istio/ztunnel/pull/731


hanxiaop:
Needs to confirm if dependencies need to be updated.

zirain (approved):

Issue #736, opened 2023-11-22
Support custom SECRET_TTL and SECRET_GRACE_PERIOD_RATIO in ztunnel


GregHanson:
ztunnel currently has `SECRET_TTL` hardcoded to 24 hours [here](https://github.com/istio/ztunnel/blob/master/src/identity/caclient.rs#L68) and `SECRET_GRACE_PERIOD_RATIO` hardcoded to 0.5 [here](https://github.com/istio/ztunnel/blob/master/src/tls/boring.rs#L187).  Sidecar supports these two environment variables to allow customizing these cert rotations settings - should ztunnel do the same?

GregHanson:
related to second item raised during review [here](https://github.com/istio/ztunnel/pull/730#issuecomment-1821649547)

> Eventually you do need to mint certificates with 'new'. Is 12hr short enough for this, or do we need something between "instant" and "12hr"? 12hr seems fine to me... plus you can always force it w/ restart

In addition to the two env vars, sidecar also has PCDS as a way to pick up some cert changes immediately.  Since ztunnel does not support PCDS, these env vars would allow users who don't want to wait up to 12 hours for cert changes to have some control over these settings

hzxuzhonghu:
To support multi root tust and root cert rotation,  I think PCDS should be supported.

It is simple to add two environments to ztunnel.

hzxuzhonghu:
@linsun PCDS response now only contains cacerts. Logging level can be changed by `istioctl proxy-config log <sidecar-pod> <sidecar-container> --level <log-level>` now.

linsun:
> @linsun PCDS response now only contains cacerts. Logging level can be changed by `istioctl proxy-config log <sidecar-pod> <sidecar-container> --level <log-level>` now.

Do you mean this cmd is also avail for ztunnel as well?

GregHanson:
@linsun best way to use the log command for ztunnel is:
```
istioctl pc log ds/ztunnel -n istio-system --level <error|warn|info|debug|trace>
```
which will configure all ztunnels at once

GregHanson:
> It is simple to add two environments to ztunnel.

@hzxuzhonghu outside of meshconfig settings (in src/config.rs), there isn't much precedence for pulling in configuration settings from environment variables.  Couldn't recall if there was any intentional design discussion around this or if it was just something that was saved for post-beta

GregHanson:
[...]

PR #742, opened 2023-12-01
support setting a custom cert ttl via SECRET_TTL env var

GregHanson:
if users do no want to wait up to 12 hours for cert rotations to take effect in ztunnel

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -40,6 +40,7 @@ const LOCAL_XDS_PATH: &str = "LOCAL_XDS_PATH";
 const XDS_ON_DEMAND: &str = "XDS_ON_DEMAND";
 const XDS_ADDRESS: &str = "XDS_ADDRESS";
 const CA_ADDRESS: &str = "CA_ADDRESS";
+const SECRET_TTL: &str = "SECRET_TTL";
 const FAKE_CA: &str = "FAKE_CA";
 const ZTUNNEL_WORKER_THREADS: &str = "ZTUNNEL_WORKER_THREADS";
 const ENABLE_ORIG_SRC: &str = "ENABLE_ORIG_SRC";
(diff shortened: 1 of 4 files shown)


GregHanson:
part of https://github.com/istio/ztunnel/issues/736

src/config.rs
```diff
@@ -335,6 +339,10 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
         xds_root_cert,
         ca_address,
         ca_root_cert,
+        secret_ttl: match parse::<i64>(SECRET_TTL)? {
```
hzxuzhonghu, line 342: Why don't you use parse_default

hzxuzhonghu: can it parse string like "24h"

hzxuzhonghu: I believe this function cannot parse time duration like `1h` `1m`

GregHanson: [duration-str](https://docs.rs/duration-str/latest/duration_str/) looks like it works.  Now accepts times in the form of `5m`, `1h` etc

hzxuzhonghu: Looks good

hzxuzhonghu (commented):
make gen to fix the ci

linsun (approved):
lgtm

GregHanson:
so `make gen` by itself did not fix the issue.  Was confused looking at the failing diff because it looked like the changes I had already made to the `Cargo.lock` file - but the lint failure was actually for `fuzz/Cargo.lock` file.  `make gen`/`cargo fmt` did not produce any changes in git.  Checking through the CI logs in the [failed test](https://storage.googleapis.com/istio-prow/pr-logs/pull/istio_ztunnel/742/test_ztunnel/1732216113577594880/build-log.txt), I had to `cd fuzz; cargo check; cd .. ; cargo fmt` to get the required changes

PR #747, opened 2023-12-08
inpod redirection mode

yuval-k:
In pod redirection mode.
for context see: https://docs.google.com/document/d/1dynKlnNgIOywv3cwuw_2RCk_SKFRs7IrESaC8_r-sm0/edit

This adds a new **optional** redirection mode to ztunnel - redirecting traffic from within the pod namespace.

A few details on the code:
- ztunnel communicates with new node agent (@bleggett will do a PR for that soon) using a unix socket
- The node agent provides that ztunnel with information and net-ns for pods local to the node
- In the code we call this protocol ZDS in the code
- Each pod get its own instance of a ztunnel `proxy::Proxy`, that starts its sockets in the pod's netns
- We've created `trait SocketFactory` to abstract the netns details from the `proxy::Proxy`
- Most of the code is in the new "inpod" module

Note that ztunnel would still work with the old redirection mode. new mode kicks in when `INPOD_ENABLED` is set to `true`.

Summary of changes:

| File  | Change |
| ------------- | ------------- |
| proto/zds.proto | ZDS protocol messages |
| scripts/ztunnel-redirect-inpod.sh | used in namespaced tests to redirect traffic |
| src/app.rs | start inpod machinary if inpod mode enabled |
| src/admin.rs | support dynamic admin handler (inpod mode only has a handler if it is enabled) |
| src/dns/server.rs | wire-in socket factory |
| src/config.rs | config options for inpod |
| src/inpod.rs | initialize the workload-proxy manager |
| src/inpod/config.rs | inpod socket factory and its config |
| src/inpod/admin.rs | admin endpoint for inpod - showing currently managed pods |
| src/inpod/netns.rs | specialized netns code suited for inpod |
| src/inpod/metrics.rs | inpod metrics |
| src/inpod/protocol.rs | ZDS implementation that parses the procol |
| src/inpod/packet.rs | SEQPACKET UDS impl |
| src/inpod/workloadmanager.rs | manages workload lifecycle. driven by events coming from protocol.rs |
| src/inpod/statemanager.rs | manages the running proxies |
| src/proxy.rs | wire-in socket factory |
| src/proxy/inbound_passthrough.rs | wire-in socket factory |
| src/proxy/inbound.rs | wire-in socket factory |
| src/proxy/socks5.rs | wire-in socket factory |
| src/proxy/outbound.rs | wire-in socket factory |
| src/socket.rs | add methods to set mark |
| src/proxyfactory.rs | factory to create proxies using a socket factory |
| src/test_helpers/app.rs | support admin_request inside the ztunnel netns |
| src/test_helpers/linux.rs | deploy ztunnel in inpod mode |
| src/test_helpers/inpod.rs | start a ZDS server, simulating the node agent |
| tests/namespaced.rs | add a test for in-pod mode |
| src/test_helpers/netns.rs | support returning value from `run()` |


yuval-k:
part of https://github.com/istio/istio/issues/48212

proto/zds.proto
```diff
@@ -0,0 +1,71 @@
...
+
+syntax = "proto3";
+
+// GRPC package - part of the URL. Service is added.
+// URL: /PACKAGE.SERVICE/METHOD
+package istio.workload.zds;
+
+option go_package="pkg/zdsapi";
+
+// Add a workload to the zutnnel. this will be accompanied by ancillary data contianing
```
PlatformLC, line 12:
```suggestion
// Add a workload to the ztunnel. this will be accompanied by ancillary data contianing
```
nit: typo

linsun:
@hzxuzhonghu or @howardjohn can you please review this so we can make progress on the PR?

cc @stevenctl

howardjohn:
> @hzxuzhonghu or @howardjohn can you please review this so we can make progress on the PR?
> 
> cc @stevenctl

@linsun yes I am looking, but this is among the largest PR (especially when considered among https://github.com/istio/istio/pull/48253) so please set expectations accordingly. It will _not_ be done within week.

howardjohn:
Also I have been reviewing the istiod side first fwiw

linsun:
Ok, just wanted to make sure folks are reviewing as we don't see much feedbacks so far.

src/inpod/workloadmanager.rs
```diff
@@ -0,0 +1,586 @@
...
+    ready: readiness::Ready,
+    // Manually drop as we don't want to mark ready if we are dropped.
+    // This can happen when the server drains.
+    block_ready: Option<std::mem::ManuallyDrop<readiness::BlockReady>>,
+}
+
+pub struct WorkloadProxyManager {
+    state: super::statemanager::WorkloadProxyManagerState,
+    networking: WorkloadProxyNetworkHandler,
+    // readiness - we are only ready when we are connected. if we get disconnected, we become not ready.
```
howardjohn, line 44: what is the meaning/use of transitioning from ready -> not ready?

Currently its only not ready -> ready.

src/inpod/admin.rs
```diff
@@ -0,0 +1,163 @@
...
+    Up,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ProxyState {
+    pub state: State,
+
+    // using refernce counts to account for possible race between the proxy task that notifies us
+    // that a proxy is down, and the proxy factory task that notifies us when it is up.
+    #[serde(skip_serializing, skip_deserializing)]
```
ymesika, line 37:
```suggestion
    #[serde(skip)]
```

src/inpod/metrics.rs
```diff
@@ -0,0 +1,65 @@
...
+            m.pending_proxy_count.clone(),
+        );
+        registry.register(
+            "inpod_proxies_started",
+            "The current number of active inpod proxies",
+            m.proxies_started.clone(),
+        );
+        registry.register(
+            "inpod_proxies_stopped",
+            "The current number of active inpod proxies",
```
ymesika, line 57: Is this description correct? Should it be inactive / stopped ?

src/proxyfactory.rs
```diff
@@ -0,0 +1,131 @@
...
+        cert_manager: Arc<SecretManager>,
+        proxy_metrics: Option<Metrics>,
+        dns_metrics: Option<dns::Metrics>,
+        drain: Watch,
+    ) -> std::io::Result<Self> {
+        let proxy_metrics = match proxy_metrics {
+            Some(metrics) => Some(Arc::new(metrics)),
+            None => {
+                if config.proxy {
+                    error!("dns proxy configured but no dns metrics provided")
```
ymesika, line 57:
```suggestion
                    error!("proxy configured but no metrics provided")
```

src/inpod/workloadmanager.rs
yuval-k: My thinking was that the user wants to be aware everytime the ztunnel is not connected to the CNI agent (as new pods will not get traffic).
The way to signal that is to make the ztunnel not-ready when it looses connectivity to the node agent.

so the meaning of "Ready" here is that ztunnel is ready to operate on new pods (i.e. connected to node agent).
WDYT?

proto/zds.proto
hzxuzhonghu, line 1: it is sent in json, so why not define struct directly

yuval-k: This struct it is sent as binary proto. Where do you see it sent as json?

linsun:
@hzxuzhonghu @howardjohn any further comments? It is been out for review for 2 weeks - is this ready to go?  Should be minimal risk - the ztunnel PR should be safe to merge and it is not used/exercised until the corresponding istio [PR](https://github.com/istio/istio/pull/48253) is merged. After the PR is merged, ztunnel can still work with the old CNI. 

This blocks Yuval from doing other works for L4 ambient:

- https://github.com/istio/istio/issues/48286 (install CNI race)
- https://github.com/istio/istio/issues/46628 (mistaken identity)

linsun:
@yuval-k should we remove the hold label?

yuval-k:
> @yuval-k should we remove the hold label?

 @howardjohn place the label, so I'll direct the question to him. 

As far as I am concerned this can be merged once reviewed.

howardjohn:
Hold was just to make sure it was reviewed by interested parties and didn't
get 1 approval and accidentally merge in if others were still in-review

proto/zds.proto
hzxuzhonghu: Ok, i see the proto api in istio pr too

src/inpod/protocol.rs
```diff
@@ -0,0 +1,409 @@
...
+        if f_type != nix::sys::statfs::PROC_SUPER_MAGIC && f_type != nix::sys::statfs::NSFS_MAGIC {
+            anyhow::bail!("Unexpected FD type for netns: {:?}", f_type);
+        }
+    }
+
+    debug!("FD {} looks like a netns", fd);
+    Ok(())
+}
+
+fn iovs_buf<'a>(slices: &'a [IoSliceMut<'a>], remaining: usize) -> IoSliceBuf<'a> {
```
howardjohn, line 263: Why do we need all of this custom implementation? Isn't there always only a single element in the slices list? Seems like we can just just directly translate to bytes::Bytes from that and remove all of this code. 

or if we really need multiple buffers for some reason,

```rust
        bytes::Bytes::from_iter(slices
            .iter()
            .flat_map(|slice| slice.as_ref())
            .map(|x| *x))
```

proto/zds.proto
costinm: Any particular reason for same-machine small message using protobuf ? GRPC and proto are great, but may be overkill - lots of debugging and possible integrations would be simpler with just json.

proto/zds.proto
```diff
@@ -0,0 +1,71 @@
...
+package istio.workload.zds;
+
+option go_package="pkg/zdsapi";
+
+enum Version {
+  NOT_USED = 0;
+  V1 = 1;
+}
+
+message ZdsHello {
```
costinm, line 14: And if we start with version - seems another NIH protocol. It doesn't seem to need all this complexity - if we need to change we can use a different address.

Not a huge deal - but unnecessary IMO

proto/zds.proto
```diff
@@ -0,0 +1,71 @@
...
+
+// Ztunnel ack message. If error is not empty, this is an error message.
+message Ack {
+  string error = 1;
+}
+
+/*
+Protocol details:
+on new connection from ztunnel to CNI, the CNI agent
+- will send all the existing payloads (that it has in its cache) to the ztunnel using AddWorkload message.
```
costinm, line 48: Same as above - seems very complicated and for no clear reason.

Just send a json with all the data - how many pods are we talking about ? 200 ? One message in, one response out.

What can the CNI do in case of error - and what could be the error that requires this ping-pong ?

CNI is way overbloated and complex - I love the recent changes of using UDS and sending descriptors, but it can be done far simpler.

costinm (commented):
Big +1 on the mechanism, we should expand on it - but the complexity on both sides is hard to understand. Or perhaps we can just merge it, and work on a simpler/cleaner implementation ?

costinm:
Not against merging - but I can't see this as a final solution, just strictly internal code consistent with the existing CNI ( which is way over complex), with some goal to replace with a cleaner version at some point.

src/inpod/statemanager.rs
```diff
@@ -0,0 +1,522 @@
...
+#[mockall_double::double]
+use crate::proxyfactory::ProxyFactory;
+
+#[mockall_double::double]
+use super::config::InPodConfig;
+
+use super::netns::InpodNetns;
+
+// Note: we can't drain on drop, as drain is async (it waits for the drain to finish).
+pub(super) struct WorkloadState {
```
howardjohn, line 30:
```suggestion
struct WorkloadState {
```

src/inpod/statemanager.rs
```diff
@@ -0,0 +1,522 @@
...
+    async fn join(self) {
+        futures::future::join_all(self.draining).await;
+    }
+}
+
+pub struct WorkloadProxyManagerState {
+    proxy_gen: ProxyFactory,
+    metrics: Arc<Metrics>,
+    // use hashbrown for extract_if
+    workload_states: hashbrown::HashMap<String, WorkloadState>,
```
howardjohn, line 59: can we document what the key is (looks like UID?)

src/inpod/statemanager.rs
```diff
@@ -0,0 +1,522 @@
...
+                        .insert(poddata.info.workload_uid.clone());
+                }
+                let netns = InpodNetns::new(self.inpod_config.cur_netns(), poddata.netns)
+                    .map_err(|e| Error::ProxyError(crate::proxy::Error::Io(e)))?;
+
+                self.add_workload(poddata.info, netns)
+                    .await
+                    .map_err(Error::ProxyError)
+            }
+            WorkloadMessage::KeepWorkload(workload_uid) => {
```
howardjohn, line 119: I don't feel very comfortable with this KeepWorkload thing.

Either our netns logic is reliable and we don't need this, or its unreliable and we need to abandon this project or make it more reliable.

(I think we can make it reliable)

src/inpod/statemanager.rs
```diff
@@ -0,0 +1,522 @@
...
+                if self.snapshot_received {
+                    return Err(Error::ProtocolError);
+                }
+                self.reconcile();
+                // mark ready
+                self.snapshot_received = true;
+                Ok(())
+            }
+        }
+        // TODO: add metrics, about how many pods we have currently, how many we added, how many we removed
```
howardjohn, line 150: I think this is done?

src/inpod/statemanager.rs
```diff
@@ -0,0 +1,522 @@
...
+                // inodes are different, we have a new netns.
+                // this can happen when there's a CNI failure (that's unrelated to us) which triggers
+                // pod sandobx to be re-created with a fresh new netns.
+                // drain the old proxy and add this one.
+                self.del_workload(&workload_info.workload_uid);
+            } else {
+                // idempotency - no error if we already have a proxy for the workload
+                // check if the inodes match. if they don't, we have a new netns
+                // we need to drain the previous proxy and add this one.
+                return Ok(()); // Err(Error::ProxyExists(workload_info.workload_uid));
```
howardjohn, line 214:
```suggestion
                return Ok(());
```

src/inpod/admin.rs
```diff
@@ -0,0 +1,163 @@
...
+                error!("Failed serializing state {:?}", e);
+                vec![]
+            }
+        }
+    }
+}
+
+impl crate::admin::AdminHandler for WorkloadManagerAdminHandler {
+    fn path(&self) -> &'static str {
+        "/workloadmanager"
```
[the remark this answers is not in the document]
yuval-k: My reasoning for the new interface:
- Its only available if inpod mode is on . It's dynamically added to the admin interface.
- Another benefit is that all the logic related to "inpod" is inside the inpod module (I feel this makes the code easier  to grok).

proto/zds.proto
yuval-k: this was in response to https://github.com/istio/istio/pull/48253#issuecomment-1863792631

src/inpod/admin.rs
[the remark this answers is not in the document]
howardjohn: I like the separation (and probably a good idea to use it in other plaecs!), but maybe we can just make it so keys under config_dump are dynamically loaded?

proto/zds.proto
yuval-k: This was done this way for simplicity.
The problem is not just the json size; its the combination of that with correlating the a pod uid in the AddWorkload message with the file descriptor.

The simplest way i could think of **making sure** we don't mess up the mapping is using PACKETSEQ socket type and send a single AddWorkload message and FD at the time.

Using a larger json will force us to use STREAM socket (instead of PACKETSEQ) and do more complex accounting of the workload stream and the FD stream...

howardjohn: Is it not just a list of UIDs (json encoding, I suppose) and a list of FDs? Then correlating those (assume they are the same length or its an error)

src/app.rs
```diff
@@ -89,15 +93,7 @@ pub async fn build_with_cert(
         None
     };
 
-    // Create and start the metrics server.
-    let metrics_server = metrics::Server::new(config.clone(), drain_rx.clone(), registry)
-        .await
-        .context("stats server starts")?;
-    let metrics_address = metrics_server.address();
-    // Run the metrics sever in the current tokio worker pool.
-    metrics_server.spawn();
```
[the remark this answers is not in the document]
yuval-k: yes, it needs to be after inpod as i need to add its metrics to the registry

src/inpod/statemanager.rs
yuval-k: I agree with the sentiment. This is my thought process that led me to add this:

The current approach we have to restore state when the node agent restarts is "eventually reliable",  so this is here to avoid temporary blips when things restart in edge cases. Let me explain in more detail:

The current method to get the FD after node agent restart is by scanning the /proc file system.

I believe this leaves us with a theoretical race:
If the node agent restarts after the CNI was invoked, but before processes belonging to the pod were created. In this case, the workload is in the ztunnel but not yet available in the node agent.

When ztunnel (re)connects to node agent, it gets a "snapshot" of workloads, so it can remove workloads that may have been removed while it was disconnected. So if that happens at the same time, the ztunnel might think that a workload was removed when it wasn't..

While extremely unlikely, i figured better safe than sorry.

howardjohn: We shouldn't be using processes in the users containers, we can utilize the pause container? That should run before CNI is invoked and never exit I believe?

proto/zds.proto
yuval-k: Conceptionally yes, it's list of uids and list of fds. But:

- If you use STREAM sockets you need to account for edge case where the list of uids and list of FDs come on different `recvmsg()` invocations. For example, `recvmsg` might return a shorter list of uids than the one sent (where the rest of the list will come in the next call to `recvmsg`).
- If we do it all in one message, we come pretty close to close to SCM_MAX_FD ( which is 253. GKE clusters can potentially go above this with a max of 256 pods per node). So we'll need to account for that and do batching in the code anyway..

costinm:
On Wed, Jan 3, 2024 at 1:10 PM Yuval Kohavi ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In proto/zds.proto
> <https://github.com/istio/ztunnel/pull/747#discussion_r1440954619>:
>
> > +}
> +
> +// Let ztunnel know that a full snapshot was sent. Ztunnel should reconcile its internal state
> +// and remove internal entries that were not sent.
> +message SnapshotSent {
> +}
> +
> +// Ztunnel ack message. If error is not empty, this is an error message.
> +message Ack {
> +  string error = 1;
> +}
> +
> +/*
> +Protocol details:
> +on new connection from ztunnel to CNI, the CNI agent
> +- will send all the existing payloads (that it has in its cache) to the ztunnel using AddWorkload message.
>
> Conceptionally yes, it's list of uids and list of fds. But:
>
>    - If you use STREAM sockets you need to account for edge case where
>    the list of uids and list of FDs come on different recvmsg()
>    invocations. For example, recvmsg might return a shorter list of uids
>    than the one sent (where the rest of the list will come in the next call to
>    recvmsg).
>    - If we do it all in one message, we come pretty close to close to
>    SCM_MAX_FD ( which is 253. GKE clusters can potentially go above this with
>    a max of 256 pods per node). So we'll need to account for that and do
>    batching in the code anyway..
>
>
Sure - send one uid and fd per message. But you don't need a protobuf to
encode the UID - and it is not clear what the blocking and request/response
provides.

If we find something meaningful to do if an error happens - we can send a
message the other way, but also simple 'uid' in a message - no need for
protos.

There are plenty of places where proto and grpc are best - it doesn't seem
like this is one of them.

costinm:
On Wed, Jan 3, 2024 at 12:24 PM Yuval Kohavi ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In src/inpod/statemanager.rs
> <https://github.com/istio/ztunnel/pull/747#discussion_r1440909790>:
>
> > +                info!(
> +                    "pod {} received netns, starting proxy",
> +                    poddata.info.workload_uid
> +                );
> +                if !self.snapshot_received {
> +                    self.snapshot_names
> +                        .insert(poddata.info.workload_uid.clone());
> +                }
> +                let netns = InpodNetns::new(self.inpod_config.cur_netns(), poddata.netns)
> +                    .map_err(|e| Error::ProxyError(crate::proxy::Error::Io(e)))?;
> +
> +                self.add_workload(poddata.info, netns)
> +                    .await
> +                    .map_err(Error::ProxyError)
> +            }
> +            WorkloadMessage::KeepWorkload(workload_uid) => {
>
> I agree with the sentiment. This is my thought process that led me to add
> this:
>
> The current approach we have to restore state when the node agent restarts
> is "eventually reliable", so this is here to avoid temporary blips when
> things restart in edge cases. Let me explain in more detail:
>
> The current method to get the FD after node agent restart is by scanning
> the /proc file system.
>
> I believe this leaves us with a theoretical race:
> If the node agent restarts after the CNI was invoked, but before processes
> belonging to the pod were created. In this case, the workload is in the
> ztunnel but not yet available in the node agent.
>
> When ztunnel (re)connects to node agent, it gets a "snapshot" of
> workloads, so it can remove workloads that may have been removed while it
> was disconnected. So if that happens at the same time, the ztunnel might
> think that a workload was removed when it wasn't..
>
> While extremely unlikely, i figured better safe than sorry.
>

The problem is that 'complex' is rarely safe - and most of the time sorry.

There are plenty of simpler ways to avoid this race - and probably scanning
the proc. We can save to a file (which is good for debug ), we can send a
'deleted' message - and I believe the operation is idempotent or should be.
Ztunnel can reject or hold connections until it knows about the state too (
which it should ) - and not at all clear what 'remove' is supposed to
do, the pod will be deleted anyways.

costinm:
I think max packet size is 128K - and I don't mean a 'big json' in one go.
Even JSON is likely overkill. \n separated lines would do, or an uint32
array.

We're not saving 1TB of data or sending over modem - the varint and
generated stubs don't buy us anything.

proto/zds.proto
yuval-k: responding to @costinm [comment](https://github.com/istio/ztunnel/pull/747#issuecomment-1876093209) (i think it was meant as a reply here)

I think there is some confusion here, this protocol is not gRPC. It's a custom protocol between ztunnel (rust) and the node agent (go). 
Note:
- The reason protobuf was used, is so that we won't have to implement serialization code twice, in go and in rust. even if we want to use json for some reason, it would still be easier to use protobuf as the common schema.
- I think swagger/openapi schema is more complex than protobuf.
- Note that i have a [follow up](https://github.com/yuval-k/ztunnel/commit/4351650c7c773535aa13832940e0609465abc3a9#diff-f573b5300ba6d69c6a6be4ca2ca2c3cd7ec437ee459302c30f731969a8eb011b) where i plan to add more information to the messages
- The reason we wait for the reply is so that pod start-up is delayed until ztunnel opened its ports in the pod's netns (similar motivation to holdApplicationUntilProxyStarts).

src/inpod/statemanager.rs
yuval-k: re @howardjohn good question about the pause container - we'll need to look if containerd and cri-o start it before they call the CNI. Assuming it checks out, If we remove this from here, are you comfortable placing a `panic()`  in the node agent (where it would otherwise send the keep-workload) so it is loud and clear if we are wrong?

costinm:
On Wed, Jan 3, 2024 at 5:13 PM Yuval Kohavi ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In proto/zds.proto
> <https://github.com/istio/ztunnel/pull/747#discussion_r1441172692>:
>
> > +}
> +
> +// Let ztunnel know that a full snapshot was sent. Ztunnel should reconcile its internal state
> +// and remove internal entries that were not sent.
> +message SnapshotSent {
> +}
> +
> +// Ztunnel ack message. If error is not empty, this is an error message.
> +message Ack {
> +  string error = 1;
> +}
> +
> +/*
> +Protocol details:
> +on new connection from ztunnel to CNI, the CNI agent
> +- will send all the existing payloads (that it has in its cache) to the ztunnel using AddWorkload message.
>
> responding to @costinm <https://github.com/costinm> comment
> <https://github.com/istio/ztunnel/pull/747#issuecomment-1876093209> (i
> think it was meant as a reply here)
>
> I think there is some confusion here, this protocol is not gRPC. It's a
> custom protocol between ztunnel (rust) and the node agent (go).
>
That's my issue - a custom protocol that is not gRPC but uses protobufs in
a strange and unique way to send effectively few strings.



> Note:
>
>    - The reason protobuf was used, is so that we won't have to implement
>    serialization code twice, in go and in rust. even if we want to use json
>    for some reason, it would still be easier to use protobuf as the common
>    schema.
>    - I think swagger/openapi schema is more complex than protobuf.
>
> Why would you need swagger/openapi to read an array of ints ? They are
meant for web and other things.

UDS has a message structure - so no need to even frame. It's a byte[] -
with a trivial set of data.
Simple string or ints - nothing more, and a private protocol - no need to
have a schema and validation and generate 100 language bindings.



>
>    - Note that i have a follow up
>    <https://github.com/yuval-k/ztunnel/commit/4351650c7c773535aa13832940e0609465abc3a9#diff-f573b5300ba6d69c6a6be4ca2ca2c3cd7ec437ee459302c30f731969a8eb011b>
>    where i plan to add more information to the messages
>
> I doubt we can find enough info to add to the messages to be worth
schema/proto/openapi vs reading directly from the buffer.


>
>    - The reason we wait for the reply is so that pod start-up is delayed
>    until ztunnel opened its ports in the pod's netns (similar motivation to
>    holdApplicationUntilProxyStarts).
>
> A pod startup should be delayed until netns is fully ready - but that
doesn't mean messages for other pods starting up need to be delayed.

Ztunnel can send messages one way and receive messages the other way, it's
a common pattern ( pipelining, async, etc).

BTW - I may be a bit confused about the flow and the components involved.
We still have a binary on the host that is called as part
of K8S CNI - we have the CNI server in golang, and ztunnel in Rust.

Any reason the binary ( which is AFAIK the main thing waiting and blocking
the pod init) can't directly talk with ztunnel, or ztunnel and CNI-server ?
It might be much simpler and avoid most issues on pod startup ( and
associated waiting ).

Beyond pod startup - not clear what 'waiting' means, existing pods are
already up and running. IF ztunnel restarts/is upgraded/etc - it
needs to recover, but waiting doesn't help.

howardjohn:
going further, why not just query CRI directly from ztunnel? it gives the
netns without guessing, no need to pass files around...?

src/inpod/statemanager.rs
yuval-k: re @costinm [comment here](https://github.com/istio/ztunnel/pull/747#issuecomment-1876096537) 

> The problem is that 'complex' is rarely safe - and most of the time sorry.

I'm aware.

> and not at all clear what 'remove' is supposed to do, the pod will be deleted anyways

Let me clarify.

When a pod is deleted, its network namespace doesn't get deleted. That's because network namespaces in linux can't be explicitly deleted. They automatically get cleaned up when nothing references them. Therefore, ztunnel needs some mechanism to know that a pod is deleted, so it can clean up its resources.

Our current approach, is that the node-agent tells ztunnel that pods were deleted. And there for, if the ztunnel was disconnected from the the node agent, when it reconnects we need to reconcile the full state of local pods on the node, to account for pods that may have been deleted while the ztunnel was disconnected.

We do this by sending a bunch of AddWorkload messages, followed by a SnapshotSent message on the initial connection. The ztunnel will clean up and workloads that it didn't see in one of these WorkloadAdded messages sent prior to the SnapshotSent message.
Aftet the SnapshotSent message, the protocol becomes delta-like with AddWorkload \ DelWorkload sent as needed.

If/When the node agent restarts, it needs to reconstruct its local state. That state is composed of local pods -> netns mapping. The node agent uses heuristics to reconstruct said state.

The (potentially non-existent [see above]) problem that KeepWorkload aims to solve, is to account for a possible temporary issue with that heuristic. Thus allowing the node agent to tell the ztunnel when it reconnects, that even though it doesn't have the netns for a certain pod, the ztunnel shouldn't remove it after the SnapshotSent message is sent.

yuval-k:
> going further, why not just query CRI directly from ztunnel? it gives the netns without guessing, no need to pass files around...?

We thought about using the CRI, we ended up not doing it for the following reasons:

First we need to know the location of the CRI socket to use, which is another knob for the user.

Second, the CRI doesn't give the netns itself. It gives the path to where netns is mounted in the host mount ns. So we'll need to know the base folder ahead of time so we can host volume mount it to the ztunnel* (another knob for the user).

Also, netns is only returned when queried in verbose mode. This means the response is not bound by the CRI schema, and is more of a _convention_. This means that we depend on the convention not changing.

The third reason we didn't want to use the CRI is that this you don't have access control on the CRI. If the ztunnel can access it for queries, it can access it for anything. This basically turns the ztunnel into the root user. So this might be worse for security posture.

---
(*) note that we currently need the host mount for the netns folder in the CNI agent, but think it will be more robust to change the CNI agent and binary to pass netns FDs, than change the ztunnel to use the host mount.

yuval-k:
> UDS has a message structure - so no need to even frame. It's a byte[] -
with a trivial set of data.
Simple string or ints - nothing more, and a private protocol - no need to
have a schema and validation and generate 100 language bindings.

I'm not sure what is the big problem with using protobufs. As I mentioned, we plan to add expand these messages to have more fields. And we need to read them from multiple languages. Why work hard and hand write serialization? I havn't needed to do that since my college days. 

> Any reason the binary ( which is AFAIK the main thing waiting and blocking the pod init) can't directly talk with ztunnel, or ztunnel and CNI-server ? It might be much simpler and avoid most issues on pod startup ( and associated waiting ).

Our initial PoC did just that. 

We ended up in the current state (where the node agent manages the state and tells about it to ztunnel)
- The desire to keep ztunnel minimal and simple. All state management and reconstruction (if anything restarts) is done in the node agent. Ztunnel only does what it has too. The node agent is even the one responsible for adding iptables rules.
- The ability to support onboarding existing pods (i.e. no CNI flow) to ztunnel. This requires a controller to watch pods and namespaces to see if existing pods enter/leave the ambient mesh.

>doesn't mean messages for other pods starting up need to be delayed.
Ztunnel can send messages one way and receive messages the other way, it's
a common pattern ( pipelining, async, etc).
...
Beyond pod startup - not clear what 'waiting' means, existing pods are
already up and running. IF ztunnel restarts/is upgraded/etc - it
needs to recover, but waiting doesn't help.

This is true, but distinguishing these cases would make the protocol more complex. See [this](https://github.com/istio/istio/pull/48253#issuecomment-1863785349) comment as well from @hzxuzhonghu  who had similar thoughts.
I'm open to doing that, but I think it should be done in a follow-up.

src/inpod/protocol.rs
```diff
@@ -0,0 +1,409 @@
...
+                res =  self.stream.readable() => res,
+            }?;
+
+            let res: Result<nix::sys::socket::RecvMsg<'_, '_, ()>, std::io::Error> =
+                self.stream.try_io(tokio::io::Interest::READABLE, || {
+                    recvmsg::<()>(
+                        raw_fd,
+                        &mut iov,
+                        Some(msgspace_ref),
+                        MsgFlags::MSG_CMSG_CLOEXEC,
```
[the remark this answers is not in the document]
yuval-k: we have to use `recvmsg` to get the control messages with the file descriptors.. iov is a vector of buffers for `recvmsg`.

Cargo.toml
```diff
@@ -75,9 +75,12 @@ priority-queue = "1.3.0"
 http-body-util = "0.1.0-rc.2"
 http-body-04 = { package = "http-body", version = "0.4" }
 http-body-1 = { package = "http-body", version = "1.0.0-rc.2" }
-tower-hyper-http-body-compat = { version = "0.2" , features = ["server", "http2"] }
+tower-hyper-http-body-compat = { version = "0.2", features = ["server", "http2"] }
 futures-util = "0.3.26"
 chrono = "0.4.23"
+nix = "0.26.2"
+mockall_double = "0.3.0"
```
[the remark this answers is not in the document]
yuval-k: fair, i'll see what i can do here

keithmattix (commented):
First pass. Very complex but necessary in the short term I think

src/inpod/workloadmanager.rs
```diff
@@ -0,0 +1,606 @@
...
+            let block_ready: Option<readiness::BlockReady> = self
+                .block_ready
+                .take()
+                .map(std::mem::ManuallyDrop::into_inner);
+
+            std::mem::drop(block_ready);
+        }
+    }
+
+    fn not_ready(&mut self) {
```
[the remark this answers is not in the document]
yuval-k: we dont "need" this per-say - as you said there's no service backing the ztunnel. But by the same logic, why have any readiness signal in the ztunnel at all?

It seems that in ztunnel the semantics of "readiness" is that ztunnel is ready to handle pods. It seems that the motivation here is to signal to the user that something requires immediate attention.

Using that semantics, if the ztunnel is not connected to the node agent, it is not "ready".
While envoy won't stop being ready, it does have a metric to indicate something similar (connection to the control-plane, called "connected_state").

If you think this will confuse users rather than help, I can remove this.

src/inpod/workloadmanager.rs
```diff
@@ -0,0 +1,606 @@
...
+    next_pending_retry: Option<std::pin::Pin<Box<tokio::time::Sleep>>>,
+}
+
+impl WorkloadProxyReadinessHandler {
+    fn new(ready: readiness::Ready) -> Self {
+        let mut r = Self {
+            ready,
+            block_ready: None,
+        };
+        r.not_ready();
```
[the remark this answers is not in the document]
yuval-k: what do you mean by upgrade? can you provide the steps to reproduce this?

src/inpod/workloadmanager.rs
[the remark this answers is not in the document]
howardjohn: Readiness does still have some meaning at startup even without a service, just not _after_ startup; you need to become ready before other instances will terminate in a rollout.

src/inpod/workloadmanager.rs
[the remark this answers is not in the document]
howardjohn: just add some label in the daemonset template or similar, it will trigger a rolling restart. Real world would probably be a more useful change, of course.

src/inpod/workloadmanager.rs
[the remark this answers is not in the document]
yuval-k: Note that this doesn't currently apply in the ztunnel case - AFAIK we cannot run ztunnel side by side with another ztunnel.. so currently other instance will have to terminate first.

Even if we had ztunnel working SxS, the current logic is consistent with the rollout semantics as well - You wouldn't want to remove an old ztunnel if the new one is not connected to the node agent.

I think this mainly comes to istio users expectation. If we think that readiness changing in back to "not ready"  will create confusion, I can replace this with a metric.

proto/zds.proto
yuval-k: PACKETSEQ makes it easier as I know that the buffer was fully received. in a STREAM socket, there's a potential case that the json buffer and the FD stream will be split across different `recvmsg`.

src/inpod/protocol.rs
yuval-k: I might be able to do that. it will require a bit restructuring to out-smart the borrow checker.

The original reason is that i couldn't pass `&[u8]` to `get_workload_data` as it was already borrowed; so i had to pass the `iov`.  I didn't want the code to look weird by me just referring only to `iov[0], hence `iovs_buf`.

howardjohn:
> > going further, why not just query CRI directly from ztunnel? it gives the netns without guessing, no need to pass files around...?
> 
> We thought about using the CRI, we ended up not doing it for the following reasons:
> 
> First we need to know the location of the CRI socket to use, which is another knob for the user.
> 
> Second, the CRI doesn't give the netns itself. It gives the path to where netns is mounted in the host mount ns. So we'll need to know the base folder ahead of time so we can host volume mount it to the ztunnel* (another knob for the user).
> 
> Also, netns is only returned when queried in verbose mode. This means the response is not bound by the CRI schema, and is more of a _convention_. This means that we depend on the convention not changing.
> 
> The third reason we didn't want to use the CRI is that this you don't have access control on the CRI. If the ztunnel can access it for queries, it can access it for anything. This basically turns the ztunnel into the root user. So this might be worse for security posture.
> 
> (*) note that we currently need the host mount for the netns folder in the CNI agent, but think it will be more robust to change the CNI agent and binary to pass netns FDs, than change the ztunnel to use the host mount.

FYI: https://github.com/containerd/containerd/issues/8085. note its not implemented, so more just a reference.

src/inpod/metrics.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+#[derive(Default)]
+pub struct Metrics {
+    pub(super) active_proxy_count: Family<(), Gauge>,
+    pub(super) pending_proxy_count: Family<(), Gauge>,
+    pub(super) proxies_started: Family<(), Counter>,
+    pub(super) proxies_stopped: Family<(), Counter>,
+
+    admin_handler: Arc<super::admin::WorkloadManagerAdminHandler>,
```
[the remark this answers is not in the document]
yuval-k: Yes, originally i had it where the metrics struct would update both; I'll fix.

src/inpod.rs
```diff
@@ -0,0 +1,229 @@
...
+pub fn init_and_new(
+    registry: &mut Registry,
+    admin_server: &mut crate::admin::Service,
+    cfg: &zconfig::Config,
+    proxy_gen: ProxyFactory,
+    ready: readiness::Ready,
+) -> anyhow::Result<WorkloadProxyManager> {
+    // verify that we have the permissions for the syscalls we need
+    WorkloadProxyManager::verify_syscalls()?;
+    let metrics = Arc::new(Metrics::new(registry));
```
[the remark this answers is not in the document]
yuval-k: I can change. I used this approach because this way i don't need to make the metrics module public.

src/app.rs
```diff
@@ -259,6 +288,47 @@ pub async fn build(config: config::Config) -> anyhow::Result<Bound> {
     build_with_cert(config, cert_manager).await
 }
 
+#[cfg(not(target_os = "linux"))]
```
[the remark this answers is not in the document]
yuval-k: Do you mean add this as the first line in the single function?
I'm not sure i can do that; the inpod module [conditionally  compiles](https://github.com/istio/ztunnel/blob/007a445f30e44181918a68d2789fa8dfc9d9246b/src/lib.rs#L23-L24) only on linux, so this needs to be at compile time.

src/inpod/workloadmanager.rs
```diff
@@ -0,0 +1,606 @@
...
+    async fn connect(&self) -> UnixStream {
+        const MAX_BACKOFF: Duration = Duration::from_secs(15);
+        let mut backoff = Duration::from_millis(10);
+
+        debug!("connecting to server: {:?}", self.uds);
+
+        loop {
+            match super::packet::connect(&self.uds).await {
+                Err(e) => {
+                    backoff = std::cmp::min(MAX_BACKOFF, backoff * 2);
```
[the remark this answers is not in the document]
yuval-k: it will, because on success we exit the function. i'll change the `break` below to `return` so its more clear

src/inpod/admin.rs
```diff
@@ -0,0 +1,163 @@
...
+}
+
+impl WorkloadManagerAdminHandler {
+    pub fn proxy_pending(&self, uid: &str) {
+        let mut state = self.state.write().unwrap();
+
+        // don't increment count here, as it is only for up and down. see comment in count.
+        match state.get_mut(uid) {
+            Some(key) => {
+                key.state = State::Pending;
```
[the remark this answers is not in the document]
yuval-k: Normally no.
But we have observed an edge case  case where the CNI flow is called twice with _different_ netns for the same pod. This happened when ADD flow failed later in flow (unrelated to us), the CNI ADD will be retried again (this is also why we added the inode checks).

In that case, we will start a new proxy and drain the old one pretty quickly, hence the reference checks. 
(this also addresses the question on the reference count below)

src/inpod.rs
```diff
@@ -0,0 +1,114 @@
+// Copyright Istio Authors
```
howardjohn, line 1: We have two data sources for "local workloads": WDS and the new UDS model.

Once we get the UID from the UDS, we start the proxy in the network namespace.
Now we can let the pod start (TBH not sure how we block it, but I think you said we do?).
Pod starts, sends an outbound request.
The proxy that is running is ONLY running for that workload, but we still (1) need to have the source workload from WDS and (2) try to guess certificates, etc based on source IP.

Inbound has similar logic.

This feels fishy in a few ways (but likely more I haven't thought of):
* If we are still doing IP based checks, we haven't really resolved the ip spoofing attacks, right?
* We block the pod running until the proxy starts, but it may not actually be ready to handle traffic since we haven't gotten the WDS response

I feel like I had more concerns but i was thinking about this late last night and didn't write it down...

WDYT? I get this PR is allow inpod and the old model, and the changes I am proposing are large and invasive. So I don't necessarily want to tackle them in this PR, if at all -- but would be good to understand the gaps now and the path to resolving them in the future?

yuval-k: here are my thoughts:
https://github.com/yuval-k/ztunnel/tree/inpod-dest-id

yuval-k: i planned to do this as a follow up; LMK what you think

yuval-k: >We block the pod running until the proxy starts, but it may not actually be ready to handle traffic since we haven't gotten the WDS response

I think the answer for this concern is on_demand

yuval-k: Note, that this also requires a change on the node agent side, as the uid from WDS is different from the uid in UDS - we may want to re-think this part

costinm:
Pod delete cleanup is not time sensitive or critical - node agent can just
send a list of namespaces that shouldn't exist. Or the cni plugin could do
this - I remember it is notified.

src/inpod/workloadmanager.rs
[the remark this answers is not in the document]
yuval-k: I just tried this scenario - I think you have 2 ztunnels running on the same time on the same node.

This is not supported as they both will try to open listening sockets in the same pod netns - and you'll get the error mentioned.

I believe running 2 ztunnels at the same time wasn't supported before AFAIK, so I don't believe this is a regression.

The second ztunnel should re-try binding, so as soon as the other ztunnel is removed it should come up.

We should probably change the daemonset to 
```
      maxSurge: 0
      maxUnavailable: 1
```
(today its the other way around)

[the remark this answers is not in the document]
howardjohn: it was supported before and even fairly graceful cut over. afaik new connections went to the new one while the old one drained old connections. Without this i don't see how an upgrade is done

yuval-k:
> Pod delete cleanup is not time sensitive or critical - node agent can just send a list of namespaces that shouldn't exist. Or the cni plugin could do this - I remember it is notified.

I agree that pod delete clean-up is not time sensitive.

What is the problem you are trying to solve? What part of the delete flow do you think is too complex?

Are able perhaps to sketch out your proposal in code / or in more technical details that include the proposed changes to the code in both the node-agent and the ztunnel?
These details will allow me to better understand and evaluate your proposal, and understand if it simplifies the current state.

src/inpod/workloadmanager.rs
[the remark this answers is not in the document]
yuval-k: I'm not sure how a graceful cutoff is possible with the old system. Do you mean old connection continued with the old ztunnel?

I'll try adding SO_REUSEPORT  to the sockets.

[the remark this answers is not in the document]
yuval-k: Re upgrade, supporting SxS upgrades will involve more than solving the ports. So probably worth a separate discussion. I was under the impression that our initial upgrade strategy was to cordon/drain node before an upgrade

costinm:
On Fri, Jan 5, 2024 at 5:29 PM Yuval Kohavi ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In src/inpod/workloadmanager.rs
> <https://github.com/istio/ztunnel/pull/747#discussion_r1443555113>:
>
> > +}
> +
> +struct WorkloadProxyManagerProcessor<'a> {
> +    state: &'a mut super::statemanager::WorkloadProxyManagerState,
> +    readiness: &'a mut WorkloadProxyReadinessHandler,
> +
> +    next_pending_retry: Option<std::pin::Pin<Box<tokio::time::Sleep>>>,
> +}
> +
> +impl WorkloadProxyReadinessHandler {
> +    fn new(ready: readiness::Ready) -> Self {
> +        let mut r = Self {
> +            ready,
> +            block_ready: None,
> +        };
> +        r.not_ready();
>
> Re upgrade, supporting SxS upgrades will involve more than solving the
> ports. So probably worth a separate discussion. I was under the impression
> that our initial upgrade strategy was to cordon/drain node before an upgrade
>

Simple cordon/drain is currently the only safe option we have.

We may find ways to upgrade without node drain - but it would be ideal to
implement that in a separate binary/module - and keep the
core simple. We already have far too much complexity - cordon just works
with no extra code and will likely remain the safest option.
Adding a lot of code to deal with some other mode is likely to add risks to
the simple mode and regular operation.

IMO it is time to do whatever needs to be done to get an initial working
version working - with cordon and roughly this PR - and
start working on a clean CNI without the legacy and complexity. Perhaps
with others - i.e. define the requirements ( 'a CNI should
allow a Pod to send/receive UDS file descriptors') and work with K8S
upstream to extend the core specs, with a minimal reference
implementation free of Istio-specific logic.

 The SDS/certificate UDS socket is another example involving creating UDS
listeners in the pod namespace - and it will be a powerful
general IPC mechanism, since it enables passing file descriptors (and
memory regions) between pods. There is vast evidence it is
a powerful and viable mehanism from /dev/binder.

So let's get this in, test it - and start the cleanup/de-legacyfication as
a long-term separate effort. Even if this PR was super simple -
the rest of istio CNI is full of complexity and far too hard.

yuval-k:
@howardjohn added SO_REUSEPORT which "solves" the upgrade problem (and it should actually drain connections gracefully). The main limitation is that new ztunnel and old ztunnel need to have the same UID (enforced by linux for port-reuse). LMK what you think.

Note that i do still think that upgrading is a bigger story than just this change.

Cargo.toml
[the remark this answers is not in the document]
bleggett: Using a concrete impl instead of a mock will also break IDEs/test runs outside of Linux FWIW - I don't much care but people have complained about that in the past.

bleggett:
> I'm not sure what is the big problem with using protobufs. As I mentioned, we plan to add expand these messages to have more fields. And we need to read them from multiple languages. Why work hard and hand write serialization? I havn't needed to do that since my college days.

+1 on this.

The fact that we have a single file declaring the schema and documenting the protocol which both the client and server explicitly refer to is far and away worth more the trivial encoding cost for intelligibility and developer discoverability.

I'm not against picking a simpler or more efficient serialization protocol/schema format than protobuf (e.g. Cap'n Proto) - but there aren't many, and not having one at all makes this much harder to understand and use.

src/inpod/admin.rs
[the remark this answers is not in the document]
yuval-k: I'm not sure how do this dynamically without an interface?
what do you mean by "keys"?

src/inpod/workloadmanager.rs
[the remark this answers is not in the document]
bleggett: > it was supported before and even fairly graceful cut over.

The existing support was "fairly graceful" in the sense that on a new ztunnel, it tore down all the node-level rules, recreated them pointing at the new ztunnel, and  tried to reconcile/catch any pods that might have spawned in the interim.

Doing this correctly will require more work, and yes, the only way to do this safely in production with current code in `master` is node cordon.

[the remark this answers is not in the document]
bleggett: > the main limitation [for SO_REUSEPORT] is that new ztunnel and old ztunnel need to have the same UID (enforced by linux for port-reuse). 

Which is not really an assumption we can make, though it may be OK for now. I'm not sure openshift will like this.

Regardless though, as before, the only current way we can recommend people to safely upgrade ztunnel in production is node cordon, so it doesn't matter.

bleggett:
> > going further, why not just query CRI directly from ztunnel? it gives the netns without guessing, no need to pass files around...?
> 
> We thought about using the CRI, we ended up not doing it for the following reasons:
> 
> First we need to know the location of the CRI socket to use, which is another knob for the user.
> 
> Second, the CRI doesn't give the netns itself. It gives the path to where netns is mounted in the host mount ns. So we'll need to know the base folder ahead of time so we can host volume mount it to the ztunnel* (another knob for the user).
> 
> Also, netns is only returned when queried in verbose mode. This means the response is not bound by the CRI schema, and is more of a _convention_. This means that we depend on the convention not changing.
> 
> The third reason we didn't want to use the CRI is that this you don't have access control on the CRI. If the ztunnel can access it for queries, it can access it for anything. This basically turns the ztunnel into the root user. So this might be worse for security posture.
> 
> (*) note that we currently need the host mount for the netns folder in the CNI agent, but think it will be more robust to change the CNI agent and binary to pass netns FDs, than change the ztunnel to use the host mount.



> > > going further, why not just query CRI directly from ztunnel? it gives the netns without guessing, no need to pass files around...?
> > 
> > 
> > We thought about using the CRI, we ended up not doing it for the following reasons:
> > First we need to know the location of the CRI socket to use, which is another knob for the user.
> > Second, the CRI doesn't give the netns itself. It gives the path to where netns is mounted in the host mount ns. So we'll need to know the base folder ahead of time so we can host volume mount it to the ztunnel* (another knob for the user).
> > Also, netns is only returned when queried in verbose mode. This means the response is not bound by the CRI schema, and is more of a _convention_. This means that we depend on the convention not changing.
> > The third reason we didn't want to use the CRI is that this you don't have access control on the CRI. If the ztunnel can access it for queries, it can access it for anything. This basically turns the ztunnel into the root user. So this might be worse for security posture.
> > (*) note that we currently need the host mount for the netns folder in the CNI agent, but think it will be more robust to change the CNI agent and binary to pass netns FDs, than change the ztunnel to use the host mount.
> 
> FYI: [containerd/containerd#8085](https://github.com/containerd/containerd/issues/8085). note its not implemented, so more just a reference.

Yeah +1 on what yuval suggested - the node-level CNI plugin that already has full node perms could definitely talk to the CRI and get the FD, and propagate it directly. That would be a good followup optimization.

src/inpod/workloadmanager.rs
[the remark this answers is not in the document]
bleggett: > Note that this doesn't currently apply in the ztunnel case - AFAIK we cannot run ztunnel side by side with another ztunnel.. so currently other instance will have to terminate first.
> 
> Even if we had ztunnel working SxS, the current logic is consistent with the rollout semantics as well - You wouldn't want to remove an old ztunnel if the new one is not connected to the node agent.

Right - if the new ztunnel is not "ready" because it cannot talk to the current CNI agent on the node, and the old one is/can, the rollout *should* stall, otherwise users end up in a bad state with no explicit signaling.

Again though, this only applies to tirekicking. Rollout replace of ztunnel isn't recommended for production upgrades.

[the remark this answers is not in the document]
howardjohn: I 100% agree it needs to sync to the agent at startup. I just don't see why it should become unready if it _later_ disconnects. Note a disconnect happens in regular operation, too.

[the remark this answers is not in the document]
bleggett: Oh I see what you're saying - yeah that's fair. I can't think of a situation during rollout where:

1. ztunnel A would have previously been able to connect to the CNI during startup.
2. ztunnel A ceases to be able to for `$SOME_REASON`.
3. A rollout happens, and the new ztunnel B _IS_ able to connect to the same CNI during startup.

hzxuzhonghu:
If ztunnel has that privs, it won't need to communicate with CNI plugin for the netNs but with CRI right?

bleggett:
> If ztunnel has that privs, it won't need to communicate with CNI plugin for the netNs but with CRI right?

Yes, but it would still need to communicate with the node agent, so all that would really do in practice is spread elevated permissions among more components, rather than fewer, which is not desirable.

yuval-k:
> If ztunnel has that privs, it won't need to communicate with CNI plugin for the netNs but with CRI right?

Giving the ztunnel access to the CRI is giving it privileges:  if ztunnel communicates with the CRI instead of the node agent, that gives it the power of root. Because you can use the CRI to spin up privileged containers with host mounts, anyone that can access the CRI is effectively root

hzxuzhonghu:
Got it, that is what i think. Recently I talked to guy from containerd. He said CRI can be used to get the netNs, I think the tricky point is how to make the start up in order without race

yuval-k:
I believe I responded to all comments; Let me know if I missed anything
cc @howardjohn @keithmattix @costinm

linsun:
@howardjohn @hzxuzhonghu - is this one ready to be approved? would be great to get it in release-1.21 before branching.

src/inpod/admin.rs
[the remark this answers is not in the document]
yuval-k: sorry for late response - just saw this comment.

I think I'm misunderstanding your intention here - Does each handler returns its key and value? wouldn't this require an interface too for the handlers?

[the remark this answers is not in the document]
howardjohn: It could return a `map<string, serde::RawValue>` or similar?

Or just a RawValue and have a `key()` method (then it can only write to 1 key, though).

(RawValue may not be the right type either, not sure)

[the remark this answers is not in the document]
yuval-k: we can do that, but if we want to use it in other places it might be useful to have it return a future, for async invocations. for example, `handle_pprof` must be async

linsun:
A friendly ping to @howardjohn when he gets a min since he said should be good to go yesterday. This is blocking a few other PRs as well.

howardjohn (approved):
Looks like a good start to iterate on

PR #751, opened 2023-12-14
xds: fix removals

howardjohn:
We were using the wrong key for the map, so removals never happened.
This is something that can be easily tested, but we need a bit of
scaffolding I don't have bandwidth to cover right now. I will open an
issue and try to get it covered.

Also, our update logic is delete+add, and we log that -- this gives
super confusing logs. Make it so we only log 'delete' on real removals.

### DIFF
--- a/src/xds.rs
+++ b/src/xds.rs
@@ -132,7 +132,6 @@ impl ProxyStateUpdater {
     }
 
     pub fn remove(&self, xds_name: &String) {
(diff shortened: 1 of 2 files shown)


bleggett (approved):
LGTM if tests are pending.

ilrudie (approved):
LGTM, Agreed on the super confusing logs point

hzxuzhonghu (approved):

PR #754, opened 2023-12-18
make sure that authorization policies XDS is never in on-demand mode

yuval-k:
As auth policies have deny rules, they must all be present in the ztunnel, even in on demand mode.

The PR changes the ztunnel behavior to never issue an on-demand XDS for auth policies

### DIFF
--- a/src/xds/client.rs
+++ b/src/xds/client.rs
@@ -144,7 +144,7 @@ impl<T: 'static + prost::Message + Default> RawHandler for HandlerWrapper<T> {
                 state.add_resource(key.type_url, key.name);
                 r
             })
-            .map(|raw| decode_proto::<T>(&raw).unwrap())
+            .map(|raw| decode_proto::<T>(raw).unwrap())
             .map(XdsUpdate::Update)
             .chain(removes.into_iter().map(XdsUpdate::Remove))
             .collect();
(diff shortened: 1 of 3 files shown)


src/xds/client.rs
```diff
@@ -141,7 +144,7 @@ impl<T: 'static + prost::Message + Default> RawHandler for HandlerWrapper<T> {
                 state.add_resource(key.type_url, key.name);
                 r
             })
-            .map(|raw| decode_proto::<T>(raw).unwrap())
+            .map(|raw| decode_proto::<T>(&raw).unwrap())
```
howardjohn, line 147: why this change?  I think this means we do a clone() on the full proto which we shouldn't need?

yuval-k: we don't need this changed now, i left there by mistake from another change I'm working on. s

It should matter in terms of `clone()`ing though. You can see that `decode_proto` doesn't do any additional clone() operations

If you prefer i can revert it from here and add it in the next PR I'm working on.

yuval-k: sorry, `decode_proto` does clone the name now; i'll revert it from here and add it back in context so it will make more sense

howardjohn (approved):

src/xds/client.rs
howardjohn: nvm, misread it

howardjohn: I put hold incase you want to fix, feel free to remove

yuval-k:
fixed and removed the hold

Issue #666, opened 2023-08-24
XDS Store locking


howardjohn:
In https://github.com/istio/ztunnel/pull/545, we accidentally moved from a per-xds-update lock to more fine grained locking.

We should be explicit about how we want to lock.

In some inprecise testing I found no difference in performance between the fine grained locking vs wide locking. However, I am a bit concerned about correctness. 

Consider a `remove ep1; add ep2` push. With coarse locking, there is always 1 ep. With fine grained locking, there is a time period where there is 0 and we may see connectivity issues. That particular case is not so bad, but can be worse in other cases perhaps

howardjohn:
https://github.com/istio/ztunnel/pull/757

PR #757, opened 2023-12-19
Make proxy cache updates consistent

yuval-k:
Make proxy cache updates consistent by taking the lock on the entire update operation.
so that things like 

```
{
state.remove()
state.insert()
}
```

are not racy.


howardjohn (approved):
Awesome, thanks!

Fixes https://github.com/istio/ztunnel/issues/666

Issue #752, opened 2023-12-14
ztunnel stuggles to reconnect to istiod


howardjohn:
I assume this happens on scale to 0 then back to 1

```
2023-12-14T19:28:45.690169Z  WARN xds{id=10}: ztunnel::xds::client: XDS client connection error: gRPC connection error:status: Unknown, message: "client error (SendRequest)", source: connection error: stream closed because of a broken pipe, retrying in 20ms
2023-12-14T19:28:45.713395Z  WARN xds{id=11}: ztunnel::xds::client: XDS client connection error: gRPC connection error:status: Unknown, message: "client error (Connect)", source: tcp connect error: Connection refused (os error 111), retrying in 40ms
2023-12-14T19:28:45.756348Z  WARN xds{id=12}: ztunnel::xds::client: XDS client connection error: gRPC connection error:status: Unknown, message: "client error (Connect)", source: tcp connect error: Connection refused (os error 111), retrying in 80ms
2023-12-14T19:28:45.839700Z  WARN xds{id=13}: ztunnel::xds::client: XDS client connection error: gRPC connection error:status: Unknown, message: "client error (Connect)", source: tcp connect error: Connection refused (os error 111), retrying in 160ms

2023-12-14T19:31:01.050492Z  WARN xds{id=14}: ztunnel::xds::client: XDS client connection error: gRPC connection error:status: Unknown, message: "client error (Connect)", source: tcp connect error: Connection timed out (os error 110), retrying in 320ms
2023-12-14T19:31:01.376257Z  INFO ztunnel::xds::client: sending initial request resources=16 type_url="type.googleapis.com/istio.workload.Address"
```

Istiod was up quickly, but takes almost 2min to reconnect. I think we just need a short tcp dial timeout

ilrudie:
I've noticed this during installation too. Probably less of a concern since we're not advising use of the ambient profile for production installations but if zt starts to come up quickly enough and fails the client connection it occasionally take a really long time for zt to recover.

ilrudie:
we may just be able to set an http connect timeout. Taking a quick look at the underlying implementation it seems like it could work and should be a small PR

hzxuzhonghu:
how about sidecar now

howardjohn:
I've never seem similar in sidecar, but I am not sure exactly what is
triggering it to not happen

ilrudie:
I have not seen a sidecar do this either

linsun:
[...]

PR #763, opened 2024-01-08
add a connect timeout to grpc_connector

ilrudie:
related #752

Add a connect timeout for grpc connector to prevent long reconnect delays

### DIFF
--- a/src/tls/boring.rs
+++ b/src/tls/boring.rs
@@ -309,7 +309,7 @@ pub fn grpc_connector(
     http.set_keepalive(Some(Duration::from_secs(300)));
     http.set_keepalive_interval(Some(Duration::from_secs(75)));
     http.set_keepalive_retries(Some(9));
-    http.set_connect_timeout(Some(Duration::from_millis(100)));
+    http.set_connect_timeout(Some(Duration::from_secs(5)));
     http.enforce_http(false);
     let mut https: HttpsConnector<HttpConnector> =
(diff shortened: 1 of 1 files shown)


src/tls/boring.rs
```diff
@@ -309,6 +309,7 @@ pub fn grpc_connector(
     http.set_keepalive(Some(Duration::from_secs(300)));
     http.set_keepalive_interval(Some(Duration::from_secs(75)));
     http.set_keepalive_retries(Some(9));
+    http.set_connect_timeout(Some(Duration::from_millis(100)));
```
howardjohn, line 312: we probably want 10s, not 100ms. 100ms is actually pretty quick for a connection, especially cross-zone

hzxuzhonghu:
make it same as sidecar xds poroxy

src/tls/boring.rs
ilrudie: per @hzxuzhonghu's comment, it looks like we may use the default from the envoy `go-control-plane` which is `5s`. I'm not supremely confident in that but don't see where we set a connect timeout. Curious to be proven wrong if I missed it though. How do we feel about `5s`?

howardjohn: Its not from that library but still 5s: https://github.com/istio/istio/blob/4482b870e91ce66b6cc776700cb4c94f1edc8dca/pkg/istio-agent/xds_proxy.go#L334. 5s is good.

howardjohn (approved):

Issue #725, opened 2023-11-14
Exponential backoff on CA  failures


howardjohn:
Currently [here ](https://github.com/istio/ztunnel/blob/71961979f08bd4ea54eb7112742092a2c16ddb3c/src/identity/manager.rs#L303) we retry after 60s. 60s can be a long time or a short time. We should probably use a backoff scheme. There may be transient issues in the CA (e.g. the SA informer cache is a bit stale) that will resolve after 1s.. but if its never going to approve, retrying 1440 times a day may be a bit much.

MorrisLaw:
Thank you Keith, I'll look into this!

MorrisLaw:
Is there a retry library that maintainers recommend @howardjohn?

hzxuzhonghu:
It is not limited, but looks https://crates.io/crates/backoff is the most popular one

MorrisLaw:
PR is ready for review again.

MorrisLaw:
PR has been merged.

keithmattix:
Had to revert; reopening

MorrisLaw:
Had to revert due to an issue noticed in an automated Istio PR that tried updating to the latest stable ztunnel master. TestIntermediateCertificateRefresh failed due to missing cert. It looks like it wasn't refreshed in time or it could be a possible race condition.

Related areas of the code to look at as it relates to the integration test that were changed before and are relevant to the backoff PR, these may require modification in order to support this change:
- https://github.com/istio/ztunnel/pull/742/files
- https://github.com/istio/istio/commit/1ee1a434356b1ba94c3e763886dc03b064ebd925

Goal is to reproduce the integration test failure with the original PR that has been reverted, then figure out how to update the integration test or the business logic itself so that the integration test no longer fails.

MorrisLaw:
This has been merged back in. Found the[ issue in an integration test](https://github.com/istio/istio/pull/50043) and fixed it. Also, confirmed that istio integration is happy with this latest version of ztunnel: https://github.com/istio/istio/pull/50079/files. We can close this for real now.

PR #768, opened 2024-01-11
Add retry backoff to cert refresh

MorrisLaw:
Add backoff retry logic as the mechanism for determining the refresh_at value for fetching certificates.

Related to issue #725.

### DIFF
--- a/src/identity/caclient.rs
+++ b/src/identity/caclient.rs
@@ -218,6 +218,24 @@ pub mod mock {
             state.fetches.push(id.to_owned());
             Ok(certs)
         }
+
+        pub async fn set_error(&mut self, error: bool) {
+            if error {
+                let mut state = self.state.write().await;
+                state.fetches.push(Identity::Spiffe {
+                    trust_domain: "error".to_string(),
+                    namespace: "error".to_string(),
+                    service_account: "error".to_string(),
+                });
+            } else {
+                let mut state = self.state.write().await;
+                state.fetches.push(Identity::Spiffe {
+                    trust_domain: "success".to_string(),
+                    namespace: "success".to_string(),
+                    service_account: "success".to_string(),
+                });
+            }
+        }
     }
 
     #[async_trait]
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -242,9 +242,10 @@ impl Worker {
             current_interval: Duration::from_secs(1),
             max_interval: Duration::from_secs(60),
             // cert_backoff.max_elapsed_time is the maximum elapsed time after instantiating
-            // [`ExponentialBackoff`](struct.ExponentialBackoff.html) or calling
-            // [`reset`](trait.Backoff.html#method.reset) after which [`next_backoff`](../trait.Backoff.html#method.reset)
-            // returns `None`. More info can be found here: https://docs.rs/backoff/0.4.0/backoff.
+            // [`ExponentialBackoff`](https://docs.rs/backoff/0.4.0/backoff/type.ExponentialBackoff.html)
+            // or calling [`reset`](https://docs.rs/backoff/0.4.0/backoff/backoff/trait.Backoff.html#method.reset)
+            // after which [`next_backoff`](https://docs.rs/backoff/0.4.0/backoff/backoff/trait.Backoff.html#method.next_backoff)
+            // returns `None`.
             //
             // We set it to 5 minutes (300 seconds) to ensure that we don't retry indefinitely.
             max_elapsed_time: Some(Duration::from_secs(300)),
@@ -331,9 +332,9 @@ impl Worker {
                             //     retry_interval * (random value in range [1 - randomization_factor, 1 + randomization_factor])
                             let refresh_at = Instant::now() + cert_backoff.next_backoff().unwrap_or(CERT_REFRESH_FAILURE_RETRY_DELAY_MAX_INTERVAL);
                             // cert_backoff.start_time is the system time. It is calculated when an
-                            // [`ExponentialBackoff`](struct.ExponentialBackoff.html) instance is created
-                            // and is reset when [`retry`](../trait.Operation.html#method.retry) is called.
-                            // More info can be found here: https://docs.rs/backoff/0.4.0/backoff.
+                            // [`ExponentialBackoff`](https://docs.rs/backoff/0.4.0/backoff/type.ExponentialBackoff.html)
+                            // instance is created and is reset when [`reset`](https://docs.rs/backoff/0.4.0/backoff/backoff/trait.Backoff.html#method.reset)
+                            // is called.
                             let start_time = cert_backoff.start_time;
                             let elapsed_duration = start_time.elapsed();
                             let time_difference = elapsed_duration;
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -240,15 +240,9 @@ impl Worker {
         let mut cert_backoff = ExponentialBackoff {
             initial_interval: Duration::from_millis(500),
             current_interval: Duration::from_secs(1),
-            max_interval: Duration::from_secs(60),
-            // cert_backoff.max_elapsed_time is the maximum elapsed time after instantiating
-            // [`ExponentialBackoff`](https://docs.rs/backoff/0.4.0/backoff/type.ExponentialBackoff.html)
-            // or calling [`reset`](https://docs.rs/backoff/0.4.0/backoff/backoff/trait.Backoff.html#method.reset)
-            // after which [`next_backoff`](https://docs.rs/backoff/0.4.0/backoff/backoff/trait.Backoff.html#method.next_backoff)
-            // returns `None`.
-            //
-            // We set it to 5 minutes (300 seconds) to ensure that we don't retry indefinitely.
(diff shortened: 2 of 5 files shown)


src/identity/manager.rs
```diff
@@ -287,6 +293,14 @@ impl Worker {
                     },
                     None => break 'main,
                 },
+
+                // Update backoff to have values more appropriate for certificate refresh failures.
+                backoff.initial_interval = CERT_REFRESH_FAILURE_RETRY_DELAY_INITIAL_INTERVAL; // start with 500ms
+                backoff.max_interval = CERT_REFRESH_FAILURE_RETRY_DELAY_MAX_INTERVAL; // max out at 60s
+                backoff.multiplier = 2.0; // double the interval each time
+                backoff.randomization_factor = 0.2; // 20% randomization
+                backoff.max_elapsed_time = Some(Duration::from_secs(300)); // max out at 5 minutes
```
howardjohn, line 302: I think this is broken. After 5min it returns None and therefor CERT_REFRESH_FAILURE_RETRY_DELAY_INITIAL_INTERVAL -- not 5 minutes. 

Experienced learned from outage in the Go equivilent https://github.com/istio/istio/blob/50fe4b86425bd106810d323280dd5f8098fb082b/pkg/backoff/exponential.go#L77 :slightly_smiling_face:

src/identity/manager.rs
```diff
@@ -296,10 +310,14 @@ impl Worker {
                     }
                     let (state, refresh_at) = match res {
                         Err(err) => {
-                            let refresh_at = Instant::now() + CERT_REFRESH_FAILURE_RETRY_DELAY;
+                            // Use the next backoff to determine when to retry the fetch and default to the constant value if the backoff has been reset.
+                            let refresh_at = Instant::now() + backoff.next_backoff().unwrap_or(CERT_REFRESH_FAILURE_RETRY_DELAY_INITIAL_INTERVAL);
```
howardjohn, line 314: Hmm. I am a little worried about this if I understand right. Is this the behavior:

1. fetch cert1. Fails, enqueue with 500ms wait
2. fetch cert2. Fails, enqueue with 1s wait
...
3. fetch cert100. Fails, enqueue with 5m wait
4. fetch cert101. success! reset backoff
5.  ...cert100 only retries in 5m...

Two possible concerns:
* Each new cert ignores the backoff
* When we start succeeding again, all of our previously failed certs are stuck for a while

WDYT?

I guess it really depends on if the issue is caused by CA connectivity in general or that specific certificate. Both are possible

src/identity/manager.rs
MorrisLaw: Ahh yes, good catch!

src/identity/manager.rs
MorrisLaw: The current parameters (specifically, max_elapsed_time) dictate that the total amount of time for all retries can't be longer than 5 minutes. But any one individual retry interval has a max of 60 seconds.

So iiuc, worst case would make it so that something retries for 60 seconds until the total time of 5 minutes has elapsed, then next_backoff starts returning none.

But I'm a bit confused now with the thing you shared before about the Go equivalent, if we set it to max_interval when it hits None. Won't it basically retry for 60 seconds indefinitely?

Maybe I should add some logic here to error out and have specific errors for what you've described:
```
I guess it really depends on if the issue is caused by CA connectivity in general or that specific certificate. Both are possible
```

At some point we'd want to just give up entirely and stop retrying. I'll have to spend some more time trying to understand this logic though. It's a bit confusing to me right now.

howardjohn: The code, AFAIK, does not stop once it hits the max. `.unwrap_or(CERT_REFRESH_FAILURE_RETRY_DELAY_INITIAL_INTERVAL)`

src/identity/manager.rs
```diff
@@ -234,20 +234,24 @@ impl Worker {
...
-        cert_backoff.randomization_factor = 0.2;
+        // The backoff strategy used for retrying operations. Sets the initial values for the backoff.
+        // The values are chosen to be reasonable for the CA client to be able to recover from transient
+        // errors.
+        let mut cert_backoff = ExponentialBackoff {
+            initial_interval: Duration::from_millis(500),
+            current_interval: Duration::from_secs(1),
+            max_interval: Duration::from_secs(60),
+            // cert_backoff.max_elapsed_time is the maximum elapsed time after instantiating
+            // [`ExponentialBackoff`](struct.ExponentialBackoff.html) or calling
```
keithmattix, line 245: Do these links work?

MorrisLaw: No. Just pushed up the FQDNs for those links in the comments. Should be good now.

src/identity/manager.rs
```diff
@@ -232,6 +234,25 @@ impl Worker {
...
+            initial_interval: Duration::from_millis(500),
+            current_interval: Duration::from_secs(1),
+            max_interval: Duration::from_secs(60),
+            // cert_backoff.max_elapsed_time is the maximum elapsed time after instantiating
+            // [`ExponentialBackoff`](https://docs.rs/backoff/0.4.0/backoff/type.ExponentialBackoff.html)
+            // or calling [`reset`](https://docs.rs/backoff/0.4.0/backoff/backoff/trait.Backoff.html#method.reset)
+            // after which [`next_backoff`](https://docs.rs/backoff/0.4.0/backoff/backoff/trait.Backoff.html#method.next_backoff)
+            // returns `None`.
+            //
+            // We set it to 5 minutes (300 seconds) to ensure that we don't retry indefinitely.
```
howardjohn, line 250: Do we want to stop? I worry if we have a 5 min outage of the CA (not unreasonable) we will never recover?

src/identity/caclient.rs
```diff
@@ -218,6 +218,24 @@ pub mod mock {
             state.fetches.push(id.to_owned());
             Ok(certs)
         }
+
+        pub async fn set_error(&mut self, error: bool) {
+            if error {
+                let mut state = self.state.write().await;
+                state.fetches.push(Identity::Spiffe {
```
keithmattix, line 225: This is a test helper function right? I would suggest moving it into a separate test file or at the very least putting it near the actual tests that use it so that its purpose is clearer

src/identity/manager.rs
keithmattix: +1 - retry indefinitely but not too frequently

src/identity/caclient.rs
MorrisLaw: I believe [clear_fetches](https://github.com/istio/ztunnel/blob/6d9a457ed533d10b3bc618d7aaa81df8ed2fef7a/src/identity/manager.rs#L819), the function right above this is also a test helper function.

src/identity/manager.rs
MorrisLaw: Should we not have a max_elapsed_time at all and just retry indefinitely w/ no panic **or** should we pick a higher time, like 30 minutes (not sure if this is too high or unreasonable) and maybe even a slightly higher max_interval and lower initial_interval?

howardjohn: Indefinitely with a 1-5min timeout (not sure exactly what, but something in that range?) seems reasonable. Anytime we 'give up' we would never recover which seems pretty bad

src/identity/caclient.rs
keithmattix: I'm not sure if it's a test helper or just a function used outside of tests. clear_fetches looks generally useful

MorrisLaw: The only use of `clear_fetches` I see is in [test_priority in the manager.rs](https://github.com/istio/ztunnel/blob/6d9a457ed533d10b3bc618d7aaa81df8ed2fef7a/src/identity/manager.rs#L819) file.

The way I'm currently setting the error in that `set_error`, I think that if I wanted this function outside of this file I'd have to copy some of the structs like ClientState, ClientConfig and CaClient and the necessary imports over to manager.rs. Is that the suggestion or are you suggesting more simple/elegant?

src/identity/manager.rs
MorrisLaw: I can remove the max_elapsed_time and changed max_interval to 2.5 minutes. The exponential retry logic will calculate new backoff intervals on error until it hits either a success or the max_interval. Once the max_interval is hit, it will continue to do 2.5 minute waits until a cert is successful which will trigger the backoff reset and the exponential calculations start back at the initial_interval next time around for each cert.


The backoff crates reset function that will be called:
```
fn reset(&mut self) {
    self.current_interval = self.initial_interval;
    self.start_time = self.clock.now();
}
```

Since we're setting refresh_at to 2.5 minutes in the worst case, that could mean that we'd still be waiting on some certs to finish their wait in the pending priority queue.

Wdyt of this @howardjohn @keithmattix ?

keithmattix: 2.5 minutes worst case in the queue is acceptable IMO ; even the k8s crashloopbackoff is higher I think

src/identity/caclient.rs
keithmattix: Fair enough; not blocking on it

keithmattix (approved):
This LGTM, but of course other ztunnel experts need to approve as well

howardjohn (approved):

Issue #311, opened 2023-01-04
Determine if RBAC updates should drain


howardjohn:
Consider a long lived connection. After it is established, we create a rule that would have denied it. Should the connection be terminated, or should only new connections start being terminated?

Prior art:
* Istio: Drains
* Cilium: operates on many layers, so needs more research
* Ztunnel currently: does not drain

keithmattix:
Definitely feels like it should drain IMO

GregHanson:
this also impact waypoints.  If a long lived connection exists, and a waypoint is deployed, ztunnel will not direct traffic through the waypoint until the original connection is closed and re-established.

steps to reproduce using [online boutique](https://github.com/GoogleCloudPlatform/microservices-demo/tree/main) microservice:
```
istioctl install --set tag=1.18.0 -y --set profile=ambient --set meshConfig.accessLogFile=/dev/stdout
kubectl label namespace default istio.io/dataplane-mode=ambient --overwrite
kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/microservices-demo/main/release/kubernetes-manifests.yaml

# deploy namespace wide waypoint
istioctl x waypoint apply -n default

# no access logs
kubectl logs deploy/namespace-istio-waypoint -f

kubectl delete pod -lapp=loadgenerator

# access logs generated per request as expected
kubectl logs deploy/namespace-istio-waypoint -f
```

kyessenov:
I don't understand what "drain" here means. There's no general drain strategy for TCP connections. You can only abruptly terminate them. Is that what's proposed?

howardjohn:
Yes, drain was poorly worded. Although fwiw I'm not proposing it, just that
we make an explicit decision - I am leaning towards not terminating

kyessenov:
I think never terminating that would be dangerous in an eventually consistent system. When workloads and policies are deployed concurrently (enabling waypoint is a policy), the order should not matter when things open connections. So it seems necessary to terminate eventually, and the challenge is defining when to terminate connections and limiting the scope to only those should be.

linsun:
Agreed, I'd expect updated RBAC become effective even with long lived connection.

cc @louiscryan for input

howardjohn:
I agree with the concerns of eventual consistency. For RBAC it makes more sense I think - we explicitly would close the connection if it was new. For Waypoint being added, though, closing the connection seems unneccesarily aggressive

bleggett:
[...]

PR #772, opened 2024-01-16
close connections which violate policy after updates

ilrudie:
related #311 

Adds:
- tracking open connections for `inbound` and `inbound_passthrough`
- notifications to trackers when auth policy changes occur
- separate task handles change notifications, asserts policy against the existing connections and closes open connections which are now denied by policy

### DIFF
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -33,7 +33,7 @@ use super::Error::{self, Spiffe};
 
 const CERT_REFRESH_FAILURE_RETRY_DELAY: Duration = Duration::from_secs(60);
 
-#[derive(Debug, PartialEq, Eq, Clone, Hash, Ord)]
+#[derive(Debug, PartialEq, Eq, Clone, Hash)]
 pub enum Identity {
     Spiffe {
         trust_domain: String,
@@ -42,12 +42,18 @@ pub enum Identity {
     },
 }
 
-impl PartialOrd for Identity {
+impl Ord for Identity {
     // Not sure this is a super legit compare but I think it should work for POC
-    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+    fn cmp(&self, other: &Self) -> Ordering {
         let s = format!("{self}");
         let o = format!("{other}");
-        s.partial_cmp(&o)
+        s.cmp(&o)
+    }
+}
+
+impl PartialOrd for Identity {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
     }
 }
 
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -42,6 +42,15 @@ pub enum Identity {
     },
 }
 
+impl PartialOrd for Identity {
+    // Not sure this is a super legit compare but I think it should work for POC
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        let s = format!("{self}");
+        let o = format!("{other}");
+        s.partial_cmp(&o)
+    }
+}
+
 impl EncodeLabelValue for Identity {
     fn encode(&self, writer: &mut LabelValueEncoder) -> Result<(), std::fmt::Error> {
         writer.write_str(&self.to_string())
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -33,7 +33,7 @@ use super::Error::{self, Spiffe};
 
 const CERT_REFRESH_FAILURE_RETRY_DELAY: Duration = Duration::from_secs(60);
 
-#[derive(Debug, PartialEq, Eq, Clone, Hash)]
+#[derive(Debug, PartialEq, Eq, Clone, Hash, Ord)]
 pub enum Identity {
     Spiffe {
         trust_domain: String,
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -33,28 +33,13 @@ use super::Error::{self, Spiffe};
 
 const CERT_REFRESH_FAILURE_RETRY_DELAY: Duration = Duration::from_secs(60);
 
-#[derive(Debug, PartialEq, Eq, Clone, Hash)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Hash)]
 pub enum Identity {
     Spiffe {
         trust_domain: String,
         namespace: String,
         service_account: String,
     },
-}
-
-impl Ord for Identity {
-    // Not sure this is a super legit compare but I think it should work for POC
-    fn cmp(&self, other: &Self) -> Ordering {
-        let s = format!("{self}");
-        let o = format!("{other}");
-        s.cmp(&o)
-    }
-}
-
-impl PartialOrd for Identity {
-    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        Some(self.cmp(other))
-    }
 }
 
 impl EncodeLabelValue for Identity {
--- a/src/proxy/connection_manager.rs
+++ b/src/proxy/connection_manager.rs
@@ -90,3 +90,83 @@ impl ConnectionManager {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
+    use drain::Watch;
+    use std::net::{Ipv4Addr, SocketAddrV4};
+    use std::time::Duration;
+
+    use super::ConnectionManager;
+    use crate::rbac::Connection;
+
+    #[tokio::test]
+    async fn test_connection_manager() {
+        // setup a new ConnectionManager
+        let connection_manager = ConnectionManager::new();
+        // ensure drains is empty
+        assert_eq!(connection_manager.drains.read().await.len(), 0);
+        assert_eq!(connection_manager.connections().await.len(), 0);
+
+        // track a new connection
+        let conn1 = Connection {
+            src_identity: None,
+            src_ip: std::net::IpAddr::V4(Ipv4Addr::new(192, 168, 0, 1)),
+            dst_network: "".to_string(),
+            dst: std::net::SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(192, 168, 0, 2), 8080)),
+        };
+        let close1 = connection_manager.clone().track(&conn1).await;
+        // ensure drains contains exactly 1 item
+        assert_eq!(connection_manager.drains.read().await.len(), 1);
+        assert_eq!(connection_manager.connections().await.len(), 1);
+        assert_eq!(connection_manager.connections().await, vec!(conn1.clone()));
+
+        // setup a second track on the same connection
+        let another_conn1 = conn1.clone();
+        let another_close1 = connection_manager.clone().track(&another_conn1).await;
+        // ensure drains contains exactly 1 item
+        assert_eq!(connection_manager.drains.read().await.len(), 1);
+        assert_eq!(connection_manager.connections().await.len(), 1);
+        assert_eq!(connection_manager.connections().await, vec!(conn1.clone()));
+
+        // track a second connection
+        let conn2 = Connection {
+            src_identity: None,
+            src_ip: std::net::IpAddr::V4(Ipv4Addr::new(192, 168, 0, 3)),
+            dst_network: "".to_string(),
+            dst: std::net::SocketAddr::V4(SocketAddrV4::new(Ipv4Addr::new(192, 168, 0, 2), 8080)),
+        };
+        let close2 = connection_manager.clone().track(&conn2).await;
+        // ensure drains contains exactly 2 items
+        assert_eq!(connection_manager.drains.read().await.len(), 2);
+        assert_eq!(connection_manager.connections().await.len(), 2);
+        assert_eq!(
+            connection_manager.connections().await,
+            vec!(conn1.clone(), conn2.clone())
+        );
+
+        // spawn tasks to assert that we close in a timely manner for conn1
+        tokio::spawn(async_close_assert(close1));
+        tokio::spawn(async_close_assert(another_close1));
+        // close conn1
+        connection_manager.close(&conn1).await;
+        // ensure drains contains exactly 1 item
+        assert_eq!(connection_manager.drains.read().await.len(), 1);
+        assert_eq!(connection_manager.connections().await.len(), 1);
+        assert_eq!(connection_manager.connections().await, vec!(conn2.clone()));
+
+        // spawn a task to assert that we close in a timely manner for conn2
+        tokio::spawn(async_close_assert(close2));
+        // close conn2
+        connection_manager.close(&conn2).await;
+        // assert that drains is empty again
+        assert_eq!(connection_manager.drains.read().await.len(), 0);
+        assert_eq!(connection_manager.connections().await.len(), 0);
+    }
+
+    // small helper to assert that the Watches are working in a timely manner
+    async fn async_close_assert(c: Watch) {
+        let result = tokio::time::timeout(Duration::from_secs(1), c.signaled()).await;
+        assert!(matches!(result, Ok(_)));
+    }
+}
--- a/src/proxy/connection_manager.rs
+++ b/src/proxy/connection_manager.rs
@@ -84,14 +84,6 @@ impl ConnectionManager {
         // potentially large copy under read lock, could require optomization
         self.drains.read().await.keys().cloned().collect()
     }
-
-    #[allow(dead_code)]
-    pub async fn drain_all(self) {
-        let mut drains = self.drains.write_owned().await;
-        for (_conn, cd) in drains.drain() {
-            cd.drain().await;
-        }
-    }
 }
 
 pub async fn policy_watcher(
@@ -120,13 +112,20 @@ pub async fn policy_watcher(
 }
 
 #[cfg(test)]
-mod test {
+mod tests {
     use drain::Watch;
     use std::net::{Ipv4Addr, SocketAddrV4};
+    use std::sync::{Arc, RwLock};
     use std::time::Duration;
+    use tokio::sync::watch;
+    use trust_dns_resolver::config::{ResolverConfig, ResolverOpts};
+
+    use crate::rbac::Connection;
+    use crate::state::{DemandProxyState, ProxyState};
+    use crate::xds::istio::security::{Action, Authorization, Scope};
+    use crate::xds::ProxyStateUpdateMutator;
 
     use super::ConnectionManager;
-    use crate::rbac::Connection;
 
     #[tokio::test]
     async fn test_connection_manager() {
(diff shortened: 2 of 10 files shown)


src/proxy/connection_manager.rs
```diff
@@ -0,0 +1,87 @@
...
+                drains.insert(c.clone(), cd);
+                rx
+            }
+        }
+    }
+
+    pub async fn drain(&self, c: &Connection) {
+        match self.drains.clone().write().await.remove(c) {
+            Some(cd) => {
+                cd.tx.drain().await;
```
howardjohn, line 67: is the lock held here? If so, is drain() possibly blocking?

src/proxy/inbound.rs
```diff
@@ -94,12 +98,35 @@ impl Inbound {
...
+            loop {
+                tokio::select! {
+                    _ = stop_rx.changed() => {
+                        return
+                    }
+                    _ = policies_changed.changed() => {
+                        let connections = connection_manager.connections().await;
+                        for conn in connections {
+                            if !state.assert_rbac(&conn).await {
+                                connection_manager.drain(&conn).await;
```
howardjohn, line 117: This definitely needs some nice logging since it will be a pretty major "WTF just happened" in some cases

howardjohn (commented):
Mostly skimmed it so far but looks reasonable overall

src/proxy/connection_manager.rs
ilrudie: I don't think the write lock should be held here. Remove totally removes a value from the HashMap returning the owned value (not a reference) so there's no writing beyond the single remove and no reference to the lock required. Now you've got me doubting that understanding a bit though 😅

ilrudie: It's definitely possible to make that more explicit though if we want.

src/proxy/inbound.rs
ilrudie: firm agree. TY

src/proxy/connection_manager.rs
ilrudie:
```rust
        let w = self.drains.clone().write().await;
        let cd = w.remove(c);
        drop(w);

        match cd {
            ...
        }
```

howardjohn: The lock should be released when the value returned from `write()` is dropped. I am just not sure if its drops when the match body (L67) is done or when the match check (L65) is done

ilrudie:
```rust
        let cd = self.drains.clone().write().await.remove(c);
        match cd {...}
```

ilrudie: probably better break it out like that to align with the standards we have for zt... prioritize understandability for our reviewers over the most idiomatic looking rustlang

ilrudie: It turns out the lock is held through the duration of the match. This lead to a deadlock in the track method where the read lock was being held for the entire match and one branch was trying to get a write lock.

bleggett: Use of a central RWLocked connection tracker pretty much means by definition that we will be blocking on add/remove ops, I think.

If that's not acceptable, inverting this via channels is probably the other way.

src/proxy/connection_manager.rs
```diff
@@ -0,0 +1,87 @@
...
+                // this is bad, possibly drain called twice
+                error!("requested drain on a Connection which wasn't initialized");
+            }
+        }
+    }
+
+    pub async fn connections(&self) -> Vec<Connection> {
+        // potentially large copy under read lock, could require optomization
+        self.drains.read().await.keys().cloned().collect()
+    }
```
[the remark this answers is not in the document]
ilrudie: We could consider that but I was a little worried about interlocking. We'd need to await locking on the PolicyStore (likely multiple times) inside this lock if we assert_rbac in here I think.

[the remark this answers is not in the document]
bleggett: We shouldn't have to lock state, once we get a change event we can (at least) clone it and provide it to the conn_manager to apply in an iterative loop.

Now whether that clone is cheaper than this clone is another question, but in general if we go with this approach I think eventually `assert_rbac` should live in `connection_manager.rs` and not `state`, anyway, and we can probably avoid cloning the entire state.

src/proxy/connection_manager.rs
ilrudie: that's fair. we could do some reworking of how policy is applied and retained to ease or eliminate interlocking problems if need be. My thinking was sort of limited to not making those kind changes which maybe is unnecessarily narrow

src/proxy/connection_manager.rs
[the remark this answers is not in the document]
ilrudie: that's fair. we could do some reworking of how policy is applied and retained to ease or eliminate interlocking problems if need be. My thinking was sort of limited to not making those kind changes which maybe is unnecessarily narrow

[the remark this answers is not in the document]
bleggett: This is fine for now. If the centralized rwlock'd conntrack becomes a problem/bottleneck, we can just remove it.

src/proxy/connection_manager.rs
```diff
@@ -0,0 +1,91 @@
...
+        // try_insert is the best solution once it's no longer a nightly-only experimental API
+        let mut drains = self.drains.write().await;
+        if let Some(w) = drains.get(c) {
+            return w.rx.clone();
+        }
+        drains.insert(c.clone(), cd);
+        rx
+    }
+
+    pub async fn drain(&self, c: &Connection) {
```
hzxuzhonghu, line 71: is there a graceful draining time?

ilrudie: That's still TBD

ilrudie: I think there are really 2 behaviors we want here:
1. drain: used when ztunnel is shutting down. sends goaway and continues to handle traffic
2. close: used when auth policy changes and existing connections become denied. stop handling traffic

ilrudie: @howardjohn and @kyessenov, does this approach seem reasonable?

kyessenov: I don't see any drain here. This is just local close?

ilrudie: As it sits now even when policy changes from allow to deny we send a goaway and continue to handle traffic. We'd originally implemented drains in this way because our only connection drains were when ztunnel was being shutdown and in that case we decided that k8s killing the pod would be our drain timeout.

From an auth deny perspective this didn't seem correct though. I'm just about ready to push some new code which has both drain (for shutdown) and close (for auth) implemented for inbound.

kyessenov: Let's not conflate HBONE goaway with application drain. HBONE goaway is an infastructure mechanism and should not be used for application intent - it's like reinstalling a cluster when redeploying an app.

For application policies, there's no drain or GOAWAY here AFAICT. You should just close the connection.

kyessenov: By close, I mean RST. I think we shouldn't do half close here.

src/proxy/connection_manager.rs
```diff
@@ -54,24 +56,37 @@ impl ConnectionManager {
...
 
-    pub async fn close(&self, c: &Connection) {
+    // releases tracking on a connection
+    // uses a counter to determine if there are other tracked connections or not so it may retain the tx/rx channels when necessary
+    pub async fn release(self, c: &Connection) {
+        let mut drains = self.drains.write().await;
+        if let Some((k, v)) = drains.remove_entry(c) {
+            if v.count.fetch_sub(1, std::sync::atomic::Ordering::SeqCst) > 1 {
+                // something else is tracking this connection, retain
+                drains.insert(k, v);
```
bleggett, line 83: We talked about this already - but `drain-rs` is a _very_ light and rather inflexible wrapper around plain old tokio channels, and I suspect there might be other ways to accomplish this by using tokio channels directly (different channel type, or something like `WeakSender`), without the manual external atomic count.

But the effect would be the same so it is probably not worth it, this will do the job AFAICT - but if `drain-rs` gives you trouble and it's easier to drop, I don't have a problem with ditching it.

ilrudie: It's a fair idea. I'm not sure we strictly need the bidirectional close functionality the crate offers for this case... in fact we may prefer a fire and forget since for our purposes right now we don't actually care if the receiver get's a signal or an error. We're sending 1 bit of information which just means do your thing... At any rate this is certainly something to consider.

ilrudie:
@howardjohn, @hzxuzhonghu and @kyessenov; if you get a chance to take another look and provide feedback that would be greatly appreciated. TYIA

src/proxy/connection_manager.rs
```diff
@@ -0,0 +1,359 @@
...
+    let mut policies_changed = state.read().policies.subscribe();
+    loop {
+        tokio::select! {
+            _ = stop_rx.changed() => {
+                break;
+            }
+            _ = policies_changed.changed() => {
+                let connections = connection_manager.connections().await;
+                for conn in connections {
+                    if !state.assert_rbac(&conn).await {
```
kyessenov, line 120: Sorry, I'm not a Rust expert - does this await individually or globally? Any time there's IO, you probably don't want to stall in a loop in a switch - you may overflow the pending events because IO is slow. It's best to fire-and-forget and then let someone reap the zombies.

ilrudie: This will await the individual assertion against the one connection but await doesn't mean block (there is a specific block function if that's what you want though). If the Future being awaited is not ready it relinquishes control and will be polled later by the executor of the task/thread.

Rusts Futures are somewhat unique in that they are lazy. If nothing drives them to completion they will not do any work so we can't totally fire and forget. We can either await them and let the executor/callback model poll them or we could collect the all futures and join them later. If we do neither no work would happen though.

ilrudie:
cc @stevenctl

src/proxy/connection_manager.rs
kyessenov: I feel like you do want to block the loop or at least debounce it to avoid edge-trigerred event spam in the runtime. But I have no idea how to express that in Rust. Maybe a separate actor that is "fast" in consuming policy events but "slow" in enacting them.

ilrudie: The code you are commenting on is _sort of_ that actor (in concept at least). It's being spun up in its own task so it should be (at least somewhat) independent of the tasks which are actually moving data.

src/proxy/connection_manager.rs
yuval-k: we can also explore using linked lists for less locking

src/identity/manager.rs
```diff
@@ -42,6 +42,21 @@ pub enum Identity {
     },
 }
 
+impl Ord for Identity {
+    // Not sure this is a super legit compare but I think it should work for POC
```
howardjohn, line 46: given this is no long POC should we do a more robust approach?

src/proxy/connection_manager.rs
```diff
@@ -0,0 +1,359 @@
...
+            drains: Arc::new(RwLock::new(HashMap::new())),
+        }
+    }
+
+    // register a connection with the manager and get a channel to receive on
+    pub async fn track(self, c: &Connection) -> drain::Watch {
+        let cd = ConnectionDrain::new();
+        let rx = cd.rx.clone();
+        let mut drains = self.drains.write().await;
+        if let Some(w) = drains.remove(c) {
```
howardjohn, line 64: nit: del+mutate+insert can be done with code like:

```
map.entry("poneyland")
   .and_modify(|e| { *e += 1 })
   .or_insert(42);
```

howardjohn: maybe not with the `return rx` aspect though

src/proxy/connection_manager.rs
```diff
@@ -0,0 +1,359 @@
...
+    // always inline, this is for convenience so that we don't forget to drop the rx but there's really no reason it needs to grow the stack
+    #[inline(always)]
+    async fn drain(self) {
+        drop(self.rx); // very important, drain cannont complete if there are outstand rx
+        self.tx.drain().await;
+    }
+}
+
+#[derive(Clone)]
+pub struct ConnectionManager {
```
howardjohn, line 48: We already have a "pool" tracking connections. It seems like there, ideally, would be a single data structure tracking connections. I guess Pool tracks outer HBONE connections while this tracks inner user connections, though. Also I don't see a feasible way to use 1 structure anyways, so mostly hypothetical.

src/proxy/connection_manager.rs
```diff
@@ -0,0 +1,359 @@
...
+        debug!("there are {} tracked connections", drains.len());
+        rx
+    }
+
+    // releases tracking on a connection
+    // uses a counter to determine if there are other tracked connections or not so it may retain the tx/rx channels when necessary
+    pub async fn release(self, c: &Connection) {
+        let mut drains = self.drains.write().await;
+        if let Some((k, v)) = drains.remove_entry(c) {
+            if v.count.fetch_sub(1, std::sync::atomic::Ordering::SeqCst) > 1 {
```
howardjohn, line 81: nit: why do we need atomic operations if we have a write lock?

src/proxy/connection_manager.rs
```diff
@@ -0,0 +1,359 @@
...
+                // something else is tracking this connection, retain
+                drains.insert(k, v);
+            }
+        }
+    }
+
+    // signal all connections listening to this channel to take action (typically terminate traffic)
```
howardjohn, line 88: comment seems confusing, or I don't understand it. Signal **all connections**? But the input is 1 connection.

Is `Connection` representing the outer hbone connection, and `drain()` waits until all inner connects close?

src/proxy/connection_manager.rs
howardjohn: If I understand the event spam, is the concern is we enqueue multiple times to trigger this code? Like while we are looping here, we get `changed()` multiple times stacked up?

If so, I think it will have at most one pending. That being said, you could still hit some spam I guess

src/state/policy.rs
```diff
@@ -57,12 +79,14 @@ impl PolicyStore {
             RbacScope::WorkloadSelector => {}
         }
         self.by_key.insert(key, rbac);
+        self.notifier.send();
```
howardjohn, line 82: Probably we want to send at most 1 per xds push. That may modify many policies at once

src/identity/manager.rs
ilrudie: Good call. I do at want to consider if this is adequate.

src/proxy/connection_manager.rs
ilrudie: You're correct. I was gunna try something a little fancier and ended going with this without removing the atomic. A lot of this logic may be just removed though if we want to switch off the drain crate onto a different type of channel as Ben suggested elsewhere.

src/proxy/connection_manager.rs
ilrudie: It's a good thought. I couldn't work out how to do it either though.

src/proxy/connection_manager.rs
ilrudie: Potentially a single rbac::connection could have multiple connections handling traffic I think. In that case they all need to be closed and this code should handle the case if it arises.

src/state/policy.rs
ilrudie: I was going to look to a follow up which adds batching and includes putting the scope onto the channel so that the receivers have less spam and can inspect the scope to determine if they need to re-assert. If we think it's critical path I could include it but if not critical perhaps a TODO would be better so it's not just in my head.

src/proxy/connection_manager.rs
ilrudie: [modified rust playground example](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=06454ee0114a4d96054c4d32f55dcb5c)

Seems like it could work. TY

ilrudie: could very well be a step too far and difficult to grok but:

```rust
    // register a connection with the manager and get a channel to receive on
    pub async fn track(self, c: &Connection) -> drain::Watch {
        let cd = ConnectionDrain::new();
        self.drains
            .write()
            .await
            .entry(c.to_owned())
            .and_modify(|cd| cd.count += 1)
            .or_insert(cd)
            .rx
            .clone()
    }
```

src/identity/manager.rs
ilrudie: Using something like this would force us to consider how Identity::X should be compared to Identity::Spiffe if/when a new identity format is added.

```rust
impl Ord for Identity {
    fn cmp(&self, other: &Self) -> Ordering {
        let s = match self {
            Identity::Spiffe {
                trust_domain,
                namespace,
                service_account,
            } => trust_domain.to_owned() + namespace + service_account,
        };
        let o = match other {
            Identity::Spiffe {
                trust_domain,
                namespace,
                service_account,
            } => trust_domain.to_owned() + namespace + service_account,
        };
        s.cmp(&o)
    }
}

```

howardjohn: why do we actually need ord?

ilrudie: nm, they can be derived... that's the play

ilrudie: Ord is required to sort which is used in the testing.

src/proxy/connection_manager.rs
ilrudie: Should be better now. Atomic removed

src/proxy/connection_manager.rs
ilrudie: [dashmap crate](https://docs.rs/dashmap/latest/dashmap/struct.DashMap.html)

src/state/policy.rs
ilrudie: added an impl for your comment. Not scoped but should reduce the number of notifications sent to policy subscribers.

src/proxy/connection_manager.rs
ilrudie: I've changed the implementation of sending to policy subscribers to be less frequent in cases where we've got batched updates. Should help

ilrudie:
> There can be a race between connection tracked and policy update.
> like:
> 
> t0: a connection established with old rbac
> t1: the rbac updated
> t2: policy watcher running drain here
> t3. the connection established at t0 get tracked
> 
> the connection actually should be closed

Good catch. Track used to be called with assert_rbac but the design changed a little and now there is a significant gap for races to occur. I'll have take a look at this in more detail.

ilrudie:
added hold to consider reworking without drain::Watch and using a tokio channel instead

hzxuzhonghu (approved):
Some tests should be added, could in a follow up

ilrudie:
Thanks @hzxuzhonghu, I've created an issue to track on follow up items (#798) but will stop making changes to this PR.

supercharge-xsy:
not familiar with Rust，have a question：If the connection is closed properly after send some packets, can the management program sense it? I don't seem to see the corresponding processing. @ilrudie

ilrudie:
> not familiar with Rust，have a question：If the connection is closed properly after send some packets, can the management program sense it? I don't seem to see the corresponding processing. @ilrudie

It's not checking the connections themselves but we are presently using the [drain crate](https://docs.rs/drain/0.1.1/drain/struct.Signal.html#method.drain) which waits for the handle(s) to be dropped when you signal. The handle drop should occur once we break out of looping on handling traffic for the connection.

PR #777, opened 2024-01-22
proxy: validate ZDS provided workload service account when asserting rbac

yuval-k:
If we get the workload service account using ZDS, use it when asserting rbac to guarantee that the workload fetched by ip is the correct one.
This is a part of fixing  https://github.com/istio/istio/issues/46628 (after this merges, a change to the istio node agent is needed to give the ztunnel the proper uid)

We introduce a new optional workload_info message in ZDS. if provided, we carry it over to the proxy, and whenever it asserts_rbac, we use it to check that the workload fetched by `fetch_workload` is indeed the one we expect.

Currently, the workload info has the name,namespace,service_account and trust_domain in it.

The previous approach to this PR (see below) used the pod UID to fetch the workload. This had some issues:
- WDS UIDs ≠ Pod UIDs. this means that ztunnel would have needed to build the WDS UID string on its own, which is fragile, as UID format is opaque and can change.
- To use the Pod's UID to fetch the workload from WDS, would add extra complexity to the ambient_index in istiod.

So current approach achieves security without the complexity. The trade-off being a small chance of a short period where ztunnel rejects rbac when it shouldn't (i.e. if pod is recycled with the same IP but WDS didn't let us know yet).

#### Old description:
This change makes it that a workload has 2 uids:
- One coming from the node agent to purposes of lifecycle.
- The other computer internally to match the uid of the ambient index, for purposes of fetching the workload to assert RBAC.

I'm not a fan of this because:
- Two UIDs where there can be one
- The invisible dependency created between ambient index and the ztunnel

A possible solution is that the UIDs the node agent gives ztunnel (over ZDS) that matches the ambient index ones (over WDS).
This however just moves the "invisible dependency problem" to the node agent (assuming that istiod and node agent have different lifecycles). If we ever change the UID format for ambient index, it may lead to upgrade complexities down the road.
But if we think that the ambient index UID is stable, then this is a reasonable solution.


proto/zds.proto
```diff
@@ -15,10 +15,17 @@ message ZdsHello {
...
+message PodInfo {
+  string name = 1;
+  string namespace = 2;
+  string uid = 4;
+}
+
 // Add a workload to the ztunnel. this will be accompanied by ancillary data contianing
 // the workload's netns file descriptor.
 message AddWorkload {
   string uid = 1;
```
bleggett, line 27: Still don't quite understand why we are passing down 2 UIDs to ztunnel for a pod add, when the one for ambient index can be computed from clusterID/name/namespace without this?

I'm fine with 2 (clearly, differently-named, commented) UIDs, where the one used for index lookup is computed internally by ztunnel on AddWorkload.

CNI agent is dealing with Actual Pods and the K8S API, and should probably traffic exclusively in K8S Pod UIDs, since it is conceivable (but not likely) that at that level we could get something in the same cluster/name/namespace but a different K8S UID, and mess ourselves up.

yuval-k: Are you referring to the `string uid = 4` ? Yes, this can be removed. 

I think the broader question is that we need to decide if computing the uid that way (clusterID/name/namespace) is acceptable, as it creates an implicit dependency between the ztunnel an ambient index code

The other options being to ambient index to pass-in pod uuids, or change the node agent to send ambient-index style uuids.

bleggett: > Are you referring to the string uid = 4 ? Yes, this can be removed.

yes, just that was confusing me, aside from the broader questions - ty!

src/inpod/statemanager.rs
```diff
@@ -233,10 +237,18 @@ impl WorkloadProxyManagerState {
         let workload_netns_inode = netns.workload_inode();
         let (drain_tx, drain_rx) = drain::channel();
 
+        // this matches istio's generatePodUID
```
howardjohn, line 240: > UID represents a globally unique **opaque** identifier for this workload.
> For k8s resources, it is **recommended** to use the more readable format:

Assuming a format is not valid here

howardjohn (changes requested):

src/inpod/statemanager.rs
yuval-k: yes, I'm aware - see my questions in the PR body.  TL,DR I think we have these options:

- Computing the UID this way (with implied dep with ambient index)
- Support fetching with the pod uid in the ambient index, so that ztunnel can fetch workloads by it
- Change the node agent to send ambient-index style uuids instead of pod UIDs.

linsun: Hi folks, can we drive some resolution here?  Which one is best - pod UID or UUID?

yuval-k: hi @howardjohn any thoughts on above options?

src/inpod.rs
```diff
@@ -69,6 +69,7 @@ impl WorkloadUid {
 #[derive(Debug, Clone)]
 struct WorkloadInfo {
     workload_uid: WorkloadUid,
+    pod_info: Option<istio::zds::PodInfo>,
```
bleggett, line 72: Can we:

1. Move `workload_uid` to a top-level field in `WorkloadData`
2. Move `pod_info` to a top-level field in `WorkloadData`

drop `WorkloadInfo` -> workloaddata.workloadinfo.blah feels unnecessary.

src/inpod/statemanager.rs
howardjohn: Option 2 basically means that each workload has multiple UID's. So while we don't depend on the format of one UID, we now have a _new_ UID and we _do_ depend on that UID's format. That doesn't feel great.

Is it crazy to not use UID at all? We could match on quite a few options.. some combination of `{node+cluster, IP, SA, name, namespace}`? We may not need to assert _all_ of them, though we could.

howardjohn: Just to clarify i mean to pass this info through PodInfo. The node+cluster is not needed in PodInfo, ztunnel knows its own node/cluster.

bleggett: > Is it crazy to not use UID at all? We could match on quite a few options.. some combination of {node+cluster, IP, SA, name, namespace}? We may not need to assert all of them, though we could.

Does that really save much in terms of the workload query API tho? That's still basically just a UID lookup but with a local key schema/alias.

howardjohn: Putting Pod UID as an alias seems like "We had a UID, but the spec said its opaque so we built a _new_ UID field and treat that as non-opaque". Now we have 2 UIDs for one resource, and we still have non-opaque UID. It feels like the worst of both worlds.

yuval-k: Pod UID is the best thing to use on as it covers all edge cases (like pod being re-used with same name but different service account).

I'm not sure what you mean by "match"? - My understanding is that whatever property we use needs to be a valid resource name so it can work with on-demand WDS?

I agree that having 2 uids is not great. why did we decide that the resource name is a UID though? i think it's very reasonable that the resource name would have a semantic meaning?

bleggett: > Putting Pod UID as an alias seems like "We had a UID, but the spec said its opaque so we built a new UID field and treat that as non-opaque". Now we have 2 UIDs for one resource, and we still have non-opaque UID. It feels like the worst of both worlds.

I think everyone is in agreement on this point, but since I'm the new guy here - why is UID opacity important here, exactly? What led to that requirement?

howardjohn: > I'm not sure what you mean by "match"? - My understanding is that whatever property we use needs to be a valid resource name so it can work with on-demand WDS?

You can lookup by IP then just verify it matches all the properties (`{node+cluster, IP, SA, name, namespace}`)?

The lookup logic is already pretty complex, I am worried about making it so we have so many aliases and lookup patterns for each resource.

bleggett: Should we just drop `fetch_workload_by_uid` from `state.rs` then?

If we don't want clients using the non-opaque internal UID format, probably shouldn't expose it at all.

(and if we don't clients people using it, why does it matter if it's opaque or not?)

howardjohn: TBH I did not know that UID lookup was added. If we do go with the approach I mentioned I think we could drop that

yuval-k:
Hi @howardjohn @bleggett 
responded to your comments. I need to do a bit of cleanup, but should be roughly what is discussed. Also udpated the code so it works with @ilrudie 's connection manager.

The updated logic is to get the properties that are important for rbac from ZDS (name, ns, svc acct, trust domain) and assert that they match when performing rbac.

src/inpod.rs
```diff
@@ -66,15 +66,11 @@ impl WorkloadUid {
...
-#[derive(Debug, Clone)]
-struct WorkloadInfo {
-    workload_uid: WorkloadUid,
-    pod_info: Option<istio::zds::PodInfo>,
-}
 #[derive(Debug)]
 pub struct WorkloadData {
     netns: std::os::fd::OwnedFd,
-    info: WorkloadInfo,
+    workload_uid: WorkloadUid,
```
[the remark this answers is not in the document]
yuval-k: This was in response to johns question here: https://github.com/istio/ztunnel/pull/747#discussion_r1440759395
The type serves as "documentation" that this is a uid.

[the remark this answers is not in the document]
bleggett: I'd rather just do that with the normal function comment/variable name since this is a bit tedious in practice and just results in a lot of arguably useless type conversion, but will defer if he feels strongly.

src/state.rs
```diff
@@ -71,10 +71,43 @@ impl fmt::Display for Upstream {
...
+    ) -> Self {
+        Self {
+            name,
+            namespace,
+            trust_domain,
+            service_account,
+        }
+    }
+
+    pub fn matches(&self, w: &Workload) -> bool {
```
bleggett, line 99: Deferrable nit - why not just implement `Eq`?

bleggett (approved):

src/inpod/protocol.rs
```diff
@@ -280,6 +280,35 @@ mod tests {
         assert!(matches!(m, WorkloadMessage::AddWorkload(_)));
     }
 
+    #[test]
+    fn test_parse_add_workload_with_info() {
+        let owned_fd: OwnedFd = std::fs::File::open("/dev/null").unwrap().into();
```
bleggett, line 285: Deferrable nit: would rather use something like [`tempfile` ](https://docs.rs/tempfile/latest/tempfile/) to generate a junk file to get FDs off of, so we have a guarantee the underlying file goes away no matter what.

src/state.rs
yuval-k: different types. i'm comparing this to `Workload` not `WorkloadInfo`

bleggett: derp - yep

src/inpod/protocol.rs
yuval-k: `/dev/null` is pretty much a constant - I don't need it to be deleted.. I'm not sure i understand the advantage of using a tempfile here?

bleggett: It's not a big deal - I'm just thinking that if we ever have to debug tests or suites leaking FDs, it's _much_ easier to find FDs dangling from deleted one-per-test tempfiles than from `/dev/null` which will always be there and which potentially many other things on the system are opening FDs against and using.

src/inpod.rs
```diff
@@ -66,14 +66,11 @@ impl WorkloadUid {
...
-#[derive(Debug, Clone)]
-struct WorkloadInfo {
-    workload_uid: WorkloadUid,
-}
 #[derive(Debug)]
 pub struct WorkloadData {
     netns: std::os::fd::OwnedFd,
-    info: WorkloadInfo,
+    workload_uid: WorkloadUid,
+    workload_info: Option<istio::zds::WorkloadInfo>,
```
howardjohn, line 73: is it `Option` forever or just because Istio doesn't yet send it?

howardjohn (approved):

src/inpod.rs
yuval-k: probably forever - it will also be an option for VMs

yuval-k: though i guess for VMs this can come from config.. so we'll see when we get there.

howardjohn: VMs don't use inpod though, right?

I think it needs to be an `Option` in the rbac enforcement but maybe not here.

Anyways, not a big deal, was mostly checking my understanding

yuval-k: VMs won't use inpod, but conceptionally may want something similar. 

yea right now it needs to be an option otherwise it will break istio integration tests.
Will soon follow up with istio/istio PR to fill in the info and we can decide if to remove the option from here once its all settled.

PR #789, opened 2024-01-24
basic sandwich

stevenctl:
When the Waypoint is captured, zTunnel sees the connection address as the Waypoint's address and the HBONE address as the actual destination. 

This impl currently uses that to detect whether the current connection is part of a sandwich, and verifies that the HBONE target actually references the connection address as a Waypoint. Instead of sending traffic to the HBONE target, we send it to the connection address. 

Depends on https://github.com/istio/ztunnel/pull/832 so the relevant commits are: https://github.com/istio/ztunnel/pull/789/files/b5af5ac0fb7e71805ee6f5b9fe56e8010511dead..a17e43e3488a4567c4c84eaae0638bf09d045c4d

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -629,11 +629,7 @@ mod tests {
             workload::gatewayaddress::Destination,
         },
     };
-    use std::{
-        collections::HashMap,
-        net::{Ipv4Addr, SocketAddrV4},
-        sync::RwLock,
-    };
+    use std::{collections::HashMap, net::Ipv4Addr, sync::RwLock};
 
     #[tokio::test]
     async fn check_gateway() {
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -29,7 +29,7 @@ use tracing::{error, trace, warn, Instrument};
 use inbound::Inbound;
 pub use metrics::*;
 
-use crate::identity::SecretManager;
+use crate::identity::{Identity, SecretManager};
 use crate::metrics::Recorder;
 use crate::proxy::connection_manager::{ConnectionManager, PolicyWatcher};
 use crate::proxy::inbound_passthrough::InboundPassthrough;
@@ -37,7 +37,8 @@ use crate::proxy::outbound::Outbound;
 use crate::proxy::socks5::Socks5;
 use crate::rbac::Connection;
 use crate::state::service::{endpoint_uid, Service, ServiceDescription};
-use crate::state::workload::{network_addr, Workload};
+use crate::state::workload::address::Address;
+use crate::state::workload::{network_addr, GatewayAddress, Workload};
 use crate::state::{DemandProxyState, WorkloadInfo};
 use crate::{config, identity, socket, tls};
 
@@ -262,6 +263,9 @@ pub enum Error {
 
     #[error("unsupported feature: {0}")]
     UnsupportedFeature(String),
+
+    #[error("ip mismatch: {0} != {1}")]
+    IPMismatch(IpAddr, IpAddr),
 }
 
 // TLS record size max is 16k. But we also have a H2 frame header, so leave a bit of room for that.
@@ -526,6 +530,60 @@ pub fn guess_inbound_service(
         .map(ServiceDescription::from)
 }
 
+// Checks if the connection's source identity is the identity for the upstream's waypoint
+async fn check_from_waypoint(
+    state: DemandProxyState,
+    upstream: &Workload,
+    src_identity: Option<&Identity>,
+) -> bool {
+    check_gateway_address(state, src_identity, upstream.waypoint.as_ref()).await
+}
+
+// Checks if the connection's source identity is the identity for the upstream's network
+// gateway
+async fn check_from_network_gateway(
+    state: DemandProxyState,
+    upstream: &Workload,
+    src_identity: Option<&Identity>,
+) -> bool {
+    check_gateway_address(state, src_identity, upstream.network_gateway.as_ref()).await
+}
+
+// Check if the source's identity matches any workloads that make up the given gateway
+// TODO: This can be made more accurate by also checking addresses.
+async fn check_gateway_address(
+    state: DemandProxyState,
+    src_identity: Option<&Identity>,
+    gateway_address: Option<&GatewayAddress>,
+) -> bool {
+    let Some(src_identity) = src_identity else {
+        return false;
+    };
+    if let Some(gateway_address) = gateway_address {
+        let from_gateway = match state.fetch_destination(&gateway_address.destination).await {
+            Some(Address::Workload(wl)) => &wl.identity() == src_identity,
+            Some(Address::Service(svc)) => {
+                for (_ep_uid, ep) in svc.endpoints.iter() {
+                    // fetch workloads by workload UID since we may not have an IP for an endpoint (e.g., endpoint is just a hostname)
+                    if state
+                        .fetch_workload_by_uid(&ep.workload_uid)
+                        .await
+                        .map(|w| w.identity())
+                        .as_ref()
+                        == Some(src_identity)
+                    {
+                        return true;
+                    }
+                }
+                false
+            }
+            None => false,
+        };
+        return from_gateway;
+    }
+    false // this occurs if gateway_address was None
+}
+
 #[cfg(test)]
 mod tests {
     use std::assert_eq;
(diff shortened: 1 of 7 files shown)


src/proxy/inbound.rs
```diff
@@ -259,138 +291,173 @@ impl Inbound {
...
-                    Err(_) => StatusCode::SERVICE_UNAVAILABLE,
-                };
+        let dst_network_addr = NetworkAddress {
+            network: conn.dst_network.to_string(), // dst must be on our network
+            address: addr.ip(),
+        };
+
+        // for sandwich, we haven't resolved the VIP
+        // otherwise, this is a workload and should have the service
+        let wl_svcs = state.fetch_workload_services(&dst_network_addr).await;
```
stevenctl, line 336: for sandwich we should really look up the service using this VIP and look up the waypoint workload and make sure they're related

src/proxy/inbound.rs
```diff
@@ -259,138 +291,173 @@ impl Inbound {
         metrics: Arc<Metrics>,
         socket_factory: Arc<dyn SocketFactory + Send + Sync>,
     ) -> Result<Response<Empty<Bytes>>, hyper::Error> {
-        match req.method() {
-            &Method::CONNECT => {
```
stevenctl, line 263: the diff might look odd here because I un-indented this to be an `if { return } ...` rather than a match with a very long clause

src/proxy/inbound.rs
```diff
@@ -259,138 +291,170 @@ impl Inbound {
...
-                    info!(%conn, "unknown destination");
-                    return Ok(Response::builder()
-                        .status(StatusCode::NOT_FOUND)
-                        .body(Empty::new())
-                        .unwrap());
-                };
-                let has_waypoint = upstream.waypoint.is_some();
-                let from_waypoint = Self::check_waypoint(state.clone(), &upstream, &conn).await;
-                let from_gateway = Self::check_gateway(state.clone(), &upstream, &conn).await;
+        let addr: SocketAddr = addr.unwrap();
```
ilrudie, line 313: **nitty**: this looks a bit odd. I think you can `let addr = match addr{...}` and it works Ok so long as arms that won't result in the desired type return.

src/proxy/inbound.rs
```diff
@@ -259,138 +291,170 @@ impl Inbound {
...
+
+        let (next_hop, inbound_req) = match sandwich {
+            false => (addr, Hbone(req)),
+            true => (
+                conn.dst,
+                Sandwich(
+                    // TODO add src_port or change conn.src_ip to src (SocketAddr)
+                    req,
+                    (SocketAddr::new(conn.src_ip, 10000), addr),
+                    conn.src_identity.unwrap().to_string(),
```
ilrudie, line 433: Is it possible to reach this point with no src identity? I don't think so but not 100% sure.

ilrudie (commented):
direction seems good

src/proxy/inbound.rs
stevenctl: Yeah, definitely need to cleanup unwraps although I'm pretty sure only auth'd conns make it here.

src/state/service.rs
```diff
@@ -39,6 +41,10 @@ pub struct Service {
     /// Maps endpoint UIDs to service [Endpoint]s.
     #[serde(default)]
     pub endpoints: HashMap<String, Endpoint>,
+    /// Cache the set of gateways in front of this Service
+    /// based on its endoints.
+    #[serde(default)]
+    pub waypoints: HashMap<Destination, usize>,
```
stevenctl, line 47: An alternative here is to denormalize the `waypoint` field of `Workload` into the `Endpoint` struct. Lookups aren't quite as fast though.

hzxuzhonghu:
@stevenctl Is there a proposal, i missed the context

stevenctl:
@hzxuzhonghu 

See [this doc](https://docs.google.com/document/d/1MFPLVTatJI4Z9a88bSo9pZ8z9ftweHUYXpH7zdrXU94/edit) (primarily the propagating L4 info section) and https://github.com/istio/istio/issues/48362

The goal is to provide alternative Waypoint implementations with an option other than supporting CONNECT.

tests/namespaced.rs
```diff
@@ -284,10 +282,8 @@ mod namespaced {
         let local = manager.deploy_ztunnel_inpod(DEFAULT_NODE, uds_default_node.clone())?;
 
         info!("sending workload to ztunnel");
-        remote_node_server.send(server_fd).await.unwrap();
-        remote_node_server_ack.recv().await.unwrap();
-        default_node_server.send(client_fd).await.unwrap();
-        default_node_server_ack.recv().await.unwrap();
+        remote_node_server.register_workload(server_fd).await;
+        default_node_server.register_workload(client_fd).await;
```
stevenctl, line 286: wip: refactoring the test harness to allow re-using the inpod stuff without as much copypaste

src/proxy/outbound.rs
```diff
@@ -470,15 +470,26 @@ impl OutboundConnection {
...
+        match (
+            from_waypoint,
+            self.pi
+                .state
+                .fetch_waypoint(&mutable_us.workload, workload_ip)
+                .await,
+        ) {
+            (_, Ok(None)) => {} // workload doesn't have a waypoint; this is fine
+            (true, _) => {}     // we already traversed the waypoint
+            (false, Ok(Some(waypoint_us))) => {
```
stevenctl, line 492: In this branch (and the one for to Service) we're using the port WDS GatewayAddress sends us. 

https://github.com/istio/ztunnel/blob/a17e43e3488a4567c4c84eaae0638bf09d045c4d/src/proxy/outbound.rs#L423

Right now, the CP happens to send us 15008 which happens to be what the ztunnel listens on.

stevenctl: Is the CP going to be smart and send the zTunnel address, or do we need to be smart and use `pi.hbone_port`?

stevenctl:
cc @bleggett

src/proxy.rs
```diff
@@ -526,6 +530,60 @@ pub fn guess_inbound_service(
...
+                for (_ep_uid, ep) in svc.endpoints.iter() {
+                    // fetch workloads by workload UID since we may not have an IP for an endpoint (e.g., endpoint is just a hostname)
+                    if state
+                        .fetch_workload_by_uid(&ep.workload_uid)
+                        .await
+                        .map(|w| w.identity())
+                        .as_ref()
+                        == Some(src_identity)
+                    {
+                        return true;
```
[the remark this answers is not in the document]
stevenctl: You're reading it right, but also this code isn't new. I moved it here because it's now also used in outbound. 

This is usually used for checking the source of traffic, we could expand it to include an address.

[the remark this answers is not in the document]
stevenctl: I ended up changing this to check identity and IP. 

> require at-most-one backing workload for gateway services

Sort of.. the from_waypoint check in outbound is deciding if we need to traverse the waypoint for sandwich. This is only relevant for sandwiched gateways which have workloads.

The from_waypoint check in inbound will is used to prevent bypass. For now this change does break waypoints that don't have a Workload, but that will eventually be replaced by RBAC rules. 

We could use different logic in each place though.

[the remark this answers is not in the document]
stevenctl: The reason for the change is that some tests were breaking. I assume the tests had things sharing identities.

ilrudie (commented):
Looking good. couple comments/questions

src/proxy/inbound.rs
```diff
@@ -277,41 +279,48 @@ impl Inbound {
...
                 let has_waypoint = upstream.waypoint.is_some();
-                let from_waypoint = Self::check_waypoint(pi.state.clone(), &upstream, &conn).await;
-                let from_gateway = Self::check_gateway(pi.state.clone(), &upstream, &conn).await;
+                let from_waypoint = proxy::check_from_waypoint(
+                    pi.state.clone(),
+                    &upstream,
+                    conn.src_identity.as_ref(),
+                    &conn.src_ip,
+                )
+                .await;
```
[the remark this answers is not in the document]
stevenctl: Yep, I just don't want to take an opinion on that in this PR.

src/proxy/inbound.rs
```diff
@@ -419,49 +428,103 @@ impl Inbound {
...
-                    false
+                    Some((conn_wl, vec![*svc]))
                 }
-                None => false,
-            };
-            return from_gateway;
+                Address::Workload(wl) => {
+                    let svc = state.services.get_by_workload(&wl);
+                    Some((*wl, svc))
+                }
```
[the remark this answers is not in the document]
stevenctl: fixed

src/proxy/inbound.rs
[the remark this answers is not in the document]
ilrudie: https://github.com/istio/ztunnel/issues/851

created this so I/we don't forget to circle back around on it until a user files a report that things don't work like we said they do

ilrudie (approved):
LGTM, thanks!

src/proxy/outbound.rs
bleggett: Right now as per the protocol 15008 is "the" HBONE port and we assume that in other spots, I suspect.

For this I would suggest maybe splitting the difference and taking what CP sends us, but explicitly checking here that `waypoint_us.port == pi.hbone_port` and throwing an error if not. If we wish to intentionally change that behavior later we can.

bleggett (approved):

src/proxy.rs
```diff
@@ -526,6 +530,61 @@ pub fn guess_inbound_service(
...
 
+// Checks that the source identiy and address match the upstream's waypoint
+async fn check_from_waypoint(
+    state: DemandProxyState,
+    upstream: &Workload,
+    src_identity: Option<&Identity>,
+    src_ip: &IpAddr,
+) -> bool {
+    let is_waypoint = |wl: &Workload| {
+        Some(wl.identity()).as_ref() == src_identity && wl.workload_ips.contains(src_ip)
```
howardjohn, line 541: Long term it would be nice to have a method more reliable than source IP. That being said, since its node-local it would be quite strange for this check to not work somehow.

howardjohn (approved):
LGTM, all comments are theoretical future improvements so good to go ahead

PR #802, opened 2024-02-07
Initial connection MDS implementation

howardjohn:
* Store each connection in a shared map
* Capture special IP address, redirect to MDS handler
* MDS handler takes 4 tuple as input.
* Response is JSON containing identity

Also has an example Golang library implementation to add HTTP middleware that extracts the identity.

Istio tests: https://github.com/istio/istio/pull/44536 (blocked by this PR, of course), which also show how to use it

Original PR in https://github.com/istio/ztunnel/pull/504 got force deleted so cannot re-open

### DIFF
--- a/go-metadata/istio.go
+++ b/go-metadata/istio.go
@@ -135,7 +135,7 @@ func FetchFromClientConnection(c net.Conn) (*ConnectionMetadata, error) {
 	return lookup(c.LocalAddr().String(), c.RemoteAddr().String())
 }
 
-// FetchFromServerConnection attempts to fetch ConnectionMetadata from a client connection.
+// FetchFromServerConnection attempts to fetch ConnectionMetadata from a server connection.
 func FetchFromServerConnection(c net.Conn) (*ConnectionMetadata, error) {
 	return lookup(c.RemoteAddr().String(), c.LocalAddr().String())
 }
--- a/go-metadata/istio.go
+++ b/go-metadata/istio.go
@@ -1,0 +1,141 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package istiometadata
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"os"
+	"time"
+)
+
+// ConnectionMetadata returns the metadata about a connection
+type ConnectionMetadata struct {
+	// Identity provides the identity of the peer.
+	// Example: spiffe://cluster.local/ns/a/sa/b.
+	Identity string `json:"identity"`
+}
+
+// metadataClient provides a default client for all metadata requests
+var metadataClient = http.Client{
+	Timeout: time.Second,
+}
+
+// metadataContextKey is the key the metadata handler will store connection metadata in
+var metadataContextKey = &contextKey{}
+
+type contextKey struct {
+	name string
+}
+
+func (k *contextKey) String() string { return "istio metadata context value" }
+
+const mdsHostEnv = "GCE_METADATA_HOST"
+
+// For now, we use a well-known IP which is intercepted.
+// TODO: consider creating a real service with node-affinity to drop redirection dependency.
+const defaultHost = "169.254.169.111"
+
+func metadataServerURL() *url.URL {
+	host := os.Getenv(mdsHostEnv)
+	if host == "" {
+		host = defaultHost
+	}
+	u, err := url.Parse("http://" + host)
+	if err != nil {
+		panic(err.Error())
+	}
+	return u
+}
+
+func lookup(src, dst string) (*ConnectionMetadata, error) {
+	u := metadataServerURL()
+	params := url.Values{}
+	params.Add("src", src)
+	params.Add("dst", dst)
+	u.RawQuery = params.Encode()
+	u.Path = "/connection"
+
+	resp, err := metadataClient.Get(u.String())
+	if err != nil {
+		return nil, fmt.Errorf("metadata lookup failed: %v", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		bdy, _ := io.ReadAll(resp.Body)
+		return nil, fmt.Errorf("metadata server returned %v: %v", resp.StatusCode, string(bdy))
+	}
+
+	mr := &ConnectionMetadata{}
+	if err := json.NewDecoder(resp.Body).Decode(mr); err != nil {
+		return nil, err
+	}
+	return mr, nil
+}
+
+// Handler is an HTTP middleware that looks up the metadata for each request, storing it in context.
+// It can later be looked up with ExtractFromRequest.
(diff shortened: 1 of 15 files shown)


src/proxy/metadata.rs
```diff
@@ -0,0 +1,105 @@
...
+use tokio::net::TcpStream;
+use tracing::{error, trace};
+
+use crate::proxy::connection_manager::{ConnectionManager, ConnectionTuple};
+
+use super::Error;
+
+/// METADATA_SERVER_IP provides the well-known metadata server IP.
+/// This is captured by the redirection.
+pub const METADATA_SERVER_IP: IpAddr = IpAddr::V4(Ipv4Addr::new(169, 254, 169, 111));
```
bleggett, line 36: Probably need to document this like we do for the other link-local addresses we reserve.

src/proxy/metadata.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    let params = url::form_urlencoded::parse(query.as_bytes())
+        .into_owned()
+        .collect::<HashMap<String, String>>();
+    let src: SocketAddr = params.get("src").ok_or(anyhow!("missing src"))?.parse()?;
+    let dst: SocketAddr = params.get("dst").ok_or(anyhow!("missing dst"))?.parse()?;
+
+    // To restrict access to sensitive metadata, ensure the client is part of the requested connection.
+    // This can be error prone if the client has multiple NICs.
+    if remote.ip() != dst.ip() && remote.ip() != src.ip() {
```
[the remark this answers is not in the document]
howardjohn: pre-inpod, we could only do IP check to know the source of a node-local client. If we do https://github.com/istio/ztunnel/pull/777 we can adopt similar here

src/proxy/metadata.rs
howardjohn: Ack. An alternative is to not do anything special and make a real service like https://github.com/istio/istio/pull/45009.. not sure what is preferred

bleggett: We could also bind another in-pod port which would make it explicitly private and effectively remove the need to do identity checks or reserve an IP or expose this in the cluster - but I don't care much.

I am fine with link-local for now if we toss it into `ARCHITECTURE.md`

howardjohn: Lets do the link local thing for now but ill make sure its documented + explicitly called out as experimental. This feature will need a lot of feedback overall

src/proxy/metadata.rs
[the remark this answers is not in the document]
ilrudie: connection manager isn't tracking all the zt's connections (to prevent unrelated tasks from blocking each other with locks). The connections stored in your instance should already be unique to your pod since.

ilrudie (commented):
I didn't envision connection manager being used in this way so we'll probably need to make more changes to ensure all outbound connections register and release correctly

src/proxy/connection_manager.rs
```diff
@@ -12,16 +12,31 @@
...
+#[derive(Clone, Copy, Hash, Debug, Eq, PartialEq)]
+pub struct ConnectionTuple {
+    pub src: SocketAddr,
+    pub dst: SocketAddr,
+}
+
+#[derive(Clone, Debug, Serialize)]
+pub struct ConnectionMetadata {
+    #[serde(default)]
+    identity: Option<Identity>,
```
hzxuzhonghu, line 37: I am confused about a connection's identity. Usually, we say the src identity or dest identity

go-metadata/istio.go
```diff
@@ -0,0 +1,141 @@
...
+	local := r.Context().Value(http.LocalAddrContextKey)
+	return lookup(r.RemoteAddr, local.(*net.TCPAddr).String())
+}
+
+// FetchFromClientConnection attempts to fetch ConnectionMetadata from a client connection.
+func FetchFromClientConnection(c net.Conn) (*ConnectionMetadata, error) {
+	return lookup(c.LocalAddr().String(), c.RemoteAddr().String())
+}
+
+// FetchFromServerConnection attempts to fetch ConnectionMetadata from a client connection.
```
hzxuzhonghu, line 138: s/client/server

hzxuzhonghu (commented):
Can you add desc of the usage

src/proxy/outbound.rs
```diff
@@ -131,22 +132,27 @@ impl OutboundConnection {
...
         &mut self,
         mut stream: TcpStream,
-        remote_addr: IpAddr,
+        remote_addr: SocketAddr,
         orig_dst_addr: SocketAddr,
         block_passthrough: bool,
     ) -> Result<(), Error> {
+        let remote_ip = remote_addr.ip();
+        if orig_dst_addr.ip() == METADATA_SERVER_IP {
+            return handle_metadata_lookup(&self.connection_manager.clone(), stream, remote_addr)
```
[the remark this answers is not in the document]
ilrudie: #804 should resolve. adds CM sharing between inbound/outbound/inbound_passthrough/socks5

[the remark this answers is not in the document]
ilrudie: It doesn't change which outbound connections register with CM

src/proxy/connection_manager.rs
ilrudie: In the implementation we've got it will be the source identity. I suppose keeping the type general would allow us to represent peer identity generically with the same type in the future though.

src/proxy/metadata.rs
```diff
@@ -0,0 +1,105 @@
...
+        .serve_connection(
+            hyper_util::rt::TokioIo::new(stream),
+            service_fn(|req: hyper::Request<hyper::body::Incoming>| {
+                let ct = ct.clone();
+                async move {
+                    let res: Result<_, Infallible> =
+                        serve_request(&ct, remote_addr, req).await.or_else(|e| {
+                            Ok(crate::hyper_util::plaintext_response(
+                                StatusCode::UNPROCESSABLE_ENTITY,
+                                e.to_string(),
```
ilrudie, line 57: Do we need to be concerned about `e` leaking something we shouldn't to the mds client?

src/proxy/metadata.rs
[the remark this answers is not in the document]
ilrudie: #777 did merge, should we adopt now or add as a follow up issue perhaps. Overall I don't think it's as critical for in-pod but if folks disagree there may be something I'm missing. At any rate using identity would resolve the multiple NICs issue I think.

src/proxy/connection_manager.rs
```diff
@@ -118,6 +133,19 @@ impl ConnectionManager {
         // potentially large copy under read lock, could require optomization
         self.drains.read().await.keys().cloned().collect()
     }
+
+    /// fetch looks up a tuple and returns the connection metadata
+    // TODO: we should key on tuple so we can more efficiently lookup
```
ilrudie, line 138: Maybe another index instead of re-keying? I suppose we could see about keying on tuple and storing the full `ProxyRbacContext` in the `ConnectionDrain` struct. Otherwise we'll need to do work to re-construct the full context when we need to assert.

ilrudie (commented):
This is looking good. I have a few comments but nothing I suspect to be a show stopper for this direction.

src/proxy/connection_manager.rs
howardjohn: I documented a bit more now

src/proxy/metadata.rs
howardjohn: I think its ok? I documented that we expose errors which could help a bit

ilrudie (approved):
LGTM

src/proxy/connection_manager.rs
howardjohn: Yeah I would probably do the index if we do it

stevenctl:
Do we still want to merge this anytime soon? Sorry I missed it.

ilrudie:
IMO it's worth doing. also, I did not realize that my review would get retroactively upgraded... 😬

PR #818, opened 2024-02-20
forget certs where appropriate during xds remove

### LINKED ISSUE #351: certificates: stop trying to refresh when identity is no longer on the node

ilrudie:
fixes #351

Adds and implementation for forgetting certificates during xds removal when no other WL is present on the node which requires the certificate.

### DIFF
--- a/src/cert_fetcher.rs
+++ b/src/cert_fetcher.rs
@@ -24,6 +24,7 @@ use tracing::{debug, error, info};
 /// Responsible for pre-fetching certs for workloads.
 pub trait CertFetcher: Send + Sync {
     fn prefetch_cert(&self, w: &Workload);
+    fn forget(&self, w: &Workload);
 }
 
 /// A no-op implementation of [CertFetcher].
@@ -31,6 +32,7 @@ pub struct NoCertFetcher();
 
 impl CertFetcher for NoCertFetcher {
     fn prefetch_cert(&self, _: &Workload) {}
+    fn forget(&self, _: &Workload) {}
 }
 
 /// Constructs an appropriate [CertFetcher] for the proxy config.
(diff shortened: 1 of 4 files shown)


src/xds.rs
```diff
@@ -171,6 +172,17 @@ impl ProxyStateUpdateMutator {
                     .remove_endpoint(&prev.uid, &endpoint_uid(&prev.uid, None));
             }
 
+            // if expected is true we are removing a workload, we should forget it
+            // if expected is false this is an update, no need to forget on update
+            if expected
```
howardjohn, line 177: maybe_remove is called on insert_workload so this seems problematic

howardjohn: Also we happen to call removes after adds (see HandlerWrapper), but is it guaranteed? Maybe we can add tests to ensure if we get a XDS update with [add SA A, remove SA A] we do not end up flipping the cert

src/cert_fetcher.rs
```diff
@@ -45,27 +47,42 @@ pub fn new(cfg: &config::Config, cert_manager: Arc<SecretManager>) -> Arc<dyn Ce
...
+                            }
+                            Err(e) => error!(
+                                "unable to prefetch cert for {:?}, skipping, {:?}",
+                                workload_identity.to_string(),
+                                e
+                            ),
+                        }
+                    }
+                    Action::Forget(workload_identity) => {
+                        cert_manager.forget_certificate(&workload_identity).await;
```
howardjohn, line 83: I kind of wonder if we are going to get in a state where we flip flop. Imagine I have a single node cluster and a 1 replica deployment and it is deleted+restarted.

Or maybe the manager handles this already

src/xds.rs
ilrudie: mabye_remove sets expected to `false`. The check for expected stops forgetting when during insert, yeah?

Agree on tests. I started adding some but wound up at a dead end so decided to PR some code before spending more time on it incase this path received a firm "no"

src/cert_fetcher.rs
ilrudie: I think if the only pod with some SA on a node is flapping then we would do a lot of fetches and forgets in this implementation.

ilrudie:
I think this might be able to be scrapped and a new version created that sets some ttl when we get specific failures, such as wl not on node, and unless we get a success somewhere which removes the ttl we forget it after the first railed retry which occurs after ttl expired.

Issue #797, opened 2024-02-02
Long term: Migrate from `boring` to `rustls`


howardjohn:
Currently, we use the `boring` library crate. This is to support the ability to use FIPS compliant crypto. This comes at the cost of simplicity and ecosystem compatibility, where rustls is king (and, I believe, there are _massive_ performance differences in our use case; likely some could be improved with `boring`, but may be a wasted effort if we do this issue).

Some background: TLS does a lot of things outside of just raw crypto operations. Only the crypto operations are relevant for compliance.

Historically, rustls uses `ring` for crypto, which is not certified. However, recently it has added a pluggable mechanism allowing alternative crypto backends. There is also ongoing  work to make `ring` itself have a FIPS mode.

We should look into migrating to Rustls. A pre-requisite to this should be the presence of a (high quality) provider to alternative crypto backends. Plausibly we could offer alternative backends users could build, but as a baseline I would expect a boringssl backend, probably as the default to match with current state and Envoy.

Relevant links:
- https://github.com/cloudflare/boring/issues/165
- https://github.com/rustls/rustls/issues/1540
- https://github.com/janrueth/boring-rustls-provider
- https://briansmith.org/lets-build-a-fips-rust-crypto

Issue #641, opened 2023-08-04
Boring build broken for non-FIPS


nmittler:
For those of us that don't build against the pre-built FIPS binary (e.g. mac users), the build is currently broken due to Boring. It's not  clear exactly what is broken, but I recently had this same issue with [quinn-boring](https://github.com/quinn-rs/quinn-boring) and fixed it by using the latest commit.

For quinn-boring I actually needed to use the latest commit. We could probably get by using the latest release (https://github.com/cloudflare/boring/tree/v3.0.2).

@howardjohn assigning to you since we're using your branch of Boring currently.

To verify it works:

1. Comment out the boring env vars in config.toml, and ...
2. Comment out the default features in cargo.toml

This should force boring to rebuild boringssl from scratch.

nmittler:
FYI, I suspect this is blocked by https://github.com/cloudflare/boring/pull/156.

bleggett:
fwiw I gave up trying to rebuild boringssl from scratch on M1 Mac and just use one of

- a local linux/arm64 VM 
- `make shell` (effectively linux/arm64) 

to do ztunnel builds.

ztunnel has platform conditionals that render non-linux builds mostly useless anyway.

nmittler:
The latest cloudflare boring build works fine for me on M1 osx. It's just that the particular version ztunnel is using is now broken (I assume something changed in upstream boringssl).

> ztunnel has platform conditionals that render non-linux builds mostly useless anyway.

While true, there's a lot of development that can still be done without requiring linux. All/most of unit tests run fine on osx, for example.

bleggett:
I do think/agree non-FIPS builds are something we need to support broadly, since as I've mentioned before FIPS is a massive pain and many orgs/users will not need it, or expressly will not want it (e.g. they will want to use newer crypto libs than the USG certifies).

That being said, without even basic "does it build" CI for $platform+$arch it's going to be effectively impossible to guard against breakage long-term for that $platform+$arch. And we are (probably) never adding `ztunnel` CI for macOS.

nmittler:
Yeah, I think the main guard against macOS breakage is simply developers for the project (I know I'm not the only one on mac).

Issue #110, opened 2022-11-11
Optimize Read/Write syscalls


howardjohn:
Aside from TLS crypto overhead, most CPU time is spent on r/w syscalls.

Brain dump from my investigations so far...

When we switch from rustls to boringssl, we saw a 50% regression in perf. 

Some analysis of the R/W pattern

BoringSSL reads: read 5/26/5/8k. Each 5 read is the TLS record header, while the 26 is the H2 data frame header. the 8k is the data frame body.
BoringSSL writes: write 5/26/5/8k

Rustls reads: all 16k
Rustls writes: 31/8k/31/8k - this is H2 header then body repeated.

So Rustls is doing a full buffering on reads, but still writing header and frame in 2 writes. Boringssl is much worse; the TLS reads are reading the TLS record header and data separately which compounds with the writes being split as well. This tanks the performance.

---

https://github.com/istio/ztunnel/pull/109 dramatically improves this all around. With this, we see:

Both Reads and writes sometimes are 5/26/5/16k, but other times multiple 16k reads in a row. Not clear why. 

There is likely room for improvement here. 

---

Research notes:
SSL_read takes in a buffer, but that isn't used as the buffer for recv syscalls; instead it handles all that internally with a growing buffer and then memcpy's it into our buffer. I couldn't find anyway to influence this behavior

The small writes are because of rust code calling SSL_write with small data. This largely seems due to The h2 library: https://github.com/hyperium/h2/blob/294000c0745c64009151a1ab39978cd6f02dfd68/src/codec/framed_write.rs#L128 will make a `Chain` with the frame header and frame data, then call write()

https://github.com/hyperium/h2/blob/294000c0745c64009151a1ab39978cd6f02dfd68/src/codec/framed_write.rs#L181-L187 will only call poll_write_vectored if the writer supports it; the BoringSSL writer does not - and neither does any other TLS library in the ecosystem that I found. So we end up on the other path; chunk() returns only the frame header, it doesn't concat them.

So if we made a writer that implements poll_write_vectored by concating the bytes together, we may see some improvements.

---

stracing methodology:

```
strace -s 0 -ff -o cmd.strace -v -s 64 -x ./out/rust/release/ztunnel
cat cmd* | rg recv | hh '\\27\\3' recvfrom '= 5' sendto 8192 8214 '\\x17\\x03\\x03' 15360
[...]

Issue #399, opened 2023-02-15
Avoid requiring manual munges of `config.toml` to switch build target architecture


bleggett:
Right now for both CI and local builds we require manual munges of `.cargo/config.toml` `BSSL_*` env var paths, to point https://github.com/cloudflare/boring at the right vendored FIPS build for the target build architecture.

We are currently unable to intelligently switch these vendored paths our own `build.rs` due to logic https://github.com/cloudflare/boring has in its own `build.rs`

We have three options here if FIPS-only builds remain the default:

1. Let https://github.com/cloudflare/boring leverage it's own `boringssl` autobuild logic, and stop shipping vendored binaries in our  repo -> this places significant FIPS-related restrictions on local build boxes. This could be mitigated with a containerized build env.

2. PR changes into upstream https://github.com/cloudflare/boring to enable it to dynamically select the correct vendored `boringssl` library path based on build target arch -> this may be too much "intelligence" to push upstream, but we can try.

3. PR changes into upstream https://github.com/cloudflare/boring to enable _our_ downstream `build.rs` to dynamically select the correct vendored `boringssl` library path based on build target arch.

stevenctl:
https://github.com/rust-lang/pkg-config-rs#external-configuration-via-target-scoped-environment-variables

Possibly viable?

bleggett:
> https://github.com/rust-lang/pkg-config-rs#external-configuration-via-target-scoped-environment-variables
> 
> ~Possibly viable?~
> 
> Misread it. But their approach might be?

Yep could be. 

Ideally I would prefer to fix upstream `boring` so it respects whatever BSSL_* vars are already set by parent crates, and do any fancy stuff (e.g. autoselection) in our own `build.rs`, as I think that's both reasonable and simple.

howardjohn:
After https://github.com/istio/ztunnel/pull/820 this should no longer be required by default. If you want to build with boringssl you would need to still, though.

PR #820, opened 2024-02-21
Migrate to `rustls`

howardjohn:
This PR moves from usage of `boring` as a crypto library _and_ TLS library, to using `rustls`. Rustls is the standard TLS library in the rust ecosystem.

Rustls offers a pluggable crypto interface. There are ~3 providers (2 in core, 1 external): ring, aws_lc_rs, and boring. The latter two provider FIPS certified builds (though I think aws_lc only has _in progress_ builds); `ring` has no FIPS certified builds though I think they are working towards one. In this model, the _TLS_ aspects (which are a large portion of the code involved) are handled purely by rustls's code, written in safe rust. Just the crypto primitives are offloaded to these various libraries (which are all ultimately varying degrees of openssl forks, as I understand it).

This offers a few benefits:
* For vendors, the ability to plug in alternative crypto modules
* Better performance (~10% improvement)
* Standardize with rest of rust ecosystem. Ever library integrates with rustls; we need to make our own shims for everything today like tonic, etc

Fixes https://github.com/istio/ztunnel/issues/797
Fixes https://github.com/istio/ztunnel/issues/641
Helps with https://github.com/istio/ztunnel/issues/149
Fixes https://github.com/istio/ztunnel/issues/110


.cargo/config.toml
```diff
@@ -1,5 +1,5 @@
 [build]
 target-dir = "out/rust"
 [env]
-BORING_BSSL_PATH = { value = "vendor/boringssl-fips/linux_x86_64", force = true, relative = true }
-BORING_BSSL_INCLUDE_PATH = { value = "vendor/boringssl-fips/include/", force = true, relative = true }
+BORING_BSSL_FIPS_PATH = { value = "vendor/boringssl-fips/linux_x86_64", force = true, relative = true }
```
bleggett, line 4:
```suggestion
BORING_BSSL_FIPS_PATH = { value = "vendor/boringssl-fips/linux_x86_64", force = true, relative = true }
```
Does this also fix: https://github.com/istio/ztunnel/issues/399 ?


I haven't checked, but one of the minor annoyances before was we couldn't conditionally set this variable in `build.rs` per-arch/platform, because `boring-sys` was doing stuff in its own `build.rs` that both 

A) didn't work with vendoring
B) defeated cargo's "recompute" logic making it impossible to override in our `build.rs`

which means the build has to `sed` this and those of us on `linux_arm64` have to locally munge this file constantly.

howardjohn: I don't think so unless you buid with `ring` instead of boring which sidesteps the issue

howardjohn: Oh I forgot a bit part of this - I am proposing we default Istio builds to ring as well.. that decision can be decoupled from the broader PR though

ilrudie:
Big plus 1 for moving to rustls!

vendor/boringssl-fips/linux_x86_64/lib/libcrypto.a
bleggett, line 1: We vendor the arm64 `boring` libs as well, they also need to be moved.

src/tls/csr.rs
```diff
@@ -0,0 +1,70 @@
...
+
+        extensions.push(subject_alternative_name)?;
+        csr.add_extensions(&extensions)?;
+        csr.sign(&pkey, MessageDigest::sha256())?;
+
+        let csr = csr.build();
+        let pkey_pem = pkey.private_key_to_pem_pkcs8()?;
+        let csr_pem = csr.to_pem()?;
+        let csr_pem = std::str::from_utf8(&csr_pem)
+            .expect("CSR is valid string")?
```
bleggett, line 44:
```suggestion
            .expect("CSR is valid string")
```

bleggett (commented):
Seems happy building locally with both FIPS and non-FIPS variants on arm64 (with the vendored lib move mentioned in comments).

README.md
```diff
@@ -21,10 +21,23 @@ If a feature is not directly used to implement the node proxy component in ambie
...
+### TLS/Crypto provider
 
-Ztunnel builds currently enable the `fips` Cargo feature by default, which in turn enables the `fips` feature
-on [BoringSSL](https://github.com/cloudflare/boring).
+Ztunnel's TLS is built on [rustls](https://github.com/rustls/rustls).
+
+Rustls has support for plugging in various crypto providers to meet various needs (compliance, performance, etc).
+
+| Name                                          | How To Enable                                  |
+|-----------------------------------------------|------------------------------------------------|
```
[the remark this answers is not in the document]
howardjohn: Probably worth 2 targets (or maybe a variable that applies to all targets, not just `build` is probably better??) and build both in CI?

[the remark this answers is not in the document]
dhawton: I think a variable might be sufficient to reduce duplication. We probably don't need to do anymore than build in CI as part of the test

Cargo.toml
```diff
@@ -25,70 +26,78 @@ name = "throughput"
 harness = false
 
 [dependencies]
-http-02 = { package = "http", version = "0.2.9" }
+# Enabled with 'tls-boring'
+boring-rustls-provider = { git = "https://github.com/janrueth/boring-rustls-provider", optional = true } #
```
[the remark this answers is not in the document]
howardjohn: I am working to get this actively maintained (TBD by who). Note this is an _off by default_ feature; the default Istio builds do not use it.

bleggett (approved):
I'm happy with this approach, LGTM.

Cargo.toml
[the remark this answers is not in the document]
hzxuzhonghu: It has nothing todo with default or not. Every dependency should be actively maintained. So I get it, you own this crate

hzxuzhonghu (approved):

ericvn (approved):

PR #821, opened 2024-02-21
Explicitly panic (so test actually fails) if we cannot run it

bleggett:
`eprintln` doesn't actually fail the test in local `cargo test` runs - resulting in the test "silently" being skipped, which is not great if it would actually fail in the required environment.

### DIFF
--- a/Makefile.overrides.mk
+++ b/Makefile.overrides.mk
@@ -1,2 +1,15 @@
 # Use the build container by default
 BUILD_WITH_CONTAINER ?= 1
+
+# Cannot run root tests in container unless the container is privileged,
+# and we cannot/should not run it as privileged by default - so communicate that.
+ifeq ($(TEST_MODE), root)
+    ifeq ($(BUILD_WITH_CONTAINER), 1)
+        $(info ***** NOTE ******)
+        $(info TEST_MODE=root and BUILD_WITH_CONTAINER=1)
+        $(info If you wish to run tests that require root privilege inside the build container,)
+        $(info you must run the container as privileged yourself.)
+        $(info Alternatively, set BUILD_WITH_CONTAINER=0 and run the tests as root outside the build container.)
+        $(info ***** NOTE ******)
+	endif
+endif
--- a/Makefile.overrides.mk
+++ b/Makefile.overrides.mk
@@ -7,9 +7,11 @@ ifeq ($(TEST_MODE), root)
     ifeq ($(BUILD_WITH_CONTAINER), 1)
         $(info ***** NOTE ******)
         $(info TEST_MODE=root and BUILD_WITH_CONTAINER=1)
-        $(info If you wish to run tests that require root privilege inside the build container,)
-        $(info you must run the container as privileged yourself.)
-        $(info Alternatively, set BUILD_WITH_CONTAINER=0 and run the tests as root outside the build container.)
+        $(info Running tests as root inside a build container requires running the container in privileged mode)
+        $(info If you are uncomfortable with this or your environment does not allow priviledged containers,)
+        $(info set BUILD_WITH_CONTAINER=0 and run the tests as root outside the build container.)
         $(info ***** NOTE ******)
+		DOCKER_RUN_OPTIONS += --privileged
+		UID = 0
 	endif
 endif
--- a/Makefile.overrides.mk
+++ b/Makefile.overrides.mk
@@ -2,7 +2,7 @@
 BUILD_WITH_CONTAINER ?= 1
 
 # Cannot run root tests in container unless the container is privileged,
-# and we cannot/should not run it as privileged by default - so communicate that.
+# so if you set both those flags we will attempt to run the build container as privileged.
 ifeq ($(TEST_MODE), root)
     ifeq ($(BUILD_WITH_CONTAINER), 1)
         $(info ***** NOTE ******)
--- a/src/inpod/config.rs
+++ b/src/inpod/config.rs
@@ -176,7 +176,7 @@ mod test {
 
     macro_rules! fixture {
         () => {{
-            if !crate::test_helpers::can_run_privilged_test() {
+            if !crate::test_helpers::can_run_privileged_test() {
                 panic!("This test requires root - rerun with TEST_MODE=root; skipping");
             }
 
--- a/src/inpod/config.rs
+++ b/src/inpod/config.rs
@@ -177,8 +177,7 @@ mod test {
     macro_rules! fixture {
         () => {{
             if !crate::test_helpers::can_run_privilged_test() {
-                eprintln!("This test requires root; skipping");
-                return;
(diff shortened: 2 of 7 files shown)


howardjohn (commented):
This was intended; in CI it panics but for local dev it does not so you don't _have_ to run these tests. Basically equivalent to `t.Skip`.

Maybe `t.Skip` is not the best path here though? WDYT? The intent was to make `cargo test` actually just work without requiring hoops

bleggett:
> This was intended; in CI it panics but for local dev it does not so you don't _have_ to run these tests. Basically equivalent to `t.Skip`.
> 
> Maybe `t.Skip` is not the best path here though? WDYT? The intent was to make `cargo test` actually just work without requiring hoops

I don't think it's great local UX - without this, the following is the result when these tests (which are actually failing for me locally in a branch when run as root) are run "normally":

````
 » cargo test test_process_add                                                                                                               
    Blocking waiting for file lock on build directory
   Compiling ztunnel v0.0.0 (/home/bleggett/Source/solo-io/ztunnel-backup)
    Finished test [unoptimized + debuginfo] target(s) in 29.04s
     Running unittests src/lib.rs (out/rust/debug/deps/ztunnel-7fd42b2982b9b6f4)

running 2 tests
test inpod::workloadmanager::tests::test_process_add_and_del ... ok
test inpod::workloadmanager::tests::test_process_add ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 156 filtered out; finished in 0.00s
```` 

They fail as root, because they're actually _running_ as root. But that's not clear at all from the output.

I'd rather them fail explicitly if they cannot be run.

**Comment by howardjohn** [comment]
Yeah it will at least never pass CI though. If you `go test ./...` in Istio you are skipping tons of tests too (all the integration ones). Local test is not necessarily fully complete.

If we do this should we at least put `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER=sudo` in the config.toml so people don't have to specify it ? And I suppose we can enumerate a variety of targets (wish you didn't need this but... )

**Comment by bleggett** [comment]
> Yeah it will at least never pass CI though. If you `go test ./...` in Istio you are skipping tons of tests too (all the integration ones). Local test is not necessarily fully complete.

That's fine - if it says `skipped` or `ignored` that's one thing - if it says `passed` when it didn't actually _run_, that's really confusing.

> If we do this should we at least put `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER=sudo` in the config.toml so people don't have to specify it ? And I suppose we can enumerate a variety of targets (wish you didn't need this but... )

We are already setting this with the `make test-root` target.

**Comment by howardjohn** [comment]
> > Yeah it will at least never pass CI though. If you `go test ./...` in Istio you are skipping tons of tests too (all the integration ones). Local test is not necessarily fully complete.
> 
> That's fine - if it says `skipped` or `ignored` that's one thing - if it says `passed` when it didn't actually _run_, that's really confusing.

Yeah, rust doesn't support this :-(

> > If we do this should we at least put `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER=sudo` in the config.toml so people don't have to specify it ? And I suppose we can enumerate a variety of targets (wish you didn't need this but... )
> 
> We are already setting this with the `make test-root` target.

Sure, but I think its a pretty good property to run `cargo test` (or `go test`) on a project and have it just work instead of remembering obscure project-specific makefile commands. plus, its needed for interactive use (maybe i want some flags, specific tests, etc). While I personally can just set CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER since I work on ztunnel a lot, its probably a bit of friction for the occasionally contributor

**Comment by howardjohn** [comment]
or at the very least maybe we can explain in the error how to set CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER/equivilent

**Comment by bleggett** [comment]
I also note that the `sudo` prompt you get with the current `makefile` export of `CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER` requires a password, and since we are defaulting to `BUILD_WITH_CONTAINER=1` - it seems to be asking for the root password of the in-container user, not the local user, which is probably confusing.

Do we not add the `build-tools` user to the container-local local `wheel` group?

**Comment by bleggett** [comment]
Aight I settled on if `TEST_MODE=root` && `BUILD_WITH_CONTAINER=1`

1. We tell you this means we have to run the build container as privileged.
2. We override the docker args to do that.
3. If you don't want that, then run with BUILD_WITH_CONTAINER=0

File: src/inpod/netns.rs [hunk 1]
```diff
@@ -118,8 +118,7 @@ mod tests {
     #[test]
     fn test_run_works() {
         if !crate::test_helpers::can_run_privilged_test() {
-            eprintln!("This test requires root; skipping");
-            return;
+            panic!("This test requires root - rerun with TEST_MODE=root; skipping");
```
MorrisLaw, line 121: Should these panic messages include the `and BULD_WITH_CONTAINER=1` part too?

src/inpod/config.rs
```diff
@@ -177,8 +177,7 @@ mod test {
     macro_rules! fixture {
         () => {{
             if !crate::test_helpers::can_run_privilged_test() {
```
MorrisLaw, line 179: **nit**: privilged is spelled wrong, maybe we can update that since we're in here. Doesn't seem to be called in a lot of places other than the files you're already touching.

src/inpod/netns.rs
bleggett: I thought about that but it's orthogonal IMO - you can run with `TEST_MODE` root if you are running the container as privileged, the test doesn't care.

MorrisLaw (approved):

Makefile.overrides.mk
```diff
@@ -1,2 +1,17 @@
 # Use the build container by default
 BUILD_WITH_CONTAINER ?= 1
+
+# Cannot run root tests in container unless the container is privileged,
+# so if you set both those flags we will attempt to run the build container as privileged.
+ifeq ($(TEST_MODE), root)
```
stevenctl, line 6: This should be the default. If you don't want to/can't do this, we should have something like Go's testing.short()

bleggett: The problem with making root tests default-enabled is that since we also default to container builds, we are by-default asking people to run privileged containers that could very well trash their box in a non-recoverable fashion.

That's not a great default, and in general one of the reasons why I don't like unit tests that require root.

I don't want to make root runs the default unit test target - if we are concerned that this means test coverage is skipped, we should rewrite those tests with mocks to stub out OS functionality, or move them to integration tests (which is arguably where most of these belong anyway)

bleggett:
https://github.com/istio/ztunnel/pull/962 just drops the root req, so this is no longer needed

PR #832, opened 2024-02-29
service addressed waypoint

ilrudie:
implementation for sending to waypoint for service-addressed traffic

### DIFF
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -35,6 +35,7 @@ use crate::proxy::{util, Error, ProxyInputs, TraceParent, BAGGAGE_HEADER, TRACEP
 
 use crate::state::service::ServiceDescription;
 use crate::state::set_gateway_address;
+use crate::state::workload::gatewayaddress::Destination;
 use crate::state::workload::{NetworkAddress, Protocol, Workload};
 use crate::{hyper_util, proxy, rbac, socket};
 
@@ -380,7 +381,54 @@ impl OutboundConnection {
             None => return Err(Error::UnknownSource(downstream)),
         };
 
+        // to Service with waypoint
+        if let Some(target_address) = self
+            .pi
+            .state
+            .fetch_destination(&Destination::Address(NetworkAddress {
+                network: self.pi.cfg.network.clone(),
+                address: target.ip(),
+            }))
+            .await
+        {
+            match target_address {
+                crate::state::workload::address::Address::Workload(_w) => {
+                    // this is to workload, fall through
+                }
+                crate::state::workload::address::Address::Service(s) => {
+                    if let Some(wp) = s.waypoint.clone() {
+                        let addr = match wp.destination {
+                            Destination::Address(a) => a.address,
+                            Destination::Hostname(_) => todo!(), // it has a waypoint with an invalid addy... error?
+                        };
+                        let gateway = SocketAddr::new(addr, wp.hbone_mtls_port);
+                        let wp_upstream = self
+                            .pi
+                            .state
+                            .fetch_upstream(&self.pi.cfg.network, gateway)
+                            .await
+                            .unwrap(); // ugh, TODO NOT OK
+                        let destination_service = ServiceDescription::try_from(&*s).ok();
+
+                        return Ok(Request {
+                            protocol: Protocol::HBONE,
+                            direction: Direction::Inbound,
+                            source: source_workload,
+                            destination: target,
+                            destination_workload: None, // this is to Service traffic with a wp... gateway will handle workload selection
+                            destination_service,
+                            expected_identity: Some(wp_upstream.workload.identity()),
+                            gateway,
+                            request_type: RequestType::ToServerWaypoint,
+                            upstream_sans: wp_upstream.sans,
+                        });
+                    }
+                }
+            }
+        }
+
         // TODO: we want a single lock for source and upstream probably...?
+        // TODO(ilrudie) where I put to svc logic probably makes this worse...
         let us = self
             .pi
             .state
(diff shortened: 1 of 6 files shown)


src/proxy/outbound.rs
```diff
@@ -380,7 +381,69 @@ impl OutboundConnection {
...
+        if let Some(target_address) = self
+            .pi
+            .state
+            .fetch_destination(&Destination::Address(NetworkAddress {
+                network: self.pi.cfg.network.clone(),
+                address: target.ip(),
+            }))
+            .await
+        {
+            match target_address {
```
stevenctl, line 394: nit: `if let`?

stevenctl: could actually just do additional pattern matching on 385

src/proxy/outbound.rs
```diff
@@ -380,7 +381,69 @@ impl OutboundConnection {
...
+                            Destination::Address(a) => a.address,
+                            Destination::Hostname(_) => todo!(), // it has a waypoint with an invalid addy... error?
+                        };
+                        let waypoint_vip = SocketAddr::new(waypoint_vip, wp.hbone_mtls_port);
+                        let waypoint_us = self
+                            .pi
+                            .state
+                            .fetch_upstream(&self.pi.cfg.network, waypoint_vip)
+                            .await
+                            .unwrap(); // ugh, TODO NOT OK
```
hzxuzhonghu, line 410: should not use unwrap

src/proxy/outbound.rs
```diff
@@ -380,7 +381,69 @@ impl OutboundConnection {
...
+                                self.pi.metrics.clone(),
+                            )
+                            .await
+                            .unwrap(); // ugh, TODO NOT OK
+
+                        let waypoint_socket_address =
+                            SocketAddr::new(waypoint_ip, waypoint_us.port);
+                        let destination_service = ServiceDescription::try_from(&*s).ok();
+
+                        return Ok(Request {
```
hzxuzhonghu, line 428: can we abstract a fine grained function to reuse

ilrudie: Do you mean some sort of "request builder" instead of returning literals all the time?

hzxuzhonghu: yes

src/proxy/outbound.rs
```diff
@@ -380,6 +381,64 @@ impl OutboundConnection {
...
+                        &source_workload,
+                        self.pi.metrics.clone(),
+                    )
+                    .await?; // if we can't load balance just return the error
+
+                let waypoint_socket_address = SocketAddr::new(waypoint_ip, waypoint_us.port);
+
+                return Ok(Request {
+                    protocol: Protocol::HBONE,
+                    direction: Direction::Inbound,
```
howardjohn, line 429: isn't it outbound?

ilrudie: probably should be. I did a copy/paste to start building the request and it seems to be a hangover from a hangover.

https://github.com/istio/ztunnel/pull/486/files#r1187432943

stevenctl:
https://docs.google.com/document/d/1DAM_6SajCyC2hD8wBUraaZWZJqUVYPyemzw0-xpHmdY/edit?disco=AAABIl-P1-k

This thread affects the decision for the bypass stuff... which waypoint _must_ we traverse.

ilrudie:
Option 4 (or maybe 2.1) expose some API for the control plane to optionally create policy at the svc owner's discretion

src/proxy/outbound.rs
ilrudie: I'd rather not address in this PR if it's all the same. I opened #846 to track work on this idea.

hzxuzhonghu:
How can we skip sending to waypoint if a pod belongs to svc disabled waypoint according to https://docs.google.com/document/d/1DAM_6SajCyC2hD8wBUraaZWZJqUVYPyemzw0-xpHmdY/edit?disco=AAABIl-P1-k

ilrudie:
> How can we skip sending to waypoint if a pod belongs to svc disabled waypoint according to https://docs.google.com/document/d/1DAM_6SajCyC2hD8wBUraaZWZJqUVYPyemzw0-xpHmdY/edit?disco=AAABIl-P1-k

I think your question is what if a user declares they want a waypoint on a pod and declares they want #none for svc. It's an interesting case and I'm not sure the answer is clear here. Should #none on a service mean bypass all waypoints for to service traffic or just do not configure a namespace-defined waypoint?

louiscryan:
> > How can we skip sending to waypoint if a pod belongs to svc disabled waypoint according to https://docs.google.com/document/d/1DAM_6SajCyC2hD8wBUraaZWZJqUVYPyemzw0-xpHmdY/edit?disco=AAABIl-P1-k
> 
> I think your question is what if a user declares they want a waypoint on a pod and declares they want #none for svc. It's an interesting case and I'm not sure the answer is clear here. Should #none on a service mean bypass all waypoints for to service traffic or just do not configure a namespace-defined waypoint?
> 
> cc @louiscryan and @stevenctl

Use-waypoint is a directive around how to route 'addressed' traffic so we should not infer anything from  "service X capture #none" other than that the user wants traffic _addressing_ that service (via VIP/host) to not transit a waypoint. The user can simultaneously say that they want traffic _addressing_ a pod which is a member of that service *to go through a waypoint*. There is no conflict here, its just important to document for users the semantics of capture on the addressing mechanism.

If users want to remove ambiguity they should just capture namespaces for the most part.

hzxuzhonghu (approved):
That is clear, a service label indicates whther to tansit traffic targeted at its clusterip. rather than targeted for its pods

PR #850, opened 2024-03-11
proxy protocol for sandwich

stevenctl:
We add `ApplicationTunnel` as an optional piece of `GatewayAddress` to allow more advanced sandwiches:

* It's possible that a sandwiched waypoint wants additional info like PROXY protocol
* It's possible that the sandwiched waypoint binds to a different port than the HBONE address targets

### DIFF
--- a/proto/workload.proto
+++ b/proto/workload.proto
@@ -157,6 +157,12 @@ message Workload {
   // * Requests *to* this workload, via the tunnel_protocol, do not need to be de-tunneled.
   bool native_tunnel = 14;
 
+  
+  // If an application, such as a sandwiched waypoint proxy, supports
+  // directly receiving information from zTunnel they can set application_protocol.
+  // This supersedes native_tunnel.
+  ApplicationTunnel application_tunnel = 23;
+
   // The services for which this workload is an endpoint.
   // The key is the NamespacedHostname string of the format namespace/hostname.
   map<string, PortList> services = 22;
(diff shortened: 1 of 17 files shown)


proto/workload.proto
```diff
@@ -212,6 +212,30 @@ enum TunnelProtocol {
...
+message ApplicationTunnel {
+  enum Protocol {
+    // Bytes are copied from the inner stream without modification.
+    NONE = 0;
+
+    // Prepend PROXY protocol headers before copying bytes
+    // Standard PROXY source and destination information
+    // is included, along with potential extra TLV headers:
+    // 0xD0 - The SPIFFE identity of the source workload
+    // 0xD1 - The FQDN or Hostname of the targeted Service
```
ilrudie, line 227: Is 0xD1 no longer used, automatically implemented by the crate or a TODO?

stevenctl: This is planned in the future to allow routing by host instead of VIP.

ilrudie: sounds good. TY

src/proxy/inbound.rs
```diff
@@ -432,26 +469,43 @@ impl Inbound {
...
-            return Ok((next_hop, us_wl, us_svc));
-        }
+            (next_hop, us_wl, us_svc)
+        } else {
+            return Err(Error::IPMismatch(conn.dst.ip(), hbone_addr.ip()));
+        };
 
-        Err(Error::IPMismatch(conn.dst.ip(), hbone_addr.ip()))
+        // Application tunnel may override the port.
+        let (upstream_addr, inbound_protocol) = match upstream.application_tunnel.clone() {
```
howardjohn, line 496: nit: just make Protocol `Copy` the you won't need the `clone()`

src/proxy/inbound.rs
```diff
@@ -551,6 +605,12 @@ pub(super) enum InboundConnect {
     DirectPath(TcpStream),
     /// Hbone is a standard HBONE request coming from the network.
     Hbone(Request<Incoming>),
+    // PROXY adds source and dest headers and source identity before forwarding bytes.
+    Proxy(
+        Request<Incoming>,
+        (SocketAddr, SocketAddr),
+        Option<Identity>,
```
howardjohn, line 612: nit: any reason to not just  pass the full `rbac::Connection`?

howardjohn (approved):

PR #859, opened 2024-03-26
replace hbone-addr with target

hzxuzhonghu:
It makes readable especially with sandwich


stevenctl:
I don't have a very strong opinion,  but FWIW I like hbone_addr or hbone_target to help differentiate what the "next hop" target is vs the inner-tunnel target. I guess the hop between inbound-zTunnel and the sandwiched-Waypoint isn't "hbone" but it's propagating hbone info.

hzxuzhonghu:
There is another reason; we have used target in outbound.rs widely.

PR #865, opened 2024-04-02
Remove gperftools, use jemalloc

bleggett:
Seems to actually work, requires no container deps, keeping as an optional build target for now.

Signed-off-by: Benjamin Leggett <benjamin.leggett@solo.io>


Cargo.toml
```diff
@@ -6,7 +6,7 @@ rust-version = "1.65"
 
 [features]
 default = ["tls-ring"]
-gperftools = ["dep:gperftools"]
+jemalloc = ["dep:tikv-jemallocator", "dep:jemalloc_pprof"]
```
howardjohn, line 9: should we default to jemalloc? and/or unconditionally use it? or do we think we need more performance testing?

src/admin.rs
```diff
@@ -406,61 +395,39 @@ fn change_log_level(reset: bool, level: &str) -> Response<Full<Bytes>> {
...
-        .unwrap();
-    Response::builder()
-        .status(hyper::StatusCode::OK)
-        .body(buffer.into())
-        .unwrap()
+#[cfg(feature = "jemalloc")]
+async fn handle_jemalloc_pprof_heapgen(_req: Request<Incoming>) -> Response<Full<Bytes>> {
+    const FILE_PATH: &str = "/tmp/jemalloc-pprof-heap.prof";
+    // make sure the file exists/is writable before handing to gprof
+    File::create(FILE_PATH).await.unwrap();
```
howardjohn, line 402: can we stream it back instead of file? is the format the "pprof" format that can be read by `go tool pprof` or some other format?

howardjohn (approved):

Cargo.toml
bleggett: For now I would rather not, this is just a drop-in replacement for what we had before that works, but we can certainly look into doing that later, yeah.

src/admin.rs
bleggett: `go tool pprof` - starting a PROFILING.md

Issue #862, opened 2024-04-02
Locality aware routing


howardjohn:
We should support some 'locality aware' load balancing in ztunnel. 

Kubernetes core Service has this now in 1.30: https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/4444-service-routing-preference#summary. We can utilize this as the API.

This gives a 'PreferClose' to strictly use the close endpoints only, if they are available.

There is also existing "use only same node" mode that we will want to implement.

We will want to make PreferClose the default for waypoints.

API design TBD

PR #868, opened 2024-04-03
Implement locality aware load balancing

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/862

### DIFF
--- a/proto/workload.proto
+++ b/proto/workload.proto
@@ -82,27 +82,43 @@ message Service {
 }
 
 message LoadBalancing {
-  enum Targets {
-    UNSPECIFIED = 0;
+  enum Scope {
+    UNSPECIFIED_SCOPE = 0;
+    // Prefer traffic in the same region.
     REGION = 1;
+    // Prefer traffic in the same zone.
     ZONE = 2;
+    // Prefer traffic in the same subzone.
     SUBZONE = 3;
+    // Prefer traffic on the same node.
     NODE = 4;
+    // Prefer traffic in the same cluster.
     CLUSTER = 5;
+    // Prefer traffic in the same network.
     NETWORK = 6;
   }
   enum Mode {
-    UNSPECIFIEDx = 0;
+    UNSPECIFIED_MODE = 0;
 
+    // In STRICT mode, only endpoints that meets all of the routing preferences will be considered.
+    // This can be used, for instance, to keep traffic ONLY within the same cluster/node/region.
+    // This should be used with caution, as it can result in all traffic being dropped if there is no matching endpoints,
+    // even if there are endpoints outside of the preferences.
     STRICT = 1;
+    // In FAILOVER mode, endpoint selection will prefer endpoints that match all preferences, but failover to groups of endpoints
+    // that match less (or, eventually, none) preferences.
+    // For instance, with `[NETWORK, REGION, ZONE]`, we will send to:
+    // 1. Endpoints matching `[NETWORK, REGION, ZONE]`
+    // 2. Endpoints matching `[NETWORK, REGION]`
+    // 3. Endpoints matching `[NETWORK]`
+    // 4. Any endpoints
     FAILOVER = 2;
-
   }
-  // routing_preference defines what scopes we want to keep traffic within. These are ANDed together
-  // For example [LOCALITY, NETWORK] means we want to keep traffic within the same locality and network.
-  // The earlier items in the list are prioritized; if the above config was used with `mode: FAILOVER`,
-  // we would prefer endpoints in the same locality before endpoints only in the same network.
-  repeated Targets routing_preference = 1;
+
+  // routing_preference defines what scopes we want to keep traffic within.
+  // The `mode` determines how these routing preferences are handled
+  repeated Scope routing_preference = 1;
+  // mode defines how we should handle the routing preferences.
   Mode mode = 2;
 }
 
--- a/proto/workload.proto
+++ b/proto/workload.proto
@@ -74,6 +74,36 @@ message Service {
   // Waypoint is the waypoint proxy for this service. When set, all incoming requests must go
   // through the waypoint.
   GatewayAddress waypoint = 7;
+
+  // Load balancing policy for selecting endpoints.
+  // Note: this applies only to connecting directly to the workload; when waypoints are used, the waypoint's load_balancing
+  // configuration is used.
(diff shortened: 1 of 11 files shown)


proto/workload.proto
```diff
@@ -74,6 +74,36 @@ message Service {
...
+    UNSPECIFIED = 0;
+    REGION = 1;
+    ZONE = 2;
+    SUBZONE = 3;
+    NODE = 4;
+    CLUSTER = 5;
+    NETWORK = 6;
+  }
+  enum Mode {
+    UNSPECIFIEDx = 0;
```
louiscryan, line 95: typo?

proto/workload.proto
```diff
@@ -74,6 +74,36 @@ message Service {
...
+
+    STRICT = 1;
+    FAILOVER = 2;
+
+  }
+  // routing_preference defines what scopes we want to keep traffic within. These are ANDed together
+  // For example [LOCALITY, NETWORK] means we want to keep traffic within the same locality and network.
```
louiscryan, line 102: no 'LOCALITY' value

proto/workload.proto
```diff
@@ -74,6 +74,36 @@ message Service {
...
   GatewayAddress waypoint = 7;
+
+  // Load balancing policy for selecting endpoints.
+  // Note: this applies only to connecting directly to the workload; when waypoints are used, the waypoint's load_balancing
+  // configuration is used.
+  LoadBalancing load_balancing = 8;
+}
+
+message LoadBalancing {
+  enum Targets {
```
louiscryan, line 85: Target is a pretty vague term, maybe 'Locality'

proto/workload.proto
```diff
@@ -74,6 +74,36 @@ message Service {
...
+    UNSPECIFIED = 0;
```
louiscryan, line 86: would ANY or NONE read more cleanly in docs.

"When the locality_preference is set to [NONE] traffic can be routed to any endpoint that is a member of the mesh"

proto/workload.proto
```diff
@@ -74,6 +74,36 @@ message Service {
...
+    ZONE = 2;
+    SUBZONE = 3;
+    NODE = 4;
+    CLUSTER = 5;
+    NETWORK = 6;
+  }
+  enum Mode {
+    UNSPECIFIEDx = 0;
+
+    STRICT = 1;
```
louiscryan, line 97: Doc.  "Only if the set of endpoints in each target defined in routing_preference is empty or every endpoint is unreachable can an endpoint in the next target in routing_preference be used"

src/state.rs
```diff
@@ -250,6 +255,60 @@ impl ProxyState {
         }
         None
     }
+
+    fn load_balance<'a>(&self, src: &Workload, svc: &'a Service) -> Option<&'a Endpoint> {
```
ilrudie, line 259: this seems to ignore failover/strict and always does a best match, yeah?

howardjohn:
```
                        // Doesn't match all, and required to. Do not select this endpoint
                        if lb.mode == LoadBalancerMode::Strict && rank != lb.targets.len() {
                            return None;
                        }
```

should do it

src/state.rs
```diff
@@ -250,6 +255,62 @@ impl ProxyState {
...
+                                    src.locality.subzone == wl.locality.subzone
+                                }
+                                LoadBalancerScopes::Node => src.node == wl.node,
+                                LoadBalancerScopes::Cluster => src.cluster_id == wl.cluster_id,
+                                LoadBalancerScopes::Network => src.network == wl.network,
+                            };
+                            if matches {
+                                rank += 1;
+                            } else {
+                                break;
```
stevenctl, line 293: We should document if we always expect the scopes to be "size-ordered"

howardjohn: Its meant to be documented here:
```
   // In FAILOVER mode, endpoint selection will prefer endpoints that match all preferences, but failover to groups of endpoints
    // that match less (or, eventually, none) preferences.
    // For instance, with `[NETWORK, REGION, ZONE]`, we will send to:
    // 1. Endpoints matching `[NETWORK, REGION, ZONE]`
    // 2. Endpoints matching `[NETWORK, REGION]`
    // 3. Endpoints matching `[NETWORK]`
    // 4. Any endpoints
```

Which is also how failover works in Istio-envoy

src/state.rs
```diff
@@ -575,8 +644,8 @@ impl DemandProxyState {
             }
         };
         let wp_socket_addr = SocketAddr::new(wp_nw_addr.address, gw_address.hbone_mtls_port);
-        match self
-            .fetch_upstream(&wp_nw_addr.network, wp_socket_addr)
+        match self // THIS IS THE WRONG WORKLOAD
```
stevenctl, line 647: just need to pass in src, right?

stevenctl:
LGTM

src/state.rs
ilrudie: I was also going to ask about this wrt the `break` being ok or not. I think it relies on the routing preferences being guaranteed ordered correctly.

src/state.rs
howardjohn: ah good catch, forgot about this todo. Yep

src/state.rs
howardjohn: My intended implementation was to make order matter here. Open to changes though

howardjohn:
@stevenctl this should be good to go, will open istio side once this lands

src/state.rs
ilrudie: I think that's fine as long as folks are aware that it matters.

stevenctl (approved):

proto/workload.proto
```diff
@@ -74,6 +74,52 @@ message Service {
...
+    // 3. Endpoints matching `[NETWORK]`
+    // 4. Any endpoints
+    FAILOVER = 2;
+  }
+
+  // routing_preference defines what scopes we want to keep traffic within.
+  // The `mode` determines how these routing preferences are handled
+  repeated Scope routing_preference = 1;
+  // mode defines how we should handle the routing preferences.
+  Mode mode = 2;
```
hzxuzhonghu, line 122: With this api, how can user set it for `strict`? 

And how can we expand to support other algorithms with this struct?

daixiang0: Package this as one algorithm if need to expand?

PR #873, opened 2024-04-03
cleanup and to-service traffic handling fixes

ilrudie:
- some adjustments to how svc-addressed traffic is handled
- some cleanup of mutable variables and less idiomatic is_none + unwrap logic


src/proxy/inbound.rs
```diff
@@ -372,14 +372,14 @@ impl Inbound {
...
-                        .unwrap());
-                }
+                // if has_waypoint && !from_waypoint {
+                //     info!(%rbac_ctx.conn, "bypassed waypoint");
+                //     connection_manager.release(&rbac_ctx).await;
+                //     return Ok(Response::builder()
+                //         .status(StatusCode::UNAUTHORIZED)
+                //         .body(Empty::new())
+                //         .unwrap());
+                // }
```
ilrudie, line 382: What do we think about this? Something similar will be necessary at some point but are we comfortable doing that before istiod is able to generate a L4 policy with similar effects?

cc @louiscryan @howardjohn @stevenctl

stevenctl: If we drop this, we should drop `debug!("request from waypoint, skipping policy")`

ilrudie: yeah, that tracks

ilrudie: I had been testing with this but forgot it wasn't pushed yet. Pushed now.

howardjohn (approved):

howardjohn (approved):

Issue #84, opened 2022-11-09
Connection Logging


stevenctl:
Cheap, in memory, stores some list of past events, similar to Envoy access logs.

There should be a cheap in-memory connection log that's accessible for debugging purposes. The runtime cost of maintaining this log should be extremely low. This could be achieved by a per-thread in-memory list of events including the 5-tuple of the connection, what happened (open, close, error, pkt/bytes sent) and a timestamp. Whenever something happens on a connection, we can add to this list a note at very low cost. There could be a configurable limit to the size of the thing.
With this, when debugging a user could run a ztctl command to explore this data. They should be able to pass in an IP or 5-tuple and see all connections that have occurred recently with that configuration. There should be a watch command that shows the connections currently flowing through the system as they come and go. The debug dump tool should dump every thread's connection log so that it can be analyzed off-box by a support engineer.

Issue #861, opened 2024-04-01
Correlate inpod UIDs with known objects


howardjohn:
We see logs like 

```
2024-03-31T03:48:39.160862Z  WARN proxy{uid=7136ad06-c86c-4879-b89f-8eedd1377582}:outbound{id=c6448ce0a846e223caca1d2a903a4c2a}: ztunnel::proxy::outbound: failed dur=106.83µs err=unknown waypoint: unable to determine waypoint upstream
```

There is no way to correlate the UID (`7136ad06-c86c-4879-b89f-8eedd1377582`) with anything; its not in config_dump, etc

daixiang0:
We can add output for `WorkloadData` in dump refer to https://github.com/istio/ztunnel/blob/af8532da131a6d7047766b3cb71eb10435afcd86/src/inpod/statemanager.rs#L108-L130

Thus the uid can connect with specified `WorkloadData` and we can see all info.

PR #883, opened 2024-04-04
admin: add connection info dump

howardjohn:
Example:
```json
{
  "dd03d501-825e-4fa9-b0c4-b31348219a11": {
    "state": "Up",
    "connections": []
  },
  "beb72dff-67bb-44f0-9edc-59d7a1dda275": {
    "state": "Up",
    "connections": [
      {
        "conn": {
          "src": "10.244.2.9:55393",
          "dst": "10.244.2.7:9090",
          "src_identity": "spiffe://cluster.local/ns/default/sa/default",
          "dst_network": ""
        }
      }
    ]
  },
  "432c28a6-4b03-4677-9777-242c4a437ef0": {
    "state": "Up",
    "connections": [
      {
        "conn": {
          "src": "10.244.2.9:53029",
          "dst": "10.244.2.10:80",
          "src_identity": "spiffe://cluster.local/ns/default/sa/default",
          "dst_network": ""
        }
      }
    ]
  }
}
```

Fixes https://github.com/istio/ztunnel/issues/84

There is a lot of changes from making the mutex use `std` mutex; tokio mutex is not needed since we never hold it across an `await` point

### DIFF
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -50,6 +50,15 @@ impl EncodeLabelValue for Identity {
     }
 }
 
+impl serde::Serialize for Identity {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        self.to_string().serialize(serializer)
+    }
+}
+
 impl FromStr for Identity {
     type Err = Error;
     fn from_str(s: &str) -> Result<Self, Self::Err> {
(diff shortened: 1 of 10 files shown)


hzxuzhonghu:
What does the uid mean? Not readable for user

howardjohn:
> What does the uid mean? Not readable for user

I agree, tracking improvement there in https://github.com/istio/ztunnel/issues/861. I didn't add the UID in this PR

hzxuzhonghu (approved):

PR #884, opened 2024-04-04
e2e: use less clients where possible

howardjohn:
This drops 16 network namespace we need to setup. They take about a
second each on my machine.


bleggett (approved):
SGTM, if we want to do a proper multi-client loadtest this is not the spot anyway.

stevenctl (approved):

Issue #770, opened 2024-01-16
Pooling key needs to take source IP


howardjohn:
https://github.com/istio/ztunnel/blob/d867379ffa457153034d0d6c911810453caa8505/src/proxy/pool.rs#L85-L91

No src pod here. So if you send request from 2 pods on the same node to some destination pod, the destination will see them as from the same src IP incorrectly

ilrudie:
Would this be a Beta blocker? Seems like it's not a security issue but more a bug that'll make telemetry and troubleshooting weird.

ilrudie:
(also, do you have code for this John? If not and we think it'll block Beta I'd be happy to add it to my up next)

howardjohn:
I forgot to mention it mostly/entirely doesn't matter for inpod. So that may be the only "Beta" part anyways. But ignoring that I would say yes. The fix should be pretty simple ,though. We should make sure `src` is not `SocketAddr`  but `IpAddr`.

> (also, do you have code for this John? If not and we think it'll block Beta I'd be happy to add it to my up next)


I haven't worked on this yet, just noticed with @srampal  and tested to confirm its broken

ilrudie:
yeah, fair point re: inpod. I agree the fix doesn't seem like it would be difficult

srampal:
@howardjohn @ilrudie If you arent working on this, I could potentially work on this as it seems a relatively simple fix for someone new to the code to get experience making a code fix. Plus its an area I've been looking into.

hzxuzhonghu:
@howardjohn Can you clarify why cannot we share connection per {src-dst} identity?

ilrudie:
@srampal, go for it. I had not begun any work on this

hzxuzhonghu:
OK, I get sharing connection per identify could be wrong in some case. But if not possible or share per IP,  i donot think the connection pool is as useful as now.

ilrudie:
I don't think this really changes the value of pooling connections much. From my understanding this only really happens when 2 or more source pods sharing an identity exist on the same node and the fix would be basically the same as if those 2 pods happened to be on different nodes in terms of pool efficiency. That still seems fine to me.

PR #886, opened 2024-04-04
pool: include source IP in key

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/770


bleggett (approved):
As mentioned in the issue this is only relevant when running ztunnel in `Shared` mode (which we do not do), but as long as shared mode exists as an option, this makes sense to fix the pooling.

stevenctl (approved):

bleggett:
...actually this defeats pooling for inpod as well, which I missed, as it takes the srcport into account.

But pooling redo coming shortly so that's probably for the best.

PR #894, opened 2024-04-05
Ensure some spawns have deadlines

bleggett:
1. TLS handshakes (tokio-rustls) should have a deadline (practically not a big deal for us, but still) 

2. During certain scenarios where workloads are removed, inbound server "graceful shutdown" can deadlock while `await`ing http2 server to finish up, leaving the spawned task to hang indefinitely, so add a deadline to that as a safety measure (still looking into why it does, the `shutdown` part of `graceful_shutdown` implies a guaranteed termination, but deadline seems safer regardless)

- Also updates to hyper `1.2.0` (if this is controversial, can drop it)
- Also drops `tokio-console` stuff (don't find it terribly useful, also our logrates @ tracelevel tend to overwhelm it, will keep if someone feels strongly)

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -148,8 +148,8 @@ impl Inbound {
                             Err(e) => {
                                 error!("inbound serve drain err: {e}");
                                 Ok(())
-                            },
-                         };
+                            }
+                        };
                         debug!("inbound serve drain done");
                         res
                     }
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -27,8 +27,9 @@ use hyper::service::service_fn;
 use hyper::{Method, Request, Response, StatusCode};
 
 use tokio::net::{TcpListener, TcpStream};
+use tokio::time::timeout;
 
-use tracing::{debug, error, info, instrument, trace, trace_span, warn, Instrument};
+use tracing::{debug, error, info, instrument, trace_span, warn, Instrument};
 
 use super::connection_manager::ConnectionManager;
 use super::{Error, SocketFactory};
@@ -82,7 +83,6 @@ impl Inbound {
     }
 
     pub(super) async fn run(self) {
-        // let (tx, rx) = oneshot::channel();
         let acceptor = InboundCertProvider {
             state: self.pi.state.clone(),
             cert_manager: self.pi.cert_manager.clone(),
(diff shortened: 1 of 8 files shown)


src/proxy/inbound.rs
```diff
@@ -134,12 +135,29 @@ impl Inbound {
                 match futures_util::future::select(Box::pin(drain.signaled()), serve).await {
                     // We got a shutdown request. Start gracful shutdown and wait for the pending requests to complete.
                     futures_util::future::Either::Left((_shutdown, mut server)) => {
+                        debug!("inbound serve got drain {:?}", server);
                         let drain = std::pin::Pin::new(&mut server);
                         drain.graceful_shutdown();
-                        server.await
+                        // There are scenarios where the http2 server never resolves after
+                        // `graceful_shutdown`, which will hang the whole task.
```
howardjohn, line 142: Interesting. So before `inpod`, drain was the entire process shutting down. Now its not, so it MUST terminate in some bounded time.

Do we want this here, or at the `inpod` layer?

src/tls/workload.rs
```diff
@@ -146,12 +150,14 @@ where
...
             let tls = acceptor.fetch_cert(&conn).await?;
-            tokio_rustls::TlsAcceptor::from(tls)
+            let tls_accept = tokio_rustls::TlsAcceptor::from(tls)
                 .accept(conn)
-                .map_err(TlsError::Handshake)
-                .await
-        })
-    }
+                .map_err(TlsError::Handshake);
+            timeout(Duration::from_secs(TLS_HANDSHAKE_TIMEOUT), tls_accept).map_err(move |e| {
```
howardjohn, line 156: is https://docs.rs/tls-listener/latest/tls_listener/struct.Builder.html#method.handshake_timeout not sufficient here? defaults to 10s already

howardjohn (commented):
lgtm

src/proxy/inbound.rs
bleggett: > Do we want this here, or at the inpod layer?

I'm not sure where else to put it, but in terms of error reporting this seems best - this task is spawned to run to completion, and the inpod layer already initiates the drain and waits for it, the most it can do is drop the drain.

(we could do spawnhandles and aborts and punch this up a layer I guess, but the effect would be the same)

src/tls/workload.rs
bleggett: Ah, I missed that, yep - should handle it, reverted that commit.

howardjohn (approved):

PR #898, opened 2024-04-09
Implement outbound drain

bleggett:
Similar to https://github.com/istio/ztunnel/pull/894 - there are some other `spawns` that we need to guarantee terminate, to avoid stuck tasks.

Most of these are in Outbound, which unlike the others never really had per-connection drain.

This adds drains such that all outbound `spawns` are guaranteed to terminate when the workload is drained, this avoids several scenarios where task leaks are possible.

What we _really_ need beyond this PR is something like a `drain_type` and more complex drain channels:

- Drain(reason: workload gone) -> workload gone, don't wait for connections to "unstick" - just drop.
- Drain(reason: proxy terminating) -> workloads are not gone, but whole proxy is going - start refusing conns and give workloads a grace period.

This PR is just a stopgap to handle some workload termination corner cases.

### DIFF
--- a/src/proxy/inbound_passthrough.rs
+++ b/src/proxy/inbound_passthrough.rs
@@ -62,11 +62,13 @@ impl InboundPassthrough {
     }
 
     pub(super) async fn run(self) {
+        let inner_drain = self.drain.clone();
         let accept = async move {
         loop {
             // Asynchronously wait for an inbound socket.
             let socket = self.listener.accept().await;
             let pi = self.pi.clone();
+            let stream_drain = inner_drain.clone();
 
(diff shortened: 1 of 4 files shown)


howardjohn (approved):

bleggett:
What we really need beyond this PR is something like a drain_type and more complex drain channels:

- Drain(reason: workload gone) -> workload gone, don't wait for connections to "unstick" - just drop.
- Drain(reason: proxy terminating) -> workloads are not gone, but whole proxy is going - start refusing conns and give workloads a grace period.

This PR is just a stopgap to handle some workload termination corner cases.

PR #903, opened 2024-04-09
README: add rust version comment

### LINKED ISSUE #902: failed to build
```
root:[ztunnel]$ cargo check
error: Package `ztunnel v0.0.0 (/root/github/ztunnel)` does not have feature `rcgen`. It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.
root:[ztunnel]$ cargo version
cargo 1.65.0 (4bc8f24d3 2022-10-20)
```

daixiang0:
Fix #902

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,10 @@ default:
 shell:
 	@$(RUN) /bin/bash
 
-.PHONY: default shell
+rust-version:
+	@$(RUN) /usr/bin/rustc -vV
+
+.PHONY: default shell rust-version
 
 else
 
--- a/Makefile
+++ b/Makefile
@@ -49,10 +49,7 @@ default:
 shell:
 	@$(RUN) /bin/bash
 
-rust-version:
-	@$(RUN) /usr/bin/rustc -vV
-
-.PHONY: default shell rust-version
+.PHONY: default shell
 
 else
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Please use the same Rust version as the [`build-tools`](https://github.com/istio
 You can determine the version that the `build-tools` image uses by running the below command:
 
 ```shell
-$ make rust-version
+$ BUILD_WITH_CONTAINER=1 make rust-version
 ```
 
 ### TLS/Crypto provider
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ The details of architecture is [here](./ARCHITECTURE.md).
 ## Building
 
 Please use the same Rust version as the below command output:
+
 ```
 $ make rust-version
 ```
--- a/README.md
+++ b/README.md
@@ -23,7 +23,10 @@ The details of architecture is [here](./ARCHITECTURE.md).
 
 ## Building
 
-Please use Rust 1.70+.
+Please use the same Rust version as the below command output:
+```
+$ make rust-version
+```
 
 ### TLS/Crypto provider
 
--- a/README.md
+++ b/README.md
@@ -23,7 +23,8 @@ The details of architecture is [here](./ARCHITECTURE.md).
 
 ## Building
 
-Please use the same Rust version as the below command output:
+Please use the same Rust version as the [`build-tools`](https://github.com/istio/tools/tree/master/docker/build-tools) image.
+You can determine the version that the `build-tools` image uses by running the below command:
 
 ```shell
 $ make rust-version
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -21,6 +21,8 @@ If a feature is not directly used to implement the node proxy component in ambie
 
 ## Building
 
+Please use Rust 1.70+.
```
howardjohn, line 24: Can we have a single source of truth here? `Cargo.toml` has a `rust-version` field already. Probably its out of date, though

bleggett: My preference here (and what I do internally) is to say

"please use the Rust version that ships with the repo's `build-tools` image"

We can automate running build-tools and dump the `rustc` version to console with a make target if we wanna get fancy.

daixiang0: I miss it, let me update doc, thanks.

Makefile
```diff
@@ -49,7 +49,10 @@ default:
 shell:
 	@$(RUN) /bin/bash
 
-.PHONY: default shell
+rust-version:
```
howardjohn, line 52: this file is copied from common-files repo and will be reverted. Only makefile.core.mk should be used. 

you could just list "common/scripts/run.sh rustc --version" or similar

daixiang0: After test, run in core.md would raise an error since no BUILD_WITH_CONTAINER value enabled, so write it in the overrides.mk.

ericvn: I believe a simple:
```
rust-version:
	rustc -vV
```
should have worked in the Makefile.core.mak but this works also.

README.md
```diff
@@ -21,6 +21,12 @@ If a feature is not directly used to implement the node proxy component in ambie
 
 ## Building
 
+Please use the same Rust version as the below command output:
```
bleggett, line 24:
```suggestion
Please use the same Rust version as the [`build-tools`](https://github.com/istio/tools/tree/master/docker/build-tools) image - you can determine this by running the below command:
```

Makefile
daixiang0: > I believe a simple:
> 
> ```
> rust-version:
> 	rustc -vV
> ```
> 
> should have worked in the Makefile.core.mk but this works also.

This need to be run in the dev container.

ericvn: `make rust-version` would start the build-tools container and then run the same command inside the container. Assuming one isn't overriding the config to not run inside the build container. Or maybe I'm misunderstanding your setup.

bleggett: `/common/scripts/run.sh /usr/bin/rustc -vV` will work regardless of BUILD_WITH_CONTAINER, which is what we want.

But @ericvn is correct - @daixiang0 can you just move this target to `Makefile.core.mk` with the other targets, the command you have works just as well there.

ericvn: But beware as running rustc outside the container will tell you your local version of rust (if you have it installed, else an error), and not the one in the build-tools container, which is what the doc says.

bleggett: > But beware as running rustc outside the container will tell you your local version of rust (if you have it installed, else an error), and not the one in the build-tools container, which is what the doc says `You can determine the version that the `build-tools` image uses...`

Yep with `run.sh` it uses the build container regardless of whether I specify `BUILD_WITH_CONTAINER=0|1` on the CLI, so we should be good on that front.

ericvn: Gotcha. So we want this to work from the command line irrespective of if the developer has BUILD_WITH_CONTAINER set to 0? That does make the problem trickier. What's the probability that people are running with BUILD_WITH_CONTAINER=0? It's probably higher than I think..

From my testing, what is in this PR right now (calling run... in the Makefile) would work if BUILD_WITH_CONTAINER=1, but will fail with BUILD_WITH_CONTAINER=0. Also if someone is inside a `make shell`, a `make rust-version` will fail as well.

My suggestion works in the shell and locally except for `BUILD_WITH_CONTAINER=0 make rust-version` as that will try and show the locally installed rust version.

Would need to try some more experimentation to find something that might work everywhere.

bleggett: Yeah I apologize - I noticed that the `build-tools` image was pulled either way, but you're correct, whether the actual command runs  is contingent on that variable.

```
rust-version:
	rustc -vV
```

is enough for this, IMO. It will work inside and outside (unless someone overrides defaults) If someone is explicitly setting `BUILD_WITH_CONTAINER=0` (not the default) - they are already on their own as far as build tooling.

We can fence the run with `BUILD_WITH_CONTAINER` checks but that's excessive.

daixiang0: @bleggett Done.

howardjohn (approved):

Issue #876, opened 2024-04-03
DNS logs are spammy


howardjohn:
```
2024-04-03T21:26:23.872731Z  INFO hickory_server::server::server_future: request:8504 src:UDP://10.244.0.80#58719 QUERY:waypoint.echo-4-12380.svc.cluster.local.echo-4-12380.svc.cluster.local.:A:IN qflags:RD response:NXDomain rr:0/0/0 rflags:RD
2024-04-03T21:26:23.872762Z  INFO hickory_server::server::server_future: request:1806 src:UDP://10.244.0.80#46816 QUERY:waypoint.echo-4-12380.svc.cluster.local.echo-4-12380.svc.cluster.local.:AAAA:IN qflags:RD response:NXDomain rr:0/0/0 rflags:RD
```

We get a log for every request. We should drop this off by default. Either make it DEBUG level somehow, or default hickory_server to warn

ilrudie:
you beat me to it :)

PR #907, opened 2024-04-09
logs: hide dns logs by default

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/876

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -759,7 +759,10 @@ mod tests {
 
         let resp = change_log_level(true, "");
         let resp_str = get_response_str(resp).await;
-        assert_eq!(resp_str, "current log level is info\n");
+        assert_eq!(
+            resp_str,
+            "current log level is hickory_server::server::server_future=off,info\n"
+        );
 
         let resp = change_log_level(true, "invalid_level");
         let resp_str = get_response_str(resp).await;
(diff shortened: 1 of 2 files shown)


bleggett (approved):
We should explicitly document how to re-enable them for debugging but I notice that we don't have any info around RUST_LOG overrides in LOCAL.md yet, and we have several PRs out for that already, so that can come after.

howardjohn:
> We should explicitly document how to re-enable them for debugging but I notice that we don't have any info around RUST_LOG overrides in LOCAL.md yet, and we have several PRs out for that already, so that can come after.

I document it in https://github.com/istio/ztunnel/pull/874 (though not specifically the obscure DNS target name. I actually wonder if we can/should overwrite it to obscure the `hickory_...` junk)

bleggett:
> > We should explicitly document how to re-enable them for debugging but I notice that we don't have any info around RUST_LOG overrides in LOCAL.md yet, and we have several PRs out for that already, so that can come after.
> 
> I document it in #874 (though not specifically the obscure DNS target name. I actually wonder if we can/should overwrite it to obscure the `hickory_...` junk)

Ah I missed that, ty.

`DEBUG_DNS` et al would be nice but as long as we document the particular `hickory` incantations required it doesn't have to be pretty.

ilrudie (approved):
LGTM, agree about documenting how to turn back on but I am ok with having that in the access log PR where that section is being written

stevenctl (approved):

PR #910, opened 2024-04-09
tls: avoid setting CN in CSR

howardjohn:
This matches sidecar behavior, and Istio CA errors out on CN >64 chars

### DIFF
--- a/src/tls/csr.rs
+++ b/src/tls/csr.rs
@@ -65,12 +65,14 @@ impl CsrOptions {
 
     #[cfg(feature = "tls-ring")]
     pub fn generate(&self) -> Result<CertSign, Error> {
-        use rcgen::{CertificateParams, SanType};
+        use rcgen::{CertificateParams, DistinguishedName, SanType};
         let kp = rcgen::KeyPair::generate_for(&rcgen::PKCS_ECDSA_P256_SHA256)?;
         let private_key = kp.serialize_pem();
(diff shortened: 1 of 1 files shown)


src/tls/csr.rs
```diff
@@ -65,12 +65,14 @@ impl CsrOptions {
...
     pub fn generate(&self) -> Result<CertSign, Error> {
-        use rcgen::{CertificateParams, SanType};
+        use rcgen::{CertificateParams, DistinguishedName, SanType};
         let kp = rcgen::KeyPair::generate_for(&rcgen::PKCS_ECDSA_P256_SHA256)?;
         let private_key = kp.serialize_pem();
         let mut params = CertificateParams::default();
         params.subject_alt_names = vec![SanType::URI(self.san.clone().try_into()?)];
         params.key_identifier_method = rcgen::KeyIdMethod::Sha256;
+        // Avoid setting CN
+        params.distinguished_name = DistinguishedName::new();
```
stevenctl, line 75: It's unclear what setting an empty DistinguishedName actually does?

howardjohn: Let me add some tests and docs. but basically it defaults to `CN=rcgen self signed cert` if we don't set this. Istio uses a CN being set to mean "Put the identity in CN", and then also rejects >64 length identity

stevenctl (approved):

PR #911, opened 2024-04-09
tls: drop certificates we don't care about anymore

howardjohn:
Taken over from https://github.com/istio/ztunnel/pull/730

Fixes https://github.com/istio/ztunnel/issues/729
Fixes https://github.com/istio/ztunnel/issues/351

### DIFF
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -424,7 +424,7 @@ async fn maybe_sleep_until(till: Option<Instant>) -> bool {
     }
 }
 
-enum Request {
+pub enum Request {
     Fetch(Identity, Priority),
     Forget(Identity),
 }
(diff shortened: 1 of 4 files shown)


src/identity/manager.rs
```diff
@@ -584,6 +584,8 @@ impl SecretManager {
     }
 
     pub async fn forget_certificate(&self, id: &Identity) {
+        // TODO: consider keeping the cert around for a minute or so to avoid churn
```
stevenctl, line 587: Not sure if this is realistic: What if we churn [delete, re-add]  around the time a fetch_certificate call is made? Isn't there a (very small chance) the cert isn't there when we look it up? 

https://github.com/istio/ztunnel/blob/3f3a1af786c4fedca1fd4afab34a4a4999baa83f/src/proxy/inbound.rs#L638

1. fetch workload
2. it's cert is cleaned up
3. we fail to get the cert

A slight delay here would eliminate that rare case.

stevenctl (approved):

PR #912, opened 2024-04-10
outbound: add request builder

### LINKED ISSUE #846: finer grained request builder
Per @hzxuzhonghu [comment](https://github.com/istio/ztunnel/pull/832#discussion_r1508386863) there are a lot of struct literals in outbound build_request. We should consider implementing a finer grained function just for building requests to remove the literals.

daixiang0:
Fix #846 

Introduce typed-builder to reduce builder codes and make all struct easy to build.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3226,6 +3226,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e421abadd41a4225275504ea4d6566923418b7f05506fbc9c0fe86ba7396114b"
 
 [[package]]
+name = "typed-builder"
+version = "0.18.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "444d8748011b93cb168770e8092458cb0f8854f931ff82fdf6ddfbd72a9c933e"
+dependencies = [
+ "typed-builder-macro",
+]
+
+[[package]]
+name = "typed-builder-macro"
+version = "0.18.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "563b3b88238ec95680aef36bdece66896eaa7ce3c0f1b4f39d38fb2435261352"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.58",
+]
+
+[[package]]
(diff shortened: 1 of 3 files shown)


howardjohn (changes requested):
I think this is a regression in safety and behavior. It bloats the code and adds indirection via macro, and makes it possible to forget to set fields - less safe than today. The only real benefit is we can omit the None fields, which I view as a downside - being explicit is important here

daixiang0:
I want to use existed crate to reduce the builder codes, here is an example without any crate:

```
struct RequestBuilder {
    method: String,
    url: String,
    headers: Vec<(String, String)>,
    body: Option<String>,
}

impl RequestBuilder {
    pub fn new(method: &str, url: &str) -> Self {
        Self {
            method: method.to_string(),
            url: url.to_string(),
            headers: Vec::new(),
            body: None,
        }
    }

    pub fn add_header(mut self, key: &str, value: &str) -> Self {
        self.headers.push((key.to_string(), value.to_string()));
        self
    }

    pub fn set_body(mut self, body: &str) -> Self {
        self.body = Some(body.to_string());
        self
    }

    pub fn build(self) -> Result<Request, &'static str> {
        // Validate and construct the actual request here
        // For demonstration purposes, let's assume we have a Request struct
        // with fields method, url, headers, and body.
        // You can replace this with your actual implementation.

        // Example:
        let request = Request {
            method: self.method,
            url: self.url,
            headers: self.headers,
            body: self.body,
        };

        // Validate the request (imaginary validation function)
        if request_is_valid(&request) {
            Ok(request)
        } else {
            Err("Invalid request")
        }
    }
}

// Example usage:
let request = RequestBuilder::new("GET", "https://api.example.com")
    .add_header("Authorization", "Bearer token")
    .set_body("Hello, world!")
    .build()?;

println!("Request: {:?}", request);
```

howardjohn:
My concern is not with the crate it's with the concept in general

daixiang0:
Sure, if the issue is not planed, feel free to close all.

daixiang0:
Add @hzxuzhonghu as the issue requester.

PR #913, opened 2024-04-10
Remove hairpin

ilrudie:
- remove hairpin logic from ztunnel
- remove hairpin test


ilrudie:
hold added since this will cause istio/istio integ tests to fail

howardjohn (approved):

bleggett (approved):

Issue #926, opened 2024-04-15
Periodic reporting on TCP metrics


howardjohn:
Currently, we report "open" at the start of connection, and "close", "sent", "recieved" at the end of the connection.

For long lived connections  this gives pretty terrible results. Notably, Kiali's graph only shows recv/sent, so its blind to these

PR #927, opened 2024-04-15
Periodic metrics reporting

howardjohn:
Implements https://github.com/istio/ztunnel/issues/926

This PR adds incremental recv/sent bytes metrics recording. This is done by a fork of the tokio copy_buf implementation to increment stats on read/write.

This has two issues currently, which I will open followup issues:
* The `splice` mode is hard to do this way. We will need to fork the library we use or get support upstream. For now, this is disabled
* Currently every r/w operation increments the metrics. This might be a bit excessive. A metric update takes 300ns. A followup could be to only record every N operations (though, ideally, we would also have a way to do it every N seconds or N operations, whichever is smaller...)

For now, I think the benefit here is worth the tradeoffs.

When testing netperf with 2mb payloads and 1byte payloads:

* Before: 10.4Gb/s and 28Mb/s
* After: 10.7Gb/s and 29Mb/s
So roughly in the ballpark. (Presumably,  I did not make things faster, just noise).

Also, pprof shows <1% on metrics reporting

### DIFF
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -12,13 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use realm_io::AsyncRawIO;
 use std::io::Error;
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::ops::AddAssign;
+use std::os::fd::{AsRawFd, RawFd};
+use std::pin::Pin;
+use std::task::{Context, Poll};
 
 use tokio::io;
-use tokio::net::TcpListener;
+use tokio::io::{AsyncRead, AsyncWrite, Interest, ReadBuf};
+use tokio::net::{TcpListener, TcpStream};
 use tokio::net::TcpSocket;
+use tracing::trace;
 
+use crate::proxy::ConnectionResult;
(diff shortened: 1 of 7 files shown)


louiscryan:
Ultimately likely to be better off with thread-local counters and periodic rollup centrally but benefit would be marginal in the grander scheme. Seems fine.

src/socket.rs
```diff
@@ -166,26 +172,102 @@ mod linux {
...
-        Err(ref e) if e.raw_os_error().map_or(false, |ec| ec == EINVAL) => {
-            tokio::io::copy_bidirectional(downstream, upstream).await
-        }
-        Err(e) => Err(e),
+// TLS record size max is 16k. But we also have a H2 frame header, so leave a bit of room for that.
+const BUFFER_SIZE: usize = 16_384 - 64;
+
+pub async fn copy_bidirectional<A, B>(
+    downstream: &mut A,
+    upstream: &mut B,
```
bleggett, line 180: We should be able to drop `realm_io` as a top-level dep, this was the only reference IIRC.

bleggett (approved):
LGTM, aside from `Cargo.toml` nit.

PR #929, opened 2024-04-16
DOC: fix In-pod mode with istiod on kind not works

yanickxia:
Default debug in kind, will get `gRPC connection error:status: Unknown, message: "client error (Connect)", source: invalid peer certificate: NotValidForName`, because rusttls check dns name with subject_name

### DIFF
--- a/Development.md
+++ b/Development.md
@@ -147,6 +147,11 @@ spec:
           path: istio-token
 EOF
 ```
+add `istiod.istio-system.svc` to `hosts` file, make [`verify_is_valid_for_subject_name`](https://crates.io/crates/rustls-webpki) works
+
+```shell
+echo "127.0.0.1 istiod.istio-system.svc" >> /etc/hosts
+```
 
 Now, port forward istiod, copy over the token and run ztunnel (under sudo):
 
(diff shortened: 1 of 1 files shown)


Development.md
```diff
@@ -147,6 +147,11 @@ spec:
           path: istio-token
 EOF
 ```
+add `istiod.istio-system.svc` to `hosts` file, make [`verify_is_valid_for_subject_name`](https://crates.io/crates/rustls-webpki) works
```
dhawton, line 150:
```suggestion

Add `istiod.istio-system.svc` to the system's `hosts` file so that [`verify_is_valid_for_subject_name`](https://crates.io/crates/rustls-webpki) passes:
```

yanickxia: miss markdown lint, careful next time

howardjohn (commented):
You can just port-forward (as the doc already says); localhost should already accepted

Development.md
```diff
@@ -148,6 +148,12 @@ spec:
 EOF
 ```
 
+Add `istiod.istio-system.svc` to the system's `hosts` file so that [`verify_is_valid_for_subject_name`](https://crates.io/crates/rustls-webpki) passes:
+
+```shell
+echo "127.0.0.1 istiod.istio-system.svc" >> /etc/hosts
```
bleggett, line 154:
```suggestion
echo "127.0.0.1 istiod.istio-system.svc" >> /etc/hosts
```

We shouldn't advise people to muck with their local `hosts` file - we shouldn't have any tests that depend on this.

yanickxia: you right.. we maybe add some flag to skip dns subject check.

yanickxia:
Seem Add some way to close dns subject check

yanickxia:
> You can just port-forward (as the doc already says); localhost should already accepted

from this mr https://github.com/istio/ztunnel/pull/820 merge, seem debug in local is break.

```bash
2024-04-17T03:12:20.744145Z     warn    xds::client:xds{id=9}   XDS client connection error: gRPC connection error:status: Unknown, message: "client error (Connect)", source: invalid peer certificate: NotValidForName, retrying in 5.12s      
```

seem
https://github.com/rustls/rustls/blob/d2e1e668aae3d9807ffa782ea29ed26f2778371d/rustls/src/webpki/server_verifier.rs#L256

PR #931, opened 2024-04-16
Better-behaved HBONE pooling

bleggett:
This works around several issues found in https://github.com/istio/ztunnel/issues/864 relating to clients (loadgen mostly) that open many connections very rapidly.

1. `hyper`'s pooling library has rather basic scheduling controls, to the point where we can rather easily overwhelm "send_request" such that it never resolves. Tweaking server/client streamlimits does not work around this, as the pooling scheduler does not respect them when selecting pooled streams, and will simply allow a single conn to get backed up.

2. `hyper`'s pooling library doesn't really have the concept of multiple HTTP/2 connections to the same dest, which makes sense as the HTTP/2 spec doesn't recommend this (the spec says SHOULD NOT instead of MUST NOT, which in spec-speak is "we can't stop you, I guess")

3. Racing to create connections creates problems in scenarios where a lot of connection attempts are made at once due to lack of throttling - many connections can actually be established or "cross the finish line at the same time", when we only need one, and this noise can compromise the ability of the local or dest ztunnel to respond to other workloads.

Notes:

- By design, this pool does not manage Hyper client state, nor is it tightly coupled with it. All it does is hold references to Hyper clients that can be reused, and evict its own held references to those Hyper clients when they cannot be reused. The rest is managing async access.

- By design, this pool is designed to create backpressure when one client opens many connections to the same destination at the same time. It does this by locking connection-mutative operations per src/dest key.

- The only write lock should be on that inner src/dest key. The only operations inside the pool that must be ordered/force ordering are 1. Ensuring that we have a key writelock in the outer map. 2. Locking the inner key writelock. The rest should not demand any sort of ordering to be safe. 

- This pool does not decrement stream counts - when a connection hits its max stream count, the pool pops the reference and removes it from the rotation, leaving the connection to die when whatever is using it drops all the refs, and will create a new connection on the next go round. This leads to simpler accounting, while still maintaining the protections against connection storms that we need.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -65,7 +65,8 @@ const DEFAULT_SELFTERM_DEADLINE: Duration = Duration::from_secs(5);
 const DEFAULT_CLUSTER_ID: &str = "Kubernetes";
 const DEFAULT_CLUSTER_DOMAIN: &str = "cluster.local";
 const DEFAULT_TTL: Duration = Duration::from_secs(60 * 60 * 24); // 24 hours
-const DEFAULT_POOL_RELEASE: Duration = Duration::from_secs(60 * 5); // 5 minutes
+const DEFAULT_POOL_UNUSED_RELEASE_TIMEOUT: Duration = Duration::from_secs(60 * 5); // 5 minutes
+const DEFAULT_POOL_MAX_STREAMS_PER_CONNECTION: u16 = 100; //Go: 100, Hyper: 200, Envoy: 2147483647 (lol), Spec recommended minimum 100
 
 const DEFAULT_INPOD_MARK: u32 = 1337;
 
@@ -134,8 +135,8 @@ pub struct Config {
     // This can be used to effect flow control for "connection storms" when workload clients
     // (such as loadgen clients) open many connections all at once.
     //
-    // Note that this will only be checked and inner conns rebalanced accordingly when a new connection
-    // is requested from the pool, and not on every stream queue on that connection.
+    // Note that this will only be checked when a *new* connection
+    // is requested from the pool, and not on every *stream* queued on that connection.
     // So if you request a single connection from a pool configured wiht a max streamcount of 200,
     // and queue 500 streams on it, you will still exceed this limit and are at the mercy of hyper's
     // default stream queuing.
@@ -339,11 +340,14 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
             .get(DNS_CAPTURE_METADATA)
             .map_or(false, |value| value.to_lowercase() == "true"),
 
-        pool_max_streams_per_conn: parse_default(POOL_MAX_STREAMS_PER_CONNECTION, 250)?,
+        pool_max_streams_per_conn: parse_default(
+            POOL_MAX_STREAMS_PER_CONNECTION,
+            DEFAULT_POOL_MAX_STREAMS_PER_CONNECTION,
+        )?,
 
         pool_unused_release_timeout: match parse::<String>(POOL_UNUSED_RELEASE_TIMEOUT)? {
-            Some(ttl) => duration_str::parse(ttl).unwrap_or(DEFAULT_POOL_RELEASE),
-            None => DEFAULT_POOL_RELEASE,
+            Some(ttl) => duration_str::parse(ttl).unwrap_or(DEFAULT_POOL_UNUSED_RELEASE_TIMEOUT),
+            None => DEFAULT_POOL_UNUSED_RELEASE_TIMEOUT,
         },
 
         window_size: 4 * 1024 * 1024,
--- a/src/config.rs
+++ b/src/config.rs
@@ -371,7 +371,7 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
             DEFAULT_READINESS_PORT, // There is no config for this in ProxyConfig currently
         ),
 
-        socks5_addr: SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 15080),
+        socks5_addr,
         inbound_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15008),
         inbound_plaintext_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15006),
         outbound_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15001),
--- a/src/config.rs
+++ b/src/config.rs
@@ -50,6 +50,7 @@ const SECRET_TTL: &str = "SECRET_TTL";
 const FAKE_CA: &str = "FAKE_CA";
 const ZTUNNEL_WORKER_THREADS: &str = "ZTUNNEL_WORKER_THREADS";
 const POOL_MAX_STREAMS_PER_CONNECTION: &str = "POOL_MAX_STREAMS_PER_CONNECTION";
+const POOL_UNUSED_RELEASE_TIMEOUT: &str = "POOL_UNUSED_RELEASE_TIMEOUT";
 const ENABLE_ORIG_SRC: &str = "ENABLE_ORIG_SRC";
 const PROXY_CONFIG: &str = "PROXY_CONFIG";
 
@@ -64,6 +65,7 @@ const DEFAULT_SELFTERM_DEADLINE: Duration = Duration::from_secs(5);
 const DEFAULT_CLUSTER_ID: &str = "Kubernetes";
 const DEFAULT_CLUSTER_DOMAIN: &str = "cluster.local";
 const DEFAULT_TTL: Duration = Duration::from_secs(60 * 60 * 24); // 24 hours
+const DEFAULT_POOL_RELEASE: Duration = Duration::from_secs(60 * 5); // 5 minutes
 
 const DEFAULT_INPOD_MARK: u32 = 1337;
 
@@ -138,6 +140,8 @@ pub struct Config {
     // and queue 500 streams on it, you will still exceed this limit and are at the mercy of hyper's
     // default stream queuing.
     pub pool_max_streams_per_conn: u16,
+
+    pub pool_unused_release_timeout: Duration,
 
     pub socks5_addr: Option<SocketAddr>,
     pub admin_addr: SocketAddr,
@@ -336,6 +340,11 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
             .map_or(false, |value| value.to_lowercase() == "true"),
 
         pool_max_streams_per_conn: parse_default(POOL_MAX_STREAMS_PER_CONNECTION, 250)?,
+
+        pool_unused_release_timeout: match parse::<String>(POOL_UNUSED_RELEASE_TIMEOUT)? {
+            Some(ttl) => duration_str::parse(ttl).unwrap_or(DEFAULT_POOL_RELEASE),
+            None => DEFAULT_POOL_RELEASE,
+        },
 
         window_size: 4 * 1024 * 1024,
         connection_window_size: 4 * 1024 * 1024,
--- a/src/config.rs
+++ b/src/config.rs
@@ -49,6 +49,7 @@ const CA_ADDRESS: &str = "CA_ADDRESS";
 const SECRET_TTL: &str = "SECRET_TTL";
 const FAKE_CA: &str = "FAKE_CA";
 const ZTUNNEL_WORKER_THREADS: &str = "ZTUNNEL_WORKER_THREADS";
+const POOL_MAX_STREAMS_PER_CONNECTION: &str = "POOL_MAX_STREAMS_PER_CONNECTION";
 const ENABLE_ORIG_SRC: &str = "ENABLE_ORIG_SRC";
 const PROXY_CONFIG: &str = "PROXY_CONFIG";
 
@@ -124,6 +125,19 @@ pub struct Config {
     pub window_size: u32,
     pub connection_window_size: u32,
     pub frame_size: u32,
+
+    // The limit of how many streams a single HBONE pool connection will be limited to, before
+    // spawning a new conn rather than reusing an existing one, even to a dest that already has an open connection.
+    //
+    // This can be used to effect flow control for "connection storms" when workload clients
+    // (such as loadgen clients) open many connections all at once.
+    //
+    // Note that this will only be checked and inner conns rebalanced accordingly when a new connection
+    // is requested from the pool, and not on every stream queue on that connection.
+    // So if you request a single connection from a pool configured wiht a max streamcount of 200,
+    // and queue 500 streams on it, you will still exceed this limit and are at the mercy of hyper's
+    // default stream queuing.
+    pub pool_max_streams_per_conn: u16,
 
     pub socks5_addr: Option<SocketAddr>,
     pub admin_addr: SocketAddr,
@@ -321,6 +335,8 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
             .get(DNS_CAPTURE_METADATA)
             .map_or(false, |value| value.to_lowercase() == "true"),
 
+        pool_max_streams_per_conn: parse_default(POOL_MAX_STREAMS_PER_CONNECTION, 250)?,
+
         window_size: 4 * 1024 * 1024,
         connection_window_size: 4 * 1024 * 1024,
         frame_size: 1024 * 1024,
@@ -342,7 +358,7 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
             DEFAULT_READINESS_PORT, // There is no config for this in ProxyConfig currently
         ),
 
-        socks5_addr,
+        socks5_addr: SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 15080),
         inbound_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15008),
         inbound_plaintext_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15006),
         outbound_addr: SocketAddr::new(IpAddr::V6(Ipv6Addr::UNSPECIFIED), 15001),
--- a/src/proxy/pool.rs
+++ b/src/proxy/pool.rs
@@ -1051,7 +1051,7 @@ mod test {
         let cfg = crate::config::Config {
             local_node: Some("local-node".to_string()),
             pool_max_streams_per_conn: 1000,
-            pool_unused_release_timeout: Duration::from_secs(5),
+            pool_unused_release_timeout: Duration::from_secs(2),
             ..crate::config::parse_config().unwrap()
         };
         let sock_fact = Arc::new(crate::proxy::DefaultSocketFactory);
@@ -1104,7 +1104,7 @@ mod test {
         );
 
         // Attempt to wait long enough for pool conns to timeout+drop
-        sleep(Duration::from_secs(6)).await;
+        sleep(Duration::from_secs(2)).await;
 
         let real_conncount = conn_counter.load(Ordering::Relaxed);
         let real_dropcount = conn_drop_counter.load(Ordering::Relaxed);
@@ -1134,6 +1134,7 @@ mod test {
         let cfg = crate::config::Config {
             local_node: Some("local-node".to_string()),
             pool_max_streams_per_conn: 50,
+            pool_unused_release_timeout: Duration::from_secs(2),
             ..crate::config::parse_config().unwrap()
         };
         let sock_fact = Arc::new(crate::proxy::DefaultSocketFactory);
@@ -1161,7 +1162,7 @@ mod test {
         let (client_stop_signal, client_stop) = drain::channel();
         let persist_res = spawn_persistent_client(pool.clone(), key1.clone(), server_addr, client_stop);
 
-        //loop thru the nonpersistent results
+        //loop thru the nonpersistent clients and wait for them to finish
         while let Some(Err(res)) = tasks.next().await {
             assert!(!res.is_panic(), "CLIENT PANICKED!");
             continue;
@@ -1179,12 +1180,12 @@ mod test {
         );
 
 
-        // Attempt to wait long enough for pool conns to timeout+drop
-        sleep(Duration::from_secs(6)).await;
+        // Attempt to wait long enough for pool conns to timeout+evict
+        sleep(Duration::from_secs(2)).await;
 
         let real_conncount = conn_counter.load(Ordering::Relaxed);
         assert!(real_conncount == 3, "actual conncount was {real_conncount}");
-        // At this point, we should still have one conn that hasn't been dropped (even though the pool has)
+        // At this point, we should still have one conn that hasn't been dropped
         // because we haven't ended the persistent client
         let real_dropcount = conn_drop_counter.load(Ordering::Relaxed);
         assert!(
@@ -1194,7 +1195,7 @@ mod test {
         client_stop_signal.drain().await;
         assert!(!persist_res.await.is_err(), "PERSIST CLIENT ERROR");
 
-        sleep(Duration::from_secs(6)).await;
+        sleep(Duration::from_secs(2)).await;
 
         let after_conncount = conn_counter.load(Ordering::Relaxed);
         assert!(after_conncount == 3, "after conncount was {after_conncount}");
@@ -1293,12 +1294,12 @@ mod test {
                 loop {
                     count += 1;
                     let res = c1.send_request(req()).await;
-
                     if res.is_err() {
                         panic!("SEND ERR: {:#?} sendcount {count}", res);
                     } else if res.unwrap().status() != 200 {
                         panic!("CLIENT RETURNED ERROR")
                     }
+
                 }
             };
 
--- a/src/proxy/pool.rs
+++ b/src/proxy/pool.rs
@@ -101,15 +101,19 @@ impl ConnSpawner {
             .then_some(key.src);
         let cert = self.cert_manager.fetch_certificate(&key.src_id).await?;
         let connector = cert.outbound_connector(key.dst_id)?;
+        debug!("got connector for key");
         let tcp_stream =
             super::freebind_connect(local, key.dst, self.socket_factory.as_ref()).await?;
+        debug!("got stream for key");
         tcp_stream.set_nodelay(true)?; // TODO: this is backwards of expectations
         let tls_stream = connector.connect(tcp_stream).await?;
+        debug!("connector connected, handshaking");
         let (request_sender, connection) = builder
             .handshake(::hyper_util::rt::TokioIo::new(tls_stream))
             .await
             .map_err(Error::HttpHandshake)?;
 
+        debug!("got TLS stream for key");
         // spawn a task to poll the connection and drive the HTTP state
         // if we got a drain for that connection, respect it in a race
         // it is important to have a drain here, or this connection will never terminate
@@ -216,7 +220,7 @@ impl PoolState {
         &self,
         workload_key: &WorkloadKey,
         pool_key: &pingora_pool::ConnectionMeta,
-    ) -> Option<ConnClient> {
+    ) -> Result<Option<ConnClient>, Error> {
         let inner_conn_lock = {
             trace!("getting keyed lock out of lockmap");
             let guard = self.established_conn_writelock.guard();
@@ -250,11 +254,11 @@ impl PoolState {
                             workload_key, pool_key
                         );
                         self.checkin_conn(client.clone(), pool_key.clone());
-                        Some(client)
+                        Ok(Some(client))
                     }
                     Err(e) => {
                         error!("could not spawn new conn, got {e}");
-                        None
+                        return Err(e)
                     }
                 }
                 // END take inner writelock
@@ -264,7 +268,7 @@ impl PoolState {
                     "did not win connlock for wl key {:#?}, something else has it",
                     workload_key
                 );
-                None
+                Ok(None)
             }
         }
     }
@@ -489,7 +493,7 @@ impl WorkloadHBONEPool {
         let res = match self
             .state
             .start_conn_if_win_writelock(&workload_key, &pool_key)
-            .await
+            .await?
         {
             Some(client) => client,
             None => {
@@ -639,6 +643,7 @@ mod test {
     use tokio::net::TcpListener;
     use tokio::task::{self};
     use tokio::time::sleep;
+    use http_body_util::BodyExt;
 
     #[cfg(tokio_unstable)]
     use tracing::Instrument;
@@ -886,8 +891,8 @@ mod test {
         server_handle.await.unwrap();
         drop(pool);
 
-        let real_conncount = conn_counter.load(Ordering::Relaxed);
-        assert!(real_conncount == 4, "actual conncount was {real_conncount}");
+        let real_conncount = conn_counter.load(Ordering::SeqCst);
+        assert!(real_conncount == 2, "actual conncount was {real_conncount}");
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
@@ -1274,7 +1279,7 @@ mod test {
                 // needs tokio_unstable, but useful
                 // .instrument(tracing::debug_span!("client_tid", tid=%tokio::task::id()))
                 .await
-                .unwrap();
+                .expect("connect should succeed");
             debug!(
                 "client spent {}ms waiting for conn",
                 start.elapsed().as_millis()
@@ -1287,10 +1292,18 @@ mod test {
 
                 if res.is_err() {
                     panic!("SEND ERR: {:#?} sendcount {count}", res);
-                } else if res.unwrap().status() != 200 {
+                } else if !res.is_ok() {
                     panic!("CLIENT RETURNED ERROR")
                 }
 
+                let mut okres = res.unwrap();
+                const HBONE_MESSAGE: &[u8] = b"hbone\n";
+                while let Some(next) = okres.frame().await {
+                    let frame = next.expect("better have a resp body");
+                    if let Some(chunk) = frame.data_ref() {
+                        assert_eq!(HBONE_MESSAGE, chunk);
+                    }
+                }
                 if count >= req_count {
                     debug!("CLIENT DONE");
                     break;
@@ -1366,6 +1379,7 @@ mod test {
                     Ok(upgraded) => {
                         let (mut ri, mut wi) =
                             tokio::io::split(hyper_util::rt::TokioIo::new(upgraded));
+                        // wi.write_all(b"hbone\n").await.unwrap();
                         wi.write_all(b"hbone\n").await.unwrap();
                         tcp::handle_stream(tcp::Mode::ReadWrite, &mut ri, &mut wi).await;
                     }
--- a/src/proxy/pool.rs
+++ b/src/proxy/pool.rs
@@ -221,6 +221,7 @@ impl PoolState {
     //
     // This function should ALWAYS return a connection if it wins the writelock for the provided key.
     // This function should NEVER return a connection if it does not win the writelock for the provided key.
+    // This function should ALWAYS propagate Error results to the caller
     //
     // It is important that the *initial* check here is authoritative, hence the locks, as
     // we must know if this is a connection for a key *nobody* has tried to start yet
@@ -288,6 +289,7 @@ impl PoolState {
     //
     // This function should ALWAYS return a connection if a writelock exists for the provided key.
     // This function should NEVER return a connection if no writelock exists for the provided key.
+    // This function should ALWAYS propagate Error results to the caller
     //
     // It is important that the *initial* check here is authoritative, hence the locks, as
     // we must know if this is a connection for a key *nobody* has tried to start yet
--- a/src/proxy/pool.rs
+++ b/src/proxy/pool.rs
@@ -526,7 +526,7 @@ impl WorkloadHBONEPool {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 // A sort of faux-client, that represents a single checked-out 'request sender' which might
 // send requests over some underlying stream using some underlying http/2 client
 pub struct ConnClient {
@@ -576,6 +576,24 @@ impl Drop for ConnClient {
             self.stream_count,
             self.stream_count_max
         )
+    }
+}
+
+// This is currently only for debugging
+impl Clone for ConnClient {
+    fn clone(&self) -> Self {
+        trace!(
+            "cloning ConnClient for key {:#?} with streamcount: {:?} / {:?}",
+            self.wl_key,
+            self.stream_count,
+            self.stream_count_max
+        );
+        ConnClient {
+            sender: self.sender.clone(),
+            stream_count: self.stream_count.clone(),
+            stream_count_max: self.stream_count_max,
+            wl_key: self.wl_key.clone(),
+        }
     }
 }
 
--- a/src/proxy/pool.rs
+++ b/src/proxy/pool.rs
@@ -211,6 +211,7 @@ impl PoolState {
         }
     }
 
+
     async fn start_conn_if_win_writelock(
         &self,
         workload_key: &WorkloadKey,
@@ -324,7 +325,18 @@ impl PoolState {
                             Some(e_conn)
                         }
                     }
-                    None => None,
+                    None => {
+                        trace!("checkout - no existing conn for key {:#?}, adding one", workload_key);
+                        let pool_conn = self.spawner.new_pool_conn(workload_key.clone()).await;
+                        let r_conn = ConnClient {
+                            sender: pool_conn?,
+                            stream_count: Arc::new(AtomicU16::new(0)),
+                            stream_count_max: self.max_streamcount,
+                            wl_key: workload_key.clone(),
+                        };
+                        self.checkin_conn(r_conn.clone(), pool_key.clone());
+                        Some(r_conn)
+                    },
                 };
 
                 Ok(result)
@@ -427,9 +439,7 @@ impl WorkloadHBONEPool {
             return Ok(existing_conn.unwrap());
         }
 
-        // We couldn't get a conn. This means either nobody has tried to establish any conns for this key yet,
-        // or they have, but no conns are currently available
-        // (because someone else has checked all of them out and not put any back yet)
+        // We couldn't get a conn. This means nobody has tried to establish any conns for this key yet,
         //
         // So, we will take a nonexclusive readlock on the lockmap, to see if an inner lock
         // exists for our key.
@@ -828,7 +838,7 @@ mod test {
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn test_pool_100_clients_streamexhaust() {
-        crate::telemetry::setup_logging();
+        // crate::telemetry::setup_logging();
 
         let (server_drain_signal, server_drain) = drain::channel();
 
--- a/src/proxy/pool.rs
+++ b/src/proxy/pool.rs
@@ -594,6 +594,7 @@ mod test {
     use hyper::{Request, Response};
     use std::sync::atomic::AtomicU32;
     use std::time::Duration;
+    use tokio::time::sleep;
     use tokio::io::AsyncWriteExt;
     use tokio::net::TcpListener;
     use tokio::task::{self};
@@ -611,7 +612,9 @@ mod test {
 
         let (server_drain_signal, server_drain) = drain::channel();
 
-        let (server_addr, server_handle) = spawn_server(server_drain).await;
+        let conn_counter: Arc<AtomicU32> = Arc::new(AtomicU32::new(0));
+        let conn_drop_counter: Arc<AtomicU32> = Arc::new(AtomicU32::new(0));
+        let (server_addr, server_handle) = spawn_server(server_drain, conn_counter.clone(), conn_drop_counter.clone()).await;
 
         let cfg = crate::config::Config {
             local_node: Some("local-node".to_string()),
@@ -639,7 +642,8 @@ mod test {
 
         server_drain_signal.drain().await;
         drop(pool);
-        let real_conncount = server_handle.await.unwrap();
+        server_handle.await.unwrap();
+        let real_conncount = conn_counter.load(Ordering::Relaxed);
         assert!(real_conncount == 1, "actual conncount was {real_conncount}");
 
         assert!(client1.is_ok());
@@ -650,7 +654,10 @@ mod test {
     #[tokio::test]
     async fn test_pool_does_not_reuse_conn_for_diff_key() {
         let (server_drain_signal, server_drain) = drain::channel();
-        let (server_addr, server_handle) = spawn_server(server_drain).await;
+
+        let conn_counter: Arc<AtomicU32> = Arc::new(AtomicU32::new(0));
+        let conn_drop_counter: Arc<AtomicU32> = Arc::new(AtomicU32::new(0));
+        let (server_addr, server_handle) = spawn_server(server_drain, conn_counter.clone(), conn_drop_counter.clone()).await;
 
         // crate::telemetry::setup_logging();
 
@@ -681,7 +688,10 @@ mod test {
 
         server_drain_signal.drain().await;
         drop(pool);
-        let real_conncount = server_handle.await.unwrap();
+
+        server_handle.await.unwrap();
+
+        let real_conncount = conn_counter.load(Ordering::Relaxed);
         assert!(real_conncount == 2, "actual conncount was {real_conncount}");
 
         assert!(client1.is_ok());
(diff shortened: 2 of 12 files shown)


src/proxy/pool.rs
```diff
@@ -12,80 +12,398 @@
...
+            // HOWEVER. Here we know this connection will either establish or timeout
+            // and we WANT other tasks to go back to sleep if there is an outstanding lock.
+            // So the downsides are actually useful (we WANT task contention -
+            // to block other parallel tasks from trying to spawn a connection if we are already doing so)
+            //
+            // BEGIN take outer readlock
+            let map_read_lock = self.established_conn_writelock.read().await;
+            let exist_conn_lock = map_read_lock.get(&hash_key).unwrap();
+            // BEGIN take inner writelock
+            let found_conn = match exist_conn_lock.as_ref().unwrap().try_lock() {
```
howardjohn, line 186: The locking seems problematic. Talking about the outer lock here.

1. We open connection with readlock. Connection can take ~20s
2. New connection starts with different key. It tries to writelock. This blocks until all readers  (1) are done.
3. New connection starts for an existing key. This tries to readlock. This blocks until writer (2) are done.

End result: all new connections blocked on one connection
?

bleggett: Yes, this will block on net-new connections. It should not on reusing established connections.

If a workload opens many connections to many different endpoints at once, it will effectively be throttled by this.

The only alternative to a pool-wide state I can think of is race-discard, which throttles everybody, including neighbors - this just throttles the workload.

howardjohn: In the scenario about even entirely different pool keys can block.

A -> B can block A -> C. B and C could be different nodes entirely as well

src/proxy/pool.rs
```diff
@@ -12,80 +12,398 @@
...
-#[derive(Debug, Clone)]
-struct Client(http2::SendRequest<Empty<Bytes>>);
+                debug!("getting conn for key {:#?} and hash {:#?}", key, hash_key);
+                self.connected_pool.get(&hash_key).and_then(|e_conn| {
+                    debug!("got existing conn for key {:#?}", key);
+                    if e_conn.at_max_streamcount() {
+                        debug!("got conn for key {:#?}, but streamcount is maxed", key);
+                        None
+                    } else {
+                        self.connected_pool.put(pool_key, e_conn.clone());
```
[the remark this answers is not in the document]
bleggett: `get` actually removes the conn from the pool - it is not a get-by-ref.

We take it out, and (if it's still good) clone it and put it back in.

src/proxy/pool.rs
howardjohn: Taking a step back, here is what I see as the requirements

There should be a Pool, per proxy instance (pod, for inpod mode)

It should expose, something like
`  fn send_request(key, Request) -> Response`

With the following behavior:
If there is an existing connection open with available streams for the key, send the request
   Because hyper doesn't expose active streams and max streams, this will need to
   increment a per-connection "active streams" counter at the start, and make a Drop impl decrement.
Else if there is an existing connection pending for the key, and the number of pending+active requests is < max streams, wait for it to complete 
Else, establish a new connection and put it in the pool

After the last request on a connection is finished, it is marked idle. Connections idle for more than ~90s should be closed

State of connections for other keys than the one requested should not impact

Not strictly required, but a good design sign: No IO should be done with locks held

---

Hyper's pool falls short in two ways here:
* We don't have a way to not re-use an existing steam because `reserve()` is infallible. This could probably be fixed with a `try_reserve()` extension (upstream)
* We don't have a way to say not to re-use a connect because its going to be full when we get it

howardjohn: I am not sure I follow why we need to chose between "open excessive connections" and "unrelated keys can block each other"?

Do you mean even in abstract these are in contention, or its challenging to code it that way?

I think it is conceptually possible, though all of the pooling logic is tricky (this, pingora, hyper-util -- the general concept, really)

bleggett: A third way it falls short:

- We don't have a natural way to manage/track more than one active connection in a hyper pool to the same dest.

> If there is an existing connection open with available streams for the key, send the request

> Because hyper doesn't expose active streams and max streams, this will need to
> increment a per-connection "active streams" counter at the start, and make a Drop impl decrement.

If we remove/deschedule full streams from the pool, the decrement is probably not required, but yes.

> Else if there is an existing connection pending for the key, and the number of pending+active requests is < max streams, wait for it to complete

Yes, the key here being that we need to synchronize the spawning of a new connection, such that at-most-one task "wins" the right to spawn - especially in scenarios where 100s of tasks might spin  up trying to do this.

> Else, establish a new connection and put it in the pool

Yep.

> After the last request on a connection is finished, it is marked idle. Connections idle for more than ~90s should be closed

Yep, that's a todo here. This doesn't strictly have to be managed by the pool though, it can just be client/server timeouts. 

> State of connections for other keys than the one requested should not impact
> Not strictly required, but a good design sign: No IO should be done with locks held

For the outer locking, a more advanced 3rd party concurrent hashmap impl makes sense here to minimize that outer lock.

howardjohn: I think we may have some terminology mismatch here, specifically around "full stream" .Here's my understanding, lmk if you disagree with anything and the names

A (outer) connection is a TCP/TLS connection 
On a connection we have HTTP 2 streams
1 stream and is 1 req+response pair

Each inner connection is 1:1 with a request (and thus 1:1 with a stream)

A typical pool will return a connection for a key. 
For our use cases, pool users are guaranteed to only send a single request on each usage - so we can specialize here, if desired, and expose an interface to send a single request/stream instead of connection level access

A connection is full of it has active streams equal to max streams 

a connection is idle if it has 0 streams on it 

---

If the pool doesn't close idle connection who else would? In theory the server could but this is racy (client should always have lower timeout to avoid sending on a closed connection) and insecure (we cannot trust someone else to manage our memory usage)

bleggett: > I think we may have some terminology mismatch here, specifically around "full stream" .Here's my understanding, lmk if you disagree with anything and the names

No, you're correct - I meant to type "full conn" and not "full stream".

>  For our use cases, pool users are guaranteed to only send a single request on each usage - so we can specialize here, if desired, and expose an interface to send a single request/stream instead of connection level access

Yeah - I considered that (use channels to send reqs and just expose that), but did not go down that road because it doesn't really solve any of the contention problems AFAICT. The pool still has to do pretty much the same locking/backpressure.

> If the pool doesn't close idle connection who else would? In theory the server could but this is racy (client should always have lower timeout to avoid sending on a closed connection) and insecure (we cannot trust someone else to manage our memory usage)

Probably several spots. Our server (in inbound.rs) should try to close idle connections. IIRC the `hyper` client should have its own idle timer, and we can just treat a closed client like a stream-maxed client, in terms of the pool - throw it out and start a new one.

howardjohn: If you only remove when you try to use it then you will end up with zombie connections if they are not used anymore. The server closing it shouldn't be relied on (untrusted)

---

for the "1 request per checkout" I was thinking it might help track the active streams.

bleggett: > > If you only remove when you try to use it then you will end up with zombie connections if they are not used anymore. The server closing it shouldn't be relied on (untrusted)
> 
> I'm assuming both client and server can each independently configure close timeouts on their respective ends (and probably should).

Uh, I am not sure what happened to this comment, I was on mobile. I will revert it.

bleggett: > If you only remove when you try to use it then you will end up with zombie connections if they are not used anymore. The server closing it shouldn't be relied on (untrusted)

Yes, the client can't rely on the server doing it, and the server can't rely on the client doing it. So they both have to do it.

Clientside pool should ensure it drops references to clients that are closed, agreed.

ilrudie: > Probably several spots, to be correct. Our server (in inbound.rs) should try to close idle connections. IIRC the hyper client should have its own idle timer, and we can just treat a closed client like a stream-maxed client, in terms of the pool - throw it out and start a new one, if we pull it out of the bag.

I agree with John that server-side close is likely to cause issues because it'll be racy. We should try to close on the client side if at all possible. Server-side closes should be last resort to manage resources where it's likely the client vanished without closing for some reason.

bleggett: > > Probably several spots, to be correct. Our server (in inbound.rs) should try to close idle connections. IIRC the hyper client should have its own idle timer, and we can just treat a closed client like a stream-maxed client, in terms of the pool - throw it out and start a new one, if we pull it out of the bag.
> 
> I agree with John that server-side close is likely to cause issues because it'll be racy. We should try to close on the client side if at all possible. Server-side closes should be last resort to manage resources where it's likely the client vanished without closing for some reason.

Agree - just saying we have to do server-side timeouts too, since that "last resort where the client vanishes without closing" is pretty easy to induce.

src/proxy/pool.rs
```diff
@@ -12,80 +12,398 @@
...
+    // one connection is created, before deciding if they should create more or just use that one.
+    pub async fn connect(&mut self, key: WorkloadKey) -> Result<Client, Error> {
+        // TODO BML this may not be collision resistant/slow. It should be resistant enough for workloads tho.
+        let mut s = DefaultHasher::new();
+        key.hash(&mut s);
+        let hash_key = s.finish();
+        let pool_key = pingora_pool::ConnectionMeta::new(
+            hash_key,
+            GLOBAL_CONN_COUNT.fetch_add(1, Ordering::Relaxed),
+        );
```
ilrudie, line 112: According to ConnectionMeta documentation every connection that shares this hash_key is considered the same for the purposes of connection reuse. This suggests that hash collision could be pretty problematic, potentially even a security vulnerability, yeah? I'm not sure we can TODO something like that. It seems like we need an implementation which is better than just collision resistant sooner rather than later.

src/proxy/pool.rs
ilrudie: ah, I misunderstood. yeah, we have to have some server-side time out but we need to be careful that it is strictly longer than all client timeouts.

ilrudie (commented):
I might be missing something but it seems like correctly resolving hash collisions is critical for the pool's functionality and not yet implemented or tested.

src/proxy/pool.rs
```diff
@@ -12,80 +12,398 @@
...
-                },
-                TokioExec,
-                Some(TokioTimer::new()),
-            ),
+            pool_notifier: Arc::new(tx),
+            pool_watcher: rx,
+            max_streamcount: cfg.pool_max_streams_per_conn,
+            // the number here is simply the number of unique src/dest keys
+            // the pool is expected to track before the inner hashmap resizes.
+            connected_pool: Arc::new(pingora_pool::ConnectionPool::new(50000)),
```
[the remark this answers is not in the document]
bleggett: Yes, this is arbitrary, it's just to avoid resizing in the inner pool. I agree this is too high.

src/proxy/pool.rs
bleggett: It's a TODO for me - hash functions grow on trees, so I'm not very worried about dropping in a better one.

bleggett: I'm going to look at making Pingora's pool take an `impl Hash, Eq` but for now I decided to stick with this hasher and just bind ConnClients to workload keys.

Then do a final check before returning to prove that the conn we're giving you belongs to the key you gave us.

This should be strictly safe, and I think the odds of collision are very low for us here, the workload would likely crash the box long before it would open enough conns to create a kind of collision.

src/proxy/pool.rs
```diff
@@ -12,80 +12,398 @@
...
+            // hold it over does not resolve.
+            //
+            // HOWEVER. Here we know this connection will either establish or timeout
+            // and we WANT other tasks to go back to sleep if there is an outstanding lock.
+            // So the downsides are actually useful (we WANT task contention -
+            // to block other parallel tasks from trying to spawn a connection if we are already doing so)
+            //
+            // BEGIN take outer readlock
+            let map_read_lock = self.established_conn_writelock.read().await;
+            let exist_conn_lock = map_read_lock.get(&hash_key).unwrap();
```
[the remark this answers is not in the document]
bleggett: Sorry, if you mean "are we guaranteed this map isn't cleaned up/entries purged" - yes, we manage it, and it is not currently truncated.

bleggett:
Added a form of conn timeout to evict connections from the pool that haven't been used in a while.

Once evicted, they will be closed by hyper when clients close, due to `streamcount || refcount` being == 0

bleggett:
Switched to a proper concurrent hashmap that actually has a decent pedigree, which should allow for nonblocking reads (at the cost of using a little more memory, TANSTAAFL)

src/proxy/pool.rs
```diff
@@ -12,80 +12,539 @@
...
+            let guard = self.state.established_conn_writelock.guard();
+
+            let exist_conn_lock = self
+                .state
+                .established_conn_writelock
+                .get(&hash_key, &guard)
+                .unwrap();
+            trace!("fallback attempt - got keyed lock out of lockmap");
+            exist_conn_lock.as_ref().unwrap().clone()
+        };
```
ilrudie, line 322: If the mutex contained a canonical WorkloadKey that the guard is for we could detect here that a collision had occurred, which I think would be safer than allowing any pool manipulation under unchecked collision.

src/proxy/pool.rs
```diff
@@ -12,80 +12,539 @@
...
+                let pool_conn = self.spawn_new_pool_conn(workload_key.clone()).await;
+                let r_conn = ConnClient {
+                    sender: pool_conn?,
+                    stream_count: Arc::new(AtomicU16::new(0)),
+                    stream_count_max: self.max_streamcount,
+                    wl_key: workload_key.clone(),
+                };
+                self.state.checkin_conn(r_conn.clone(), pool_key.clone());
+                Ok(r_conn)
+            }
```
ilrudie, line 435: If we are under hash collision AND we happened to be at max streams we can now potentially pollute the pool with a connection to a different dst I think. Even MORE unlikely but I think technically possible

ilrudie (commented):
I'm not fully through the PR but this seemed like it was worth commenting on early perhaps so here goes:

src/config.rs
```diff
@@ -117,6 +120,21 @@ pub struct Config {
     pub connection_window_size: u32,
     pub frame_size: u32,
 
+    // The limit of how many streams a single HBONE pool connection will be limited to, before
+    // spawning a new conn rather than reusing an existing one, even to a dest that already has an open connection.
+    //
+    // This can be used to effect flow control for "connection storms" when workload clients
+    // (such as loadgen clients) open many connections all at once.
+    //
+    // Note that this will only be checked and inner conns rebalanced accordingly when a new connection
```
howardjohn, line 129: What does "rebalanced" mean? that seems impossible to me.

src/config.rs
```diff
@@ -307,6 +325,13 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
             .get(DNS_CAPTURE_METADATA)
             .map_or(false, |value| value.to_lowercase() == "true"),
 
+        pool_max_streams_per_conn: parse_default(POOL_MAX_STREAMS_PER_CONNECTION, 250)?,
```
howardjohn, line 328: So we fail over if we happen to set this above the H2 level max stream setting. We don't have visiblity into this, so we guess.

I looked around, and defaults are:

Go: 100
Hyper: 200
Envoy: 2147483647 (lol)
Spec recommended minimum 100

This seems like if we are at 250 we are going to break even for ztunnel internal communication.

I would recommend we set this to 100.

Long term we can be adaptive to the actual limit, of course.

src/proxy/pool.rs
```diff
@@ -12,80 +12,539 @@
...
-    fn reserve(self) -> Reservation<Self> {
-        let b = self.clone();
-        let a = self;
-        Reservation::Shared(a, b)
+    pub fn send_request(
+        &mut self,
+        req: Request<Empty<Bytes>>,
+    ) -> impl Future<Output = hyper::Result<Response<Incoming>>> {
+        // TODO should we enforce streamcount per-sent-request? This would be slow.
+        self.stream_count.fetch_add(1, Ordering::Relaxed);
```
howardjohn, line 531: am I missing something or do we never decrement this?

BTW -- we need to decrement not when the send_request is done, since the streaming body persists beyond the `Response`

src/proxy/pool.rs
```diff
@@ -12,80 +12,539 @@
...
+                // If we get here, it means the following are true:
+                // 1. At one point, there was a preexisting conn in the pool for this key.
+                // 2. When we checked, we got nothing for that key.
+                // 3. We could not get the exclusive inner writelock to add a new one for this key.
+                // 4. Someone else got the exclusive inner writelock, and is adding a new one for this key.
+                //
+                // So, loop and wait for the pool_watcher to tell us a new conn was enpooled,
+                // so we can pull it out and check it.
+                loop {
+                    match self.pool_watcher.changed().await {
```
howardjohn, line 363: hm.. Should the watchers be per-key? Seems like we are going to wake up a lot of things unnecessarily.

src/proxy/pool.rs
```diff
@@ -12,80 +12,539 @@
...
+                        Ok(_) => {
+                            trace!(
+                                "notified a new conn was enpooled, checking for hash {:#?}",
+                                hash_key
+                            );
+                            // The sharded mutex for this connkey is already locked - someone else must be making a conn
+                            // if they are, try to wait for it, but bail if we find one and it's got a maxed streamcount.
+                            let existing_conn = self.state.connected_pool.get(&hash_key);
+                            match existing_conn {
```
howardjohn, line 372: I don't understand this. 

This loop should get triggered by a pool_watcher event. pool_watcher can come from _anyone_ notifying (i.e. different keys).

Then we read from the pool immediately. This is not a mutex that is held for a long time (basically just to read and immediately unlock, not held across IO.

So the "try to wait for it" comment confuses me

src/proxy/pool.rs
```diff
@@ -12,80 +12,539 @@
...
+                                None => {
+                                    trace!("woke up on pool notification, but didn't find a conn for {:#?} yet", hash_key);
+                                    continue;
+                                }
+                                Some(e_conn) => {
+                                    debug!("found existing conn after waiting");
+                                    // We found a conn, but it's already maxed out.
```
howardjohn, line 379: I am not sure this is what we want. Consider max steams=3 (for simple example).

Here is the desired behavior I think:
1. I open 2 streams. One wins, and makes a connection and opens a stream. The other waits, and opens a stream on the connection.
1. I open 2 more streams. One checks out the existing connection, _and the other opens a new connection_ .
1. (now I have 1 conn w/ 3, 1 conn w/ 1 stream)
2. Now I open 6 more streams. I should have 2 checkout the existing connection, and at the same time _open 2 more connections concurrently_

IIUC here, we are saying "You can never have more than `max_streamcount` _at all_". But that is what we are doing today. What we want is "you can never have more than max_streamcount on a connection (because that is how HTTP2 works), so if you go over we spin up more connections).

howardjohn: Upon rereading, I misunderstood (which I was pretty sure going into it :slightly_smiling_face: ).

I do think on (4) we are not doing the new connections concurrently but

src/proxy/pool.rs
```diff
@@ -12,80 +12,539 @@
...
+            // Clone the underlying client, return a copy, and put the other back in the pool.
+            Some(f_conn) => {
+                self.state.checkin_conn(f_conn.clone(), pool_key.clone());
+                Ok(f_conn)
+            }
+
+            // After waiting, we found an available conn, but for whatever reason couldn't use it.
+            // (streamcount maxed, etc)
+            // Start a new one, clone the underlying client, return a copy, and put the other back in the pool.
+            None => {
```
howardjohn, line 424: If we have multiple waiters, can they all end up here at once and spam a bunch of new connections?

howardjohn (commented):
Ill review more closely tomorrow, only had a chance for a quick look. As such, many of my comments are probably not accurate

src/proxy/pool.rs
```diff
@@ -12,80 +12,539 @@
...
+        let existing_conn = self
+            .state
+            .first_checkout_conn_from_pool(&workload_key, &pool_key)
+            .await;
+
+        // Early return, no need to do anything else
+        if existing_conn.is_some() {
+            debug!("initial attempt - found existing conn, done");
+            return Ok(existing_conn.unwrap());
+        }
```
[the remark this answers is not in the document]
bleggett: Yes, good catch, I twiddled the indent to fast-exit (used to be one return).

src/proxy/pool.rs
```diff
@@ -12,80 +12,539 @@
...
+// PoolState is effectively the gnarly inner state stuff that needs thread/task sync, and should be wrapped in a Mutex.
+struct PoolState {
+    pool_notifier: watch::Sender<bool>, // This is already impl clone? rustc complains that it isn't, tho
+    timeout_tx: watch::Sender<bool>, // This is already impl clone? rustc complains that it isn't, tho
+    timeout_rx: watch::Receiver<bool>,
+    // this is effectively just a convenience data type - a rwlocked hashmap with keying and LRU drops
+    // and has no actual hyper/http/connection logic.
+    connected_pool: Arc<pingora_pool::ConnectionPool<ConnClient>>,
+    // this must be an atomic/concurrent-safe list-of-locks, so we can lock per-key, not globally, and avoid holding up all conn attempts
+    established_conn_writelock: HashMap<u64, Option<Arc<Mutex<()>>>>,
```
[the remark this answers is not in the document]
bleggett: What does? The use of locks?

src/proxy/pool.rs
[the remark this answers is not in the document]
bleggett: Yes, I outdented and early-returned and forgot to check here too, good catch ty.

src/proxy/pool.rs
```diff
@@ -12,80 +12,539 @@
...
+    pub async fn connect(&mut self, workload_key: WorkloadKey) -> Result<ConnClient, Error> {
+        debug!("pool connect START");
+        // TODO BML this may not be collision resistant/slow. It should be resistant enough for workloads tho.
+        // We are doing a deep-equals check at the end to mitigate any collisions, will see about bumping Pingora
+        let mut s = DefaultHasher::new();
+        workload_key.hash(&mut s);
+        let hash_key = s.finish();
+        let pool_key = pingora_pool::ConnectionMeta::new(
+            hash_key,
+            GLOBAL_CONN_COUNT.fetch_add(1, Ordering::Relaxed),
```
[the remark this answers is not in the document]
bleggett: Yep, forgot to move this in.

src/proxy/pool.rs
bleggett: It's part of the ConnClient, which is as close to the horse's mouth as you can get. And where we get a lock, we get a ConnClient, so I think it will be the same, but I'll check if it simplifies the validation.


There would be no difference in safety. Popping a conn is not harmful either way, as long as we do not return it to the caller if there's a mismatch - collision doesn't matter.

src/proxy/pool.rs
bleggett: Ah, that's a good call. Yep. TY!

src/config.rs
bleggett: Old wording, will remove.

src/proxy/pool.rs
bleggett: We never do - once we hit the max we pop the conn out of the pool, start a new one, and let this one expire naturally.

src/proxy/pool.rs
bleggett: Yes, but only if there's a lot of backpressure in the first place, so probably not a common problem.

I considered doing it per-key, but it complicates things considerably (we have another parallel keyed map, we need to synchronize it, etc etc)

src/proxy/pool.rs
bleggett: `wait` in the sense that we will wait for pool watcher to resolve, and if it resolves and we still don't have anything, loop again - not in the sense that there's a `sleep` here (though basically we are `await`ing the watcher which is functionally a `yield` from the runtime perspective)

src/proxy/pool.rs
bleggett: > What we want is "you can never have more than max_streamcount on a connection (because that is how HTTP2 works), so if you go over we spin up more connections).

yes, that is what we have here.

That being said, we only check the max streamcount _on conn checkout attempts_ - so we cannot stop someone from checking out 1x and sending 300 streams over a conn we configured with `max_streamcount = 250`

Our code doesn't do that, and I did not want to update atomics on stream send as that's a perf hit, and we really don't need that granularity here.

src/proxy/pool.rs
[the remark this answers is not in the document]
ilrudie: Just locks which don't wrap the thing they're locking...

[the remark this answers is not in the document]
bleggett: Ah, yep. It's not very cosmetic, agree.

But the kind of locking we have to do is not something the pools out there will do for us, and pingora_pool has _some_ use, at least.

src/proxy/pool.rs
howardjohn: Yeah that's fine, we can just say "1 checkout == 1 stream" and trust we follow that.

But incrementing 1 atomic takes ~1ns. We don't need to worry about that in the context of something run once per connection (O(milliseconds))

src/proxy/pool.rs
bleggett: Reworded, this was an old, out of place comment anyway

src/proxy/pool.rs
bleggett: > But incrementing 1 atomic takes ~1ns. We don't need to worry about that in the context of something run once per connection (O(milliseconds))

Yep, that's fair - I wasn't sure how long it would take (and it seems to be platform-dependent too) and I don't strictly need to do it here, so I didn't.

howardjohn:
I am seeing unexpected behavior. Open up 99 connections at once, I got 9 connections. Only 1 of them closes after the idle tieout, the rest are zombies. Also if I then open up 99 more, I get the same - 8 more zombies, 16 total.

src/proxy/pool.rs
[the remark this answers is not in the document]
ilrudie: Its moot unless we want to fork or contribute to pingora (or I guess lock the entire pool, which we don't care to do) but the point of wrapping the mutex around the thing being accessed is more than cosmetic. It ties holding the lock and accessing the thing together in a way that we can't do with pingora. We'll need to be more careful since we can't do take advantage of that.

bleggett:
> I am seeing unexpected behavior. Open up 99 connections at once, I got 9 connections. Only 1 of them closes after the idle tieout, the rest are zombies. Also if I then open up 99 more, I get the same - 8 more zombies, 16 total.

May have broken it then, needs more tests.

bleggett:
> I am seeing unexpected behavior. Open up 99 connections at once, I got 9 connections. Only 1 of them closes after the idle tieout, the rest are zombies. Also if I then open up 99 more, I get the same - 8 more zombies, 16 total.

I now have enough tests in place to be pretty confident that the pool connection eviction logic works correctly, which means the only other way this can happen is if `outbound.rs` isn't letting go of the conn.

bleggett:
@howardjohn The pool is evicting conns properly and not holding onto any references (unit tests, and logs, show this), and that `outbound.rs` is in fact evicting (and ultimately releasing) all the conns it opens, e.g the following with `kubectl exec -it deploy/captured-v1 -n echo-1-77948 -- client http://httpbin.default.svc:8000/headers --count 200 --new-connection-per-request` + 1m evict timeout + 5stream-per-conn:

~/Source/istio/istio (master*) » cat ztunnel-t6vxw-outbound.log| grep "outbound spawn START" | wc -l     
200
~/Source/istio/istio (master*) » cat ztunnel-t6vxw-outbound.log| grep "outbound spawn DONE" | wc -l
200
~/Source/istio/istio (master*) » cat ztunnel-hq82p-inbound.log| grep "accepted connection" | wc -l 
40
~/Source/istio/istio (master*) » cat ztunnel-t6vxw-outbound.log| grep "starting a connection driver" | wc -l
40
~/Source/istio/istio (master*) » cat ztunnel-t6vxw-outbound.log| grep "done with HBONE " | wc -l  
40


However, it looks as if `inbound.rs` is not properly letting go, even though the client has dropped the connection objects: 
~/Source/istio/istio (master*) » cat ztunnel-hq82p-inbound.log| grep "inbound serve done" | wc -l
11

bleggett:
@howardjohn and @ilrudie PTAL at the changes - added tests, reorganized a bit, and moved all the inner connlocking to `PoolState` so it's a bit easier to reason about.

For what John saw in https://github.com/istio/ztunnel/pull/931#issuecomment-2067188843 -> that seems to be an `inbound.rs` problem, and not related to what the pool is doing, so I'm going to look at that separately.

howardjohn:
Given the inbound.rs issue is introduced in this pr, i suspect we are not `Drop`ing one of the SendRequests.

I am not sure that dropping the driver is enough

bleggett:
> Given the inbound.rs issue is introduced in this pr, i suspect we are not `Drop`ing one of the SendRequests.
> 
> I am not sure that dropping the driver is enough

Unless `outbound.rs` is hanging on to a SendRequest (and it isn't, according to logs), I don't believe there are any unaccounted for - the pool unit tests prove the pool isn't holding on to them, at least - we'd see exactly the same problem with the test server in that case, and we do not.

howardjohn:
> > Given the inbound.rs issue is introduced in this pr, i suspect we are not `Drop`ing one of the SendRequests.
> > I am not sure that dropping the driver is enough
> 
> Unless `outbound.rs` is hanging on to a SendRequest (and it isn't, according to logs), I don't believe there are any unaccounted for - the pool unit tests prove the pool isn't holding on to them, at least - we'd see exactly the same problem with the test server in those tests in that case, and we do not.

I haven't looked at the tests much to see, but I did verify holding onto it will block:

```diff
diff --git a/src/proxy/inbound.rs b/src/proxy/inbound.rs
index e449471..37f3277 100644
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -167,7 +167,7 @@ impl Inbound {
                     }
                     // Serving finished, just return the result.
                     futures_util::future::Either::Right((server, _shutdown)) => {
-                        debug!("inbound serve done {:?}", server);
+                        error!("inbound serve done {:?}", server);
                         server
                     }
                 }
diff --git a/src/proxy/pool.rs b/src/proxy/pool.rs
index f274617..affef8b 100644
--- a/src/proxy/pool.rs
+++ b/src/proxy/pool.rs
@@ -42,7 +42,7 @@ impl Pool {
         Self {
             pool: HyperPool::new(
                 hyper_util::client::legacy::pool::Config {
-                    idle_timeout: Some(Duration::from_secs(90)),
+                    idle_timeout: Some(Duration::from_secs(1)),
                     max_idle_per_host: std::usize::MAX,
                 },
                 TokioExec,
@@ -66,11 +66,11 @@ where
 }
 
 #[derive(Debug, Clone)]
-struct Client(http2::SendRequest<Empty<Bytes>>);
+struct Client(Option<http2::SendRequest<Empty<Bytes>>>);
 
 impl Poolable for Client {
     fn is_open(&self) -> bool {
-        self.0.is_ready()
+        self.0.as_ref().map(|r| r.is_ready()).unwrap_or_default()
     }
 
     fn reserve(self) -> Reservation<Self> {
@@ -100,11 +100,17 @@ pub struct Key {
 pub struct Connection(Pooled<Client, Key>);
 
 impl Connection {
-    pub fn send_request(
+    pub async fn send_request(
         &mut self,
         req: Request<Empty<Bytes>>,
-    ) -> impl Future<Output = hyper::Result<Response<Incoming>>> {
-        self.0 .0.send_request(req)
+    ) -> hyper::Result<Response<Incoming>> {
+        let Some(mut sender) = self.0.0.take() else {
+            panic!("wtf");
+        };
+        let res = sender.send_request(req).await;
+        // UNCOMMENT ME TO NOT CLOSE
+        // std::mem::forget(sender);
+        res
     }
 }
 
@@ -122,7 +128,7 @@ impl Pool {
                 // Return an error so
                 return Err(Error::PoolAlreadyConnecting);
             };
-            let pc = Client(connect.await?);
+            let pc = Client(Some(connect.await?));
             let pooled = self.pool.pooled(connecting, pc);
             Ok::<_, Error>(pooled)
         };

```

which now I realize is obvious, and in that case the driver will not drop either. In this case, the driver does close.

So probably not useful, but still keeping the diff in a comment in case it helps

bleggett:
> So probably not useful, but still keeping the diff in a comment in case it helps

Yeah it's a little opaque when hyper/h2 decide to close, I think it's when the underlying client they manage have streamcount == 0 && refcount == 0, or something like that, IIRC

howardjohn:
From logs, we never close the stream. So it cannot be an issue on inbound.rs, it must be in the pool. Not sure where though

bleggett:
> From logs, we never close the stream. So it cannot be an issue on inbound.rs, it must be in the pool. Not sure where though

What are the logs you're keying that off of, just for ref? I'll dig into what `inbound.rs` is doing more tomorrow.

It's possible we are not doing something in the order hyper/h2 expect.

If the pool were holding onto a reference, I would expect `test_pool_1000_clients_3_srcs_drops_after_timeout` to fail.

howardjohn:
it's hyper or h2 debug logs (forgot which, I think h2)

howardjohn:
the test makes all the connections sequentially.. maybe the difference

bleggett:
> the test makes all the connections sequentially.. maybe the difference

The tests that validate eviction and conn close spawn sequentially 100 and 1000 independent tokio connection tasks which drive themselves to completion - that should be almost exactly equivalent to what `echo client` does with `--count 200 --new-connection-per-request`

Also, it's clear from the logs above that 40 conns are started, and the `connection` driver task h2 gives us self-completes without error on the clientside 40 times - that's `cat ztunnel-t6vxw-outbound.log| grep "done with HBONE " | wc -l  -> 40`

If pool was holding a ref, the h2-provided poll task would not self-complete without error for every connection, as you pointed out above. But it does. 

So it's likely something outside the pool that is misbehaving.

Either h2/hyper aren't properly following rust drop semantics and need additional clientside incantations to properly coerce servers to close connections (which I doubt), or the server needs a keepalive/conn timeout.

howardjohn:
I believe the client sends the request in parallel.

It cannot be the server at fault. An http 2 server is supposed to keep the
connection alive until signaled to close by the client. The client is not
doing so

bleggett:
> I believe the client sends the request in parallel. It cannot be the server at fault. An http 2 server is supposed to keep the connection alive until signaled to close by the client. The client is not doing so

Then one way or another it must be a hyper/h2 bug.

bleggett:
@howardjohn 

Another bit of data - since the `impl Drop for ConnClient` has a debug log:

```cat ztunnel-75mqd-outbound.log| grep "dropping ConnClient" | wc -l 

240
```

200 sends, 40 actual conns -> 240 total drops - that's all of them accounted for.

If we are dropping all the senders and the `connection` poll driver hyper hands us self-resolves without error, I don't know what else we should be doing with `hyper` clients that isn't an inner Hyper state bug.

I added https://github.com/istio/ztunnel/pull/931/commits/e63093624d598310151d05151dd156c8e21b2291

a serverside keepalive, which "fixes" the zombie conns. We will need that anyway, as we cannot assume all clients are well-behaved.


I am still digging into why `hyper` is not cleanly closing things.

src/proxy/pool.rs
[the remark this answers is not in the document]
bleggett: If we wanted to make this simpler, I would advocate we throw out pingora and implement our own inner data structure from scratch, what we want to do is non-standard and pingora is not likely to want or need what we would have to change to make it work like this - nor is any other "generic" HTTP2 pool. 

Which is why for now I just use it and then do the other stuff outside.

bleggett:
> > The one issue is the hyper/h2 client bug, which is demonstrably unrelated to the pool, needs to be addressed seperately, and is worked around by doing something we needed to do anyway, which is the server side keep-alive, which I added here.
 
> I am not sure about this. I am very very worried about leaking connections left and right as we do with this pool, this is going to lead to a huge performance regression? 

Where are we leaking connections? If you believe we are, please put that in a review comment, when we have time, with specifics. 

From https://github.com/istio/ztunnel/pull/931#issuecomment-2072995039 it's clear that we are leaking no connections at all, which is what I would expect. We create N references and we drop N references. Unless `drop` or the Rust compiler are lying to us (in which case we have bigger problems), we are not.

It's no different than the previous pool in this regard - hyper creates the connections. We hand out references to consumers. If consumers do not drop the references, the connection never closes. If they do, it does - as managed by Hyper (not the pool).

> Whether its a hyper bug or a pool bug is irrelevant -- we have 3 pooling implementations, and it _only_ exists in this one, so the end result to users is the same...

It's demonstrably not a pool bug. Hyper closes connections, the pool does not. And Hyper's client is closing the connections successfully in all cases. I agree we should chase down whether it's a hyper or h2 client bug as a separate issue, but either way it is obviated by the serverside keepalive added here, which we must have anyway. I don't think it makes sense to ship a worse pooling impl with fewer tests because we have a generic Hyper/h2 client bug with a workaround in place.

howardjohn:
> From https://github.com/istio/ztunnel/pull/931#issuecomment-2072995039 it's clear that we are leaking no connections at all, which is what I would expect. We create N references and we drop N references. Unless drop or the Rust compiler are lying to us (in which case we have bigger problems), we are not.

bleggett:
> > From [#931 (comment)](https://github.com/istio/ztunnel/pull/931#issuecomment-2072995039) it's clear that we are leaking no connections at all, which is what I would expect. We create N references and we drop N references. Unless drop or the Rust compiler are lying to us (in which case we have bigger problems), we are not.
> 
> Yes, in the synthetic unit test we claim to drop them all. But in real world testing, simply opening up a few connections concurrently is enough to trigger them to leak and never get closed. This is not a rare obscure occurance either, happens ~75% of the time i open up new connections.
> 
> Watching `Drops` is not enough because we clone the structure all over the place. I don't think its unreasonable to think we are cloning it somewhere unexpected and keeping it alive. BTW, pingora uses `ManuallyDrop` in a few places which is another plausible source of errors if they never actually trigger the drop.

Update on this - just to be safe I tracked `clones` and matched them up to `drops`:
This is testing with `kubectl exec -it deploy/captured-v1 -n echo-1-50455 -- client http://httpbin.default.svc:8000/headers --count 200 --new-connection-per-request`, with `RUST_LOG=trace`, `POOL_MAX_STREAMS_PER_CONNECTION=5`, `POOL_UNUSED_RELEASE_TIMEOUT = 5m`:

After the 5m eviction timeout evicts the final conn from the pool:

```
» cat ztunnel-fjtnx-outbound.log| grep "starting a connection driver" | wc -l
40

» cat ztunnel-fjtnx-outbound.log| grep "done with HBONE" | wc -l
40

cat ztunnel-fjtnx-outbound.log| grep "dropping ConnClient" | wc -l         
240

at ztunnel-fjtnx-outbound.log| grep "cloning ConnClient" | wc -l          
200
```
Which is again what I would expect. 40 conns (200 / 5), 200 clones, 240 drops

That's every single thing that `hyper` gives us accounted for by the pool and `outbound.rs`, unless Rust's drop semantics are lying.

bleggett:
Hyper bug seems to have been chased down to https://github.com/hyperium/hyper/pull/3647 - leaving it here for tracking.

The serverside keepalive that works around it should do in the meantime, and even after it's fixed I think it's worth keeping the keepalive to guard against badly-behaved or otherwise buggy clients.

src/proxy/pool.rs
```diff
@@ -12,241 +12,1413 @@
...
+                break;
+            }
+        }
+
+        let (client_stop_signal, client_stop) = drain::channel();
+        let persist_res =
+            spawn_persistent_client(pool.clone(), key1.clone(), server_addr, client_stop);
+
+        //loop thru the nonpersistent clients and wait for them to finish
+        while let Some(Err(res)) = tasks.next().await {
```
howardjohn, line 1198: are we intentionally running through sequentially? if not we can use `futures::future::try_join_all(tasks).await.unwrap();`

howardjohn: NVM - we spawn them so it doesn't matter

src/proxy/pool.rs
```diff
@@ -12,241 +12,1413 @@
...
+            assert!(!res.is_panic(), "CLIENT PANICKED!");
+            continue;
+        }
+
+        let before_conncount = conn_counter.load(Ordering::Relaxed);
+        let before_dropcount = conn_drop_counter.load(Ordering::Relaxed);
+        assert!(
+            before_conncount == 4,
```
howardjohn, line 1206: Why is there 4? We have 50 max streams, 100 clients. Shouldn't there be 2?

src/proxy/pool.rs
```diff
@@ -12,241 +12,1413 @@
...
+            "actual before conncount was {before_conncount}"
+        );
+        assert!(
+            before_dropcount == 0,
```
howardjohn, line 1210: shouldn't there be 1?

1 connection closes immediately, the other should idle out

src/proxy/pool.rs
```diff
@@ -12,241 +12,1413 @@
...
+            dst_id: vec![Identity::default()],
+            src: IpAddr::from([127, 0, 0, 2]),
+            dst: server_addr,
+        };
+        let client_count = 100;
+        let mut count = 0u32;
+        let mut tasks = futures::stream::FuturesUnordered::new();
+        loop {
+            count += 1;
+            tasks.push(spawn_client(pool.clone(), key1.clone(), server_addr, 100));
```
howardjohn, line 1186: Why send 100 requests? We know we will only send 1 in real world (and probably can enforce this in the library in a follow up), and it makes the test pretty slow

src/proxy/pool.rs
```diff
@@ -12,241 +12,1413 @@
...
+                // .instrument(tracing::debug_span!("client_tid", tid=%tokio::task::id()))
+                .await
+                .unwrap();
+            debug!(
+                "client spent {}ms waiting for conn",
+                start.elapsed().as_millis()
+            );
+
+            let mut count = 0u32;
+            // send forever, once we get a conn, until someone signals us to stop
```
howardjohn, line 1322: Probably don't need to infinite loop on sending requests, but rather have one request we keep open forever

src/proxy/pool.rs
```diff
@@ -12,241 +12,1413 @@
...
+        assert!(real_conncount == 3, "actual conncount was {real_conncount}");
+        assert!(real_dropcount == 3, "actual dropcount was {real_dropcount}");
+
+        server_drain_signal.drain().await;
+        server_handle.await.unwrap();
+        drop(pool);
+    }
+
+    #[tokio::test]
+    async fn test_pool_100_clients_evicts_but_does_not_close_active_conn() {
```
howardjohn, line 1151: This is pretty slow, about 8s on my machine. Seems like we can drop this down a ton. I played around with it a bit:

```
diff --git a/src/proxy/pool.rs b/src/proxy/pool.rs
index c2a4294..cb6f744 100644
--- a/src/proxy/pool.rs
+++ b/src/proxy/pool.rs
@@ -629,6 +629,7 @@ mod test {
     use tokio::net::TcpListener;
     use tokio::task::{self};
     use tokio::time::sleep;
+    use tracing::info;
 
     #[cfg(tokio_unstable)]
     use tracing::Instrument;
@@ -1149,7 +1150,7 @@ mod test {
 
     #[tokio::test]
     async fn test_pool_100_clients_evicts_but_does_not_close_active_conn() {
-        // crate::telemetry::setup_logging();
+        crate::telemetry::setup_logging();
 
         let (server_drain_signal, server_drain) = drain::channel();
 
@@ -1165,7 +1166,7 @@ mod test {
         let cfg = crate::config::Config {
             local_node: Some("local-node".to_string()),
             pool_max_streams_per_conn: 50,
-            pool_unused_release_timeout: Duration::from_secs(2),
+            pool_unused_release_timeout: Duration::from_millis(200),
             ..crate::config::parse_config().unwrap()
         };
         let sock_fact = Arc::new(crate::proxy::DefaultSocketFactory);
@@ -1181,58 +1182,63 @@ mod test {
         let client_count = 100;
         let mut count = 0u32;
         let mut tasks = futures::stream::FuturesUnordered::new();
+        info!("run0");
         loop {
             count += 1;
-            tasks.push(spawn_client(pool.clone(), key1.clone(), server_addr, 100));
+            tasks.push(spawn_client(pool.clone(), key1.clone(), server_addr, 1));
 
             if count == client_count {
                 break;
             }
         }
 
+        info!("run1");
         let (client_stop_signal, client_stop) = drain::channel();
         let persist_res =
             spawn_persistent_client(pool.clone(), key1.clone(), server_addr, client_stop);
+        info!("run12");
 
+        futures::future::try_join_all(tasks).await.unwrap();
         //loop thru the nonpersistent clients and wait for them to finish
-        while let Some(Err(res)) = tasks.next().await {
-            assert!(!res.is_panic(), "CLIENT PANICKED!");
-            continue;
-        }
+        // while let Some(Err(res)) = tasks.next().await {
+        //     assert!(!res.is_panic(), "CLIENT PANICKED!");
+        //     continue;
+        // }
+        info!("run3");
 
         let before_conncount = conn_counter.load(Ordering::Relaxed);
         let before_dropcount = conn_drop_counter.load(Ordering::Relaxed);
         assert!(
-            before_conncount == 4,
+            before_conncount == 2,
             "actual before conncount was {before_conncount}"
         );
         assert!(
-            before_dropcount == 0,
+            before_dropcount == 1,
             "actual before dropcount was {before_dropcount}"
         );
 
         // Attempt to wait long enough for pool conns to timeout+evict
-        sleep(Duration::from_secs(2)).await;
+        sleep(Duration::from_millis(200)).await;
 
         let real_conncount = conn_counter.load(Ordering::Relaxed);
-        assert!(real_conncount == 4, "actual conncount was {real_conncount}");
+        assert!(real_conncount == 3, "actual conncount was {real_conncount}");
         // At this point, we should still have one conn that hasn't been dropped
         // because we haven't ended the persistent client
         let real_dropcount = conn_drop_counter.load(Ordering::Relaxed);
-        assert!(real_dropcount == 3, "actual dropcount was {real_dropcount}");
+        assert!(real_dropcount == 2, "actual dropcount was {real_dropcount}");
         client_stop_signal.drain().await;
         assert!(persist_res.await.is_ok(), "PERSIST CLIENT ERROR");
 
-        sleep(Duration::from_secs(2)).await;
+        sleep(Duration::from_millis(200)).await;
 
         let after_conncount = conn_counter.load(Ordering::Relaxed);
         assert!(
-            after_conncount == 4,
+            after_conncount == 3,
             "after conncount was {after_conncount}"
         );
         let after_dropcount = conn_drop_counter.load(Ordering::Relaxed);
         assert!(
-            after_dropcount == 4,
+            after_dropcount == 3,
             "after dropcount was {after_dropcount}"
         );
         server_drain_signal.drain().await;

```
Gets to 500ms

The concerning bit is the assertions are different... but they seem more correct to me somehow. Probably I don't understand the test fully.

src/proxy/pool.rs
bleggett: 50 max streams, 100 clients, 100 reqs-per-client

src/proxy/pool.rs
bleggett: Just to exercise the logic, I agree we won't actually do this.

src/proxy/pool.rs
howardjohn: How does that end up at 4? 1 req or 100 should not be relevant since they are sequential. Each client has 0 or 1 active streams at a time

src/proxy/pool.rs
bleggett: Yeah the conn count changes because with streamcount = 100, when each client "takes" the connection and we check the streamcount (and what it happens to be at that time based on currently-sending clients) might vary.

Dropping this down to 1 stream per will make this test more deterministic. Unfortunately, doing so for me causes the test server to start refusing conns, probably because we are making too many connection attempts too quickly.

src/proxy/pool.rs
bleggett: Yep, this was just making some tests more confusing/flaky anyway.

src/proxy/pool.rs
bleggett: Yep, this is a flaky check.

src/proxy/pool.rs
bleggett: I halved most of these.

> test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 125 filtered out; finished in 2.53s

The test rig is crude (especially the fake server) and it could be made faster.

src/proxy/pool.rs
bleggett: Sorry, I realize what I said here is uselessly cryptic:

The client spawns are not sequential, in the test. With >1 streams per conn, and unordered spawns, the actual streamcount for a given conn when a client actually requests a con is not deterministic.

Which is just a test flake waiting to happen.

Fixed this by doing what you suggested and dropping the count down to 1 stream per spawned client, and removing the constant request loop for the persistent client, which gives us 3 - which makes more sense.

src/proxy/pool.rs
bleggett: Yeah this is largely just a rollup to see if any panicked.

bleggett:
1. The hyper bug in https://github.com/hyperium/hyper/pull/3647 is worked around for now with server keepalives. This will do until we take a bumped hyper, and we need server keepalives anyway, as a safety valve to keep misbehaving clients (even if they are ours) from hanging connections.

2. I am not 100% happy with the test rig - it is crude, it is synthetic, but it does the job and makes this that much harder to break, and it's certainly covering more behavior than before.

howardjohn (approved):

Issue #928, opened 2024-04-16
Allow Prometheus -> ztunnel scraping to be HBONE


howardjohn:
We should ideally allow ztunnel _itself_ to be a HBONE accepting workload, so traffic directly to "Ztunnel the workload" accepts HBONE.

Notably here is Prometheus, though others probably apply

This may be more complex than its worth, so lets hold this unless there is a clean approach

howardjohn:
Currently you actually cannot even call ztunnel at all. Sending a fix

PR #935, opened 2024-04-17
block illegal calls in inpod mode

howardjohn:
Part of https://github.com/istio/ztunnel/issues/928

This removes the obsolete call protections for inpod, and adds
equivilents

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashSet;
 use std::fmt::Debug;
 use std::net::{IpAddr, SocketAddr};
 use std::sync::Arc;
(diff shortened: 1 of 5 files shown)


src/proxy.rs
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::collections::HashSet;
```
ilrudie, line 15: Is hash necessary? Could we just use a vec, sort it once during init and then efficiently search it without calculating hashes on the u16. Because its small we could probably just iterate for a match instead of sort/binary search even

ilrudie (commented):
I don't have data to back it but I'm suspicious that ~4 elements just using a vec might be the cheaper option.

Otherwise looks good

src/proxy.rs
howardjohn: Probably not, but its just 1 instance with like 4 keys and we Arc it anyways so doesn't seem so bad?

bleggett (approved):
Smarter illegal port registration would be nice, but that's largely a cosmetic concern ATM.

src/proxy.rs
ilrudie: Yeah, if I start feeling froggy maybe I'll test it and let ya know.

Issue #932, opened 2024-04-16
DNS metrics have unbounded cardinality


howardjohn:
Our DNS metrics have "hostname" as a label. This is unbounded, so we can generate infinite metrics. This needs to be bound

PR #936, opened 2024-04-17
dns: remove unbounded cardinality in metrics

howardjohn:
This removes all destination attributes, but keeps source. We don't
report source in sidecar, but implicitly get it because its per-pod.

Fixes https://github.com/istio/ztunnel/issues/932

Prior art: cilium does not have src/dst meta:
https://docs.cilium.io/en/latest/observability/metrics/#dns


bleggett (approved):
This LGTM and I don't know of any specific need for the removed fields.

If we find a need, we can do something a little nicer to support them.

PR #940, opened 2024-04-19
metrics: make traffic metrics faster

howardjohn:
Before, each r/w we lookup the metric and increment it.

Now, we look it up once and then just increment. The increment is just
Atomic.increment, which on x86 is the same as a simple `add`, which is
very fast (measured 1ns, but presumably its so fast benchmarking isn't
meaningful).

Since this is on the hot path, this is meaningful.

Given this, I don't think we will ever need to do
https://github.com/istio/ztunnel/pull/927#issuecomment-2060434505.

Also, addressed
https://github.com/istio/ztunnel/pull/927#discussion_r1569021128 and
cleaned up inbound which broke a function up into 2 which made things
more complex, not simpler.

### DIFF
--- a/src/proxy/inbound_passthrough.rs
+++ b/src/proxy/inbound_passthrough.rs
@@ -231,6 +231,6 @@ impl InboundPassthrough {
         };
 
         let res = conn_guard.handle_connection(send).await;
-        result_tracker.record(res.map(|_| ()));
+        result_tracker.record(res);
     }
 }
--- a/src/proxy/inbound_passthrough.rs
+++ b/src/proxy/inbound_passthrough.rs
@@ -231,6 +231,6 @@ impl InboundPassthrough {
         };
 
         let res = conn_guard.handle_connection(send).await;
-        result_tracker.record(res);
+        result_tracker.record(res.map(|_| ()));
     }
 }
--- a/src/proxy/metrics.rs
+++ b/src/proxy/metrics.rs
@@ -14,11 +14,12 @@
 
 use std::fmt::Write;
 use std::net::SocketAddr;
-use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::{atomic, Arc};
 use std::time::Instant;
 
 use prometheus_client::encoding::{EncodeLabelSet, EncodeLabelValue, LabelValueEncoder};
-use prometheus_client::metrics::counter::Counter;
+use prometheus_client::metrics::counter::{Atomic, Counter};
 use prometheus_client::metrics::family::Family;
 use prometheus_client::registry::Registry;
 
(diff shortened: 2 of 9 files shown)


src/proxy/outbound.rs
```diff
@@ -258,7 +258,7 @@ impl OutboundConnection {
                     .await
             }
         };
-        result_tracker.record(res)
+        result_tracker.record(res.map(|_| ()))
```
bleggett, line 261: Ditto?

src/proxy/inbound_passthrough.rs
```diff
@@ -245,6 +245,6 @@ impl InboundPassthrough {
             }
             _signaled = close.signaled() => Err(Error::AuthorizationPolicyLateRejection)
         };
-        result_tracker.record(res);
+        result_tracker.record(res.map(|_| ()));
```
bleggett, line 248: What's with the empty closure?

src/proxy/metrics.rs
```diff
@@ -449,35 +459,54 @@ impl ConnectionResult {
...
         let tl = &self.tl;
 
         // Unconditionally record the connection was closed
         self.metrics.connection_close.get_or_create(tl).inc();
 
         // Unconditionally write out an access log
         let mtls = tl.connection_security_policy == SecurityPolicy::mutual_tls;
-        let bytes = res.as_ref().ok();
+        let bytes = (
+            self.recv.load(Ordering::SeqCst),
```
bleggett, line 507: Presumably this would be much slower than the writes, but is out of the hotpath so should be OK.

src/proxy/inbound_passthrough.rs
howardjohn: `res` it `Result<(u64,u64)>` but record takes `Result<()>`

src/proxy/metrics.rs
howardjohn: On x86 Relaxed is just a `mov` (read: atomic is ~free). SeqCst is `xchg`. Still very cheap, and we do this once per connection -- this is actually probably the cheapest line in this function

src/proxy/inbound_passthrough.rs
howardjohn: We could also drop all the send functions to stop tracking the bytes. Just moving this really.

bleggett: Oh, I see. We aren't using those retvals at all anymore - yeah the send funcs should probably not send them at all, it's a bit messy to throw them up to another layer which drops them in an empty closure.

Either add a TODO here explaining what's going on, or fix it, I am fine with either.

howardjohn: done

bleggett (approved):

PR #939, opened 2024-04-18
Track outbound connections

howardjohn:
- **Refactor connection tracker**
- **Track outbound connections as well**

Ultimately this enables istioctl to give us:

```
WORKLOAD                                      DIRECTION LOCAL                                              REMOTE                                             REMOTE TARGET
cartservice-bffc98bf4-bmqsg.anthos            Inbound   cartservice-bffc98bf4-bmqsg.anthos:7070            checkoutservice-c58746cd5-prxg9.anthos:41903
cartservice-bffc98bf4-bmqsg.anthos            Inbound   cartservice-bffc98bf4-bmqsg.anthos:7070            frontend-86c4744d65-f6xwb.anthos:50955
cartservice-bffc98bf4-bmqsg.anthos            Outbound  cartservice-bffc98bf4-bmqsg.anthos:50320           redis-cart-7ff8f4d6ff-lmlwg.anthos:6379            redis-cart.anthos.svc.cluster.local:6379
cartservice-bffc98bf4-bmqsg.anthos            Outbound  cartservice-bffc98bf4-bmqsg.anthos:50330           redis-cart-7ff8f4d6ff-lmlwg.anthos:6379            redis-cart.anthos.svc.cluster.local:6379
productcatalogservice-64898b8c5f-pqx6h.anthos Inbound   productcatalogservice-64898b8c5f-pqx6h.anthos:3550 recommendationservice-85846457-79l2f.anthos:58205
productcatalogservice-64898b8c5f-pqx6h.anthos Inbound   productcatalogservice-64898b8c5f-pqx6h.anthos:3550 checkoutservice-c58746cd5-prxg9.anthos:46941
productcatalogservice-64898b8c5f-pqx6h.anthos Inbound   productcatalogservice-64898b8c5f-pqx6h.anthos:3550 frontend-86c4744d65-f6xwb.anthos:39685
frontend-86c4744d65-f6xwb.anthos              Inbound   frontend-86c4744d65-f6xwb.anthos:8080              loadgenerator-866ccfdfc9-9mhz7.anthos:48071
frontend-86c4744d65-f6xwb.anthos              Outbound  frontend-86c4744d65-f6xwb.anthos:44436             checkoutservice-c58746cd5-prxg9.anthos:5050        checkoutservice.anthos.svc.cluster.local:5050
[...]

PR #942, opened 2024-04-19
Pass original service over HBONE for debugging

howardjohn:
Depends on https://github.com/istio/ztunnel/pull/939

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -483,9 +483,19 @@ pub async fn freebind_connect(
 // as a best guess.
 pub fn guess_inbound_service(
     conn: &Connection,
+    for_host_header: &Option<String>,
     upstream_service: Vec<Service>,
     dest: &Workload,
 ) -> Option<ServiceDescription> {
+    // First, if the client told us what Service they were reaching, look for that
+    // Note: the set of Services we look for is bounded, so we won't blindly trust bogus info.
+    if let Some(found) = upstream_service
+        .iter()
+        .find(|s| for_host_header.as_ref() == Some(&s.hostname))
+        .map(ServiceDescription::from)
+    {
+        return Some(found);
+    }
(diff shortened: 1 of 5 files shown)


src/proxy.rs
```diff
@@ -483,9 +483,19 @@ pub async fn freebind_connect(
 // as a best guess.
 pub fn guess_inbound_service(
     conn: &Connection,
+    for_host_header: &Option<String>,
     upstream_service: Vec<Service>,
     dest: &Workload,
 ) -> Option<ServiceDescription> {
+    // First, if the client told us what Service they were reaching, look for that
+    // Note: the set of Services we look for is bounded, so we won't blindly trust bogus info.
```
bleggett, line 491:
```suggestion
    // Note: the set of Services we look for is bounded, so we won't blindly trust bogus info.
```

Bounded by what?

src/proxy/connection_manager.rs
```diff
@@ -123,6 +122,22 @@ pub struct OutboundConnection {
...
+
+#[derive(Debug, Clone, Eq, Hash, Ord, PartialEq, PartialOrd, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct InboundConnection {
+    #[serde(flatten)]
+    pub ctx: ProxyRbacContext,
+    pub dest_service: Option<String>,
+}
+
 impl ConnectionManager {
```
bleggett, line 141: at some point we should probably rename this to `InboundConnectionManager`

src/proxy.rs
howardjohn: From XDS. The problem is there are N services for 1 pod, so we don't know which one. They tell us a arbirtry service, and we use that IFF its in the list of services for the workload

bleggett (approved):

PR #943, opened 2024-04-19
Make socks5 optional, off by default

howardjohn:
This is a super security sensitive codepath, and mostly and experiment.
Make it opt in so we don't stumble into a CVE when we ship beta.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -50,6 +50,8 @@ const ZTUNNEL_WORKER_THREADS: &str = "ZTUNNEL_WORKER_THREADS";
 const ENABLE_ORIG_SRC: &str = "ENABLE_ORIG_SRC";
 const PROXY_CONFIG: &str = "PROXY_CONFIG";
 
+const UNSTABLE_ENABLE_SOCKS5: &str = "UNSTABLE_ENABLE_SOCKS5";
+
 const DEFAULT_WORKER_THREADS: u16 = 2;
 const DEFAULT_ADMIN_PORT: u16 = 15000;
 const DEFAULT_READINESS_PORT: u16 = 15021;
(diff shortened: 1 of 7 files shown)


src/config.rs
```diff
@@ -50,6 +50,8 @@ const ZTUNNEL_WORKER_THREADS: &str = "ZTUNNEL_WORKER_THREADS";
 const ENABLE_ORIG_SRC: &str = "ENABLE_ORIG_SRC";
 const PROXY_CONFIG: &str = "PROXY_CONFIG";
 
+const UNSTABLE_ENABLE_SOCKS5: &str = "UNSTABLE_ENABLE_SOCKS5";
```
bleggett, line 53:
```suggestion
const ENABLE_SOCKS5: &str = "ENABLE_SOCKS5";
```

would prefer to not use the UNSTABLE_ prefix, env vars do not have stability guarantees anyway, nor does anyone setting them by hand on their own ztunnel instance.

src/proxy.rs
```diff
@@ -168,8 +168,13 @@ impl Proxy {
         illegal_ports.insert(inbound_passthrough.address().port());
         let outbound = Outbound::new(pi.clone(), drain.clone()).await?;
         illegal_ports.insert(outbound.address().port());
-        let socks5 = Socks5::new(pi.clone(), drain.clone()).await?;
-        illegal_ports.insert(socks5.address().port());
+        let socks5 = if pi.cfg.socks5_addr.is_some() {
```
bleggett, line 171: nit: `.and_then()` is a tad more concise here I think.

stevenctl (approved):
LGTM, hold if you want to address Ben's comments before merging.

PR #946, opened 2024-04-22
metrics: Don't log 'hbone_target' as a string.

justinpettit:
The other SocketAddr fields (src.addr and dst.addr) don't log as strings, so make them all consistent.


ilrudie (approved):
LGTM, off topic should we consider swapping the dst.addr to be a non-hbone address? from zt's perspective this makes sense but it might be less clear for users consuming the logs what is going on here...

howardjohn (approved):
/cherrypick release-1.22

PR #970, opened 2024-04-24
Implement better "namespaced" tests

howardjohn:
Big change with a lot of improvements...
* Add support for InPod mode into the namespaced tests. Make most test use them, with some core ones running in the legacy SharedNode mode.
* Allow dynamic updates to workloads/services/policies during tests
* Make inpod iptables match istio
* Add assertions against logs, helpful to verify access logs. Obviously, we will want to make sure this isn't used in silly ways
* Bug fix: report waypoint as destination for workload waypoints, as we do for service waypoints

The new tests have some different names to try to make things more clear, but there is effectively a 1:1 port of the old tests; we should have the same coverage other than additions of InPod/SharedNode variants.

Namespaced tests are pretty quick now, taking about 1-2s to run them all for me


scripts/ztunnel-redirect-inpod.sh
```diff
@@ -4,46 +4,43 @@
...
 :FORWARD ACCEPT [0:0]
 :OUTPUT ACCEPT [0:0]
 :POSTROUTING ACCEPT [0:0]
-
--A PREROUTING -m mark --mark 1337/0xfff -j CONNMARK --set-xmark 0x111/0xfff
--A PREROUTING -p tcp -m tcp --dport $POD_INBOUND -m mark ! --mark 1337/0xfff -j TPROXY --on-port $POD_INBOUND --on-ip 127.0.0.1 --tproxy-mark 0x111/0xfff
--A PREROUTING -p tcp -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
--A PREROUTING ! -d 127.0.0.1/32 -p tcp -m mark ! --mark 1337/0xfff -j TPROXY --on-port $POD_INBOUND_PLAINTEXT --on-ip 127.0.0.1 --tproxy-mark 0x111/0xfff
--A OUTPUT -m connmark --mark 0x111/0xfff -j CONNMARK --restore-mark --nfmask 0xffffffff --ctmask 0xffffffff
+:ISTIO_OUTPUT - [0:0]
```
bleggett, line 17: Eventually, if we keep this around, I want to see if we can replace this with a direct invocation of `istio-iptables` so we don't have to keep these in sync.

howardjohn: I generally like that, though it is tricky since today all of our repos are a DAG in dependencies. This would introduce a cycle. Not impossible, certainly

src/proxy/outbound.rs
```diff
@@ -653,12 +648,6 @@ struct Request {
     upstream_sans: Vec<String>,
 }
 
-#[derive(Debug)]
-enum Direction {
```
bleggett, line 657: Dunno why we added this here, yeah.

bleggett (approved):

ericvn (approved):

PR #971, opened 2024-04-24
waypoint bypass detection and authz skip

stevenctl:
An alternative to https://github.com/istio/istio/pull/50475.

```
if has_waypoint && !from_waypoint {
  deny();
else if !has_waypoint {
  check_rbac();
}
```

We had a similar check in the past. One key difference in this version is that `has_waypoint` and `from_waypoint` checks both the destination workload and all it's services for waypoint attachment.

### DIFF
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -24,6 +24,8 @@ use crate::config::ProxyMode;
 use async_trait::async_trait;
 
 use prometheus_client::encoding::{EncodeLabelValue, LabelValueEncoder};
+use serde::de::Visitor;
+use serde::Deserializer;
 use tokio::sync::{mpsc, watch, Mutex};
 use tokio::time::{sleep_until, Duration, Instant};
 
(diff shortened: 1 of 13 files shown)


louiscryan:
I think we want the following

```javascript
if is_mesh_traffic {
  if has_waypoint { 
     if from_waypoint {
       allow();
     } else if src_can_skip_waypoint {  #skip waypoint API TBD
       check_rbac();
     } else {
       deny("must go through waypoint")
     }
  } else {
    check_rbac();
  }
} else {
  check_permissive();
} 
```

I've included a hypothetical future 'skip-waypoint' API which I think will be useful, particularly to enable something like ingress which may systematically _skip_ waypoints but for which we have not done design work. 

I think the above (or the original) make it clear that authz is _delegated_ to the waypoint when one is _used_ by a workload. This lines up conceptually with how the APIs and compositional patterns are defined so I think this is overall good and better than the 'defaultAllow' alternatives. Its also pretty similar to the code that was there before so its not that disruptive.

howardjohn:
@stevenctl this is obsolete now, right?

PR #972, opened 2024-04-24
admin: support setting log of a specific scope

howardjohn:


src/admin.rs
```diff
@@ -368,26 +368,39 @@ fn list_loggers() -> Response<Full<Bytes>> {
...
-                Err(e) => plaintext_response(
-                    hyper::StatusCode::BAD_REQUEST,
-                    format!("Failed to set new level: {}\n{}", e, HELP_STRING),
-                ),
+fn validate_log_level(level: &str) -> anyhow::Result<()> {
+    for clause in level.split(',') {
+        // We support 2 forms, compared to the underlying library
+        // <level>: supported, sets the default
+        // <scope>:<level>: supported, sets a scope's level
+        // <scope>: sets the scope to 'trace' level. NOT SUPPORTED.
```
bleggett, line 376: Where do we document the scopes?

howardjohn: We don't unfortunately, they are all basically the rust module name

bleggett: :+1:  Might be good to add an admin endpoint that dumps the available/registered scopes (if that's possible, I imagine it is) eventually.

bleggett (approved):

PR #979, opened 2024-04-25
Verify workload consistency with CNI for outbound

howardjohn:
We only did this for inbound. This double check provides a stronger
guarantee than IP address matching

### DIFF
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -454,6 +454,12 @@ impl OutboundConnection {
             Some(wl) => wl,
             None => return Err(Error::UnknownSource(downstream)),
         };
+        if let Some(ref wl_info) = self.pi.proxy_workload_info {
+            // make sure that the workload we fetched matches the workload info we got over ZDS.
+            if !wl_info.matches(&source_workload) {
+                return Err(Error::MismatchedSource(downstream, wl_info.clone()));
+            }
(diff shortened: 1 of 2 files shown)


src/proxy/outbound.rs
```diff
@@ -454,6 +454,12 @@ impl OutboundConnection {
             Some(wl) => wl,
             None => return Err(Error::UnknownSource(downstream)),
         };
+        if let Some(ref wl_info) = self.pi.proxy_workload_info {
+            // make sure that the workload we fetched matches the workload info we got over ZDS.
```
MorrisLaw, line 458: what's ZDS? is that different than xDS?

MorrisLaw: Nvm. Just discovered the zds.proto.

MorrisLaw (approved):

src/proxy/outbound.rs
```diff
@@ -454,6 +454,12 @@ impl OutboundConnection {
             Some(wl) => wl,
             None => return Err(Error::UnknownSource(downstream)),
         };
+        if let Some(ref wl_info) = self.pi.proxy_workload_info {
```
bleggett, line 457: nit: we need an actual unit test for this.

bleggett (approved):

PR #985, opened 2024-04-26
performance: reduce and test size of futures

howardjohn:
Prior to this PR, each outbound connection future was 22k. This is
really big.

The reason behind this is how Rust sizes Futures; the stack size is
allocated based on the maximum required size through any branch. This
means even obscure code paths can bloat up the entire future for
everyone.

This PR optimizes a bunch of these. This is done in a few ways:
* Box::pin expensive futures. While we have to allocate these when we
  hit them, of course, when we don't hit these codepaths they are now
"free". We have a lot of uncommon expensive code paths. Especially HBONE
vs passthrough; hbone is way more expensive. But also on demand XDS,
DNS, etc
* Arc the Config which is large
* A variety of other micro-optimizations

A new `assertions` module is added to verify we do not regress.

Spawning 10k connections before this PR used about 660Mb, now we are at 300mb

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -36,6 +36,7 @@ pub async fn build_with_cert(
     config: config::Config,
     cert_manager: Arc<SecretManager>,
 ) -> anyhow::Result<Bound> {
+    let config = Arc::new(config);
     // Start the data plane worker pool.
     let data_plane_pool = new_data_plane_pool(config.num_worker_threads);
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -33,10 +33,9 @@ use crate::{admin, config, metrics, proxy, readiness, signal};
 use crate::{dns, xds};
 
 pub async fn build_with_cert(
-    config: config::Config,
+    config: Arc<config::Config>,
     cert_manager: Arc<SecretManager>,
 ) -> anyhow::Result<Bound> {
-    let config = Arc::new(config);
     // Start the data plane worker pool.
     let data_plane_pool = new_data_plane_pool(config.num_worker_threads);
 
@@ -281,7 +280,7 @@ fn new_data_plane_pool(num_worker_threads: usize) -> mpsc::Sender<DataPlaneTask>
     tx
 }
 
-pub async fn build(config: config::Config) -> anyhow::Result<Bound> {
+pub async fn build(config: Arc<config::Config>) -> anyhow::Result<Bound> {
     let cert_manager = if config.fake_ca {
         mock_secret_manager()
     } else {
--- a/src/assertions.rs
+++ b/src/assertions.rs
@@ -1,0 +1,47 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Below helper functions are used to help make about the size of types.
+// There are some compile time ways to do this, but they don't work in the way we need for the most part;
+// analyzing the size of Futures which we don't have explicit declarations for.
+// Future size is determined by the max required stack size for the async function. This means deeply
+// branched code can create huge Future's, leading to high per-connection memory usage in ztunnel.
+// Debugging these usages can be done by `RUSTFLAGS=-Zprint-type-sizes cargo +nightly build -j 1`,
+// or by logging with the functions below.
+
+#[cfg(all(any(test, feature = "testing"), debug_assertions))]
+pub fn size_between_ref<T>(min: usize, max: usize, t: &T) {
+    let size = std::mem::size_of_val(t);
+    if size < min || size > max {
+        // If it is too small: that is good, we just want to update the assertion to be more aggressive
+        // If it is too big: that is bad. We may need to increase the limit, or consider refactors.
+        panic!(
+            "type {} size is unexpected, wanted {min}..{max}, got {size}",
+            std::any::type_name::<T>(),
+        )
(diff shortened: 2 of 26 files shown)


src/proxy/inbound_passthrough.rs
```diff
@@ -49,7 +49,11 @@ impl InboundPassthrough {
 
         let transparent = super::maybe_set_transparent(&pi, &listener)?;
         // Override with our explicitly configured setting
-        pi.cfg.enable_original_source = Some(transparent);
+        if pi.cfg.enable_original_source.is_none() {
+            let mut cfg = (*pi.cfg).clone();
+            cfg.enable_original_source = Some(transparent);
```
bleggett, line 54: IDK why we bother to mutate the primary config with this TBH - the override is only used here. We might as well keep config R/O and set a private field, and skip cloning the whole thing.

src/proxy/outbound.rs
```diff
@@ -93,25 +100,25 @@ impl Outbound {
...
-                                debug!(dur=?start_outbound_instant.elapsed(), id=%oc.id, "outbound spawn DONE");
-                            })
-                            .instrument(span),
-                        );
+                                _ = oc.proxy(stream) => {}
+                            }
+                            debug!(dur=?start_outbound_instant.elapsed(), id=%oc.id, "outbound spawn DONE");
+                        }).instrument(span);
+
+                        assertions::size_between_ref(1000, 1500, &serve_outbound_connection);
```
bleggett, line 120: ooh can we selectively enable size assertions via subcomponent logging level toggles?

howardjohn: Its not from logging levels, just conditional compiling when we are in test vs release mode

howardjohn: I think you can do conditional based on logs, but its a compile time thing so you cannot dynamically change

src/app.rs
```diff
@@ -36,6 +36,7 @@ pub async fn build_with_cert(
     config: config::Config,
     cert_manager: Arc<SecretManager>,
 ) -> anyhow::Result<Bound> {
+    let config = Arc::new(config);
```
bleggett, line 39: Can we just `Arc` wrap this in `construct_config`?

src/assertions.rs
```diff
@@ -0,0 +1,47 @@
...
+
+// Below helper functions are used to help make about the size of types.
+// There are some compile time ways to do this, but they don't work in the way we need for the most part;
+// analyzing the size of Futures which we don't have explicit declarations for.
+// Future size is determined by the max required stack size for the async function. This means deeply
+// branched code can create huge Future's, leading to high per-connection memory usage in ztunnel.
+// Debugging these usages can be done by `RUSTFLAGS=-Zprint-type-sizes cargo +nightly build -j 1`,
+// or by logging with the functions below.
+
+#[cfg(all(any(test, feature = "testing"), debug_assertions))]
```
bleggett, line 23: (I don't know why they would be but) `debug` doesn't affect the struct sizes, does it?

src/proxy/inbound.rs
```diff
@@ -67,7 +67,11 @@ impl Inbound {
             .map_err(|e| Error::Bind(pi.cfg.inbound_addr, e))?;
         let transparent = super::maybe_set_transparent(&pi, &listener)?;
         // Override with our explicitly configured setting
-        pi.cfg.enable_original_source = Some(transparent);
+        if pi.cfg.enable_original_source.is_none() {
+            let mut cfg = (*pi.cfg).clone();
```
bleggett, line 71: Rather than create a mutable clone of the whole config we might as well just hoist `let enable_original_source = pi.cfg.enable_original_source;`?

We don't need to put the value back in the config, AFAIK.

src/proxy/inbound.rs
```diff
@@ -302,30 +309,30 @@ impl Inbound {
...
-        let connection_metrics = ConnectionOpen {
-            reporter: Reporter::destination,
-            source,
-            derived_source: Some(derived_source),
-            destination: Some(upstream),
-            connection_security_policy: metrics::SecurityPolicy::mutual_tls,
-            destination_service: ds,
-        };
-        let result_tracker = metrics::ConnectionResult::new(
+        let result_tracker = Arc::new(metrics::ConnectionResult::new(
```
bleggett, line 312: Can we just construct + wrap the instantiated struct with `Arc` from inside `ConnectionResult::new`?

src/proxy/outbound.rs
```diff
@@ -78,11 +84,12 @@ impl Outbound {
         //
         // So use a drain to nuke tasks that might be stuck sending.
         let (sub_drain_signal, sub_drain) = drain::channel();
+        let pi = Arc::new(self.pi);
```
bleggett, line 87: See, this makes me wonder if ProxyInputs is just a heterogeneous bag of junk we should split up, or simply always wrap in an `Arc` (not just here).

src/assertions.rs
howardjohn: Shouldn't, this is just so it triggers in tests or if you run in not-release mode. I.E. anywhere a developer will run, but not a real user.

`test` might imply `debug_assertions` tbh

src/proxy/inbound.rs
howardjohn: yeah we use it somewhat all over though.. I was originally thinking that then went with this approach. But I hate it as well, I can change it up

src/proxy/inbound.rs
howardjohn: yep, will change

howardjohn: NVM, we use a Box in other codepaths but require atomic here. IMO its best to let the callers decide

src/app.rs
howardjohn: Yeah I moved it up pretty high. The earlier in the call stack we move it, the less flexibility usages have so I didn't remove the possibility to construct a raw Config without an arc

howardjohn:
Hyper bug is fixed, this is ready to go

src/proxy/inbound.rs
bleggett: This still feels suboptimal but it would probably take more refactoring to fix, and it's an improvement in some cases at least.

bleggett (approved):

PR #991, opened 2024-04-30
Make connection tracking resiliant to connection reuse

howardjohn:
We have a race where if I open a connection with the same 4 tuple too
quickly, we would occasionally fail to track. This is due to ordering:

* New connection register
* Old connection close
* New connection track -- fails

Now, we make sure we account (in count) for the new connection in
register so we don't have this race.

Hard to test in unit tests or even integration; I used netperf TCP_CRR
and its pretty easy to trigger.

### DIFF
--- a/src/proxy/connection_manager.rs
+++ b/src/proxy/connection_manager.rs
@@ -175,16 +175,15 @@ impl ConnectionManager {
             ctx: ctx.clone(),
             dest_service,
         };
-        self.register(&conn);
-        if !state.assert_rbac(ctx).await {
-            self.release(&conn);
-            return Err(Error::AuthorizationPolicyRejection);
-        }
-        let Some(watch) = self.track(&conn) else {
+        let Some(watch) = self.register(&conn) else {
             warn!("failed to track {conn:?}");
             debug_assert!(false, "failed to track {conn:?}");
             return Err(Error::AuthorizationPolicyRejection);
         };
+        if !state.assert_rbac(ctx).await {
+            self.release(&conn);
+            return Err(Error::AuthorizationPolicyRejection);
+        }
         Ok(ConnectionGuard {
             cm: self.clone(),
             conn,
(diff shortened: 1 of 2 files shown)


src/proxy/connection_manager.rs
```diff
@@ -208,13 +215,12 @@ impl ConnectionManager {
             .write()
```
bleggett, line 215: Do we even need `track` as a discrete call here, if it's only returning `Option<watch>`?

It's only ever called right after `register` (which now `tracks`) so `register` might as well return the `Option<watch>`?

howardjohn: I've thought about this a bit. I'm not sure. The difference is we do assert RBAC between register and track. So we would track and immediately drop if they fail RBAC. That's probably ok though

howardjohn:
/retest

unrelated

src/proxy/connection_manager.rs
bleggett: Yes, I think it comes out to the same thing from the perspective of the caller, and the internal state.

howardjohn: Ok merged them, I think its fine

src/proxy/connection_manager.rs
```diff
@@ -191,33 +194,35 @@ impl ConnectionManager {
...
+            Entry::Vacant(entry) => {
+                let drain = ConnectionDrain::new();
+                let rx = drain.rx.clone();
+                entry.insert(drain);
+                Some(rx)
+            }
         }
     }
 
+    // // get a channel to receive close on for your connection
```
bleggett, line 213: nit: remove commented code

bleggett (commented):
LGTM aside from the commented code.

src/proxy/connection_manager.rs
howardjohn: oops, done

bleggett (approved):

bleggett:
/test test

hmm this one might be a flake

PR #992, opened 2024-05-01
pool: do not indefinitely leak timers

howardjohn:
We are storing these in a list and never removing. We don't use them, so
no need to store.

### DIFF
--- a/src/proxy/pool.rs
+++ b/src/proxy/pool.rs
@@ -29,7 +29,6 @@ use std::sync::atomic::{AtomicI32, AtomicU16, Ordering};
 use std::sync::Arc;
 
 use tokio::sync::watch;
-use tokio::task;
 
 use tokio::sync::Mutex;
 use tracing::{debug, error, trace};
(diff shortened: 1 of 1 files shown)


howardjohn:
/retest 
known flake

src/proxy/pool.rs
```diff
@@ -64,7 +63,6 @@ struct PoolState {
     connected_pool: Arc<pingora_pool::ConnectionPool<ConnClient>>,
     // this must be an atomic/concurrent-safe list-of-locks, so we can lock per-key, not globally, and avoid holding up all conn attempts
     established_conn_writelock: flurry::HashMap<u64, Option<Arc<Mutex<()>>>>,
-    close_pollers: futures::stream::FuturesUnordered<task::JoinHandle<()>>,
```
bleggett, line 67: Ah yeah this does nothing ATM.

Was originally `await`-ing all the join handles in `impl Drop for PoolState` after signaling to make sure all conns evict, but I don't think that's strictly necessary.

bleggett (approved):

PR #997, opened 2024-05-01
h2: poll ready before using

howardjohn:
This is "required" per the docs. I don't think it really is, but good to
be safe and follow the docs

### DIFF
--- a/src/proxy/h2_client.rs
+++ b/src/proxy/h2_client.rs
@@ -258,6 +258,9 @@ impl H2ConnectClient {
         &mut self,
         req: Request<()>,
     ) -> Result<(SendStream<SendBuf>, h2::RecvStream), Error> {
+        // "This function must return `Ready` before `send_request` is called"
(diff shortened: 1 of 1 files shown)


bleggett (approved):

src/proxy/h2_client.rs
```diff
@@ -258,6 +258,9 @@ impl H2ConnectClient {
         &mut self,
         req: Request<()>,
     ) -> Result<(SendStream<SendBuf>, h2::RecvStream), Error> {
+        // "This function must return `Ready` before `send_request` is called"
+        // We should always be ready though, because we make sure we don't go over the max stream limit out of band.
```
bleggett, line 262: Yeah I saw this in the docs at some point but didn't check if we were doing this in the DIY client.

Better to be correct.

howardjohn:
/retest

flake is fixed in another pr

howardjohn:
/retest
 https://github.com/istio/ztunnel/pull/998

PR #998, opened 2024-05-01
deps: use h2 with race condition fix

howardjohn:
Pulls in https://github.com/hyperium/h2/pull/772. We might want to wait
on this, not sure how fast it will go


bleggett (approved):
We'll get flakes until the fix is merged, and the fix a single (relatively small) diff against upstream.

I'm ok with it

bleggett:
xref with https://github.com/istio/ztunnel/issues/1000

PR #1002, opened 2024-05-02
Miscellanuous optimizations to reduce memory footprint

howardjohn:
For the test workload I was running this shaves off roughly 100mb of allocations of 1.3gb (this is total, not active). Active memory dropped about 60mb->50mb. 

Each optimization is in its own commit.

Reference:

Service: 312bytes
DemandProxyState: 272bytes
Workload: 608 bytes
ProxyInputs: 344 bytes

----

* Move some lookups on the hotpath to stay in an arc to make them ~free
* Use references to DemandProxyState; full ownership is not needed
* Avoid double lookup when not using on demand; our logic was `lookup; fetch; lookup`. If on demand is off, fetch does nothing, so its silly to look up twice.

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -355,7 +355,7 @@ impl Store {
                     // Didn't find a service, try a workload.
                     if let Some(wl) = state.workloads.find_hostname(&search_name_str) {
                         return Some(ServerMatch {
-                            server: Address::Workload(Box::new(wl)),
+                            server: Address::Workload(wl),
                             name: search_name,
                             alias,
                         });
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -347,7 +347,7 @@ impl Store {
                         .unwrap();
 
                     return Some(ServerMatch {
-                        server: Address::Service(Box::new(service)),
+                        server: Address::Service(Arc::new(service)),
                         name: search_name,
                         alias,
                     });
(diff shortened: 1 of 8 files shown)


src/dns/server.rs
```diff
@@ -347,15 +347,15 @@ impl Store {
                         .unwrap();
 
                     return Some(ServerMatch {
-                        server: Address::Service(Box::new(service)),
+                        server: Address::Service(Arc::new(service)),
```
bleggett, line 350: In general for most of these heavy types we should probably just implement them with constructors that prewrap them, or over time we will probably end up with inconsistent usage.

src/proxy/inbound.rs
```diff
@@ -181,8 +184,7 @@ impl Inbound {
                     }
                 }
             };
-            // This is pretty obscene right now. Fortunately with pooling this is less problematic than outbound.
-            assertions::size_between_ref(10_000, 12_000, &serve_client);
+            assertions::size_between_ref(1500, 2500, &serve_client);
```
bleggett, line 187: Much better

src/dns/server.rs
howardjohn: We actually are storing them as Arc and then clone them in the state.rs. So most of this is stoppnig to do that. I didn't tackle every codepath yet but presumably we could

bleggett (approved):

Issue #988, opened 2024-04-29
String optimizations for xDS


howardjohn:
Ztunnel has two primary sources of memory usage:
1. data plane traffic, largely buffers we hold in the IO flow
2. Data passed over xds originally

There is essentially no other sources. While we store things for lookup, metrics, etc -- this data is all _from_ XDS.

This largely is metadata. Strings like `default`, etc are copied likely 1000s of times. We can explore ways to optimize these by string interning and/or optimized string representations.

Quick numbers:
* `String` 24 bytes (ptr, len, cap)
* `Arc<str>` 16 bytes

## Interning

Interning allows us to basically store all of our unique strings in one place, and then reference them. There are a variety of options here.

Requirements:
1) We must have a way to bound growth (cleanup strings on last reference)
2) We need to be able to intern all strings from xDS
3) Interned strings should be ergonomic to use

Interning crates vary by whether they support multiple threads and if they require globals or locals.
In our case, the string _creation_ is all single threaded in XDS.
However, it needs to be `Send` (because we are in async).
Additionally, the usage and `Drop`ing of strings is done in separate threads.

For cleanup, we can:
1. return a type with a custom `Drop` implementation that removes things. This requires multithreading on the intern backing map (b/c now we are dropping from a different thread than creation); this could _maybe_ be hyper-optimized to have drops queue up in some data structure and periodically flushed, but ultimately its still multithreaded. If we do not use a global store, we also need to reference to the map to cleanup, which adds 8 bytes
2. Periodically cleanup. We know that things will pretty much only be dropped when an xds removal happens, so we could do cleanup then. This probably isn't 100% robust though, since somethings will hold on slightly past the removal, so we would need a timer as well. This would be pretty solid, but scanning the entire set may be expensive and lead to issues like "Every 5min my cpu spikes".

Some libraries return u64 (or even u32) identifiers that can be used to lookup a str. This is extreme efficient since we use only 4/8 bytes. However, this makes usage slower (need a lookup, which needs a lock) and possible less ergonomic (depending on implementation). This also only really makes sense with a global table, else we are spending half our bytes pointing back to the table.

## String type

`String` is a bad fit for us; we do lots of cloning and no mutability. String has mutability (which we pay for in size) and slow clones. `Arc<str>` is 8 bytes less (16bytes) and has ~free clones; we could get a lot of wins there.

[...]

PR #1016, opened 2024-05-03
performance: optimized string and various other XDS improvements

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/988

This PR introduces a new string representation. See issue above for some more background.

I initially implemented this using `internment::ArcIntern<str>`. This showed improvements over master, but I compared to using ArcStr and found actually better memory utilization with ArcStr, and much less complexity and risk. Conveniently, due to the `strng` module abstracting things, the swap was pretty much trivial (just a few small changes due to different conversion helpers in the libraries). So if we do want to switch things in the future, it shouldn't be too bad.

This optimizes the XDS path (and, presumably, others -- but I only measured XDS much).

My test covered 30k pods with 2k services. I loaded this up and checked memory usage (total and peak).

Before: 1.8GB total, 700mb retained
After: 1.1GB total, 520mb retained

Roughly 100mb of improvement came from switching the string representation, and 80mb came from other optimizations. Not confident in that breakdown as I was going back and forth moving things over + optimizing. 

All optimizations here are guided by profiling. End profile (github

Not much low hanging fruit left, most of the memory is in our hashmaps.

Optimizations aside from string changes:
* Optimizing string concat in hotpaths
* Avoid buffering up a vector of all XDS updates; stream it
* Avoid buffering up a copy of all endpoints; just iterate it
* Avoid a clone on the XDS workload

### DIFF
--- a/src/strng.rs
+++ b/src/strng.rs
@@ -1,9 +1,8 @@
 use std::fmt::Error;
 use std::ops::Deref;
-use std::rc::Rc;
-use std::sync::Arc;
+
 use arcstr::ArcStr;
-use prometheus_client::encoding::{EncodeLabelKey, LabelSetEncoder, LabelValueEncoder};
+use prometheus_client::encoding::LabelValueEncoder;
 
 pub type Strng = ArcStr;
 
@@ -33,7 +32,9 @@ impl Deref for RichStrng {
 }
 
 impl<T> From<T> for RichStrng
-    where T: Into<Strng>{
+where
+    T: Into<Strng>,
+{
     fn from(value: T) -> Self {
         RichStrng(value.into())
     }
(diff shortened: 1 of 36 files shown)


bleggett:
```
Before: 1.8GB total, 700mb retained
After: 1.1GB total, 520mb retained
```

that's more than I thought we'd save, nice.

src/strng.rs
```diff
@@ -0,0 +1,92 @@
...
+use arcstr::ArcStr;
+use prometheus_client::encoding::LabelValueEncoder;
+
+/// 'Strng' provides a string type that has better properties for our use case:
+/// * Cheap cloning (ref counting)
+/// * Efficient storage (8 bytes vs 24 bytes)
+/// * Immutable
+/// This is mostly provided by a library, ArcStr, we just provide a very thin wrapper around it
+/// for some flexibility.
+pub type Strng = ArcStr;
```
bleggett, line 27: Strictly speaking we don't need this type alias, we could just have ArcStr + RichStrng(ArcStr)?

That's a little more obvious and it's also easy to look at a type named `ArcStr` and know what it does.

But that's mostly a stylistic nit I guess, and if we ever want to use a different ArcStr impl, this insulates from that a bit.

bleggett (approved):
relatively easy win

src/strng.rs
howardjohn: I was on the fence about the `Strng` thing. It was more useful when doing the interning library since their type name was unwieldy

PR #1018, opened 2024-05-03
misc: remove Default identity from prod codepath

howardjohn:
We didn't use it anywhere which is good -- but it was a risk


bleggett (approved):
:+1:

PR #1021, opened 2024-05-03
Fix cleaning up identities

howardjohn:
This fixes a bug where identities were being dropped wrongly. Basically we didnt track the identity properly, so ANY pod removal caused all ztunnels to forget that identity. This wasnt detected because ztunnel will on-demand fetch certificates its missing.

### DIFF
--- a/src/state/workload.rs
+++ b/src/state/workload.rs
@@ -597,12 +597,12 @@ pub struct WorkloadStore {
     by_uid: HashMap<Strng, Arc<Workload>>,
     /// byHostname maps workload hostname to workloads.
     by_hostname: HashMap<Strng, Arc<Workload>>,
-    // Identity->Set of UIDs
+    // Identity->Set of UIDs. Only stores local nodes
     by_identity: HashMap<Identity, HashSet<Strng>>,
 }
 
 impl WorkloadStore {
-    pub fn insert(&mut self, w: Arc<Workload>) {
+    pub fn insert(&mut self, w: Arc<Workload>, track_identity: bool) {
         // First, remove the entry entirely to make sure things are cleaned up properly.
         self.remove(&w.uid);
 
@@ -614,6 +614,13 @@ impl WorkloadStore {
             self.by_hostname.insert(w.hostname.clone(), w.clone());
         }
         self.by_uid.insert(w.uid.clone(), w.clone());
+        // Only track local nodes to avoid overhead
+        if track_identity {
+            self.by_identity
+                .entry(w.identity())
+                .or_default()
+                .insert(w.uid.clone());
+        }
     }
 
     pub fn remove(&mut self, uid: &Strng) -> Option<Workload> {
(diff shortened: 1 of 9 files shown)


src/state/workload.rs
```diff
@@ -614,6 +614,13 @@ impl WorkloadStore {
             self.by_hostname.insert(w.hostname.clone(), w.clone());
         }
         self.by_uid.insert(w.uid.clone(), w.clone());
+        // Only track local nodes to avoid overhead
+        if track_identity {
+            self.by_identity
+                .entry(w.identity())
+                .or_default()
+                .insert(w.uid.clone());
+        }
```
ilrudie, line 623: I think this comment describes the intended use for track_identity rather than what this code does.  Should we consider adding this comment to the insert method itself and/or changing `track_identity` to be `is_node_local` perhaps?

tests/namespaced.rs
```diff
@@ -713,6 +714,78 @@ mod namespaced {
...
+            "removing local node shouldn't impact anything if I still have some running",
+        )
+        .await;
+        manager.delete_workload("id1-a-same-node").await?;
+        // TODO: this should be vec![], but our testing setup doesn't exercise the real codepath
+        check(
+            vec![id1s.clone()],
+            "removing final workload should clear things out",
+        )
+        .await;
```
ilrudie, line 785: Is this something we're looking to address before merging? Is it something covered elsewhere so we're not that worried?

Kind of odd to assert a final state that isn't actually the state it should be

ilrudie (commented):
Over all this looks good. I understand there's some limitations to the testing setup but asserting a state we don't actually want seems odd to me.

tests/namespaced.rs
howardjohn: I would love to before merging though it will take a lot of work I may not have time to in the timelines this would be good to merge in (1.22.0). I think asserting the final state in the meantime is better than no assertion so if/when we fix it, it is remembered

ilrudie: yeah, odd but probably the least bad option I suppose

src/state/workload.rs
bleggett: `should_track_certificates_for_removal()` probably belongs in `workload` and should be checked as part of `insert`, and not in `cert_mgr` - it's being called before `workload.insert` pretty much everywhere.

howardjohn: The reason I did it this way is workload doesn't have access to the proxy mode, local node, etc at all, so it cannot make the call itself. We could plumb it down, but then its inconsistent with the `should_prefetch_certificate` and requires even more plumbing around

bleggett: > The reason I did it this way is workload doesn't have access to the proxy mode, local node, etc at all,

Figured.

> We could plumb it down, but then its inconsistent with the should_prefetch_certificate and requires even more plumbing around

`should_prefetch` and `should_track` are almost the same check and it feels like _both_ should probably go in `workload` but yeah that's getting OOS for this.

bleggett (approved):

bleggett:
`should_prefix` and `should_track` are mostly the same check and both probably belong as workload properties, but will do for 1.22

PR #1023, opened 2024-05-06
logging: ignore duplicate fields from direct `log` usage

howardjohn:
This is copied from `tracing` directly

Otherwise we get stuff like `pprof::profiler stopping cpu profiler   log.target="pprof::profiler" log.module_path="pprof::profiler" log.file="/home/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pprof-0.13.0/src/profiler.rs" log.line=439`

### DIFF
--- a/src/telemetry.rs
+++ b/src/telemetry.rs
@@ -164,6 +164,8 @@ impl field::Visit for Visitor<'_> {
 
     fn record_debug(&mut self, field: &field::Field, val: &dyn std::fmt::Debug) {
         self.res = match field.name() {
+            // Skip fields that are actually log metadata that have already been handled
+            name if name.starts_with("log.") => Ok(()),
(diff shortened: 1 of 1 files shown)


ilrudie:
so the log entry will now be simply `pprof::profiler stopping cpu profiler`?

howardjohn:
@ilrudie exactly. This mirrors how it would show up if the crate used `tracing`, but it happens to use `log` (as do a few others like rustls)

ilrudie (approved):
lgtm

ilrudie:
> @ilrudie exactly. This mirrors how it would show up if the crate used `tracing`, but it happens to use `log` (as do a few others like rustls)

Thanks for confirming

bleggett (approved):

PR #1024, opened 2024-05-06
performance: make copy buffer sizes dynamic

howardjohn:
Today, we give each buffer 16k. There are 3 per con. This leads to a lot
of per connection memory usage.

This PR changes each buffer to start with only 1k. This dynamically
scales up.

Currently, the algorithm to scale up is very simple: once we have sent
128k total on the connection, we resize the buffer. This very likely
could be improved.

On a test load opening 10k connections:
Before: 90mb
After: 38mb

### DIFF
--- a/src/copy.rs
+++ b/src/copy.rs
@@ -1,0 +1,263 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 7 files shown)


src/copy.rs
```diff
@@ -0,0 +1,263 @@
...
+        }
+        Poll::Ready(Ok(&me.buf[*me.pos..*me.cap]))
+    }
+
+    fn consume(self: Pin<&mut Self>, amt: usize) {
+        let me = self.project();
+        *me.pos = cmp::min(*me.pos + amt, *me.cap);
+    }
+
+    fn resize(self: Pin<&mut Self>) {
```
bleggett, line 222: Do we have any info on how expensive resizes are in practice?
Is it something we could measure? There's clearly a tradeoff but the buffer usage is only part of it.

howardjohn: Cost to do 1 malloc (16k) + 1 memcopy (up to 1k). Its basically not noticeable -- its slow to do lots of memcpy, but 1 per connection is pretty tiny.

I did actually try to measure it by seeing if I saw a spike in latency when the resize kicked in; I wasn't able to detect one which gave me the impression it is negligible in the grand scheme of the request flow

bleggett (approved):

PR #1029, opened 2024-05-08
send HBONE address in PROXY

stevenctl:
Not sure how this was working in https://gist.github.com/stevenctl/ec2ddae79734f8f9367400961ca9c48f and other tests. This change is tested e2e with envoy that actually has filter chain matches on IPs. 

Cherry-pick to 1.22 would be nice, but I understand it might be a bit late.


howardjohn (commented):
might be good to add the PROXY case into namespaced.rs. We have one for non-PROXY sandwich already so probably isn't a ton of work

howardjohn (approved):
LGTM, I put a hold if you have time to add a test

howardjohn (approved):
nice!

PR #1030, opened 2024-05-08
Bump rust version to 1.78

howardjohn:
This matches the new docker image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "ztunnel"
 version = "0.0.0"
 edition = "2021"
-rust-version = "1.77"
+rust-version = "1.78"
 
 [features]
 default = ["tls-ring"]


howardjohn:
maybe we shouldn't dothis since we don't require 1.78, and this will make it fail on 1.77?

bleggett:
Yeah I think we need to track our CI version/`build-tools` version exactly in the cargo.toml

howardjohn:
rust-version is the 'The minimal supported Rust version.' We run fine on 1.77, do we really want to say you MUST have 1.78?

Build tools is fine since that is managed. I don't personally care since I always use `latest`. But seems a bit excessive

bleggett:
> rust-version is the 'The minimal supported Rust version.' We run fine on 1.77, do we really want to say you MUST have 1.78?

I think our min version can only be the min version we actually build and test with.

People are free to try and change it and set it lower, but our floor sort of has to be what we use in CI.

PR #1031, opened 2024-05-08
combine `proxymode = shared` and `inpod_mode = true`

bleggett:
We currently have `ProxyMode::Dedicated | Shared` and `InpodEnabled: true|false` and I can't think of any scenarios where those are not basically indicating the same thing.

We also sort of don't use `ProxyMode` at all, and I think we should - the tenancy model is more important to express than how we do the tenancy model (currently we have only one multitenant model anyway, so why make it optional?).

### DIFF
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -60,13 +60,13 @@ impl Outbound {
             transparent,
             "listener established",
         );
-        let inpod = pi.cfg.inpod_enabled;
+        let mode = pi.cfg.proxy_mode;
         Ok(Outbound {
             pi,
             listener,
             drain,
             // Do not need to spoof with inpod mode for outbound
-            enable_orig_src: transparent && !inpod,
+            enable_orig_src: transparent && mode != ProxyMode::Shared,
         })
     }
 
@@ -167,7 +167,6 @@ impl OutboundConnection {
         // For in-pod, this isn't an issue and is useful: this allows things like prometheus scraping ztunnel.
         if self.pi.cfg.proxy_mode == ProxyMode::Shared
             && Some(dest_addr.ip()) == self.pi.cfg.local_ip
-            && !self.pi.cfg.inpod_enabled
         {
             metrics::log_early_deny(source_addr, dest_addr, Reporter::source, Error::SelfCall);
             return;
@@ -268,7 +267,7 @@ impl OutboundConnection {
     ) -> Result<(), Error> {
         // Create a TCP connection to upstream
         // We do not need spoofing for inbound
-        let local = if self.enable_orig_src && !self.pi.cfg.inpod_enabled {
+        let local = if self.enable_orig_src && self.pi.cfg.proxy_mode != ProxyMode::Shared {
             super::get_original_src_from_stream(&stream)
         } else {
             None
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -165,7 +165,8 @@ impl OutboundConnection {
 
         // Block calls to ztunnel directly, unless we are in "in-pod".
         // For in-pod, this isn't an issue and is useful: this allows things like prometheus scraping ztunnel.
-        if self.pi.cfg.proxy_mode == ProxyMode::Shared
+        // TODO is this actually wrong for ProxyMode::Dedicated? Seems like a valid case
+        if self.pi.cfg.proxy_mode != ProxyMode::Shared
             && Some(dest_addr.ip()) == self.pi.cfg.local_ip
         {
             metrics::log_early_deny(source_addr, dest_addr, Reporter::source, Error::SelfCall);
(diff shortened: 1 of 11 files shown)


bleggett:
slapping a hold on until @keithmattix @jaellio @justinpettit are OK with this.

howardjohn (commented):
Hmm, how does direct.rs work still?

This seems to break local development flows which I didn't realize would happen when initially discussing

bleggett:
> Hmm, how does direct.rs work still?
> 
> This seems to break local development flows which I didn't realize would happen when initially discussing

By setting `proxy_mode: config::ProxyMode::Dedicated` in the test config it uses.

howardjohn:
Doesn't dedicated not allow multiple clients with unique source though? do we just not test that

bleggett:
> Doesn't dedicated not allow multiple clients with unique source though? do we just not test that

I don't believe we test that. Probably because we have too many flags that influence how tenancy works.

That's kind of what I'm trying to say with this PR, we really only have two tenancy modes as far as I can see - multi tenant and single tenant.

Multitenant is only supported, tested, exercised and fully functional with inpod.

And if we don't test it and we don't use it, I don't know why we carry it.

howardjohn:
> And if we don't test it and we don't use it, I don't know why we carry it/am waiting for an argument as to why we should carry (and thus test) it, or not.

Well we do test it, this PR removed the tests :slightly_smiling_face:. This currently breaks how develop and ~exclusively run ztunnel. Mind if I take a bit to poke around and see if there are viable alternatives with this change to develop locally?

bleggett:
> Well we do test it, this PR removed the tests 🙂. This currently breaks how develop and ~exclusively run ztunnel. Mind if I take a bit to poke around and see if there are viable alternatives with this change to develop locally?

Fair, we _lightly_ test it in one spot, and don't support the mode. Yep, please - I don't develop with standalone local ztunnel so I'm not the best testbed for that.

howardjohn:
ok I think my dev ex is solved easily: `PROXY_MODE=dedicated`

bleggett:
> ok I think my dev ex is solved easily: `PROXY_MODE=dedicated`

Oh good - this one?

> Doesn't dedicated not allow multiple clients with unique source though? do we just not test that

howardjohn:
It does currently allow multiple sources which I think is broken. This PR would enable us to move towards a better architecture IMO, where a Proxy is never instantiated without a `source_workload: <...>`. This could be a WorkloadDescription like we have today, but could even be a full Workload (wait over xds) and/or full certificate as well.

We might not want that to block startup though, so maybe its something like a `Lazy<(Workload, Certificate)>` where we start it when we first launch, and then block first request on it until it loads.

For local dev, can just run ztunnel twice if need to test different sources

tests/namespaced.rs
```diff
@@ -114,39 +114,6 @@ mod namespaced {
         .await
     }
 
-    // Intentionally, we do not have a 'local_captured_sharednode'
-    // This is not currently supported since https://github.com/istio/ztunnel/commit/12d154cceb1d20eb1f11ae43c2310e66e93c7120
-
-    #[tokio::test]
-    async fn server_uncaptured_sharednode() -> anyhow::Result<()> {
```
howardjohn, line 121: should we make this test dedicated mode?

bleggett:
@howardjohn should be g2g

tests/namespaced.rs
```diff
@@ -666,34 +674,9 @@ mod namespaced {
         .await
     }
 
-    #[tokio::test]
-    async fn malicious_calls_sharednode() -> anyhow::Result<()> {
```
howardjohn, line 670: nit: might be a good test for Dedicated mode still? Not quite sure.

src/proxy/outbound.rs
```diff
@@ -214,9 +214,9 @@ impl OutboundConnection {
 
         // Block calls to ztunnel directly, unless we are in "in-pod".
         // For in-pod, this isn't an issue and is useful: this allows things like prometheus scraping ztunnel.
-        if self.pi.cfg.proxy_mode == ProxyMode::Shared
+        // TODO is this actually wrong for ProxyMode::Dedicated? Seems like a valid case
```
howardjohn, line 217: Dedicated now implies ztunnel IP == app IP, so I don't think this is relevant anymore (?)

howardjohn (approved):

tests/namespaced.rs
bleggett: I don't think so - https://github.com/istio/ztunnel/pull/1031#discussion_r1659425358 was the only place where we had a "malicious calls" rule for Dedicated, and we don't need that - so there aren't any Dedicated-specific rules to test anymore.

Given the many different ways in which `ztunnel` _could_ be deployed and used in `Dedicated` mode, I dunno if it even makes sense to have any restrictions in that mode, since for all of them I can imagine valid exceptions.

src/proxy/outbound.rs
bleggett: https://github.com/istio/ztunnel/pull/1197

PR #1035, opened 2024-05-09
adopt linting rules for rustc 1.78

howardjohn:
Well, turn one off.. I think its a bad rule and upstream may revert it


bleggett (approved):

bleggett:
Yeah - it's a potential perf benefit if you impl `clone_from()` yourself, I guess. But we don't, nobody really does, we can still do that on a case-by-case basis if we want to, and if you're worried about micro-optimizing `clone`, maybe don't `clone`?

PR #1040, opened 2024-05-10
Cleanup `illegal_ports` and ARC all of PI

bleggett:
This is mostly just pushing some minor local cleanups.

- `Arc` all of ProxyInputs. We can probably de-arc some of the bits inside so they aren't double-arc'd eventually, but handlers shouldn't be able to get mut-ref to anything in PI - if they need to, it doesn't belong in PI.

- Put `illegal_ports` in config - it is effectively config anyway - all the values are sourced from config, and everywhere we use `illegal_ports` we already have config.

General goal is to make sure handlers keep a clean distinction between runtime-mutable state (their inner state/explicit mutable args) and state that should _not_ be runtime-mutable (static config, proxyinputs)

We have tests that (ab)use this lack of immutability which this also twiddles a bit, mostly just enough to prevent us from writing more tests like that.

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -185,32 +185,34 @@ impl Proxy {
         let mut tasks = vec![
             tokio::spawn(self.inbound_passthrough.run().in_current_span()),
             tokio::spawn(self.policy_watcher.run().in_current_span()),
+            tokio::spawn(self.inbound.run().in_current_span()),
+            tokio::spawn(self.outbound.run().in_current_span()),
         ];
 
-        // TODO hbone_port is ALWAYS (by necessity) fixed in read-only config -
-        // except in (some) contrived integ test cases (direct), where we bind to random non-well-known ports,
-        // and cannot rely on a well-known port - which is why this exists, so we can ignore
-        // the config object and sidechannel the non-fixed runtime-bound port.
-        #[cfg(any(test, feature = "testing"))]
-        {
-            let hbone_port = self.inbound.address().port();
-            tasks.push(tokio::spawn(
-                self.outbound.run(hbone_port).in_current_span(),
-            ));
-            tasks.push(tokio::spawn(self.inbound.run().in_current_span()));
-            if let Some(socks5) = self.socks5 {
-                tasks.push(tokio::spawn(socks5.run(hbone_port).in_current_span()));
-            };
-        }
+        // // TODO hbone_port is ALWAYS (by necessity) fixed in read-only config -
+        // // except in (some) contrived integ test cases (direct), where we bind to random non-well-known ports,
+        // // and cannot rely on a well-known port - which is why this exists, so we can ignore
+        // // the config object and sidechannel the non-fixed runtime-bound port.
+        // #[cfg(any(test, feature = "testing"))]
+        // {
+        //     let hbone_port = self.inbound.address().port();
+        //     tasks.push(tokio::spawn(
+        //         self.outbound.run(hbone_port).in_current_span(),
+        //     ));
+        //     tasks.push(tokio::spawn(self.inbound.run().in_current_span()));
+        //     if let Some(socks5) = self.socks5 {
+        //         tasks.push(tokio::spawn(socks5.run(hbone_port).in_current_span()));
+        //     };
+        // }
 
-        #[cfg(not(any(test, feature = "testing")))]
-        {
-            tasks.push(tokio::spawn(self.inbound.run().in_current_span()));
-            tasks.push(tokio::spawn(self.outbound.run().in_current_span()));
+        // #[cfg(not(any(test, feature = "testing")))]
+        // {
+            // tasks.push(tokio::spawn(self.inbound.run().in_current_span()));
+            // tasks.push(tokio::spawn(self.outbound.run().in_current_span()));
             if let Some(socks5) = self.socks5 {
                 tasks.push(tokio::spawn(socks5.run().in_current_span()));
             };
-        }
+        // }
 
         futures::future::join_all(tasks).await;
     }
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -188,31 +188,9 @@ impl Proxy {
             tokio::spawn(self.inbound.run().in_current_span()),
             tokio::spawn(self.outbound.run().in_current_span()),
         ];
-
-        // // TODO hbone_port is ALWAYS (by necessity) fixed in read-only config -
-        // // except in (some) contrived integ test cases (direct), where we bind to random non-well-known ports,
-        // // and cannot rely on a well-known port - which is why this exists, so we can ignore
-        // // the config object and sidechannel the non-fixed runtime-bound port.
-        // #[cfg(any(test, feature = "testing"))]
-        // {
-        //     let hbone_port = self.inbound.address().port();
-        //     tasks.push(tokio::spawn(
-        //         self.outbound.run(hbone_port).in_current_span(),
-        //     ));
-        //     tasks.push(tokio::spawn(self.inbound.run().in_current_span()));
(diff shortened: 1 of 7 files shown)


howardjohn (commented):
How can this work in tests where we use dynamic ports?

howardjohn:
ah we do the test in `namespaced` which doesn't use dynamic ports

howardjohn:
@bleggett in `direct` we do, we just don't care about the illegal ports there

bleggett:
> @bleggett in `direct` we do, we just don't care about the illegal ports there

Yeah I'm taking a look at those.

bleggett:
> > @bleggett in `direct` we do, we just don't care about the illegal ports there
> 
> Yeah I'm taking a look at those.

What it boils down to is we break normal/prod behavior just for the purpose of some integration tests, and we have to add punchthrus to support providing a latebound HBONE port, when in every case outside of those few tests doing so would be a bug/a bad thing.

Proper solution is to move everything in `direct` to `namespaced` and drop the random HBONE port assignment to avoid port conflicts without relying on test-only structural hacks, but for now I've just prevented the use of said hack outside test code.

src/proxy.rs
```diff
@@ -145,68 +141,69 @@ impl Proxy {
...
             tokio::spawn(self.policy_watcher.run().in_current_span()),
         ];
-        if let Some(socks5) = self.socks5 {
-            tasks.push(tokio::spawn(socks5.run().in_current_span()));
+
+        // TODO hbone_port is ALWAYS (by necessity) fixed in read-only config -
+        // except in (some) contrived integ test cases (direct), where we bind to random non-well-known ports,
+        // and cannot rely on a well-known port - which is why this exists, so we can ignore
+        // the config object and sidechannel the non-fixed runtime-bound port.
+        #[cfg(any(test, feature = "testing"))]
```
howardjohn, line 188: Adding these conditional paths is risky - we effectively are now never testing the production codepath.

Why do we even need to split it? the test path will work for prod, no?

if we really want to verify we could just make an assertion conditional compiled in that `hbone_port == 15008` or whatever want to check?

src/proxy/inbound.rs
```diff
@@ -52,22 +50,24 @@ use crate::tls::TlsError;
...
             .tcp_bind(pi.cfg.inbound_addr)
             .map_err(|e| Error::Bind(pi.cfg.inbound_addr, e))?;
         let transparent = super::maybe_set_transparent(&pi, &listener)?;
         // Override with our explicitly configured setting
-        if pi.cfg.enable_original_source.is_none() {
-            let mut cfg = (*pi.cfg).clone();
-            cfg.enable_original_source = Some(transparent);
-            pi.cfg = Arc::new(cfg);
-        }
+        let enable_orig_src = if pi.cfg.enable_original_source.is_none() {
```
howardjohn, line 65: nit: `pi.cfg.enable_original_source.unwrap_or_else(transparent)`

src/proxy.rs
bleggett: We don't test the production codepath now - the production codepath can never use a dynamic port override, and we were making large chunks of (critical) state mutable just to enable this. Only tests can make use of it.

To be fair, I don't like the conditionals either though, and may just rewrite the tests to stop using dynamic port overrides.

Creating a runtime source of truth just to inject a duplicative static config value into it (but only for test flows) is both confusing to anyone that doesn't know why we are doing it, and a smell.

howardjohn: We do test both today because we have code running dynamic and static ports (direct and namespaced). With this PR, the tests doesn't even compile the production codepath.

bleggett: > With this PR, the tests doesn't even compile the production codepath.

Right, and those tests still pass, because they were never testing that particular production codepath to begin with.

The only tests that have ever hit this particular codepath relied on doing stuff you can't/shouldn't do in production. It's a catch 22. I'll move the `direct` tests to `namespaced` and then we should be able to drop the test-only assumptions.

bleggett: Namespacing every test is expensive, so I picked something that should be a decent compromise.

src/test_helpers.rs
```diff
@@ -101,14 +116,15 @@ pub fn test_config_with_port_xds_addr_and_root_cert(
...
-        inbound_addr: SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), 0),
+        inbound_addr: SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), hbone_port),
         socks5_addr: Some(SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 0)),
         admin_addr: SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 0),
         readiness_addr: SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 0),
         stats_addr: SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 0),
         outbound_addr: SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 0),
         inbound_plaintext_addr: SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 0),
         dns_proxy_addr: SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 0),
+        illegal_ports: HashSet::new(), // for "direct" tests, since the ports are latebound, we can't test illegal ports
```
howardjohn, line 127: If we are not setting this, do we need to fix inbound_addr to a non-zero port?

The current approach will flake if you run multiple tests concurrently like `cargo-nextest` or others (I run it to detect flakes, etc)

howardjohn: Ah we need the hbone port

howardjohn: How bad is it to pass the hbone port still but keep the illegal_ports?

howardjohn: Sorry I feel a bit like https://xkcd.com/1172/...

bleggett: Still gross, but gross in one signposted spot. Should do for `nextest` and friends. Does for me.

src/proxy.rs
```diff
@@ -151,69 +147,63 @@ impl Proxy {
...
+    pub(super) async fn from_inputs(mut pi: Arc<ProxyInputs>, drain: Watch) -> Result<Self, Error> {
         // We setup all the listeners first so we can capture any errors that should block startup
         let inbound = Inbound::new(pi.clone(), drain.clone()).await?;
-        pi.hbone_port = inbound.address().port();
-        //  HBONE doesn't have redirection, so we cannot have loops, but this would allow multiple layers of HBONE.
-        // This might be desirable in the future, but for now just ban it.
-        illegal_ports.insert(inbound.address().port());
+
+        // This exists for `direct` integ tests, no other reason
+        if pi.cfg.fake_self_inbound {
```
howardjohn, line 166: nit: could `#[cfg(any(test, feature = "testing"))]` it. Do we want to use tracing instead of println?

howardjohn (approved):

src/proxy.rs
bleggett: > could #[cfg(any(test, feature = "testing"))]

Yeah I considered that. In this case it shouldn't hurt because it's already a config-based opt-in just for tests.

howardjohn: Ah thats fair. I was thinking this was like fake_ca where we do it to avoid someone doing something dumb in prod -- but that one is configurable by env var; this is not, so it should be impossible

bleggett: I'm adding it anyway tho, might as well save the check in real code.

PR #1041, opened 2024-05-10
Deflake test_cache_refresh test

howardjohn:
This currently is slow and flaky.

Old logic: fetch cert, make sure we get exactly 1 new ones within N seconds. Flaky if our timing is off due to load.

New logic: make sure we refresh the cert, and the new cert is valid.

### DIFF
--- a/src/identity/manager.rs
+++ b/src/identity/manager.rs
@@ -702,19 +702,12 @@ mod tests {
         }
     }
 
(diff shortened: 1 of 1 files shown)


bleggett (approved):

howardjohn:
/retest

fixing this flake already in another PR :upside_down_face:

Issue #888, opened 2024-04-04
idemepotency_add_workload_fails test flake


howardjohn:
https://prow.istio.io/view/gs/istio-prow/logs/test_ztunnel_postsubmit/1776030111359832064

```
thread 'inpod::statemanager::tests::idemepotency_add_workload_fails' panicked at src/inpod/statemanager.rs:434:9:
assertion failed: !state.have_pending()
```

Locally: 
```
287362 runs so far, 0 failures (100.00% pass rate). 4.54875ms avg, 47.098639ms max, 2.663018ms min
```

so it must be pretty obscure...

MorrisLaw:
Investigating another flake in [logs/test_ztunnel_postsubmit](https://prow.istio.io/job-history/gs/istio-prow/logs/test_ztunnel_postsubmit) and came across a recent version of this particular flake. Sharing it here as another more recent data point for whoever picks this up: https://prow.istio.io/view/gs/istio-prow/logs/test_ztunnel_postsubmit/1782869865959264256

howardjohn:
oof the test wasn't running locally since requires root

PR #1042, opened 2024-05-10
Fix flakes in idempotency test

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/888

### DIFF
--- a/src/inpod/statemanager.rs
+++ b/src/inpod/statemanager.rs
@@ -342,7 +342,9 @@ mod tests {
     use super::*;
     use crate::inpod::test_helpers::{self, create_proxy_confilct, new_netns, uid};
     use crate::inpod::WorkloadData;
+
     use std::sync::Arc;
+    use std::time::Duration;
 
     struct Fixture {
         state: WorkloadProxyManagerState,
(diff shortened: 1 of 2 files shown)


src/inpod/test_helpers.rs
```diff
@@ -44,17 +45,21 @@ pub struct Fixture {
...
     fn default() -> Fixture {
         crate::test_helpers::helpers::initialize_telemetry();
-        unshare(CloneFlags::CLONE_NEWNET).unwrap();
-        let lo_set = std::process::Command::new("ip")
-            .args(["link", "set", "lo", "up"])
-            .status()
-            .unwrap()
-            .success();
-        assert!(lo_set);
+        Lazy::force(&UNSHARE);
```
howardjohn, line 62: Not strictly needed, but before each test (there is like 10) would do this. So we have like 10 randomly nested namespaces

src/inpod/statemanager.rs
```diff
@@ -431,6 +433,10 @@ mod tests {
         assert!(state.have_pending());
 
         std::mem::drop(sock);
+        // Unfortunate but necessary. When we close a socket in listener, the port is not synchronously freed.
```
bleggett, line 436: Ugh, this is not the first time we've hit this type of issue.

bleggett (approved):

PR #1047, opened 2024-05-11
Fix and improve slow test

howardjohn:
This test was 3s long and not testing anything. The "set error" was not
actualyl doing anything, and since we didn't mock time it was really
slow.

### DIFF
--- a/src/identity/caclient.rs
+++ b/src/identity/caclient.rs
@@ -136,6 +136,7 @@ pub mod mock {
     #[derive(Default)]
     struct ClientState {
         fetches: Vec<Identity>,
+        error: bool,
         gen: tls::mock::CertGenerator,
     }
 
@@ -217,6 +218,9 @@ pub mod mock {
             let not_after = not_before + self.cfg.cert_lifetime;
 
             let mut state = self.state.write().await;
+            if state.error {
+                return Err(Error::Spiffe("injected test error".into()));
+            }
             let certs = state
                 .gen
                 .new_certs(&id.to_owned().into(), not_before, not_after);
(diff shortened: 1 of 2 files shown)


src/identity/caclient.rs
```diff
@@ -225,21 +229,8 @@ pub mod mock {
...
-                });
-            } else {
-                let mut state = self.state.write().await;
-                state.fetches.push(Identity::Spiffe {
-                    trust_domain: "success".into(),
-                    namespace: "success".into(),
-                    service_account: "success".into(),
-                });
-            }
+            let mut state = self.state.write().await;
```
bleggett, line 232: If we are going to skirt the use of proper mocks in this way, let's at least add conditional compilation to the test-only methods.

src/identity/caclient.rs
```diff
@@ -217,6 +218,9 @@ pub mod mock {
             let not_after = not_before + self.cfg.cert_lifetime;
 
             let mut state = self.state.write().await;
+            if state.error {
```
bleggett, line 221: Ditto - this can be in a conditionally-compiled block, which both serves to document why the logic is here, and is safer.

src/identity/caclient.rs
howardjohn: This isn't skirting it; the old logic was literally not doing anything. It was supposed to be setting an error (so the mock would return an error) but actually just pushed it onto a list that was unused

src/identity/caclient.rs
howardjohn: the entire block is conditionally compile (`mod mock`)

bleggett: Ugh, I didn't expand everything. My bad.

bleggett (approved):

PR #1054, opened 2024-05-13
allow missing svc port for sandwich

stevenctl:
Currently, if you create a sandwiched-waypoint with

```yaml
kind: Gateway
spec:
  listeners:
  - port: 15123 # not 15008!
```

The `hboneMtlsPort` on a gateway address will still be 15008 (we want it to be whatever the zTunnel port is).
When we `find_upstream` for this waypoint's service, the port is irrelevant. Currently I have to create a bogus 15008 listener as a workaround to make this check pass in zTunnel. 

The way this PR is written could possibly break things in the future though.. it would assume inbound_tunnel does port mapping. It _sort of_ does that via application_tunnel.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -45,6 +45,8 @@ use std::net::{IpAddr, SocketAddr};
 use std::sync::{Arc, RwLock, RwLockReadGuard};
 use tracing::{debug, error, trace, warn};
 
+use self::workload::ApplicationTunnel;
+
 pub mod policy;
 pub mod service;
 pub mod workload;
@@ -249,14 +251,6 @@ impl ProxyState {
             .services
             .get_by_vip(&network_addr(network.clone(), addr.ip()))
         {
-            let Some(target_port) = svc.ports.get(&addr.port()) else {
-                debug!(
-                    "found VIP {}, but port {} was unknown",
-                    addr.ip(),
-                    addr.port()
-                );
-                return None;
-            };
             // Randomly pick an upstream
             // TODO: do this more efficiently, and not just randomly
             let Some(ep) = self.load_balance(source_workload, &svc) else {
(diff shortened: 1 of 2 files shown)


src/state.rs
```diff
@@ -267,11 +261,29 @@ impl ProxyState {
                 debug!("failed to fetch workload for {}", ep.workload_uid);
                 return None;
             };
+
+            let target_port = svc.ports.get(&addr.port());
             // If endpoint overrides the target port, use that instead
-            let target_port = ep.port.get(&addr.port()).unwrap_or(target_port);
+            let target_port = ep.port.get(&addr.port()).or(target_port);
+            let target_port = match (target_port, &wl.application_tunnel) {
```
MorrisLaw, line 268: Based on the description of this PR:
```
The way this PR is written could possibly break things in the future though.. it would assume inbound_tunnel does port mapping. It sort of does that via application_tunnel.
```

Does it make sense to have a test to go with this so in the case that what you describe is actually an issue, it may be less subtle and easier to debug with a failing test to reference. (e.g. application tunnel eventually does something different than what we assume here about port mapping). Wdyt?

My main worry here is that this becomes one of those weird bugs that ppl spend a long time trying to figure out because it's not so obvious in the code.

src/state.rs
```diff
@@ -267,11 +261,29 @@ impl ProxyState {
                 debug!("failed to fetch workload for {}", ep.workload_uid);
                 return None;
             };
+
+            let target_port = svc.ports.get(&addr.port());
```
bleggett, line 265: Nit: can probably avoid 3x allocs to the same binding name with `or_else` or something.

A debug log would be nice if we mutate the port based on the tunnel setting.

src/state.rs
bleggett: +1 weird behavior is fine as long as there's a unit test that describes it exactly.

howardjohn (approved):

howardjohn:
@stevenctl I approved with hold since I saw outstanding comments btw

stevenctl:
Added the UT, and made this codepath used in the sandwich_proxy_protocol namespaced test

src/state.rs
stevenctl: Test was added.

src/state.rs
stevenctl: ended up using some if  {} else if {} so I could `return` (can't with the closures)

Issue #909, opened 2024-04-09
DNS incorrect treats all services without VIP as headless


howardjohn:
A config like:
```
apiVersion: networking.istio.io/v1alpha3
kind: ServiceEntry
metadata:
  name: se-0
  annotations:
    istio.io/use-waypoint: default
spec:
  hosts:
  - address.baz
  location: MESH_INTERNAL
  ports:
  - name: http-0
    number: 80
    protocol: HTTP
  resolution: STATIC
  workloadSelector:
    labels:
      app: echo
```

will resolve to all "echo" Pod IPs. This is not expected. It should NXDOMAIN (assuming no auto allocation enabled)

MorrisLaw:
I was going to take this up, but I've decided not to work on this right now. This is open for anyone to pickup.

Issue #554, opened 2023-06-19
Add headless service support


nmittler:
Ztunnel doesn't currently support headless services. This will be needed by the DNS proxy, however.

The initial DNS proxy PR (#536) ignores headless services for now.

hzxuzhonghu:
Why not supported even without dns proxy?

DNS is resolved by core-dns and ztunnel  fetches Workload(Address) by dst ip

nmittler:
> Why not supported even without dns proxy?

@hzxuzhonghu I'm not sure what your point is here, exactly. If you're just saying that we should implement headless services in ztunnel independent of the DNS proxy work, that's fine. However, DNS proxy is one concrete use case where it's needed.

howardjohn:
I think we can change the proto to stop using VIP and start using service key.

 cc @kdorosh

hzxuzhonghu:
Sorry I mean IIUC ztunnel can send request to headless service currently. For ztunnel there is no much diff,  today ztunnel outbound is based on upstream workload/svc.  And in headless case, definitely it is based on workload.

PR #1056, opened 2024-05-15
dns: fix headless service handling

howardjohn:
Before, we treated any service without a VIP as headless. This is not
correct. Consider

```yaml
apiVersion: networking.istio.io/v1
kind: ServiceEntry
metadata:
  name: external
  hosts:
  - httpbin.org
  location: MESH_EXTERNAL
  ports:
  - number: 80
    name: http
  resolution: DNS
```

We should NOT treat this as headless.

The implementation is a total hack, but a better fix requires WDS
changes so for now a short approach is best

Fixes https://github.com/istio/ztunnel/issues/909
Fixes https://github.com/istio/ztunnel/issues/554

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -349,8 +349,10 @@ impl Store {
                     .filter(|service| {
                         // Domain will be like `.svc.cluster.local.` (trailing .), so ignore the last character.
                         let domain = self.svc_domain.to_utf8();
-                        !service.vips.is_empty()
-                            || service.hostname.ends_with(&domain[..domain.len() - 1])
+                        let domain = domain
+                            .strip_suffix('.')
+                            .expect("the svc domain must have a trailing '.'");
+                        !service.vips.is_empty() || service.hostname.ends_with(domain)
                     })
                     // Get the service matching the client namespace. If no match exists, just
                     // return the first service.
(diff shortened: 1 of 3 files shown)


src/dns/server.rs
```diff
@@ -333,34 +334,43 @@ impl Store {
...
+                    // TODO: we should have a reliable way to distinguish these. In sidecars, we use
+                    // `svc.Attributes.ServiceRegistry`, but we don't pass anything similar over WDS.
+                    // For now, checking the domain is good enough.
+                    // This does mean a `.svc.cluster.local` ServiceEntry will use these semantics, but
+                    // its better than *ALL* ServiceEntry doing this
+                    .filter(|service| {
+                        // Domain will be like `.svc.cluster.local.` (trailing .), so ignore the last character.
+                        let domain = self.svc_domain.to_utf8();
+                        !service.vips.is_empty()
+                            || service.hostname.ends_with(&domain[..domain.len() - 1])
```
ilrudie, line 353: Should we use strip_suffix(...).expect("the svc domain must have a trailing '.'") which is a firmer assertion of the form of the svc_domain since malformed will cause a panic but won't quietly behave badly if our assumptions were false.

ilrudie (commented):
This mostly seems ok... not sure about blind trimming 1 character though

ilrudie (approved):
lgtm

PR #1057, opened 2024-05-15
h2/fix dropper

howardjohn:
- **With compiler bug**
- **h2: remove drop counter from server**

### DIFF
--- a/src/proxy/h2.rs
+++ b/src/proxy/h2.rs
@@ -71,21 +71,41 @@ async fn do_ping_pong(
 
 // H2Stream represents an active HTTP2 stream. Consumers can only Read/Write
 pub struct H2Stream {
-    pub read: H2StreamReadHalf,
-    pub write: H2StreamWriteHalf,
+    read: H2StreamReadHalf,
+    write: H2StreamWriteHalf,
 }
 
 pub struct H2StreamReadHalf {
-    pub recv_stream: h2::RecvStream,
-    pub buf: Bytes,
-    pub half_dropped: Arc<()>,
-    pub active_count: Arc<AtomicU16>,
+    recv_stream: h2::RecvStream,
+    buf: Bytes,
+    dropped: Option<DropCounter>,
 }
 
 pub struct H2StreamWriteHalf {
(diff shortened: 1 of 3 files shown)


src/proxy/h2.rs
```diff
@@ -106,36 +126,16 @@ impl H2StreamWriteHalf {
...
-        }
-    }
-}
-
-impl Drop for H2StreamWriteHalf {
+impl Drop for DropCounter {
     fn drop(&mut self) {
         let mut half_dropped = Arc::new(());
         std::mem::swap(&mut self.half_dropped, &mut half_dropped);
         if Arc::into_inner(half_dropped).is_none() {
```
ilrudie, line 133: I know you aren't changing this but it seems like this logic isn't correct... If we get `none` back it means we are **not** the last strong reference so the other half isn't dropped. If we get the inner value back we are the last strong reference.

src/proxy/h2.rs
```diff
@@ -71,21 +71,41 @@ async fn do_ping_pong(
...
-    pub half_dropped: Arc<()>,
-    pub active_count: Arc<AtomicU16>,
+    send_stream: h2::SendStream<SendBuf>,
+    _dropped: Option<DropCounter>,
+}
+
+struct DropCounter {
+    // Whether the other end of this shared counter has already dropped.
+    // We only decrement if they have, so we do not double count
+    half_dropped: Arc<()>,
```
bleggett, line 92: Could we just use an AtomicBool here?

howardjohn: Yeah this is basically the same thing but it makes me feel smart :slightly_smiling_face:

howardjohn: I think the reason I didn't do this was it would need to be Arc<AtomicBool>. Would work though

bleggett: If it's gotta be Arc'd anyway then nevermind yea

bleggett (approved):
Nit is that I'm pretty sure there's a simpler way to do the same thing without the 

`new + swap` stuff, but that's mostly cosmetic, the fix logic is good.

PR #1075, opened 2024-05-18
Fully remove public internet dependency on DNS tests

howardjohn:
Make a common fake DNS server. Replace all instances of using the system
DNS resolver from tests and instead use it.

Really we already had a fake one, so mostly just moved it to a common
place and made it a bit more flexible.

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -1410,13 +1410,11 @@ mod tests {
 
         let resp = send_request(&mut tcp_client, n("large.com."), RecordType::A).await;
         assert!(!resp.truncated(), "TCP should not truncate");
-        assert_eq!(244, resp.answers().len());
+        assert_eq!(256, resp.answers().len());
 
         let resp = send_request(&mut udp_client, n("large.com."), RecordType::A).await;
-        //resp.answers()[0].
-        //let resp = send_request(&mut tcp_client, n("large.com."), RecordType::A).await;
+        // UDP is truncated
         assert!(resp.truncated());
-        // Not sure we can rely on the truncation being exactly 75, but for now assert
         assert_eq!(75, resp.answers().len(), "expected UDP to be truncated");
     }
 
@@ -1448,7 +1446,7 @@ mod tests {
 
     fn new_large_response() -> Vec<IpAddr> {
         let mut out = Vec::new();
-        for i in 0..244 {
+        for i in 0..256 {
             out.push(ip(format!("240.0.0.{i}")));
         }
         out
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -86,15 +86,18 @@ impl Server {
         metrics: Arc<Metrics>,
         drain: Watch,
         socket_factory: &(dyn SocketFactory + Send + Sync),
+        allow_unknown_source: bool,
     ) -> Result<Self, Error> {
         // Create the DNS server, backed by ztunnel data structures.
-        let handler = dns::handler::Handler::new(Arc::new(Store::new(
+        let mut store = Store::new(
             domain,
             network.as_ref().to_string(),
             state,
             forwarder,
             metrics,
-        )));
+        );
+        store.allow_unknown_source = allow_unknown_source;
+        let handler = dns::handler::Handler::new(Arc::new(store));
         let mut server = ServerFuture::new(handler);
         info!(
             address=%addr,
@@ -165,6 +168,7 @@ struct Store {
     domain: Name,
     svc_domain: Name,
     metrics: Arc<Metrics>,
+    allow_unknown_source: bool,
 }
 
 impl Store {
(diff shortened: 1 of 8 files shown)


src/dns/server.rs
```diff
@@ -1263,8 +1275,9 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn system_forwarder() {
+    async fn forward_to_server() {
```
MorrisLaw, line 1278: nit: wdyt about `forward_to_local_server`, if the fact that it is local is truly significant as suggested in the below comment

src/dns/server.rs
```diff
@@ -1353,27 +1375,50 @@ mod tests {
...
+        let mut tcp_client = new_tcp_client(tcp_addr).await;
+        let mut udp_client = new_udp_client(udp_addr).await;
+
+        let resp = send_request(&mut tcp_client, n("large.com."), RecordType::A).await;
+        assert!(!resp.truncated(), "TCP should not truncate");
+        assert_eq!(244, resp.answers().len());
+
+        let resp = send_request(&mut udp_client, n("large.com."), RecordType::A).await;
+        //resp.answers()[0].
+        //let resp = send_request(&mut tcp_client, n("large.com."), RecordType::A).await;
```
MorrisLaw, line 1417: leftover?

src/dns/server.rs
```diff
@@ -1353,27 +1375,50 @@ mod tests {
...
+        assert!(resp.truncated());
+        // Not sure we can rely on the truncation being exactly 75, but for now assert
```
MorrisLaw, line 1419: > Not sure we can rely on the truncation being exactly 75

Why? Genuinely curious.

Also, might be good to add any extra meaningful context in case we have to come back and fix something here with the assert.

src/telemetry.rs
```diff
@@ -517,8 +517,10 @@ pub mod testing {
         let mock_writer = MockWriter::new(global_buf());
         let (non_blocking, _guard) = tracing_appender::non_blocking::NonBlockingBuilder::default()
             .lossy(false)
-            .buffered_lines_limit(10)
+            .buffered_lines_limit(1)
```
MorrisLaw, line 520: wdyt about making this a constant instead of a magic number?

src/test_helpers/dns.rs
```diff
@@ -206,3 +207,164 @@ pub fn ipv4<S: AsRef<str>>(addr: S) -> Ipv4Addr {
...
+
+pub struct TestDnsServer {
+    tcp: SocketAddr,
+    udp: SocketAddr,
+    resolver: Arc<dyn Resolver>,
+    _drain: drain::Signal,
+}
+
+impl TestDnsServer {
+    /// resolver_config gets a config that can be used passed to Ztunnel to make this the resolver
```
MorrisLaw, line 219: nit: wording seems a little off `can be used passed to Ztunnel`

src/test_helpers/dns.rs
```diff
@@ -206,3 +207,164 @@ pub fn ipv4<S: AsRef<str>>(addr: S) -> Ipv4Addr {
...
+pub async fn run_dns(responses: HashMap<Name, Vec<IpAddr>>) -> anyhow::Result<TestDnsServer> {
+    let test_metrics = {
+        let mut registry = Registry::default();
+        let istio_registry = metrics::sub_registry(&mut registry);
+        Arc::new(Metrics::new(istio_registry))
+    };
+    let (signal, drain) = drain::channel();
+    let factory = crate::proxy::DefaultSocketFactory {};
+
+    //HashMap::from([(n("www.bing.com."), vec![ip("1.1.1.1")])]),}
```
MorrisLaw, line 254: leftover?

src/dns/server.rs
```diff
@@ -1401,13 +1446,13 @@ mod tests {
...
-    // fn new_large_response() -> Vec<Record> {
-    //     let mut out = Vec::new();
-    //     for i in 0..64 {
-    //         out.push(a(n("aaaaaaaaaaaa.aaaaaa."), ipv4(format!("240.0.0.{i}"))));
-    //     }
-    //     out
-    // }
+    fn new_large_response() -> Vec<IpAddr> {
+        let mut out = Vec::new();
+        for i in 0..244 {
```
MorrisLaw, line 1451: what is the significance of 244?

src/dns/server.rs
howardjohn: Its not significant to the test, just to general testing philosophy (make tests hermetic)

src/dns/server.rs
howardjohn: Wasn't sure if the response size was full deterministic, etc. Seems to be, though

src/dns/server.rs
howardjohn: Just needs to be a large number. I made it 256 to make it a bit less random (now its the max we can do in the right octet)

bleggett (approved):

PR #1079, opened 2024-05-21
add logging for xds nack errors

### LINKED ISSUE #1077: Ztunnel should log on NACK
When we NACK, we do not log. NACK is a critical error, so we ought to log

ilrudie:
fixes #1077 

should we cherrypick to 1.22?


bleggett (approved):

howardjohn (commented):
nice! thanks

Issue #1009, opened 2024-05-02
Consider constructors for heavy custom types to make heap allocation consistent


bleggett:
Spinoff of https://github.com/istio/ztunnel/pull/1002#discussion_r1587902910

Rather than `Box`ing or `Arc`ing at point of use, we could simply add `new`s/constructors to our heavier custom types so that they are heap-allocated by default as an opinion at time of declaration, to keep things consistent for all usages.

PR #1082, opened 2024-05-21
Refactor build_requests

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/846
Fixes https://github.com/istio/ztunnel/issues/1009

This does a general cleanup of a lot of accumulated cruft. Lots of minor changes, but main ones:
* Use Arc in more places
* Rework `Request`  and `pstream` type: rename fields to make it clear what are attributes of original workload/waypoint/network gatways. No more "gateway_address", especially not the mutable aspect
* Collapse the fetch upstream, pick_workload_destination, compute various identity sets/SocketAddrs. Now one function does it all for us.
* `State` has metrics now so we don't need to pass it in most functions

### DIFF
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -24,7 +24,7 @@ use hyper::header::FORWARDED;
 
 use tokio::net::TcpStream;
 
-use tracing::{debug, error, info, info_span, trace_span, warn, Instrument};
+use tracing::{debug, error, info, info_span, trace, trace_span, warn, Instrument};
 
 use crate::config::ProxyMode;
 use crate::identity::Identity;
@@ -158,8 +158,7 @@ impl OutboundConnection {
         let source_addr =
             socket::to_canonical(source_stream.peer_addr().expect("must receive peer addr"));
         let dst_addr = socket::orig_dst_addr_or_default(&source_stream);
-        self.proxy_to(source_stream, source_addr, dst_addr, false)
-            .await;
+        self.proxy_to(source_stream, source_addr, dst_addr).await;
     }
 
     // this is a cancellable outbound proxy. If `out_drain` is a Watch drain, will resolve
@@ -174,7 +173,6 @@ impl OutboundConnection {
         stream: TcpStream,
         remote_addr: SocketAddr,
         orig_dst_addr: SocketAddr,
-        block_passthrough: bool,
         out_drain: Option<Watch>,
     ) {
         match out_drain {
@@ -183,12 +181,11 @@ impl OutboundConnection {
                         _ = drain.signaled() => {
                             info!("drain signaled");
                         }
-                        res = self.proxy_to(stream, remote_addr, orig_dst_addr, block_passthrough) => res
+                        res = self.proxy_to(stream, remote_addr, orig_dst_addr) => res
                 }
             }
             None => {
-                self.proxy_to(stream, remote_addr, orig_dst_addr, block_passthrough)
-                    .await;
+                self.proxy_to(stream, remote_addr, orig_dst_addr).await;
             }
         }
     }
@@ -198,7 +195,6 @@ impl OutboundConnection {
         mut source_stream: TcpStream,
         source_addr: SocketAddr,
         dest_addr: SocketAddr,
-        block_passthrough: bool,
     ) {
         let start = Instant::now();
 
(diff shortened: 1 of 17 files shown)


src/proxy/outbound.rs
```diff
@@ -212,42 +205,24 @@ impl OutboundConnection {
...
-        let hbone_target = if req.protocol == Protocol::HBONE {
-            Some(req.destination)
-        } else {
-            None
-        };
+        let hbone_target = req.hbone_target_destination;
         let result_tracker = Box::new(ConnectionResult::new(
             source_addr,
-            req.gateway,
+            req.actual_destination,
```
bleggett, line 225: :+1:

src/proxy/outbound.rs
```diff
@@ -329,6 +278,14 @@ impl OutboundConnection {
             .body(())
             .expect("builder with known status code should not fail");
 
+        let pool_key = Box::new(pool::WorkloadKey {
+            src_id: req.source.identity(),
+            // Clone here shouldn't be needed ideally, we could just take ownership of Request.
```
bleggett, line 283: Yeah - why construct the poolkey here at all? We can just chuck the whole request + remote_addr at `send_request_pooled`.

src/proxy/outbound.rs
```diff
@@ -363,91 +320,54 @@ impl OutboundConnection {
...
-                let waypoint_socket_address = SocketAddr::new(waypoint_ip, waypoint_us.port);
-                let id = waypoint_workload.identity();
-                return Ok(Box::new(Request {
+            if let Some(waypoint) = state
+                .fetch_service_waypoint(&target_service, &source_workload)
+                .await?
+            {
+                let upstream_sans = waypoint.workload_and_services_san();
+                let actual_destination = waypoint.workload_socket_addr();
+                trace!("built request to service waypoint proxy");
```
bleggett, line 361:
```suggestion
                debug!("built request to service waypoint proxy");
```

Mega nit: slight pref for logs indicating request processing decisions to be `debug!` - trace is a lot  of noise.

src/proxy/outbound.rs
```diff
@@ -456,105 +376,97 @@ impl OutboundConnection {
...
-                    destination_service: None,
-                    expected_identity: None,
-                    gateway: target,
-                    request_type: RequestType::Passthrough,
-                    upstream_sans: vec![],
-                }));
-            }
+            .await?
+        else {
+            trace!("built request as passthrough; no upstream found");
```
bleggett, line 383: ditto nit

src/proxy/outbound.rs
```diff
@@ -568,32 +480,28 @@ fn baggage(r: &Request, cluster: String) -> String {
...
-    source: Workload,
-    destination: SocketAddr,
-    // The intended destination workload. This is always the original intended target, even in the case
-    // of other proxies along the path.
-    destination_workload: Option<Workload>,
-    destination_service: Option<ServiceDescription>,
-    // The identity we will assert for the next hop; this may not be the same as destination_workload
+    // Source workload sending the request
+    source: Arc<Workload>,
+    // The actual destination workload we are targeting. When proxying through a waypoint, this is the waypoint,
```
bleggett, line 487: :+1:  for the comments/naming here, extreme obviousness is good.

src/state.rs
```diff
@@ -243,12 +252,12 @@ impl ProxyState {
...
     }
 
-    pub fn find_upstream(
+    fn find_upstream(
         &self,
         network: Strng,
         source_workload: &Workload,
         addr: SocketAddr,
-    ) -> Option<Upstream> {
+    ) -> Option<(Arc<Workload>, u16, Option<Arc<Service>>)> {
```
bleggett, line 260: nit: maybe `type` this (`FoundUpstream`)?

bleggett (commented):
LGTM, mostly nits, will put a hold which you can self-remove.

bleggett (approved):

src/proxy/outbound.rs
howardjohn: We have the same clone problem either way I think (actually, _maybe_ not since we Box::pin the future).

If we pass owned `Request` here we can easy deconstruct the request and not clone `upstream_sans`. But then we end up paying for the stack cost of Request twice (and I think it somehow actually becomes 4x due to some async stuff?). If we have a reference we don't need to, but then we need to clone this. If we kick this down further into `send_request_pooled` we have the same problem, just possibly more stack usage I think?

I guess maybe an owned `Box<Request>` wouldn't have that problem...

bleggett: > I guess maybe an owned Box<Request> wouldn't have that problem...

That would be my intuition, but IDK for sure. Eh, it's a small thing either way.

PR #1086, opened 2024-05-22
mertics: derive identity from TLS handshake

howardjohn:
Today we do an IP lookup. I think this should only fail when we are going through middle proxies or have `!inpod && !src_ip_spoofing` which is ~never.

For the middle proxy case we really should consider getting the other attributes from them as well (baggage) vs lookup the IP.. but at least TLS identity seems critical to get right + has a clear right answer on where to get the attribute from

### DIFF
--- a/src/proxy/metrics.rs
+++ b/src/proxy/metrics.rs
@@ -147,17 +147,11 @@ impl CommonTrafficLabels {
         self.source_canonical_service = w.canonical_name.clone().into();
         self.source_canonical_revision = w.canonical_revision.clone().into();
         self.source_workload_namespace = w.namespace.clone().into();
-        self.source_principal = w.identity().into();
+        // We explicitly do not set source_principal here. This is set only with with_derived_source
+        // based on the real mTLS identity.
         self.source_app = w.canonical_name.clone().into();
         self.source_version = w.canonical_revision.clone().into();
         self.source_cluster = w.cluster_id.to_string().into();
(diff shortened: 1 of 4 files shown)


src/proxy/metrics.rs
```diff
@@ -153,16 +153,24 @@ impl CommonTrafficLabels {
         self
     }
 
+    fn with_derived_source_identity(mut self, w: Option<&DerivedWorkload>) -> Self {
+        let Some(w) = w else { return self };
+        // This is the identity from the TLS handshake; this is the most trustworthy source so use it
+        self.source_principal = w.identity.clone().into();
```
ilrudie, line 159: should this handle if w is Some but w.identity is None? Seems like we may not want to overwrite with None.

howardjohn: We actually might want to -- if the request is not over TLS I don't think we want to report the principal?

Let me check how envoy does it

ilrudie: That would make sense. I'll add a hold so you can address or not but LGTM otherwise

ilrudie (approved):
LGTM aside from the question of proper source_principal behavior. feel free to drop the hold

src/proxy/metrics.rs
howardjohn: Envoy does NOT report the principal in metrics when doing plaintext

PR #1092, opened 2024-05-28
Fix outbound race condition where we don't have the source workload from XDS yet

bleggett:
Fixes: https://github.com/istio/istio/issues/51193

~WIP as I need to add tests (and an inpod mode conditional)~

Q: Why didn't we use `demand` here?

A: We don't need to send a new XDS request here - this only (potentially) happens very early in the pod lifecycle, when the pod is making a "startup" outbound request, but XDS hasn't gotten around to pushing us the soure workload yet. It always does/will (or we *should* fail), we just have to wait for it, so sending a demand request for something that's already coming seems chatty - when all we really need to do is stall.

In `inpod` mode, it is always safe to stall-or-fail, as we by-definition should only be making outbound requests for Ambient workloads (which definitely have to be in the CP)

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -686,28 +686,31 @@ impl DemandProxyState {
         // Take a watch listener *before* checking state (so we don't miss anything)
         let mut wl_sub = self.state.read().unwrap().workloads.new_subscriber();
 
-        match self.fetch_workload(addr).await {
-            Some(wl) => Some(wl),
-            None => {
-                // We didn't find the workload we expected, so
-                // loop until the subscriber wakes us on new workload,
-                // or we hit the deadline timeout and give up
-                let timeout = tokio::time::sleep(deadline);
-                let subscriber = wl_sub.changed();
-                tokio::pin!(timeout);
-                tokio::pin!(subscriber);
-                loop {
-                    tokio::select! {
-                        _ = &mut timeout => break None,
-                        _ = &mut subscriber => {
-                            match self.fetch_workload(addr).await {
-                                Some(found) => break Some(found),
-                                None => continue,
-                            }
-                        }
-                    };
+        debug!(%addr, "got sub, waiting for workload");
+
+        if let Some(wl) = self.fetch_workload(addr).await {
+            return Some(wl);
+        }
+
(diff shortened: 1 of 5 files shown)


src/state/workload.rs
```diff
@@ -621,6 +646,10 @@ impl WorkloadStore {
                 .or_default()
                 .insert(w.uid.clone());
         }
+
+        // We have stored a newly inserted workload, notify watchers
+        // (if any) to wake.
+        self.insert_notifier.send_replace(());
```
howardjohn, line 652: On a large XDS update, this is going to trigger N times which will then wake up all watchers and have them do a lookup (lookup is cheap, but locks so its not that cheap). Essentially we have an `O(UpdateSize*Waiters)` operation which could be not great.

I don't think this is a big deal in practice, though. But it would be better to operate on the xds message granularity vs per-resource

src/state.rs
```diff
@@ -672,6 +673,47 @@ impl DemandProxyState {
...
+
+        if let Some(wl) = self.fetch_workload(addr).await {
+            return Some(wl);
+        }
+
+        // We didn't find the workload we expected, so
+        // loop until the subscriber wakes us on new workload,
+        // or we hit the deadline timeout and give up
+        let timeout = tokio::time::sleep(deadline);
+        let subscriber = wl_sub.changed();
```
howardjohn, line 699: I think we could do something a bit more precise by re-using the on-demand infrastructure without actually using on-demand. Which basically amounts to using the `pending` map in XDS client to register a watch for a specific resource rather than anything.

I don't mind this either though

howardjohn (approved):

src/state/workload.rs
bleggett: Yeah I gestured at sending the Workload thru the channel as a future optimization, to avoid a hard lookup after insert, can do that if it creates problems - but yeah this is uncommon to begin with.

src/state.rs
bleggett: Yep, could do per-resource watches easily enough. If this were (or becomes) more of a hotpath then we can do that.

PR #1094, opened 2024-05-28
Do not expose internal rustls errors to users

howardjohn:
Like `io error: peer closed connection without sending TLS close_notify: https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof`

TBH its *possible* this is a bug in our own code in not gracefully
shutting down. I am not entirely sure. Regardless, it is amateur hour to
expose an error like this to users.


ilrudie (approved):
lgtm

PR #1095, opened 2024-05-28
access logs: consistently quote fields

howardjohn:
Before we were semi-randomly quoting or not quoting fields. Now we
always quote strings.

### DIFF
--- a/src/proxy/metrics.rs
+++ b/src/proxy/metrics.rs
@@ -24,15 +24,15 @@ use prometheus_client::metrics::family::Family;
 use prometheus_client::registry::Registry;
 
 use tracing::event;
+use tracing_core::field::Value;
 
 use crate::identity::Identity;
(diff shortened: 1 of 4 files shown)


src/proxy/metrics.rs
```diff
@@ -545,3 +513,12 @@ impl ConnectionResult {
...
 }
+
+fn display_field_owned<T: ToString>(t: T) -> impl Value {
+    t.to_string()
+}
+
+fn display_field<T: AsRef<str>>(t: &T) -> impl Value + '_ {
+    let v: &str = t.as_ref();
+    v
+}
```
ilrudie, line 524: **nit/bikeshed**: should these be `to_value` rather than display_field?

ilrudie (approved):
lgtm, added one minor/optional comment and a hold, feel free to drop the hold if you think the suggestion doesn't make sense.

PR #1096, opened 2024-05-28
socks5: add initial hostname lookup support

howardjohn:
This isn't great (the DNS usage is wonky), but works well enough and
socks5 is experimental and requires explicit opt-in.

With this, you can hookup a browser to ztunnel and it works.

### DIFF
--- a/src/proxy/socks5.rs
+++ b/src/proxy/socks5.rs
@@ -16,13 +16,19 @@ use anyhow::Result;
 use byteorder::{BigEndian, ByteOrder};
 use drain::Watch;
 
+use hickory_proto::op::{Message, MessageType, Query};
+use hickory_proto::rr::{Name, RecordType};
+use hickory_proto::serialize::binary::BinDecodable;
+use hickory_server::authority::MessageRequest;
+use hickory_server::server::{Protocol, Request};
 use std::net::{IpAddr, SocketAddr};
 use std::sync::Arc;
 
+use crate::dns::resolver::Resolver;
 use tokio::io::AsyncReadExt;
 use tokio::io::AsyncWriteExt;
 use tokio::net::TcpStream;
-use tracing::{error, info};
+use tracing::{debug, error, info};
 
 use crate::proxy::outbound::OutboundConnection;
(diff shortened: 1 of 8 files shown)


src/proxy/socks5.rs
```diff
@@ -229,3 +249,53 @@ async fn handle(
     });
     Ok(())
 }
+
+async fn dns_lookup(
```
bleggett, line 253: Feels like this (or some of it) eventually belongs on the `resolver` itself, and not in `socks5`, but fine for now I guess.

src/proxy/socks5.rs
```diff
@@ -229,3 +249,53 @@ async fn handle(
...
+
+    /// Creates a AAAA-record [Request] for the given name.
+    fn aaaa_request(name: Name, client_addr: SocketAddr, protocol: Protocol) -> Request {
+        server_request(&new_message(name, RecordType::AAAA), client_addr, protocol)
+    }
+
+    // TODO: do we need to do the search?
+    let name = Name::from_utf8(hostname)?;
+
+    // TODO: we probably want to race them or something. Is there something higher level that can handle this for us?
```
bleggett, line 287: Should probably be:

- IF client_addr == ipv6

- build both, `tokio::select`

yea

src/proxyfactory.rs
```diff
@@ -80,6 +80,25 @@ impl ProxyFactory {
         let mut result: ProxyResult = Default::default();
         let drain = proxy_drain.unwrap_or_else(|| self.drain.clone());
 
+        let mut resolver = None;
+        // Optionally create the DNS proxy.
+        if self.config.dns_proxy {
```
bleggett, line 85: We probably need to redo this config flag logic at some point - with this you can't really provide a resolver without enabling the whole DNS proxy.

bleggett (approved):

src/proxy/socks5.rs
bleggett: Actually AFAICT the go dns code waits for both to fail, so maybe we should `join` and not `select`

PR #1098, opened 2024-05-28
dns: drop (unneeded) on-demand DNS caching

howardjohn:
The client we use already has its own cache, probably better than ours.
However, currently we construct it on each request, dropping the cache
(and also probably making connection re-use, etc, not happen). We then
add our own cache on top.

This change stores the client so we can use it as a cache, and drops our
bespoke cache.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -37,7 +37,7 @@ use itertools::Itertools;
 use rand::prelude::IteratorRandom;
 use rand::seq::SliceRandom;
 use serde::Serializer;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::convert::Into;
 use std::default::Default;
 use std::fmt;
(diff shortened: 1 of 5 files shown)


src/state.rs
```diff
@@ -526,127 +503,37 @@ impl DemandProxyState {
...
-            .unwrap()
-            .resolved_dns
-            .by_hostname
-            .get(hostname)
-            .filter(|rdns| {
-                rdns.initial_query.is_some()
-                    && rdns.initial_query.unwrap().elapsed() < rdns.dns_refresh_rate
-            })
-            .cloned()
+        warn!(%hostname,"dns lookup complete {resp:?}");
```
ilrudie, line 526: warn seems a bit high for this, probably info or maybe even lower is more appropriate

ilrudie (commented):
This looks good other than the one warn level message which probably should be lowered.

src/state.rs
howardjohn: thanks, probably was a debug log while I was testing

ilrudie (approved):
lgtm, thanks!

dhawton (approved):

Issue #1074, opened 2024-05-17
dns: add "autopath" mode


howardjohn:
Like https://coredns.io/plugins/autopath/ or the same code we have in istio-agent

This is actually a correctness issue not just performance IMO

Issue #555, opened 2023-06-19
Support per-pod DNS settings


nmittler:
The DNS proxy needs a way to identify per-pod DNS settings when running as a shared (multi-tenant) proxy. The [initial implementation](https://github.com/istio/ztunnel/pull/536) will use the settings defined the ztunnel DaemonSet's `/etc/resolv.cof` for all clients running on the node. While this will likely work for most use cases, we also need to respect per-pod settings.

MorrisLaw:
This seems like a good issue for me to get deeper on Ztunnel to help further my understanding. Can I assign this to myself @nmittler @bleggett ?

MorrisLaw:
Also, I think `initial implementation` in the description should link to this: https://github.com/istio/ztunnel/pull/536

nmittler:
@MorrisLaw sure, go for it!

MorrisLaw:
RFC link: https://docs.google.com/document/d/1P2q7VnqFDZF4Ght913wtvVRuXHF3keJItbd8UENsvu4

MorrisLaw:
As per the discussion in today's Ambient meeting. We'll be pausing work on this for now.

PR #1099, opened 2024-05-29
dns: fix search lookup

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/1074 (issue was incorrect
basically).

Before our search domain was always ztunnel's (istio-system). This is incorrect; we
should know the search domain of the request.

We don't know the per-pod custom settings
(https://github.com/istio/ztunnel/issues/555), but we can use the
correct defaults.


src/dns/server.rs
```diff
@@ -756,8 +803,17 @@ impl SystemForwarder {
...
-        self.search_domains.clone()
+    fn search_domains(&self, wl: &Workload) -> Vec<Name> {
+        // TODO: https://github.com/istio/ztunnel/issues/555 really get this from pods
+        // Today we hardcode the default search assumptions
+        match &self.search_domains {
+            SearchDomains::Static(s) => s.clone(),
+            SearchDomains::Dynamic(cluster_domain) => vec![
+                Name::from_utf8(format!("{}.svc.{cluster_domain}", wl.namespace)).unwrap(),
+                Name::from_utf8(format!("svc.{cluster_domain}")).unwrap(),
+                Name::from_utf8(cluster_domain).unwrap(),
```
ilrudie, line 814: unwrap seems like it's probably not particularly unsafe but we should probably use expect or explicitly handle sad paths.

ilrudie (commented):
looks good but probably should remove the unwraps before we merge it

src/dns/server.rs
bleggett: :+1: `.expect` preferred simply because it's easier to read - the justification for why it should never fail is right there.

src/dns/server.rs
```diff
@@ -723,12 +740,17 @@ pub fn forwarder_for_mode(proxy_mode: ProxyMode) -> Result<Arc<dyn Forwarder>, E
...
 
+enum SearchDomains {
+    Static(Vec<Name>),
+    Dynamic(String),
+}
+
 impl SystemForwarder {
-    fn new() -> Result<Self, Error> {
+    fn new(per_pod: bool, cluster_domain: String) -> Result<Self, Error> {
         // Get the resolver config from /etc/resolv.conf.
```
bleggett, line 754:
```suggestion
        // Get the resolver config from `ztunnel's` /etc/resolv.conf.
```

ilrudie (approved):

Issue #1119, opened 2024-06-06
Fixes around workload.hostname


howardjohn:
Currently, Istiod will only generate a hostname workload for SE/WE of DNS type. In ztunnel, we try to match DNS requests to it: https://github.com/istio/ztunnel/blob/2e415386382734b3f7be6882b460db39980e3bfe/src/dns/server.rs#L390-L396.

This is broken.

```yaml
apiVersion: networking.istio.io/v1
kind: ServiceEntry
metadata:
  name: external-service
  namespace: echo
spec:
  addresses:
  - 240.240.240.239
  endpoints:
  - address: external.external.svc.cluster.local
  exportTo:
  - .
  hosts:
  - fake.external.com
```

This config should NOT trigger responses for 'external.external.svc.cluster.local'.

What we should be allowing is looking up pods in headless services. The logic is tricky, see https://github.com/istio/istio/blob/f69047b03842074eb2f56fef5d9ad7ba326de4cc/pkg/dns/server/name_table.go#L58. We don't do that today.

Finally, WorkloadStore has a by_hostname->Workload. This is not right; a hostname is not unique (definitely not across clusters!).

So a few changes needed
* Turn off DNS lookups for SE/WE endpoints (ill do this)
* Turn on DNS lookups for headless pods (not super high priority, we will forward them anyways)
* Make by_hostname correct, assuming we still need it

Most of by_hostname's usage beyond DNS is in gateway addresses. I am not sure actually these should allow workload hostnames, only service ones. So maybe we can remove it entirely. However, we will probably need it to do headless pods... but we should make sure its not used in all those other cases anyways. So, regardless, we need to clean it up

PR #1120, opened 2024-06-06
Drop improper workload DNS lookups

howardjohn:
Part 1 of https://github.com/istio/ztunnel/issues/1119. This removes the
broken stuff. TODO is to add it back properly, but for now this PR
strictly removes broken code.


bleggett (approved):
I don't know of any uses of this stuff off the top of my head, so no issues with removing it.

PR #1125, opened 2024-06-08
logs: do not allocate absurd buffers

howardjohn:
This is 4mb which ends up as 8mb RAM in k8s reporting. Drop it down to
1000 - that is plenty and drops to 2mb RAM usage total. Yay, 4x decrease
for free


ilrudie (approved):
lgtm

PR #1136, opened 2024-06-13
Chill out at the proto level as well as the socket level if we have issues connecting to the ambient node agent.

bleggett:
Related to: https://github.com/istio/istio/issues/51519

Context:

We already had a backoff for connecting to the node agent's socket - that's good.

But if we successfully connect to the socket and the connection then fails for some other, protocol-related reason (proto mismatch, server expects a different message, etc etc), we immediately reconnect and try again, which basically defeats the inner backoff.

~So, just have an inner and outer backoff. We could get away with just the outer one, but having two is a little more granular.~

EDIT: Rather than backing off, just treat protocol errors as terminal, and quit.

Connection/stream error -> retry
Protocol error -> bail, shutdown, should never happen.

### DIFF
--- a/src/inpod/workloadmanager.rs
+++ b/src/inpod/workloadmanager.rs
@@ -155,14 +155,6 @@ impl WorkloadProxyManager {
     async fn run_internal(&mut self, drain: Watch) {
         // for now just drop block_ready, until we support knowing that our state is in sync.
         debug!("workload proxy manager is running");
-
-        // the UDS socket `connect` has its own retry+backoff logic, but
-        // we also need to backoff at the protocol level if we successfully connect to the socket,
-        // in case we can't agree with the server on the protocol/message format
-        // (so we don't keep succesfully connecting, and spamming them with bad messages)
-        const MAX_BACKOFF: Duration = Duration::from_secs(15);
-        let mut backoff = Duration::from_millis(10);
-
         // hold the  release shutdown until we are done with `state.drain` below.
         let _rs = loop {
             // Accept a connection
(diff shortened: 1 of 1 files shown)


src/inpod/workloadmanager.rs
```diff
@@ -184,6 +192,9 @@ impl WorkloadProxyManager {
             };
             debug!("workload proxy manager is NOT ready");
             self.readiness.not_ready();
+            backoff = std::cmp::min(MAX_BACKOFF, backoff * 2);
```
howardjohn, line 195: We may need some way to reset the backoff. Otherwise, eventually over a ztunnel running for a long time it will always hit the 15s backoff (?)

bleggett: If there are sustained protocol errors we'll never get marked as ready/will stay unready until k8s kills us. Which is about all we can do.

If there are intermittent/transient protocol errors that then somehow resolve themselves(?) yep, though that's not likely to manifest as anything terribly severe or happen.

I can tie the backoff reset to the readiness toggle, so it resets when we go ready.

howardjohn: I was thinking like ztunnel runs for a month and cni restarts daily or something. Not sure how realistic that is though

bleggett: > I was thinking like ztunnel runs for a month and cni restarts daily or something. Not sure how realistic that is though

That would be handled by the inner socket backoff, and we wouldn't get to this point.

This backoff is just for protocol errors, not transport errors.

bleggett: Actually, maybe we should just die on protocol errors. Probably over-complicating this.

src/inpod/workloadmanager.rs
```diff
@@ -186,15 +184,17 @@ impl WorkloadProxyManager {
                 Ok(()) => {
                     info!("process stream ended with eof");
                 }
+                Err(Error::ProtocolError) => {
+                    error!("protocol mismatch error while processing stream, shutting down");
+                    return
```
howardjohn, line 189: do we need to bubble up an error so https://github.com/istio/ztunnel/blob/c84404d0456a1c6677966277d7f84622e655cdc7/src/app.rs#L335 hits?

howardjohn (approved):

PR #1142, opened 2024-06-14
Properly handle named targetPort services

howardjohn:
When we connect to a service with a named targetPort, we should exclude
endpoints that do not have that port defined. Today, we send to it which
doesn't work (port 0 attempted).

This is mostly straightforward but a bit of quirks around app_tunnel.

Note I also made load_balance directly return the workload. This avoids
2 lookups, but also I have 3(!) WIPs locally that need it, so good to
get it in.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -247,6 +247,7 @@ impl ProxyState {
         network: Strng,
         source_workload: &Workload,
         addr: SocketAddr,
+        resolution_mode: ServiceResolutionMode,
     ) -> Option<(Arc<Workload>, u16, Option<Arc<Service>>)> {
         if let Some(svc) = self
             .services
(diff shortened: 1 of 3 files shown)


src/state.rs
```diff
@@ -247,28 +247,27 @@ impl ProxyState {
...
         source_workload: &Workload,
         addr: SocketAddr,
+        resolution_mode: ServiceResolutionMode,
     ) -> Option<(Arc<Workload>, u16, Option<Arc<Service>>)> {
         if let Some(svc) = self
             .services
             .get_by_vip(&network_addr(network.clone(), addr.ip()))
         {
             // Randomly pick an upstream
             // TODO: do this more efficiently, and not just randomly
```
bleggett, line 257: We aren't doing it randomly anymore?

src/state.rs
```diff
@@ -336,14 +383,14 @@ impl ProxyState {
                         {
                             return None;
                         }
-                        Some((rank, ep))
+                        Some((rank, ep, wl))
                     })
                     .collect::<Vec<_>>();
-                let max = *ranks.iter().map(|(rank, _ep)| rank).max()?;
+                let max = *ranks.iter().map(|(rank, _ep, _wl)| rank).max()?;
```
bleggett, line 389: Probably a cute way to chain these `map()`s in one statement but I suspect it would be a cosmetic difference.

src/state.rs
```diff
@@ -763,6 +814,14 @@ impl DemandProxyState {
     }
 }
 
+#[derive(Eq, PartialEq, Clone, Copy, Debug)]
+pub enum ServiceResolutionMode {
```
bleggett, line 818: Comment here explaining why we need to modes and what we use this for would be good.

(I see why but in 6 months local context will help)

bleggett (approved):

src/state.rs
howardjohn: It is still random there is just some filtering before we make the random choice

PR #1143, opened 2024-06-14
Drop IP spoofing for outbound traffic

howardjohn:
This is not needed and breaks dualstack + adds complexity

### DIFF
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -50,19 +50,21 @@ impl Outbound {
             .socket_factory
             .tcp_bind(pi.cfg.outbound_addr)
             .map_err(|e| Error::Bind(pi.cfg.outbound_addr, e))?;
-        let enable_orig_src = super::maybe_set_transparent(&pi, &listener)?;
+        let transparent = super::maybe_set_transparent(&pi, &listener)?;
 
         info!(
             address=%listener.local_addr(),
             component="outbound",
-            transparent=enable_orig_src,
+            transparent,
             "listener established",
         );
+        let inpod = pi.cfg.inpod_enabled;
         Ok(Outbound {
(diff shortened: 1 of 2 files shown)


src/proxy/outbound.rs
```diff
@@ -79,7 +79,8 @@ impl Outbound {
         let (sub_drain_signal, sub_drain) = drain::channel();
         let pool = proxy::pool::WorkloadHBONEPool::new(
             self.pi.cfg.clone(),
-            self.enable_orig_src,
+            // Do not need to spoof with inpod mode for outbound
+            self.enable_orig_src && !self.pi.cfg.inpod_enabled,
```
bleggett, line 83: Can we just do this check in `new` and set `enable_orig_src = false` once-for-all?

src/proxy/socks5.rs
```diff
@@ -81,7 +81,8 @@ impl Socks5 {
                 // but ProxyInfo is overloaded and only `outbound` should ever use the pool.
                 let pool = crate::proxy::pool::WorkloadHBONEPool::new(
                     self.pi.cfg.clone(),
-                    self.enable_orig_src,
+                    // Do not need to spoof with inpod mode for outbound
+                    self.enable_orig_src && !self.pi.cfg.inpod_enabled,
```
bleggett, line 85: Ditto - hoist the config check to `new` and just turn it off at the beginning.

src/proxy/outbound.rs
howardjohn: good call, done

src/proxy/socks5.rs
howardjohn: :+1:

bleggett (approved):

PR #1144, opened 2024-06-14
Hide error logs when peer shut's down for us

howardjohn:
See https://github.com/istio/istio/discussions/51586 for a real world
example

### DIFF
--- a/src/copy.rs
+++ b/src/copy.rs
@@ -115,6 +115,7 @@ fn ignore_shutdown_errors(res: Result<(), io::Error>) -> Result<(), io::Error> {
             if e.kind() == io::ErrorKind::NotConnected
                 || e.kind() == io::ErrorKind::UnexpectedEof =>
         {
+            trace!(err=%e, "failed to shutdown peer, they already shutdown");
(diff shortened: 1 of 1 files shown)


jsievenpiper:
Thank you!

src/copy.rs
```diff
@@ -107,6 +107,20 @@ where
...
 
+// During shutdown, the other end may have already disconnected. That is fine, they shutdown for us.
+// Ignore it.
+fn ignore_shutdown_errors(res: Result<(), io::Error>) -> Result<(), io::Error> {
+    match &res {
+        Err(e)
+            if e.kind() == io::ErrorKind::NotConnected
+                || e.kind() == io::ErrorKind::UnexpectedEof =>
+        {
+            Ok(())
```
bleggett, line 118: Would we still get a `trace` in this case? Should probably still be able to see this is being ignored @ `trace` level.

bleggett (approved):

src/copy.rs
howardjohn: good idea; done

PR #1150, opened 2024-06-14
Remove mutex on copy path

howardjohn:
I don't see any noticeable delta in perf tests, but in flamegraph this is ~0.2% CPU usage range. Very easy to remove it and drop a mutex on the hotpath

### DIFF
--- a/src/copy.rs
+++ b/src/copy.rs
@@ -22,6 +22,8 @@ use std::pin::Pin;
 use std::task::{ready, Context, Poll};
 use tokio::io;
 use tokio::io::{AsyncRead, AsyncWrite};
+use tokio::net::tcp::{OwnedReadHalf, OwnedWriteHalf};
+use tokio::net::TcpStream;
 use tracing::trace;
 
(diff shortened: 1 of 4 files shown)


src/copy.rs
```diff
@@ -45,6 +47,21 @@ where
     }
 }
 
+// TcpStreamSplitter is a specialized BufferedSplitter for TcpStream, which is more efficient than the generic
+// `tokio::io::split`. The generic method involves locking to access the read and write halves
```
bleggett, line 51: If I'm reading the upstream code correctly, we are saving the mutex but losing (a bit) on the heap allocation side?

Since we have to  use `into_split` which the generic `tokio:io:split` func avoids.

howardjohn: Choices are:

* tcp_stream.split: no lock,no heap
* tcp-stream.into_split: no lock, heap
* tokio::io::split: lock, heap

bleggett: Right, and former produces splits that can't be dropped into a spawn. Fair enough.

bleggett (approved):

Issue #1153, opened 2024-06-18
Support dual-stack localhost binding for DNS


howardjohn:
Like https://github.com/istio/istio/blob/bfe4e0f2dcda98400e7fdc2c5883662fcaa385d7/pkg/dns/client/dns.go#L155

PR #1154, opened 2024-06-18
dns: bind to ipv4 and ipv6

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/1153

Without this, you cannot have IPv4 and IPv6 work for dns -- you have to
pick one or the other

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -39,7 +39,6 @@ use tracing::{debug, info, instrument, trace, warn};
 use crate::proxy::SocketFactory;
 
 use crate::config::ProxyMode;
-use crate::dns;
 use crate::dns::metrics::{
     DnsRequest, ForwardedDuration, ForwardedFailure, ForwardedRequest, Metrics,
 };
@@ -52,6 +51,7 @@ use crate::state::workload::address::Address;
 use crate::state::workload::{NetworkAddress, Workload};
 use crate::state::DemandProxyState;
 use crate::strng::Strng;
+use crate::{config, dns};
 
 const DEFAULT_TCP_REQUEST_TIMEOUT: u64 = 5;
 const DEFAULT_TTL_SECONDS: u32 = 30;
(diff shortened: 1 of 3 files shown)


src/dns/server.rs
```diff
@@ -102,32 +102,36 @@ impl Server {
...
-        let udp_socket = socket_factory
-            .udp_bind(addr)
-            .map_err(|e| Error::Bind(addr, e))?;
-        let udp_addr = udp_socket.local_addr().unwrap();
-        server.register_socket(udp_socket);
+            // Bind and register the UDP socket.
+            let udp_socket = socket_factory
+                .udp_bind(addr)
+                .map_err(|e| Error::Bind(addr, e))?;
+            udp_addr = Some(udp_socket.local_addr().unwrap());
```
ilrudie, line 127:
```suggestion
            udp_addr = Some(udp_socket.local_addr().expect("bound udp socket must have a local address"));
```

src/dns/server.rs
```diff
@@ -102,32 +102,36 @@ impl Server {
...
-            .map_err(|e| Error::Bind(addr, e))?;
-        // Save the bound address.
-        let tcp_addr = tcp_listener.local_addr();
-        server.register_listener(
-            tcp_listener.inner(),
-            Duration::from_secs(DEFAULT_TCP_REQUEST_TIMEOUT),
-        );
+        let mut tcp_addr = None;
+        let mut udp_addr = None;
+        for addr in address.into_iter() {
```
ilrudie, line 111: **non-blocking** I think we could get the same result using map on the iter and then take last of the resulting iter so we don't need to use mutable variables. I am not sure it is worth doing because the code may be less understandable and we are doing more than mapping one type onto another.

ilrudie (commented):
looks good, one suggestion to replace unwrap and a thought about removing mutability although since it is quite localized mutability it may not be worth doing.

src/dns/server.rs
howardjohn: yeah I was thinking of that but seemed more complex without much gain

ilrudie: lets leave it as is then

ilrudie (approved):

Issue #899, opened 2024-04-09
Support providing per-workload proxy tasks with drain context


bleggett:
Prompted by https://github.com/istio/ztunnel/pull/898

What we really need beyond that PR with inpod is more granular drains than `ztunnel` is equipped with currently, at a minimum I can imagine the following:

- Drain(reason: workload gone) -> workload gone, don't wait for connections to "unstick" - just drop.
- Drain(reason: proxy terminating) -> workloads are not gone, but whole proxy is going - start refusing conns and give workloads a grace period.

This is relatively easily accomplished with regular `tokio::sync` channels, we should probably just replace our use of https://github.com/linkerd/drain-rs (which is a very light wrapper around those) with use of the tokio sync primitives directly.

ALSO - as part of this we should have draintests for EVERY handler.

- `inbound`
- `outbound`
- `inbound_passthrough`
- `socks5`
- `dns`

that test that they are all guaranteed to drain down to 0 tasks when signaled.

daixiang0:
I do not see the drain-rs crate in the cargo.toml

bleggett:
> I do not see the drain-rs crate in the cargo.toml

It's just called `drain`: https://github.com/istio/ztunnel/blob/master/Cargo.toml#L43

howardjohn:
I think this is mostly done and just 1 final task tracked in https://github.com/istio/ztunnel/issues/1191

PR #1157, opened 2024-06-18
drain: give 5s (configurable) time for existing connections to gracefully close

howardjohn:
This was removed in https://github.com/istio/ztunnel/commit/fa7b1aad000234330388fb4c2f392232e47f9e11#diff-22c9338ddb21f85ab7b656447d453c0d0e0644c647054ca18c4d7d6d43fc8716R100-R113

### DIFF
--- a/src/proxy/inbound_passthrough.rs
+++ b/src/proxy/inbound_passthrough.rs
@@ -18,8 +18,10 @@ use std::time::Instant;
 
 use drain::Watch;
 use tokio::net::TcpStream;
+use tokio::sync::watch;
+use tokio::time::timeout;
 
-use tracing::{error, info, trace, Instrument};
+use tracing::{debug, error, info, trace, warn, Instrument};
 
 use crate::config::ProxyMode;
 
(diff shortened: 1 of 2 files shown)


src/proxy/inbound_passthrough.rs
```diff
@@ -64,21 +66,36 @@ impl InboundPassthrough {
     }
 
     pub(super) async fn run(self) {
+        let (sub_drain_signal, sub_drain) = drain::channel();
+        let deadline = self.pi.cfg.self_termination_deadline;
```
bleggett, line 70: At some point we may want separate vars rather than reusing this for everything - it's overloaded with different meanings already.

Also, we probably should have a more robust form of shutdown signaling with a _single_ channel, rather than creating 2-3 different shutdown channels and using them in concert.

That's probably better handled as part of https://github.com/istio/ztunnel/issues/899 tho.

bleggett (approved):
Since we can't actually signal conn termination in either of these cases, this remains incredibly optimistic and in a lot of cases is probably just a waste of cycles - connection is just as likely to self-terminate in 48 hours as it is in 5 mins, from our perspective.

Doesn't hurt tho.

howardjohn:
> Since we can't actually signal conn termination in either of these cases, this remains incredibly optimistic and in a lot of cases is probably just a waste of cycles - connection is just as likely to self-terminate in 48 hours as it is in 5 mins, from our perspective.

There is IMO a huge difference between 0 and 5s: with 0s, even short lived connections are broken.

if I just curl httpbin.org, its taking ~1s. Without this change it fails, with it it succeeds

bleggett:
> > Since we can't actually signal conn termination in either of these cases, this remains incredibly optimistic and in a lot of cases is probably just a waste of cycles - connection is just as likely to self-terminate in 48 hours as it is in 5 mins, from our perspective.
> 
> There is IMO a huge difference between 0 and 5s: with 0s, even short lived connections are broken.
> 
> if I just curl httpbin.org, its taking ~1s. Without this change it fails, with it it succeeds


If that is behavior we care about, it would probably be best to fence it with a unit test so it doesn't get dropped, and you don't have to manually/personally make sure it's still there 6 months from now.

howardjohn:
Yep I am working on improving tests here

PR #1159, opened 2024-06-18
Drop unused field in WDS

howardjohn:
This is cleanup from an earlier change on Istio side. This is deadcode.

workload.proto changes are copy+paste from Istio


ilrudie (approved):
lgtm

Issue #1152, opened 2024-06-17
dualstack: DNS always returns both IP families regardless of service configuration


howardjohn:
This is a WDS limitation. There is no current way to encode the info.

We have a Service with no vips, and a workload with 2 IPs. In Kubernetes, we have an additional bit of metadata: the ipFamilies.

Without this, we cannot properly return an empty response.

Reproduce with `dig AAAA echo-headless.default.svc.cluster.local`
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: echo-headless
spec:
  selector:
    matchLabels:
      app: echo-headless
  template:
    metadata:
      labels:
        app: echo-headless
    spec:
      securityContext:
        sysctls:
        - name: net.ipv4.ip_unprivileged_port_start
          value: "0"
      containers:
      - name: echo
        image: gcr.io/istio-testing/app:latest
        imagePullPolicy: IfNotPresent
        args:
        - --tcp=9090
        - --tcp=9091
        - --server-first=9091
        - --port=80
        - --grpc=7070
        - --port=443
        - --tls=443
        - --crt=/cert.crt
        - --key=/cert.key
        env:
        - name: INSTANCE_IP
          valueFrom:
            fieldRef:
              fieldPath: status.podIP
---
apiVersion: v1
kind: Service
metadata:
  name: echo-headless
spec:
  selector:
    app: echo-headless
  clusterIP: None
  ports:
  - name: http
    port: 80
  - name: https
    port: 443
  - name: tcp
    port: 9090
  - name: tcp-sf
    port: 9091
  - name: grpc
    port: 7070
```

PR #1161, opened 2024-06-18
Support ip_families field in service

howardjohn:
Istio side https://github.com/istio/istio/pull/51606

This implements the new ip_families field for services:

Add a new IPFamilies field to service. This is used for two things:
  1. DNS lookups for headless services MUST not return unsupported families. Without this field, it is impossible to know which workload IPs to filter
  2. When we select a workload IP after resolving a service, we must pick a supported family. This is _possible_ to do without for non-headless services (by inferring from the VIPs field), but easier to be explicit.

Fixes https://github.com/istio/ztunnel/issues/1152

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -12,14 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::HashSet;
-use std::fmt::{Display, Formatter};
-use std::net::{IpAddr, SocketAddr};
-use std::ops::Deref;
-use std::str::FromStr;
-use std::sync::Arc;
-use std::time::Duration;
-
 use drain::Watch;
 use hickory_proto::error::ProtoErrorKind;
 use hickory_proto::op::ResponseCode;
@@ -34,6 +26,13 @@ use itertools::Itertools;
 use once_cell::sync::Lazy;
 use rand::seq::SliceRandom;
 use rand::thread_rng;
+use std::collections::HashSet;
+use std::fmt::{Display, Formatter};
+use std::net::{IpAddr, SocketAddr};
+use std::ops::Deref;
+use std::str::FromStr;
+use std::sync::Arc;
+use std::time::Duration;
 use tracing::{debug, info, instrument, trace, warn};
 
 use crate::proxy::SocketFactory;
@@ -426,11 +425,15 @@ impl Store {
             Address::Service(service) => {
                 if service.vips.is_empty() {
                     // Headless service. Use the endpoint IPs.
+                    let family = service.ip_families;
                     service
                         .endpoints
                         .iter()
                         .filter_map(|(_, ep)| match &ep.address {
                             Some(addr) => {
+                                if let Some(false) = family.map(|f| f.accepts_ip(addr.address)) {
+                                    return None;
+                                }
                                 if is_record_type(&addr.address, record_type) {
                                     Some(addr.address)
                                 } else {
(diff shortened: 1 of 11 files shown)


src/dns/server.rs
```diff
@@ -426,11 +425,15 @@ impl Store {
             Address::Service(service) => {
                 if service.vips.is_empty() {
                     // Headless service. Use the endpoint IPs.
+                    let family = service.ip_families;
                     service
                         .endpoints
                         .iter()
                         .filter_map(|(_, ep)| match &ep.address {
                             Some(addr) => {
+                                if let Some(false) = family.map(|f| f.accepts_ip(addr.address)) {
```
hzxuzhonghu, line 434: @howardjohn I am not quite understanding this, checking record_type has already filtered unmatched ipfamily.

src/dns/server.rs
```diff
@@ -1627,23 +1656,31 @@ mod tests {
                 "headless.pod0.ns1.svc.cluster.local",
                 &NW1,
                 &[format!("{}/{}", NS1, kube_fqdn("headless", NS1)).as_str()],
-                &[ip("30.30.30.30")],
+                &[ip("30.30.30.30"), ip("2001:db8::30")],
```
MorrisLaw, line 1659: For my own context, where do these IPs come from?

src/proxy/outbound.rs
```diff
@@ -1009,6 +1012,105 @@ mod tests {
...
+                    },
+                    XdsNetworkAddress {
+                        network: "".to_string(),
+                        address: "::3".parse::<Ipv6Addr>().unwrap().octets().into(),
+                    },
+                ],
+                ports: vec![
+                    Port {
+                        service_port: 80,
+                        target_port: 0, // named port
```
MorrisLaw, line 1052: wdym by named port?

src/dns/server.rs
howardjohn: example is a user queries A for a service with IP family Ipv6. We need to move sure we do NOT return ipv4 addresses

bleggett: If this is a clientside defense against a server bug, it needs an error log.

(And a test, this is basically covering a scenario where a v4 is returned for an AAAA record, which is violently against the DNS spec)

howardjohn: This is not protection against a bug. Its perfectly valid for a client to send a A request for a service that is specified as IPv6 only, we just need to return no IPs.

Additionally ,there is a test that covers this exact scenario that I added alongside the PR.

> (And a test, this is basically covering a scenario where a v4 is returned for an AAAA record, which is violently against the DNS spec)

No, I think there is a confusion. This is:
* Client sends A request for Service X
* Service X is IPv6 only, selects pods that are IPv4 AND IPv6
* Before: we would return IPv4 address. Expected/with this PR: we do not respond 

In no scenario are were (with or without this PR) sending IPv4 responses to AAAA or IPv6 responses to A

howardjohn: That being said I think I can short-circuit this check to be at the service level instead of the pod level to give the same behavior.

src/dns/server.rs
howardjohn: the 30.30.30.30 is just random garbage though

src/dns/server.rs
howardjohn: Ok, changed to a service level check. Should be more clear now hopefully

src/proxy/outbound.rs
howardjohn: Named port means in the service.spec.ports.targetPort it is a name instead of a number. This was copy+pastyed from another test though, not needed on this one - I cleaned it up. Thanks

src/dns/server.rs
howardjohn: LMK if that addresses the concerns

hzxuzhonghu (approved):

Issue #1131, opened 2024-06-11
Issue trying ambient mode on an ipv4-only k8s cluster


aek-dsk:
### Is this the right place to submit this?

- [X] This is not a security vulnerability or a crashing bug
- [X] This is not a question about how to use Istio

### Bug Description

Trying ambient mode on the k8s cluster with ipv4 only results in a failing installation via Helm of ztunnel pods (crashing) producing the following error in the logs:

2024-06-04T10:35:38.591582Z	info	ztunnel	version: version.BuildInfo{Version:"909bf991d01edc4db51265bc633acfe303555ef5", GitRevision:"909bf991d01edc4db51265bc633acfe303555ef5", RustVersion:"1.77.2", BuildProfile:"release", BuildStatus:"Clean", GitTag:"1.22.0-beta.1-6-g909bf99", IstioVersion:"unknown"}	
2024-06-04T10:35:38.591778Z	info	ztunnel	running with config: proxy: true
dnsProxy: false
windowSize: 4194304
connectionWindowSize: 4194304
frameSize: 1048576
poolMaxStreamsPerConn: 100
poolUnusedReleaseTimeout:
  secs: 300
  nanos: 0
socks5Addr: null
adminAddr: 127.0.0.1:15000
statsAddr: '[::]:15020'
readinessAddr: '[::]:15021'
inboundAddr: '[::]:15008'
inboundPlaintextAddr: '[::]:15006'
outboundAddr: '[::]:15001'
dnsProxyAddr: 127.0.0.1:15053
network: ''
localNode: f74ac986-14da-431f-91b0-9bd7f0fdb6e8
proxyMode: Shared
localIp: 11.32.10.9
clusterId: Kubernetes
clusterDomain: cluster.local
caAddress: https://istiod.ccp-istio-system.svc:15012
caRootCert: !File ./var/run/secrets/istio/root-cert.pem
xdsAddress: https://istiod.ccp-istio-system.svc:15012
xdsRootCert: !File ./var/run/secrets/istio/root-cert.pem
secretTtl:
  secs: 86400
  nanos: 0
xdsOnDemand: false
fakeCa: false
selfTerminationDeadline:
  secs: 5
  nanos: 0
proxyMetadata:
  DNS_AUTO_ALLOCATE: 'false'
  ISTIO_VERSION: 1.22.0
  CLUSTER_ID: Kubernetes
  DNS_CAPTURE: 'false'
  ENABLE_HBONE: 'true'
[...]

PR #1163, opened 2024-06-20
Add option for disabling IPv6

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/1131

Testing was manual. I don't see any way we can test this in CI, since it
only works if the kernel is booted with IPv6 turned off - the sysctl to
dynamically set it will NOT break.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -56,6 +56,7 @@ const POOL_UNUSED_RELEASE_TIMEOUT: &str = "POOL_UNUSED_RELEASE_TIMEOUT";
 const CONNECTION_TERMINATION_DEADLINE: &str = "CONNECTION_TERMINATION_DEADLINE";
 const ENABLE_ORIG_SRC: &str = "ENABLE_ORIG_SRC";
 const PROXY_CONFIG: &str = "PROXY_CONFIG";
+const IPV6_DISABLED: &str = "IPV6_DISABLED";
 
 const UNSTABLE_ENABLE_SOCKS5: &str = "UNSTABLE_ENABLE_SOCKS5";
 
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -609,25 +614,28 @@ impl IntoIterator for Address {
...
         }
     }
 }
 
-impl FromStr for Address {
-    type Err = anyhow::Error;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
+impl Address {
+    fn from_str(ipv6_enabled: bool, s: &str) -> anyhow::Result<Self> {
```
ilrudie, line 634: **bikeshed:** I wonder if we should use different name entirely since we're not implementing the FromStr trait anymore and it's not really making an address from a string... Maybe just `new`

ilrudie (approved):
Looks good generally. I left a comment/bikeshed idea but I don't feel strongly enough that the current state is wrong to block. I'll add a hold in case you want to do something with the naming but also feel free to remove it if you think the current name is good enough.

src/config.rs
howardjohn: :+1:

src/config.rs
```diff
@@ -56,6 +56,7 @@ const POOL_UNUSED_RELEASE_TIMEOUT: &str = "POOL_UNUSED_RELEASE_TIMEOUT";
 const CONNECTION_TERMINATION_DEADLINE: &str = "CONNECTION_TERMINATION_DEADLINE";
 const ENABLE_ORIG_SRC: &str = "ENABLE_ORIG_SRC";
 const PROXY_CONFIG: &str = "PROXY_CONFIG";
+const IPV6_DISABLED: &str = "IPV6_DISABLED";
```
bleggett, line 59: To keep the semantics similar with https://github.com/istio/istio/blob/master/manifests/charts/istio-cni/values.yaml#L55 (and the rest of Istio), I'd prefer to invert this.

These two settings will 90% of the time be kept in lockstep, and having one be (disable) and the other be (enable) is needlessly confusing.

bleggett: (if we want to make both `disable` that's fine too - but they should be consistent, and changed together)

PR #1165, opened 2024-06-20
admin: bind to both IP families

howardjohn:


bleggett (commented):
We should conditionally bind based on https://github.com/istio/ztunnel/pull/1163, no?

howardjohn:
> We should conditionally bind based on #1163, no?

```
        admin_addr: Address::Localhost(
            ipv6_enabled,
```

handles that

bleggett:
> > We should conditionally bind based on #1163, no?
> 
> ```
>         admin_addr: Address::Localhost(
>             ipv6_enabled,
> ```
> 
> handles that

Missed it, ty!

bleggett (approved):

PR #1166, opened 2024-06-20
Add option for disabling IPv6 (#1163)

howardjohn:
* Add option for disabling IPv6

Fixes https://github.com/istio/ztunnel/issues/1131

Testing was manual. I don't see any way we can test this in CI, since it
only works if the kernel is booted with IPv6 turned off - the sysctl to
dynamically set it will NOT break.

* rename

(cherry picked from commit 3a41b096746580a6a7f1c956717b2b38024dd48c)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -54,6 +54,7 @@ const POOL_MAX_STREAMS_PER_CONNECTION: &str = "POOL_MAX_STREAMS_PER_CONNECTION";
 const POOL_UNUSED_RELEASE_TIMEOUT: &str = "POOL_UNUSED_RELEASE_TIMEOUT";
 const ENABLE_ORIG_SRC: &str = "ENABLE_ORIG_SRC";
 const PROXY_CONFIG: &str = "PROXY_CONFIG";
+const IPV6_DISABLED: &str = "IPV6_DISABLED";
 
 const UNSTABLE_ENABLE_SOCKS5: &str = "UNSTABLE_ENABLE_SOCKS5";
 
@@ -264,6 +265,12 @@ fn parse_proxy_config() -> Result<ProxyConfig, Error> {
 }
 
 pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
+    let ipv6_enabled = !parse::<bool>(IPV6_DISABLED)?.unwrap_or_default();
+    let bind_wildcard = if ipv6_enabled {
+        IpAddr::V6(Ipv6Addr::UNSPECIFIED)
+    } else {
+        IpAddr::V4(Ipv4Addr::UNSPECIFIED)
+    };
     let default_istiod_address = if env::var(KUBERNETES_SERVICE_HOST).is_ok() {
         "https://istiod.istio-system.svc:15012".to_string()
     } else {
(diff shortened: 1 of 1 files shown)


howardjohn:
Fixes https://github.com/istio/ztunnel/issues/1164

kfaseela:
does this have any impact on https://github.com/istio/test-infra/pull/5401

howardjohn:
@kfaseela not really, this is about making it so you can run on a pure IPv4 kernel. The other flag is about making IPv6 work. Actually they are really for opposite purposes

bleggett:
> @kfaseela not really, this is about making it so you can run on a pure IPv4 kernel. The other flag is about making IPv6 work. Actually they are really for opposite purposes

Are they? I'm not sure how? 

- If you want pure IPv4 you set `ambient.enableIPv6=false`
- If you want (v4 && v6 or just V6) `ambient.enableIPv6=true`

That's all the cases?

howardjohn:
@bleggett the majority of users run on ipv4 only platforms which have kernel support for IPv6. So there is really 3 cases:
1. No IPv6 at all
2. IPv6 kernel but only using ipv4 in kubernetes
3.  Full IPv6+IPv4

We just collapse (2) and (3) under one flag. Meaningfully, we do _not_ do this in sidecars, but that is mostly legacy decisions.

bleggett:
> @bleggett the majority of users run on ipv4 only platforms which have kernel support for IPv6. So there is really 3 cases:
> 
>     1. No IPv6 at all
> 
>     2. IPv6 kernel but only using ipv4 in kubernetes
> 
>     3. Full IPv6+IPv4
> 
> 
> We just collapse (2) and (3) under one flag. Meaningfully, we do _not_ do this in sidecars, but that is mostly legacy decisions.

I get that, I am asking why we need to support a configurable distinction between 1 and 2 there. Simpler to treat them as the same case?

You either want to use V6 or you do not (using it in concert with V4 is sort of unrelated to that). If you do not, do we need to make a distinction between (kernel support || kube support) in our config API surface? I don't see that we need to.

zirain (approved):

howardjohn:
We don't, I am agreeing with 2 config knobs. The disagreement (if there is one? I am not sure?) is whether it is "1 or {2+3}" or "{1+2} or {3}".

Given (1) is legacy niche poorly-configured OS, I am very inclined to make that a niche opt-out and let the rest of the world live in a modern state.

bleggett:
@howardjohn Agree with that, but am mostly concerned that until we can safely enable-v6-by-default in the rest of Istio as well, we will have to carry two flags which

- do the same thing
- both need to be synchronized
- will (confusingly) need to have inverted values

howardjohn:
The plan was to enable IPv6 by default in 1.23, and send another PR to have
to fix this up to invert the bool to be consistent.

howardjohn:
There is basically no flags unless you are in the 0.01% of users that
disable IPv6 in the kernel. Then there is a consistent flag to set on
ztunnel and CNI

aek-dsk:
> We don't, I am agreeing with 2 config knobs. The disagreement (if there is one? I am not sure?) is whether it is "1 or {2+3}" or "{1+2} or {3}".
> 
> Given (1) is legacy niche poorly-configured OS, I am very inclined to make that a niche opt-out and let the rest of the world live in a modern state.

@howardjohn @bleggett : Ehmm...It would be possible to argue that it is not a legacy poorly-configured OS, but actually a well hardened system that opts out for all unused modules to minimize the attack surface.

A recommended security measure compared to the most of the 'out of the box' installations.

howardjohn:
> > We don't, I am agreeing with 2 config knobs. The disagreement (if there is one? I am not sure?) is whether it is "1 or {2+3}" or "{1+2} or {3}".
> > Given (1) is legacy niche poorly-configured OS, I am very inclined to make that a niche opt-out and let the rest of the world live in a modern state.
> 
> @howardjohn @bleggett : Ehmm...It would be possible to argue that it is not a legacy poorly-configured OS, but actually a well hardened system that opts out for all unused modules to minimize the attack surface.
> 
> A recommended security measure compared to the most of the 'out of the box' installations.

Regardless of my personal commentary on anti-IPv6 folks, it is fully supported in the product now :slightly_smiling_face:

PR #1169, opened 2024-06-21
Optimize throughput 

howardjohn:
This has two main optimizations:
* Allow larger buffers, up to 256k. Before we maxed out at 16k. Note this is dynamic, so we aren't just using more memory.
* Avoid copy of all data on the tcp -> Hbone send side. This requires us to use Bytes everywhere instead of `&[u8]` so we can give h2 a copy for free (ref count) vs a clone

Highlight:
```
throughput/hbone        
                        thrpt:  [12.506 Gb/s 12.629 Gb/s 12.768 Gb/s]
                 change:thrpt:  [+41.815% +43.964% +46.422%]
```

### DIFF
--- a/src/copy.rs
+++ b/src/copy.rs
@@ -396,3 +396,53 @@ where
         AsyncWriteBuf::poll_shutdown(Pin::new(me.a), cx)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tokio::io::{AsyncWriteExt};
+    use tokio::io::{AsyncReadExt};
+    use crate::test_helpers::helpers::initialize_telemetry;
+
+    #[tokio::test]
+    async fn copy() {
+        initialize_telemetry();
+        let (mut client, ztunnel_downsteam) = tokio::io::duplex(32000);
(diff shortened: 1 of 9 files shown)


src/copy.rs
```diff
@@ -239,38 +310,18 @@ impl<R: AsyncRead> BufReader<R> {
...
-            *me.pos = 0;
-        }
-        Poll::Ready(Ok(&me.buf[*me.pos..*me.cap]))
-    }
-
-    fn consume(self: Pin<&mut Self>, amt: usize) {
-        let me = self.project();
-        *me.pos = cmp::min(*me.pos + amt, *me.cap);
+        // Give us enough space to read a full chunk
+        me.buf.reserve(*me.buffer_size);
```
ilrudie, line 317: If the buffer size was increased it looks like allocation would be done here, at reserve during this poll, rather than during the resize. Is that on purpose? I suppose it should mean that the buffer is empty when we allocate to reduce any potential data movement.

bleggett (approved):

ilrudie (commented):
mostly looks good, want to understand the timing around allocations for resize a bit better though.

PR #1171, opened 2024-06-24
benchmarks: benchmark various levels of concurrency

howardjohn:
Additionally, properly shutdown ztunnel in tests. This ensures we do not have a bunch of stale ztunnels when we have multiple tests (noticed since now we have a lot of them)


ilrudie (approved):
lgtm

PR #1176, opened 2024-06-25
Abstract and fix draining

howardjohn:
* Centralize draining logic in one helper function
* Fix inbound draining (HBONE). Before, we did not shut down the
  listener upon draining. This meant new connections would go to the old
ztunnel on a ztunnel restart.
* Simplify inbound draining; do not re-create the force shutdown logic,
  and instead let the common abstraction do it (which does it slightly
better)
* socsk5: add propery draining with force shutdown. Remove double-spawn,
  which adds some complexity around the proxy_to_cancellable.

This is primarily tested in https://github.com/istio/istio/pull/51710,
which sends a large stream of requests and restarts ztunnel and the
backend app (2 different tests). With this change, these tests pass.

It would be good to get more isolated tests in this repo in the future
as well

### DIFF
--- a/src/drain.rs
+++ b/src/drain.rs
@@ -17,39 +17,43 @@ use std::time::Duration;
 use tokio::sync::watch;
 use tracing::{debug, info, warn};
 
-#[derive(Debug)]
-pub struct DrainTrigger(drain_internal::Signal);
+// #[derive(Debug)]
+// pub struct DrainTrigger(internal::Signal);
 
-impl DrainTrigger {
-    /// start_drain_and_wait initiates a draining sequence. The future will not complete until the drain
-    /// is complete (all outstanding DrainWatchers are dropped).
-    pub async fn start_drain_and_wait(self) {
-        self.0.drain().await
-    }
-}
+// impl DrainTrigger {
+//     /// start_drain_and_wait initiates a draining sequence. The future will not complete until the drain
+//     /// is complete (all outstanding DrainWatchers are dropped).
+//     pub async fn start_drain_and_wait(self) {
+//         self.0.drain().await
+//     }
+// }
 
-#[derive(Clone, Debug)]
-pub struct DrainWatcher(drain_internal::Watch);
+// #[derive(Clone, Debug)]
+pub use internal::DrainMode;
+pub use internal::ReleaseShutdown as DrainBlocker;
+pub use internal::Signal as DrainTrigger;
+pub use internal::Watch as DrainWatcher;
+// pub struct DrainWatcher(internal::Watch);
+//
+// impl DrainWatcher {
+//     /// wait_for_drain will return once a drain has been initiated.
+//     /// The drain will not complete until the returned DrainBlocker is dropped
+//     pub async fn wait_for_drain(self) -> DrainBlocker {
+//         DrainBlocker(self.0.signaled().await)
+//     }
+// }
 
-impl DrainWatcher {
-    /// wait_for_drain will return once a drain has been initiated.
-    /// The drain will not complete until the returned DrainBlocker is dropped
-    pub async fn wait_for_drain(self) -> DrainBlocker {
-        DrainBlocker(self.0.signaled().await)
-    }
-}
-
-#[allow(dead_code)]
+// #[allow(dead_code)]
 /// DrainBlocker provides a token that must be dropped to unblock the drain.
-pub struct DrainBlocker(drain_internal::ReleaseShutdown);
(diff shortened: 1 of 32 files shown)


src/proxy/util.rs
```diff
@@ -12,7 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::proxy::ProxyInputs;
+use drain::Watch;
```
bleggett, line 16: If we are going to do this abstraction + wrapper move (which I do like), I would suggest we take the opportunity kick `drain` to the curb, it's redundant wrt the drain functionality we already get with `tokio`, which was one of the points in https://github.com/istio/ztunnel/issues/899

bleggett (commented):
How much of https://github.com/istio/ztunnel/issues/899 does this resolve? Specifically allowing drain logic to distinguish between (ztunnel shutdown) || (workload shutdown)?

src/proxy/util.rs
howardjohn: I partially did this by making our own simpler/more clearly named `drain` crate. but it is implemented in terms of the existing `drain` crate we use, since I don't see a problem with their implementation. LMK what you think

howardjohn: It also makes it trivial to swap out if there is a better impl

bleggett: Yep yep, it's a definite improvement, we need our own drain crate either way. I don't have a problem with their impl I just think it's strictly redundant with/easily replaced by a handful of `tokio::sync` primitives for our uses, and we don't need it.

But that's wishlisting, not PR review.

src/drain.rs
```diff
@@ -0,0 +1,80 @@
...
+
+/// New constructs a new pair for draining
+/// * DrainTrigger can be used to start a draining sequence and wait for it to complete.
+/// * DrainWatcher should be held by anything that wants to participate in the draining. This can be cloned,
+///   and a drain will not complete until all outstanding DrainWatchers are dropped.
+pub fn new() -> (DrainTrigger, DrainWatcher) {
+    let (tx, rx) = drain_internal::channel();
+    (DrainTrigger(tx), DrainWatcher(rx))
+}
+
```
bleggett, line 40: Why do we need `DrainBlocker` and friends? `tokio::sync::watch` supports all the same semantics and has `closed().await` which already blocks until all listeners are dropped? We can just use that, or am I missing something?

howardjohn: Here is my thinking:

* Having nicely named things, with clear semantics (i.e. the ONLY thing you can do with a DrainTrigger is `start_drain_and_wait`) is a good idea, regardless of the underlying implementation. If I just have a `Watch`, its not clear what its for, and how to use it.
* If we abstract it ,we can change the underlying implementation to directly use `tokio::*` instead of `drain::*`, if there is benefit to do so

bleggett: > If we abstract it ,we can change the underlying implementation to directly use tokio::* instead of drain::*, if there is benefit to do so

Yeah my point is I think the abstraction doesn't actually provide any benefit (other than the type name alias) if we just use `tokio::sync::watch`. Additionally, doing that lets us drop a Cargo dep we really don't need, that (IIRC) we are now only using here - and dropping deps is always good, especially in `ztunnel`.

I am fine wrapping `tokio::sync::watch` with this wrapper if we want to keep the type alias, but I don't want to keep `drain` AND the type alias if regular old `tokio::sync::watch` (which is already one of our deps) will do the job.

howardjohn: Here is what I want:
* A wrapper with reasonable names and exactly one method. Watch has bad names (since they are general) and too many methods (since its general) making it hard to use
* To not spend time _right now_ migrating off `drain` since it works fine.

If someone wants to swap out the underlying implement of `drain`, now its trivial to do so -- edit the tiny new package. I am not actually convinced its better (maybe it is, just haven't looked into it), though. I do not think we can SIMPLY stick a watch in there, probably we will end up with 2, and then we have just re-inventing `drain`.Which is fine if we have reasons to, but its super low priority, and we need this PR anyways.

howardjohn:
I think most of this is good to go regardless but in light of https://github.com/istio/istio/pull/51805 I might change do some work to not drain on pod shutdown, only ztunnel. or I can do that in a followup, either one works forme

bleggett:
> I think most of this is good to go regardless but in light of [istio/istio#51805](https://github.com/istio/istio/pull/51805) I might change do some work to not drain on pod shutdown, only ztunnel. or I can do that in a followup, either one works forme

Fair enough!

bleggett (approved):

PR #1178, opened 2024-06-25
Conditionally bind to localhost for partial IPv6 support

howardjohn:
Similar to https://github.com/istio/istio/pull/51712

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -14,12 +14,14 @@
 
 use std::collections::{HashMap, HashSet};
 use std::fmt::{Display, Formatter};
+use std::fs::File;
+use std::io::Read;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
-use std::{env, fs};
+use std::{env, fs, io};
 
 use anyhow::anyhow;
 use bytes::Bytes;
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -341,9 +364,9 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
     // TODO: should we override server_ordering_strategy based on our IP support?
 
     let dns_proxy_addr: Address = match pc.proxy_metadata.get(DNS_PROXY_ADDR_METADATA) {
-        Some(dns_addr) => Address::new(ipv6_enabled, dns_addr)
+        Some(dns_addr) => Address::new(ipv6_localhost_enabled, dns_addr)
```
bleggett, line 367: How does this work if there's a mismatch between whether the zt pod has V6 disabled on `lo` and a random pod making use of DNS proxy doesn't, or vice-versa?

src/config.rs
```diff
@@ -272,8 +274,29 @@ fn parse_proxy_config() -> Result<ProxyConfig, Error> {
     construct_proxy_config(mesh_config_path, pc_env).map_err(Error::ProxyConfig)
 }
 
+const IPV6_DISABLED_LO: &str = "/proc/sys/net/ipv6/conf/lo/disable_ipv6";
```
bleggett, line 277: If we wanted to get really wild we could actually have `istio-cni` inform us of this at the app pod level via zds.

src/config.rs
howardjohn: Ugh. Forgot about that. Oddly, this fix seems to work in my environment, and the host has IPv6 but pods don't

howardjohn: Oh, duh, ztunnel is not in the host.

howardjohn: Ok I have fixed it but it comes with some complexity. Not too bad

src/config.rs
```diff
@@ -274,6 +274,15 @@ fn parse_proxy_config() -> Result<ProxyConfig, Error> {
 
 pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
     let ipv6_enabled = parse::<bool>(IPV6_ENABLED)?.unwrap_or(true);
+    let ipv6_localhost_enabled = if ipv6_enabled {
```
bleggett, line 277: I actually think `ipv6_localhost_enabled` _shouldn't_ be proxywide static config, since it's not, really - it's context-sensitive.

I actually prefer doing the check in `socket_factory` at bind-time or close to it with helpers, rather than trying to have a canonical config value for it.

Can we just do that, and remove `ipv6_localhost_enabled` from the config object?

howardjohn: `ipv6_localhost_enabled` never actually ends up into `Config` fwiw. Its just a local var. At the time I think this made a lot of sense; I did not want to check it on each access since the check is not free (its pretty quick, but still).

But with the DNS change its a bit murkier. In some cases, we have a proxy-wide config (admin), but in others we have per-proxy. In the proxy-wide cases, we don't even have a socket_factory at all

bleggett: > ipv6_localhost_enabled never actually ends up into Config fwiw. Its just a local var. At the time I think this made a lot of sense; I did not want to check it on each access since the check is not free (its pretty quick, but still).

Ah right - I was thinking it might make more sense to actually put it in `admin.rs` as a helper, since that is the only place we do a proxywide bind.

> But with the DNS change its a bit murkier. In some cases, we have a proxy-wide config (admin), but in others we have per-proxy. In the proxy-wide cases, we don't even have a socket_factory at all

Yes that's fair, I was thinking that since we don't use it as a nonrecomputed var anywhere but `admin` it maybe should live in `admin.rs` or similar as a local opinion for that, and the fact that we hard-check it again in the DNS server anyway makes me think the recomputed version might be better kept in `socket_factory` - where we could use it for pods too, right?

howardjohn: Well technically we have 4 things using Address today, just only 2 of them generally are bound to localhost. The obvious way to defer the "use ipv6" is to make into_iter() take a "supports v6" flag, but then you need to always pass that on all calls.

That is still only 3+N calls, though. I am fine either way, WDYT?

bleggett: not a big deal, yeah.

bleggett (approved):

PR #1179, opened 2024-06-25
Fix edge case in delayed workload case

howardjohn:
'`async fn` resumed after completion' panic occurs when we hit the loop
twice. it is illegal to await a completed async function. For this case,
we are supposed to call changed each time.

Make the test case more comprehensive to catch this

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -645,21 +645,19 @@ impl DemandProxyState {
         // loop until the subscriber wakes us on new workload,
         // or we hit the deadline timeout and give up
         let timeout = tokio::time::sleep(deadline);
-        let subscriber = wl_sub.changed();
         tokio::pin!(timeout);
-        tokio::pin!(subscriber);
         loop {
             tokio::select! {
                 _ = &mut timeout => {
                     warn!("timed out waiting for workload from xds");
                     break None;
                 },
-                _ = &mut subscriber => {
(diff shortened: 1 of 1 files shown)


bleggett (approved):

src/state.rs
```diff
@@ -983,8 +981,16 @@ mod tests {
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn test_wait_for_workload_eventually() {
+        initialize_telemetry();
         let state = ProxyState::default();
         let wrap_state = Arc::new(RwLock::new(state));
+        let not_delayed_wl = Arc::new(Workload {
```
bleggett, line 987: Yeah this is on me for not having more robust test cases, ty.

src/state.rs
```diff
@@ -645,21 +645,19 @@ impl DemandProxyState {
...
         tokio::pin!(timeout);
-        tokio::pin!(subscriber);
         loop {
             tokio::select! {
                 _ = &mut timeout => {
                     warn!("timed out waiting for workload from xds");
                     break None;
                 },
-                _ = &mut subscriber => {
+                _ = wl_sub.changed() => {
```
bleggett, line 655: Yeah, docs do state it's cancel-safe. Shoulda caught this.

PR #1180, opened 2024-06-25
Better proto err logging/announce retries

bleggett:
I _think_ this will fix https://github.com/istio/istio/issues/51709

Context: We used to retry on protocol errors indefinitely. https://github.com/istio/ztunnel/pull/1136 made protocol errors fatal. Note that we always have retried on _unix socket connection failures_, but until https://github.com/istio/ztunnel/pull/1136 didn't fail on _protocol_ errors once the connection is established.

https://github.com/istio/istio/pull/51574 added a test which forcibly recycles `istio-cni` and makes sure zt can recover without the dataplane being disrupted (it should).

This _mostly_ works but sometimes we can get [unlucky in that test ](https://storage.googleapis.com/istio-prow/logs/integ-ambient_istio_postsubmit/1805371660040146944/artifacts/ambient-cni-64174fcad69e4bd89a4/TestTrafficWithEstablishedPodsIfCNIMissing/jwt-claim-route/matched_with_nested_claim_using_claim_to_header:200/to_captured/_test_context/istio-state549725766/primary-0/ztunnel-kr5zs_istio-proxy.previous.log) and (I think) get disconnected

1. After we successfully "accept"
2. Before we successfully send our announcement (which is always the first thing we do, and which is not ACK'd)

Now, we could just do what we used to do and retry on any failed send/recv, waiting for a new connection attempt at that point, but that gets back into the problem of spamming the node agent with spurious connections if we happen to have a proto mismatch and the announce is doomed to fail for as long as we have that node agent on our node.

Basically, because this is a very simple DIY UDP protocol, we have no way to tell why the other end hung up before we even get started - because they shut down, or because they killed the conn due to fundamental proto mismatch.

So this just does what @howardjohn originally suggested in  #1136 - adds a backoff _if_ the first time we try to send something we fail.

We reset the backoff when our internal readiness state goes green.

This is still not ideal - we will fail infinitely (well, aside from kube readiness probe limits) if it happens to be a protocol mismatch, until we get a connection to a server that can talk to us - but will fail infinitely with a backoff and a (capped) timeout, so it's a tad nicer, and should fix the issue where we have unlucky timing around `istio-cni` termination.

### DIFF
--- a/src/inpod/workloadmanager.rs
+++ b/src/inpod/workloadmanager.rs
@@ -184,10 +184,10 @@ impl WorkloadProxyManager {
                 Ok(()) => {
                     info!("process stream ended with eof");
                 }
-                Err(Error::ProtocolError) => {
+                Err(Error::ProtocolError(e)) => {
                     error!("protocol mismatch error while processing stream, shutting down");
                     self.readiness.not_ready();
-                    return Err(anyhow::anyhow!("protocol error"));
+                    return Err(anyhow::anyhow!("protocol error {:?}", e));
                 }
                 Err(e) => {
                     // for other errors, just retry
@@ -263,7 +263,7 @@ impl<'a> WorkloadProxyManagerProcessor<'a> {
         processor
             .send_hello()
             .await
-            .map_err(|_| Error::ProtocolError)?;
+            .map_err(|_| Error::ProtocolError("could not announce to node agent".into()))?;
 
         loop {
             let msg = match self.read_message_and_retry_proxies(&mut processor).await {
--- a/src/inpod/workloadmanager.rs
+++ b/src/inpod/workloadmanager.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use crate::readiness;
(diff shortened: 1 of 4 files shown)


src/inpod/workloadmanager.rs
```diff
@@ -184,17 +201,42 @@ impl WorkloadProxyManager {
...
+                    error!("could not announce to node agent, retrying with backoff");
+                    self.readiness.not_ready();
+
+                    // This will retry infinitely for as long as the socket doesn't EOF, but not immediately.
+                    let wait = self
+                        .readiness
+                        .backoff
+                        .next_backoff()
+                        .unwrap_or(CONNECTION_FAILURE_RETRY_DELAY_MAX_INTERVAL);
+                    warn!("node agent connect failed ({e}), retrying in {wait:?}");
```
howardjohn, line 225: nit: no need to log here and on L216?

howardjohn (approved):

PR #1184, opened 2024-06-26
Drain: allow more dynamic configuration

howardjohn:
tl;dr: allow us to set one value for "draining" from the user
perspective, which sets both `spec.terminationGracePeriodSeconds` and
ztunnel's drain period.

TBH, this isn't super valuable without
https://github.com/kubernetes/kubernetes/pull/125746, since Helm could
just do the math, but its nice to future proof?.


ilrudie (approved):
lgtm

Issue #1155, opened 2024-06-18
Emit end-of-process access logs


howardjohn:
Currently, when we drain we lose access logs for dropped connections. We should emit them as we terminate them.

PR #1187, opened 2024-06-27
Keep track of logs for terminated connections

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/1155

This ensures when we drain a process we emit metrics


ilrudie (approved):
looks good to me

Issue #1183, opened 2024-06-25
Do not error if the app sends a RST


howardjohn:
Some apps shutdown like this, like postgres over SSL.

Envoy does NOT log an error in this case:

```
2024-06-25T21:47:44.375498Z     trace   envoy filter external/envoy/source/common/tcp_proxy/tcp_proxy.cc:834    [Tags: "ConnectionId":"22"] upstream connection received 24 bytes, end_stream=false     thread=27
2024-06-25T21:47:44.375501Z     trace   envoy connection external/envoy/source/common/network/connection_impl.cc:529    [Tags: "ConnectionId":"22"] writing 24 bytes, end_stream false  thread=27
2024-06-25T21:47:44.375506Z     trace   envoy connection external/envoy/source/common/network/connection_impl.cc:614    [Tags: "ConnectionId":"23"] socket event: 2     thread=27
2024-06-25T21:47:44.375508Z     trace   envoy connection external/envoy/source/common/network/connection_impl.cc:737    [Tags: "ConnectionId":"23"] write ready thread=27
2024-06-25T21:47:44.375510Z     trace   envoy connection external/envoy/source/common/network/connection_impl.cc:614    [Tags: "ConnectionId":"22"] socket event: 2     thread=27
2024-06-25T21:47:44.375511Z     trace   envoy connection external/envoy/source/common/network/connection_impl.cc:737    [Tags: "ConnectionId":"22"] write ready thread=27
2024-06-25T21:47:44.375544Z     trace   envoy connection external/envoy/source/common/network/raw_buffer_socket.cc:70   [Tags: "ConnectionId":"22"] write returns: 24   thread=27
2024-06-25T21:47:44.375548Z     trace   envoy connection external/envoy/source/common/network/connection_impl.cc:614    [Tags: "ConnectionId":"22"] socket event: 3     thread=27
2024-06-25T21:47:44.375550Z     trace   envoy connection external/envoy/source/common/network/connection_impl.cc:737    [Tags: "ConnectionId":"22"] write ready thread=27
2024-06-25T21:47:44.375551Z     trace   envoy connection external/envoy/source/common/network/connection_impl.cc:654    [Tags: "ConnectionId":"22"] read ready. dispatch_buffered_data=0        thread=27
2024-06-25T21:47:44.375553Z     trace   envoy connection external/envoy/source/common/network/raw_buffer_socket.cc:25   [Tags: "ConnectionId":"22"] read returns: 0     thread=27
[...]

PR #1188, opened 2024-06-27
Improve logging of errors in copy

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/1183.

Before, we logged basically any error during copying. It turns out this
confuses users a lot when they see "connection reset by peer" just
because the other end shutdown. There are a number of sequences of
events that can lead to non-graceful termination of copying, and
exposing those to users just brings confusion based on some feedback.

For instance, Postgresql always closes connections with a hard RST.
This, previously, meant every postgres connection was reported as an
"error" in Ztunnel - no good.

Looking at Envoy codebase, they do NOT report these as errors: once the
connection is started, it can close for any reason silently (*except at
debug level, of course).

We take the same approach here. Also, add more context on where we are
failing (shutdown, write, etc)

### DIFF
--- a/src/copy.rs
+++ b/src/copy.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::proxy;
 use crate::proxy::ConnectionResult;
 use crate::proxy::Error::{BackendDisconnected, ClientDisconnected, ReceiveError, SendError};
 use bytes::{Buf, Bytes, BytesMut};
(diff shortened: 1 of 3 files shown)


src/copy.rs
```diff
@@ -149,42 +150,68 @@ where
 {
     let (mut rd, mut wd) = downstream.split_into_buffered_reader();
     let (mut ru, mut wu) = upstream.split_into_buffered_reader();
-
     let downstream_to_upstream = async {
-        let res = copy_buf(&mut rd, &mut wu, stats, false).await;
+        let report = |e: io::Error| {
```
ilrudie, line 154: report as a name was pretty confusing to me. It holds a fn which converts io::ErrorKind into proxy::Error, yeah? I think something like `into_proxy_error` or `into_senderror` would clarify.

src/copy.rs
```diff
@@ -149,42 +150,68 @@ where
...
-        .map_err(|e| ReceiveError(Box::new(e)))?;
+    let sent = sent?;
+    let received = received?;
     trace!(sent, received, "copy complete");
     Ok(())
 }
 
+// During copying, we may encounter errors from either side closing their connection. Typically, we
+// get a fully graceful shutdown with no errors on either end, but can if one end sends a RST directly,
+// or if we have
```
ilrudie, line 198: intentional? comment reads like it is incomplete

src/copy.rs
```diff
@@ -149,42 +150,68 @@ where
...
+fn ignore_io_errors<T: Default>(res: Result<T, io::Error>) -> Result<T, io::Error> {
+    use io::ErrorKind::*;
+    match &res {
+        Err(e) => match e.kind() {
+            NotConnected | UnexpectedEof | ConnectionReset | BrokenPipe => {
+                trace!(err=%e, "io terminated ungracefully");
+                // Returning Default here is vey hacky, but the data we are returning isn't critical so its no so bad to lose it.
```
ilrudie, line 205: sp

```suggestion
                // Returning Default here is very hacky, but the data we are returning isn't critical so its no so bad to lose it.
```

ilrudie (commented):
I like it. Minor objection to naming of things though. Hardest problem in computer science...

src/copy.rs
```diff
@@ -149,42 +150,68 @@ where
 {
     let (mut rd, mut wd) = downstream.split_into_buffered_reader();
     let (mut ru, mut wu) = upstream.split_into_buffered_reader();
-
     let downstream_to_upstream = async {
-        let res = copy_buf(&mut rd, &mut wu, stats, false).await;
+        let report = |e: io::Error| {
+            SendError(Box::new(match e.kind() {
+                io::ErrorKind::NotConnected => BackendDisconnected,
```
MorrisLaw, line 156: `downstream_to_upstream` matches errors `NotConnected` and `WriteZero` to `BackendDisconnected` whereas `upstream_to_downstream` does the opposite and matches `NotConnected` and `WriteZero` to `ClientDisconnected`

Can you provide context for why it's different?

howardjohn: Yes, these are opposite ends of the connection. So they indicate opposite sides. its bidirectional

MorrisLaw: Ah ok that makes sense, thanks for the clarification!

src/copy.rs
howardjohn: ty

src/copy.rs
```diff
@@ -149,42 +150,72 @@ where
 {
     let (mut rd, mut wd) = downstream.split_into_buffered_reader();
     let (mut ru, mut wu) = upstream.split_into_buffered_reader();
-
     let downstream_to_upstream = async {
-        let res = copy_buf(&mut rd, &mut wu, stats, false).await;
+        let translate_error = |e: io::Error| {
```
ilrudie, line 154: I like it! TY

ilrudie (approved):
lgtm

PR #1205, opened 2024-07-12
More XDS metrics

### LINKED ISSUE #1104: Add more XDS operational metrics
* Number of messages
* Total message size (counter) by type

Stevenjin8:
Resolves #1104.

### DIFF
--- a/src/xds/metrics.rs
+++ b/src/xds/metrics.rs
@@ -15,7 +15,7 @@
 use prometheus_client::encoding::{EncodeLabelSet, EncodeLabelValue};
 use prometheus_client::metrics::counter::Counter;
 use prometheus_client::metrics::family::Family;
-use prometheus_client::registry::Registry;
+use prometheus_client::registry::{Registry, Unit};
 
 use crate::metrics::Recorder;
 
@@ -57,16 +57,17 @@ impl Metrics {
         let message_count = Family::default();
 
         registry.register(
-            "message_count",
+            "xds_message",
             "Total number of messages received (unstable)",
             message_count.clone(),
         );
 
         let total_messages_size = Family::default();
 
-        registry.register(
-            "total_messages_size",
+        registry.register_with_unit(
+            "xds_message",
             "Total number of bytes received (unstable)",
+            Unit::Bytes,
             total_messages_size.clone(),
         );
 
--- a/src/xds/metrics.rs
+++ b/src/xds/metrics.rs
@@ -24,7 +24,7 @@ use super::service::discovery::v3::DeltaDiscoveryResponse;
 pub struct Metrics {
     pub connection_terminations: Family<ConnectionTermination, Counter>,
     pub message_types: Family<TypeUrl, Counter>,
-    pub total_messages_size: Counter,
+    pub total_messages_size: Family<TypeUrl, Counter>,
 }
 
 #[derive(Clone, Hash, Debug, PartialEq, Eq, EncodeLabelSet)]
@@ -54,15 +54,15 @@ impl Metrics {
             connection_terminations.clone(),
         );
 
-        let message_types = Family::default();
+        let message_count = Family::default();
 
         registry.register(
-            "message_types",
+            "message_count",
             "Total number of messages received (unstable)",
-            message_types.clone(),
+            message_count.clone(),
         );
 
-        let total_messages_size = Counter::default();
+        let total_messages_size = Family::default();
 
         registry.register(
             "total_messages_size",
@@ -72,7 +72,7 @@ impl Metrics {
 
         Self {
             connection_terminations,
-            message_types,
+            message_types: message_count,
             total_messages_size,
         }
     }
@@ -91,12 +91,18 @@ impl Recorder<DeltaDiscoveryResponse, ()> for Metrics {
         let type_url = TypeUrl {
             url: response.type_url.clone(),
         };
-        self.message_types.get_or_create(&type_url).inc_by(1);
+        self.message_types.get_or_create(&type_url).inc();
 
-        let mut message_size: usize = 0;
+        let mut total_message_size: u64 = 0;
         for resource in &response.resources {
-            message_size += resource.resource.as_ref().unwrap().value.len();
+            total_message_size += resource
+                .resource
+                .as_ref()
+                .map(|v| v.value.len())
+                .unwrap_or_default() as u64;
         }
-        self.total_messages_size.inc_by(message_size as u64);
+        self.total_messages_size
+            .get_or_create(&type_url)
+            .inc_by(total_message_size);
     }
 }
--- a/src/xds/metrics.rs
+++ b/src/xds/metrics.rs
@@ -19,8 +19,12 @@ use prometheus_client::registry::Registry;
 
 use crate::metrics::Recorder;
 
+use super::service::discovery::v3::DeltaDiscoveryResponse;
+
 pub struct Metrics {
     pub connection_terminations: Family<ConnectionTermination, Counter>,
+    pub message_types: Family<TypeUrl, Counter>,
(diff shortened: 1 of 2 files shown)


src/xds/metrics.rs
```diff
@@ -19,8 +19,12 @@ use prometheus_client::registry::Registry;
 
 use crate::metrics::Recorder;
 
+use super::service::discovery::v3::DeltaDiscoveryResponse;
+
 pub struct Metrics {
     pub connection_terminations: Family<ConnectionTermination, Counter>,
+    pub message_types: Family<TypeUrl, Counter>,
```
Stevenjin8, line 26: Open to better names

src/xds/metrics.rs
```diff
@@ -58,3 +85,18 @@ impl Recorder<ConnectionTerminationReason, u64> for Metrics {
             .inc_by(count);
     }
 }
+
+impl Recorder<DeltaDiscoveryResponse, ()> for Metrics {
+    fn record(&self, response: &DeltaDiscoveryResponse, _: ()) {
+        let type_url = TypeUrl {
+            url: response.type_url.clone(),
```
Stevenjin8, line 92: Feels like there should be a way of doing this without cloning

Stevenjin8: Also, should I be getting the type_url from the response, or for each resource in response.resources?

src/xds/metrics.rs
```diff
@@ -58,3 +85,18 @@ impl Recorder<ConnectionTerminationReason, u64> for Metrics {
             .inc_by(count);
     }
 }
+
+impl Recorder<DeltaDiscoveryResponse, ()> for Metrics {
```
howardjohn, line 89:
```suggestion
impl Recorder<DeltaDiscoveryResponse, u64> for Metrics {
```

The intent here is this is ~always a u64. I am not even sure why we make it generic, I think we maybe used to have some float usage.

Then `IncrementRecorder` auto-implements and gives `.increment()` btw

src/xds/metrics.rs
```diff
@@ -58,3 +85,18 @@ impl Recorder<ConnectionTerminationReason, u64> for Metrics {
...
+        let type_url = TypeUrl {
+            url: response.type_url.clone(),
+        };
+        self.message_types.get_or_create(&type_url).inc_by(1);
+
+        let mut message_size: usize = 0;
+        for resource in &response.resources {
+            message_size += resource.resource.as_ref().unwrap().value.len();
+        }
+        self.total_messages_size.inc_by(message_size as u64);
```
howardjohn, line 100: I think we should have this keyed by type_url

src/xds/metrics.rs
```diff
@@ -58,3 +85,18 @@ impl Recorder<ConnectionTerminationReason, u64> for Metrics {
...
+impl Recorder<DeltaDiscoveryResponse, ()> for Metrics {
+    fn record(&self, response: &DeltaDiscoveryResponse, _: ()) {
+        let type_url = TypeUrl {
+            url: response.type_url.clone(),
+        };
+        self.message_types.get_or_create(&type_url).inc_by(1);
+
+        let mut message_size: usize = 0;
+        for resource in &response.resources {
+            message_size += resource.resource.as_ref().unwrap().value.len();
```
howardjohn, line 98: unwrap here is not safe, we should just skip if its empty

```suggestion
            message_size += resource.resource.as_ref().map(|v| v.value.len()).unwrap_or_default();
```

src/xds/metrics.rs
Stevenjin8: But which metric would this increment? The number of bytes or the resource type? My thought process is to have one `impl` to handle all metrics for a single request. That way, we aren't indexing the metric by the metric's type, which will cause collisions if we have two metrics of type, say, `u64`.

howardjohn: The type of the metric isn't u64, its the type of the value.

But it is maybe a bit weird since someone could do `record(10)`, which is nonsense -- you can only really meaningfull `increment()` it

Stevenjin8: I think we have different understandings of the code.

My impression of the current implementation of `Recorder` and `IncrementRecorder` is that they deal with only one metric at a time.
For example, `fn record(&self, reason: &ConnectionTerminationReason, count: u64)` updates the `Metrics.connection_termination` entry corresponding to `reason` by `count`.
But if I used `u64` instead of `()` for `Recorder`, then the signature for `record` would be `record(&self, response: &DeltaDiscoveryResponse, x: u64)`. This `record` implementation would update two metrics: `Metrics.message_types`, and `Metrics.total_message_size`. In this case. what would `x` be? Is it the number of bytes? The number of resources of a certain type? If so, of which type? Regardless, this information is already in `response`. Even in the `increment` case (which is really just `record(response, 1)`), I still find the second argument to be confusing. I just don't see a need for the second argument, hence the `()` type.

I could create multiple `record` functions, one for each metric, but I feel like we run into the same problem.
If I key `total_message_size` by `TypeUrl` and have the function signature be `record(&self, response: &TypeUrl, count: u64)`, it isn't clear which metric would be updated by `count`. Is it `message_types` or `total_message_size`? Regardless, having one `record` for each `DeltaDiscoveryResponse` seems better to me because we only have to make one call per response instead of however many metrics we have.

We could also just ignore the second argument of `record` (`fn record(response: &DeltaDiscoveryResponse, _: u64)`) and then just do `increment(response)`, but that feels like the same thing with more steps.

I realize that this probably doesn't reflect what you're thinking and I probably don't really understand what you're saying, but I hope my thought process is more clear now.

Stevenjin8:
Also, do we have tests for these?

src/xds/metrics.rs
```diff
@@ -45,8 +54,26 @@ impl Metrics {
...
+        registry.register(
+            "message_count",
+            "Total number of messages received (unstable)",
+            message_count.clone(),
+        );
+
+        let total_messages_size = Family::default();
+
+        registry.register(
+            "total_messages_size",
```
howardjohn, line 68: See https://prometheus.io/docs/practices/naming/#metric-names for general guidance.

I think we need to make this:
```
        registry.register_with_unit(
            "xds_message",
            "Total number of bytes received (unstable)",
            Unit::Bytes,
            total_messages_size.clone(),
        );
```
. This will automatically become "istio_xds_message_bytes_total" - the istio_ prefix and bytes_total suffix are automatically added

Stevenjin8: I think it will add the units, but not `total`. This is what I see when I curl the metrics endpoint

```
# HELP istio_xds_connection_terminations The total number of completed connections to xds server (unstable).
# TYPE istio_xds_connection_terminations counter
# HELP istio_xds_message Total number of messages received (unstable).
# TYPE istio_xds_message counter
istio_xds_message_total{url="type.googleapis.com/istio.workload.Address"} 1
istio_xds_message_total{url="type.googleapis.com/istio.security.Authorization"} 1
# HELP istio_xds_message_bytes Total number of bytes received (unstable).
# TYPE istio_xds_message_bytes counter
# UNIT istio_xds_message_bytes bytes
istio_xds_message_bytes_total{url="type.googleapis.com/istio.workload.Address"} 2527
istio_xds_message_bytes_total{url="type.googleapis.com/istio.security.Authorization"} 0

```

howardjohn (approved):

src/xds/metrics.rs
howardjohn: # UNIT istio_xds_message_bytes bytes
istio_xds_message_bytes_**total**{url="type.googleapis.com/istio.workload.Address"} 2527

That is the part that matters. Its mismatched from the HELP since this is OpenMetrics spec (which is a bit weird...)

PR #1209, opened 2024-07-16
Disable IP spoofing on outbound

howardjohn:
This is not needed at all, as the ztunnel now always runs in the same
network as the workload (both inpod and dedicated).

The main part of this effort was the testing. The 'dedicated' mode is
not really testing 'dedicated mode', but rather than legacy geneve
stuff. This PR moves it closer to testing a real world setup (which runs
like an Istio sidecar/VM).

The challenge here is setting up iptables. For real world, we just run
the process and ztunnel as different namespaces. I couldn't find a way
to do that in our case, since we have one thread and some user namespace
magic (the latter was the bigger problem). To workaround this, I use a
mark, which is not compiled in by default.

The end result is we have two modes (in tests and real world):
* In pod mode
* Dedicated mode, with ztunnel running directly next to the app in the
  same network, with iptables rules doing traffic capture

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -226,6 +226,10 @@ pub struct Config {
     pub inpod_uds: PathBuf,
     pub inpod_port_reuse: bool,
     pub inpod_mark: u32,
+
+    // Mark to assign to all packets. Only allowed in `--feature testing` mode.
+    // This is to enable the testing of 'dedicated' mode.
+    pub testing_dedicated_mark: u32,
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -492,6 +496,7 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
         inpod_port_reuse: parse_default(INPOD_PORT_REUSE, true)?,
         inpod_mark: parse_default(INPOD_MARK, DEFAULT_INPOD_MARK)?,
         fake_self_inbound: false,
+        testing_dedicated_mark: 0,
     })
 }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -37,7 +37,7 @@ const KUBERNETES_SERVICE_HOST: &str = "KUBERNETES_SERVICE_HOST";
 const NETWORK: &str = "NETWORK";
 const NODE_NAME: &str = "NODE_NAME";
 const PROXY_MODE: &str = "PROXY_MODE";
-const INPOD_MARK: &str = "INPOD_MARK";
+const PACKET_MARK: &str = "PACKET_MARK";
 const INPOD_UDS: &str = "INPOD_UDS";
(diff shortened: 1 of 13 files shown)


src/config.rs
```diff
@@ -492,6 +496,7 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
         inpod_port_reuse: parse_default(INPOD_PORT_REUSE, true)?,
         inpod_mark: parse_default(INPOD_MARK, DEFAULT_INPOD_MARK)?,
         fake_self_inbound: false,
+        testing_dedicated_mark: 0,
```
bleggett, line 499: I am relatively sure that even non-testing usages of `dedicated` mode might need this mark, so TBH I am fine just making it `dedicated_mark` and setting it either via env or via `DEFAULT_DEDICATED_MARK` like we do with inpod above.

Alternatively - we just rename `inpod_mark` to `proxy_mark`, default to `1337` (current inpod default + also the current sidecar/envoy default) and use the same config val under both modes.


tl;dr I think this will be necessary for `dedicated` mode outside of just tests so we might as well formalize it? thoughts?

src/proxy/outbound.rs
```diff
@@ -257,15 +249,8 @@ impl OutboundConnection {
         req: &Request,
         connection_stats: &ConnectionResult,
     ) -> Result<(), Error> {
-        // Create a TCP connection to upstream
-        // We do not need spoofing for inbound
-        let local = if self.enable_orig_src && self.pi.cfg.proxy_mode != ProxyMode::Shared {
```
bleggett, line 262: :+1:

src/test_helpers/linux.rs
```diff
@@ -142,16 +145,26 @@ impl WorkloadManager {
             local_ip: Some(ns.ip()),
             inpod_uds,
             proxy_mode,
+            testing_dedicated_mark: if proxy_mode == ProxyMode::Dedicated {
```
bleggett, line 148: Yeah - in ref to above comments our `inpod` and `dedicated_testing` marks are both 1337 - so maybe it should just be a single (overridable) "mark" config value, and we just use it differently according to ProxyMode

src/config.rs
howardjohn: I was on the fence, went with test only since I wasn't sure a concrete use case where running as a different UID/GID wasn't viable. Not opposed if there is (or if we think its simpler to just make it a real thing -- with the inpod_mark collapsing it may just be quite a bit simpler to do so)

howardjohn: ok, I made the change

howardjohn:
I added a cherrypick label since there is a small chance this fixes the obscure AWS issues we have seen.

bleggett (approved):

jacob-delgado (approved):

PR #1212, opened 2024-07-17
add context to inpod::workloadmager failed to connect log message

ilrudie:
Adds context to `inpod::workloadmager failed to connect...` log message with the hope that it makes troubleshooting easier. Without this you need to have a pretty good idea of the design of inpod  to begin to understand what is failing.

### DIFF
--- a/src/inpod/workloadmanager.rs
+++ b/src/inpod/workloadmanager.rs
@@ -114,7 +114,7 @@ impl WorkloadProxyNetworkHandler {
                     backoff =
                         std::cmp::min(CONNECTION_FAILURE_RETRY_DELAY_MAX_INTERVAL, backoff * 2);
                     warn!(
-                        "failed to connect to server {:?}: {:?}. retrying in {:?}",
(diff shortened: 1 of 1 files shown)


src/inpod/workloadmanager.rs
```diff
@@ -114,7 +114,7 @@ impl WorkloadProxyNetworkHandler {
                     backoff =
                         std::cmp::min(CONNECTION_FAILURE_RETRY_DELAY_MAX_INTERVAL, backoff * 2);
                     warn!(
-                        "failed to connect to server {:?}: {:?}. retrying in {:?}",
+                        "failed to connect to Istio CNI Node Agent over {:?}, is the Istio CNI healthy? details: {:?}. retrying in {:?}",
```
bleggett, line 117:
```suggestion
                        "failed to connect to the Istio CNI node agent over {:?}, is the node agent healthy? details: {:?}. retrying in {:?}",
```

bleggett (approved):

src/inpod/workloadmanager.rs
bleggett: If you want, otherwise approved, so pull hold when you want.

PR #1216, opened 2024-07-19
Add support for hostnetwork pods

howardjohn:
This goes along with https://github.com/istio/istio/pull/52191. Most of
the change is in Istiod, the only real change here is to stop indexing
the workload address if it is a host network pod.

In terms of user facing behavior, this means that if I call a
host-network pod directly (NOT via service), then it will not associate
with the workload. If this were to happen, we would effectively be
associating it with an arbitrary host network pod on the same node;
there is no way to distinguish these. In practice, hostnetwork pods
cannot be HBONE, so this ensures that we consistently get "unknown" as
the telemetry info rather then a random pod. This is encoded in the new
test.

There is additional misc trace level logging improvements that helped me
debug/develop this.


bleggett:
This is maybe a dumb question, but why track hostNetwork pods as Workloads at all?

Why have "hostNW/not-hostNW" flags on Workloads?

Why not just treat them as "no upstream found, passthrough" and have no Workload for them at all - just pass the traffic through as we would to any other unknown/un-attestable dest? That is effectively what they are, right?

howardjohn:
> This is maybe a dumb question, but why track hostNetwork pods as Workloads at all?

I can have a service with backends that have hostNetwork pods. We must send to those pods

> Why have "hostNW/not-hostNW" flags on Workloads?

This is simply to avoid attempting to reverse lookup IP to workload for these. Traffic will work fine even if we don't do this, but telemetry for direct-to-pod traffic for these is wonky since it will pick a random one on the node.

> Why not just treat them as "no upstream found, passthrough" and have no Workload for them at all - just pass the traffic through as we would to any other unknown/un-attestable dest? That is effectively what they are, right?

This is fine for direct-to-pod but not for Service load balancing.

bleggett:
> This is fine for direct-to-pod but not for Service load balancing.

ah, and Services must have Workloads behind them, not naked IPs. ok.

bleggett (approved):

Issue #1198, opened 2024-07-08
Refactoring improvements now that we have a proxy-per-workload approach


howardjohn:
The original ztunnel design was a fully multi-tenant design. In the current architecture, we build up a Proxy per workload, so while we run multi-tenant binary, we have pretty siloed code execution.

Given this, we can make a few improvements IMO. Here is what I am thinking:

* Each Proxy gets a `WorkloadContext`

```
struct WorkloadContext {
  workload_info: Fetch<Workload>,
  certificate: Fetch<Certificate>,
}
```

* A new `Fetch` type (note: probably there is a better name, and existing implementation) wraps these, which exposes async getters. All usages calls these, which ensures they get updates if they change, and that they wait if they are not yet available

This is a cleaner model than passing around a CA that can fetch arbitrary certs but hoping we select the right one (to be clear: we have many checks to ensure we do, its just awkward), and similar for workload lookup.

The tricky part here is dedicated mode. In inpod, we have a workloaddescription to match up the workload and WDS. For dedicated we don't really.

Some options:
* Make the user provide some info... what info? name, namespace, ...?
* Automatically infer the IP based on local interfaces, match that up
* Some combination of ^

bleggett:
> The original ztunnel design was a fully multi-tenant design. In the current architecture, we build up a Proxy per workload, so while we run multi-tenant binary, we have pretty siloed code execution.
> 
> Given this, we can make a few improvements IMO. Here is what I am thinking:
> (snip)
> This is a cleaner model than passing around a CA that can fetch arbitrary certs but hoping we select the right one (to be clear: we have many checks to ensure we do, its just awkward), and similar for workload lookup.

Yep, :+1: that in general would be a bit tidier.

> The tricky part here is dedicated mode. In inpod, we have a workloaddescription to match up the workload and WDS. For dedicated we don't really.
> 
> Some options:
> 
>     * Make the user provide some info... what info? name, namespace, ...?
> 
>     * Automatically infer the IP based on local interfaces, match that up
> 
>     * Some combination of ^

1. How is this accomplished in sidecar mode? How does a proxy know what workload it should be proxying?

2. 

> Make the user provide some info... what info? name, namespace, ...?

I'd be inclined to make users specify the fields in Dedicated mode that we would get from the node agent in Shared mode - that is,   

  ```
  string name = 1;
  string namespace = 2;
  string service_account = 3;
```

howardjohn:
[...]

PR #1218, opened 2024-07-19
Fetch local workload directly, rather than based on IP

howardjohn:
This aligns our logic with the reality of how Ztunnel is deployed today. In its initial inception, we had one `Proxy`, and dynamically associated requests with pods by IP.

Now, we have one `Proxy` instance per workload, whether that is shared/inpod or dedicated.

To align with this, this PR stops using IP as a means to associate traffic to a workload - we always use the workload the Proxy is deployed for.

For inpod, this is clear: the CNI tells us specifically what pod we should be starting the proxy for. Today, we already do a second-check that the workload we get from the IP address matches.

For dedicated, its less clear (and not yet solved, hence why its WIP). I think we should just add a new config like `DEDICATED_WORKLOAD=some-ns/some-workload-name` or whatever.

Aside from being more clear, less error prone, and more efficient (one lookup per pod vs per connection).

One specific bug this fixes is usage of socks5, which is currently basically broken. The source IP is always 127.0.0.1 when connecting to ztunnel over socks, so it always gets 'unknown source'

Open question:
* Should we validate the IP address at all? is there a case where some arbitrary external traffic hits our outbound listener somehow that we were protected against before but no longer are?

Blockers:
* Add a way to make it work with dedicate mode
* Ensure if the workload changes we get notified for future connections

Fixes https://github.com/istio/ztunnel/issues/1198


bleggett:
> For dedicated, its less clear (and not yet solved, hence why its WIP). I think we should just add a new config like DEDICATED_WORKLOAD=some-ns/some-workload-name or whatever.

xref https://github.com/istio/ztunnel/issues/1198

bleggett:
> * Should we validate the IP address at all? is there a case where some arbitrary external traffic hits our outbound listener somehow that we were protected against before but no longer are?

Given that by-itself IP simply isn't an identity (address != identity, for both machines and people, though it's a factor) I doubt it.

> For inpod, this is clear: the CNI tells us specifically what pod we should be starting the proxy for. Today, we already do a second-check that the workload we get from the IP address matches.

Yeah I think this is fine for inpod, the secondary check doesn't hurt.

There's a couple of ways we can do this for dedicated mode tho

- just have `DEDICATED_ALLOWED_IP=whatever`
- have an array of env vars as mentioned in #1198 describing the singular workload identity
- require outbound stuff to have a packet/connmark applied out of band by e.g. iptables before we accept it (this is more of a "you solve this in your env, not our problem" approach)

For dedicated, we could _only_ do the IP address match and leave it at that (first option) - I think that's fine for now?

howardjohn:
> Given that by-itself IP simply isn't an identity (address != identity, for both machines and people, though it's a factor) I doubt it.

My concern was more: A packet comes *into* the pod to `pod:15001` and somehow is confused into sending outbound. It probably cannot send outbound since it would likely not have an orig_dst (unless it hit a collision?) and instead, at worst, infinite loop; our 'illegal ports' solves that?

bleggett:
> > Given that by-itself IP simply isn't an identity (address != identity, for both machines and people, though it's a factor) I doubt it.
> 
> My concern was more: A packet comes _into_ the pod to `pod:15001` and somehow is confused into sending outbound. It probably cannot send outbound since it would likely not have an orig_dst (unless it hit a collision?) and instead, at worst, infinite loop; our 'illegal ports' solves that?

We TPROXY everything coming in - so that isn't possible unless the inpod rules are defeated.

And if they inpod rules are defeated traffic can just egress without hitting ztunnel at all - so yea, I don't think it's worth doing the check here.

src/state/workload.rs
```diff
@@ -656,13 +679,34 @@ impl WorkloadStore {
...
+    pub fn find_by_info(&self, wl: &WorkloadInfo) -> Option<Arc<Workload>> {
+        // We do not have an index directly on the full workload info, but we can narrow it down
+        // to only workloads on the same node with the same identity -- a tiny set to iterate over
+        self.node_local_by_identity
+            .get(&WorkloadIdentity {
+                namespace: strng::new(&wl.namespace),
+                service_account: strng::new(&wl.service_account),
+            })?
+            .iter()
+            .find_map(|uid| self.by_uid.get(uid).cloned())
```
ilrudie, line 692: It seems to me `find_map`'s function is missing a check to ensure that the result of `by_uid.get(uid)` actually matches wl, not just shares a node and identity and was first in the HashSet.iter().

src/state.rs
```diff
@@ -627,17 +635,17 @@ impl DemandProxyState {
     // when we get it, or nothing if the timeout is exceeded, whichever happens first
     pub async fn wait_for_workload(
         &self,
-        addr: &NetworkAddress,
+        wl: &WorkloadInfo,
```
bleggett, line 638: The more I think about this, the more I think we can't allow XDS cache lookups by WorkloadInfo directly.

XDS `Workloads` are XDS, and ZDS `WorkloadInfos` are ZDS. They aren't directly convertable, and for inpod you need _both_ pieces to make an authoritative decision about a workload's identity.

Could we maybe use/repurpose `identity/manager.rs`'s `Identity` struct here as the "glue struct"?

- You need a Workload _and_ a WorkloadInfo to _fully_ construct an Identity struct.
- XDS State lookups by Identity struct only (not WLInfo)

howardjohn: If you need a Workload to get an Identity, and an Identity to get XDS state, and Workload is XDS state, this is circular -- where do we break the loop (or where do I misunderstand)?

I do think `LocalWorkloadInformation.get_workload` does need to always do a lookup (since it can change at anytime) for sure, instead of caching once at Proxy startup.

bleggett: Sorry - I should have clarified/still thinking.

`no cache lookups by WorkloadInfo directly for local workloads in the _routing path_`

So this is something like

1. on WorkloadInfo (zds), subscribe/get matching XDS entry (matching == same Name/NS/SA)
2. Construct Identity from union of ZDS + XDS (NS/SA/TD to start with, since that's more or less the "important" data union here between these types)
3. Use this to start cert prefetch, assert that any cert we get back has the correct NS/SA/TD we got from the identity/union.
4. From this point on, we can establish mTLS connections safely, and WLInfo basically drops out of use in favor of Identity.
5. Routing path lookups for Workloads keyed are by Identity only

something like that.

bleggett: (this is out of scope of this PR, I am mostly just thinking out loud)

src/proxy.rs
```diff
@@ -148,54 +150,93 @@ pub struct Proxy {
 #[derive(Clone)]
 pub struct ScopedSecretManager {
     cert_manager: Arc<SecretManager>,
-    allowed: Option<Arc<WorkloadInfo>>,
+    allowed: Arc<WorkloadInfo>,
```
bleggett, line 153: (probably out of context thought) ScopedSecretManager could probably be removed or mostly mooted if `allowed` were just an Identity instead of a WorkloadInfo.

This would require waiting to create the proxy until we got both the XDS push and the WorkloadInfo, so we could construct and propagate the Identity tho.

src/proxy.rs
```diff
@@ -148,54 +150,93 @@ pub struct Proxy {
...
+        trust_domain: Strng,
+    ) -> Result<Arc<tls::WorkloadCertificate>, identity::Error> {
+        let id = &Identity::Spiffe {
+            trust_domain,
+            namespace: (&self.wi.namespace).into(),
+            service_account: (&self.wi.service_account).into(),
+        };
+        self.full_cert_manager.fetch_certificate(id).await
+    }
+
```
bleggett, line 226: @Maybe a `pub fn get_identity()` here?

src/proxy/inbound.rs
```diff
@@ -221,7 +221,8 @@ impl Inbound {
 
         let rbac_ctx = crate::state::ProxyRbacContext {
             conn,
-            dest_workload_info: pi.proxy_workload_info.clone(),
+            dest_workload_info: Some(pi.local_workload_information.workload_info()),
```
bleggett, line 224: (musing/aside/future work) One thing I still don't like is that we are carrying these two things as state everywhere for a bunch of decisions, it still doesn't seem correct.

Carrying on from https://github.com/istio/ztunnel/pull/1218#discussion_r1722283016 - I don't think `WorkloadInfo` should ever be used to make RBAC decisions? `Workload`/`Identity` should be all we use, by the point we check RBAC we should have gotten all we need from WorkloadInfo and either converted it into an Identity or done whatever Identity-based checks we needed.

src/proxy/inbound.rs
```diff
@@ -470,7 +471,10 @@ impl crate::tls::ServerCertProvider for InboundCertProvider {
             %identity,
             "fetching cert"
         );
```
bleggett, line 473: This is probably a bit easier to read if we just do `IF identity(from sock IP) == self.local_workload.get_identity() then self.local_workload.fetch_certificate() ELSE error`

src/state/workload.rs
bleggett: This probably should be the invert of `identity/manager.rs`'s `matches` - e.g. we just add another check in the `find_map` closure to assert that `wl.Name == xds_wl.Name` and it's As Good As that existing check

bleggett: (or we just use that func, actually)

src/state.rs
howardjohn: > Routing path lookups for Workloads are keyed by Identity/WL only (not WLInfo)


Do we need any lookup at all? Isn't it just `proxy.get_local_workload()`?

bleggett: Probably not, once we have established that `local_workload == local_workloadinfo` - when I say "lookup" I am mostly talking about certs.

howardjohn: Part of the goal with this is that you cannot do `lookup(identity)` -- then you can get the wrong one!.

A `Proxy` is scoped to one specific workload. So we already know the identity of it. I want to expose `proxy.get_local_cert()` and `proxy.get_local_workload()` exclusively, so we never do any IP -> cert or IP -> Workload lookups (to identify what we are proxying for; of course for outbound we need a check like that)

src/proxy/inbound.rs
howardjohn: Good news -- we literally do not use it at all so it was trivial to just remove it. Technically we could just get the policies + namespace (vs storing the whole workload), but i don't want to make a clone so its easier to just pass the WL around

src/state.rs
bleggett: :+1: yep fair enough - 100% of this is me just not wanting to propagate multiple structs that can be/might be used for identity/RBAC lookup purposes in the hotpath.

I'm imagining something where WorkloadInfo might have information used to establish/validate the `local_workload`'s identity or validity - we already sort of do this in RBAC checks where we compare WLInfo and Workload and effectively use the union of both to make a decision. I'd rather keep away from stuff like that and just convert to something authoritative as early in the local workload's lifecycle as possible.

(but this is orthogonal to what you're doing here so out-of-scope)

src/state/workload.rs
howardjohn: great call :+1: updated

src/proxy/pool.rs
```diff
@@ -996,10 +1001,35 @@ mod test {
             ..crate::config::parse_config().unwrap()
         };
         let sock_fact = Arc::new(crate::proxy::DefaultSocketFactory);
-        let cert_mgr = proxy::ScopedSecretManager::new(identity::mock::new_secret_manager(
-            Duration::from_secs(10),
+
+        let mut state = ProxyState::new(None);
```
bleggett, line 1005: nit: probably should be a series of helpers/we may already have a helper that does some or all of this.

src/state/workload.rs
```diff
@@ -685,13 +708,34 @@ impl WorkloadStore {
...
     }
 
+    /// Finds the workload by workload information, as an arc.
+    pub fn find_by_info(&self, wl: &WorkloadInfo) -> Option<Arc<Workload>> {
+        // We do not have an index directly on the full workload info, but we can narrow it down
+        // to only workloads on the same node with the same identity -- a tiny set to iterate over
+        self.node_local_by_identity
+            .get(&WorkloadIdentity {
+                namespace: strng::new(&wl.namespace),
+                service_account: strng::new(&wl.service_account),
```
bleggett, line 718: I'm still not 100% sure what use `WorkloadIdentity` is here, it's only used in this file, it's a strict subset of `WorkloadInfo`, and it's _not_ an identity, as it could apply to N workloads, it's more of a query filter.

Either remove it, or include `wl.name` in `WorkloadIdentity` so `get` can filter down to a unique workload in all cases and we don't need the `filter` clause below?

bleggett (commented):
Vast majority LGTM, minor comments

src/state/workload.rs
howardjohn: It is for what it says on the tin:
```
/// WorkloadIdentity provides information about a workloads identity. This is used in place of Identity
/// in places where we **do not have the full identity (no trust domain)** and know we are working with workloads specifically.
```

We don't have TD so we cannot use `Identity`. We want a mapping of `ThingThatIsAKeyForCertificate -> Workloads` (which is what `node_local_by_identity` is for), so if we add `name` to it we are no longer doing that.

The issue here really stems from the fact there are two things using node_local_by_identity. One is find_by_info which would prefer to have a name, and one is `was_last_identity_on_node` which **cannot** have `name`. We could make two indexes, but IMO its simpler and more efficient to have `find_by_info` do the filtering (its a cheap filter, there is at most 100 entries to filter and likely fewer)

bleggett: `and one is was_last_identity_on_node which cannot have name`

ugh. I missed `was_last_identity_on_node`. You're right, carry on.

bleggett (approved):

dhawton (approved):

PR #1224, opened 2024-07-23
Support hostname waypoints

howardjohn:
This PR adds support for sending traffic to waypoints identified by hostnames. Currently, this code is not exercised by the control plane, so it should never be hit. However, we need this to be present 1 release _before_ we turn it on for istiod for compatibility purposes. This gives us that capability.

This is preferred long term as the implementation is simpler in both the control plane and ztunnel, and hostname is already the unique key for a service, so its a better lookup key. Additionally, it makes dual stack usage more clear (i.e. we don't have to pick an arbitrary IP to put, we can just reference the hostname).

Related: https://github.com/istio/istio/issues/52318
Istiod change: https://github.com/istio/istio/pull/52294

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -694,7 +694,7 @@ impl DemandProxyState {
             return None;
         }
         self.fetch_on_demand(uid.clone()).await;
-        self.state.read().unwrap().workloads.find_uid(uid)
+        self.read().workloads.find_uid(uid)
     }
 
     pub async fn fetch_upstream(
@@ -707,12 +707,8 @@ impl DemandProxyState {
         self.fetch_address(&network_addr(network.clone(), addr.ip()))
             .await;
         let upstream = {
-            self.state.read().unwrap().find_upstream(
-                network,
-                source_workload,
-                addr,
-                resolution_mode,
-            )
+            self.read()
+                .find_upstream(network, source_workload, addr, resolution_mode)
             // Drop the lock
         };
         self.finalize_upstream(source_workload, addr, upstream)
@@ -766,7 +762,7 @@ impl DemandProxyState {
                 )
             }
             Destination::Hostname(host) => {
-                let state = self.state.read().unwrap();
+                let state = self.read();
                 match state.find_hostname(host) {
                     Some(Address::Service(s)) => state.find_upstream_from_service(
                         source_workload,
--- a/src/state.rs
+++ b/src/state.rs
@@ -254,35 +254,12 @@ impl ProxyState {
             .services
             .get_by_vip(&network_addr(network.clone(), addr.ip()))
         {
-            // Randomly pick an upstream
-            // TODO: do this more efficiently, and not just randomly
-            let Some((ep, wl)) = self.load_balance(source_workload, &svc, addr, resolution_mode)
-            else {
-                debug!("VIP {} has no healthy endpoints", addr);
-                return None;
-            };
-
-            let svc_port = svc.ports.get(&addr.port()).copied().unwrap_or_default();
-            let target_port = if let Some(&port) = ep.port.get(&addr.port()) {
-                // prefer endpoint port mapping
-                port
-            } else if svc_port > 0 {
-                // otherwise, see if the service has this port
-                svc_port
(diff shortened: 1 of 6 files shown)


src/state.rs
```diff
@@ -724,36 +751,59 @@ impl DemandProxyState {
...
-            Destination::Hostname(_) => {
-                return Err(Error::UnsupportedFeature(
-                    "hostname lookup not supported yet".to_string(),
-                ));
+        // Waypoint can be referred to by an IP or Hostname.
+        // Hostname is preferred as it is a more stable identifier.
+        let res = match &gw_address.destination {
+            Destination::Address(ip) => {
+                let addr = SocketAddr::new(ip.address, gw_address.hbone_mtls_port);
+                self.state.read().unwrap().find_upstream(
```
ilrudie, line 761: **nit**: expect instead of unwrap

src/state.rs
```diff
@@ -724,36 +751,59 @@ impl DemandProxyState {
...
+                    ip.network.clone(),
+                    source_workload,
+                    addr,
+                    ServiceResolutionMode::Waypoint,
+                )
+            }
+            Destination::Hostname(host) => {
+                let state = self.state.read().unwrap();
```
ilrudie, line 769: **nit**: expect instead of unwrap

src/state.rs
```diff
@@ -698,18 +706,37 @@ impl DemandProxyState {
...
         self.fetch_address(&network_addr(network.clone(), addr.ip()))
             .await;
-        let Some((wl, port, svc)) = self.state.read().unwrap().find_upstream(
-            network,
-            source_workload,
-            addr,
-            resolution_mode,
-        ) else {
+        let upstream = {
+            self.state.read().unwrap().find_upstream(
```
ilrudie, line 710: **nit**: expect instead of unwrap

ilrudie (commented):
The code seems fine. Just a couple of nits from your friendly neighborhood unwrap nag.

What is the intended timeline for rolling this out? 1.23 shipping in pretty much imminent so guessing 1.24 ships with this in ztunnel in preparation for 1.25 in istiod?

What is the plan for knowing this works as intended if actual usability is that far out? It seems like timing wise this could complicate maintaining ztunnel in the short term and might make more sense to merge closer to shipping 1.24 so that any fixes for ztunnel that come up as folks get their hands on 1.23 in the wild won't need to contend with these changes as we try to backport them.

src/state.rs
bleggett: (can we just make clippy flag `unwrap`? might not be terrible)

ilrudie: quick bit of work on this, we have something in the range of 60 uses of unwrap outside of tests and build stuff so non-trivial to just yeet it. Happy to do the work to cleanup and enable the lint if we want to go this route though.

ilrudie: https://github.com/istio/ztunnel/issues/1245

howardjohn:
> The code seems fine. Just a couple of nits from your friendly neighborhood unwrap nag.
> 
> What is the intended timeline for rolling this out? 1.23 shipping in pretty much imminent so guessing 1.24 ships with this in ztunnel in preparation for 1.25 in istiod?
> 
> What is the plan for knowing this works as intended if actual usability is that far out? It seems like timing wise this could complicate maintaining ztunnel in the short term and might make more sense to merge closer to shipping 1.24 so that any fixes for ztunnel that come up as folks get their hands on 1.23 in the wild won't need to contend with these changes as we try to backport them.

I think the key to this is just testing... which was lacking in the initial PR -- I have built it out a lot more now. WDYT?

FWIW I have also been manually testing with Istiod PR, just hard to turn that into automation

ilrudie (approved):
LGTM but I do think we should reconsider carrying the diff if we don't get the cherrypick accepted and we won't have the istiod change in 1.24

thedebugger:
@ilrudie are you saying we should hold off on this change in 1.23 since related istiod changes aren't certain in 1.24?

ilrudie:
@thedebugger, The opposite, the istiod changes are ready, or nearly so, I think we should cherry-pick this to 1.23 so that we can merge the istiod changes soon and begin full testing.

PR #1226, opened 2024-07-25
Enable DNS server by default

howardjohn:
This enables the DNS server by default. This has no impact on the
*behavior* of applications out of the box -- they still need to opt-in
to redirecting traffic to it. However, by making this on by default, we
require only one knob to opt-in to DNS, rather than 2 which need to be
kept in sync.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -400,10 +400,12 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
 
     validate_config(Config {
         proxy: parse_default(ENABLE_PROXY, true)?,
+        // Enable by default; running the server is not an issue, clients still need to opt-in to sending their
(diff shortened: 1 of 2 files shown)


keithmattix:
Any baseline perf impact from the enablement?

bleggett:
> Any baseline perf impact from the enablement?

bleggett:
The external/visible effects of this will mostly be 

- A new DNS port binding in all pods at 15053

that port won't get any DNS traffic unless you toggle the sibling flag in `istio-cni` however.

howardjohn:
> Any baseline perf impact from the enablement?

Shouldn't be, we have all the same control plane logic whether its used or not. there is a near-zero impact of opening up a socket that never gets connections I suppose

bleggett (approved):

PR #1231, opened 2024-07-26
Add support for `publishNotReadyAddresses`

howardjohn:
Istio side: https://github.com/istio/istio/pull/52359

This implements support for `publishNotReadyAddresses`. This is done by inserting a new field into Service.loadbalancer which specifies we should allow unhealthy things.

The implementation is _almost_ trivial, but made more complex by how we reconcile service endpoints. There is a bug I intentionally left, where if you change from publishNotReadyAddresses=false to publishNotReadyAddresses=true, the endpoints are not added. This could be supported, however, the cost is we need to totally change how we many endpoints. IMO, given the field is niche, and changing the field is even nicher, it is safer to not support this. This is tested + commented in the code.

### DIFF
--- a/proto/workload.proto
+++ b/proto/workload.proto
@@ -137,12 +137,20 @@ message LoadBalancing {
     // 4. Any endpoints
     FAILOVER = 2;
   }
+  enum HealthPolicy {
+    // Only select healthy endpoints
+    ONLY_HEALTHY = 0;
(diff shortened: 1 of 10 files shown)


ilrudie:
`should_include_endpoint` sounds good. It does clarify that this is a check

PR #1235, opened 2024-07-26
[Direct to 1.23] Future proof LoadBalancerPolicy

howardjohn:
Currently, the implementation of the LoadBalancer policy is flawed; it
assumes if the message is set, we will use failover. This actually
mostly works since failover against 'no routing preferences' is the same
as no policy, but its wasteful performance and confusing.

This ensures that if LBPolicy is used for new fields (which it will be,
see https://github.com/istio/ztunnel/pull/1231), we do not accidentally
start doing failover routing.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -349,8 +349,7 @@ impl ProxyState {
         });
 
         match svc.load_balancer {
-            None => endpoints.choose(&mut rand::thread_rng()),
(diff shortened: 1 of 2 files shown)


mikemorris (approved):
Reviewed smaller change here and context over in #1231, makes sense to me.

Issue #1247, opened 2024-08-02
AuthorizationPolicy still take effect by namespace scope after a workloadSelector added to it


xocodergit:
Bug Description:
1. Create a namespace-scoped AuthorizationPolicy
2. Change it to workload-scoped by add WorkloadSelector to it
3. AuthorizationPolicy still take effect by namespace scope

Istio Version:
1.22.3

bleggett:
Can you add an example for number 3 that shows exactly what you're talking about? 

My understanding is adding workload selectors should further refine the existing namespace scope, not replace it.

PR #1248, opened 2024-08-02
remove authorization from "by_namespace" when scope change to workload

xocodergit:
As described in https://github.com/istio/ztunnel/issues/1247

The root cause of this bug is that Authorization not removed from 'by_namespace' when change to workload scope.

### DIFF
--- a/src/state/policy.rs
+++ b/src/state/policy.rs
@@ -56,28 +56,28 @@ impl PolicyStore {
             .collect()
     }
 
-    pub fn insert(&mut self, rbac: Authorization) {
-        let key: Strng = rbac.to_key();
+    pub fn insert(&mut self, xds_name: Strng, rbac: Authorization) {
+        self.remove(xds_name.clone());
         match rbac.scope {
             RbacScope::Global => {
                 self.by_namespace
                     .entry(strng::EMPTY)
                     .or_default()
-                    .insert(key.clone());
+                    .insert(xds_name.clone());
             }
             RbacScope::Namespace => {
                 self.by_namespace
                     .entry(strng::new(&rbac.namespace))
                     .or_default()
-                    .insert(key.clone());
+                    .insert(xds_name.clone());
             }
             RbacScope::WorkloadSelector => {}
         }
-        self.by_key.insert(key, rbac);
+        self.by_key.insert(xds_name.clone(), rbac);
     }
 
-    pub fn remove(&mut self, name: Strng) {
-        let Some(rbac) = self.by_key.remove(&name) else {
+    pub fn remove(&mut self, xds_name: Strng) {
+        let Some(rbac) = self.by_key.remove(&xds_name) else {
             return;
         };
         if let Some(key) = match rbac.scope {
@@ -86,7 +86,7 @@ impl PolicyStore {
             RbacScope::WorkloadSelector => None,
         } {
             if let Some(pl) = self.by_namespace.get_mut(&key) {
-                pl.remove(&name);
+                pl.remove(&xds_name);
                 if pl.is_empty() {
                     self.by_namespace.remove(&key);
                 }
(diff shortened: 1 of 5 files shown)


bleggett (commented):
I might be missing something (which is why I asked for an example of the issue you're facing in the linked bug) - but I don't think *removing* the rbac rule from the namespace scope when the scope is namespace+workload is correct?

If it was, we would do that for policies moving from Global to Namespace scope as well, and we don't?

howardjohn (commented):
I think this makes sense but can you add a unit test?

xocodergit:
@howardjohn @bleggett Thank you for your review, unit test added.

src/state/policy.rs
```diff
@@ -71,7 +71,23 @@ impl PolicyStore {
                     .or_default()
                     .insert(key.clone());
             }
-            RbacScope::WorkloadSelector => {}
+            RbacScope::WorkloadSelector => {
+                let existing = self.by_key.get(&key);
+                match existing {
+                    Some(e) => {
+                        match e.scope {
+                            RbacScope::Global => {
```
bleggett, line 79: Actually, looking at this and the surrounding code, I think maintaining multiple lists for some cases and not others is what I find really troublesome - I don't see the point in `by_namespace` if we don't also have `by_workload` or `by_global`.

It leads to this need to keep things synced in some cases but not others, which feels like a footgun. If it's intended a perf optimization, there are better ways to do sharding.


(all of that predates this fix tho, I am just looking at this code now).

bleggett (approved):

src/state/policy.rs
howardjohn: I 100% agree with the fragileness, and it applies to the rest of `state` as well. But we cannot fix that by removing all indexes and having O(N) lookups everywhere (which quickly become O(N^2) when composed).

This is exactly why we built `krt`. Not sure a good lightweight approach that solves it though

bleggett: Yep I get that - maybe better to have `by_workload` and `by_namespace` then - right now we just have `everything && by_namespace` which probably isn't going to work well for either lookups or footguns.

howardjohn:
Apologies for the late review here, I didn't think it through enough on my initial review. How we avoid this in the rest of the code (workload/service) is to make "insert" do "remove + insert". WDYT of doing similar here?

```diff
diff --git a/src/xds.rs b/src/xds.rs
index c9ebefe..d7a7e60 100644
--- a/src/xds.rs
+++ b/src/xds.rs
@@ -273,12 +273,15 @@ impl ProxyStateUpdateMutator {
     pub fn insert_authorization(
         &self,
         state: &mut ProxyState,
+        name: Strng,
         r: XdsAuthorization,
     ) -> anyhow::Result<()> {
         info!("handling RBAC update {}", r.name);

         let rbac = rbac::Authorization::try_from(r)?;
         trace!("insert policy {}", serde_json::to_string(&rbac)?);
+        // Remove the policy to clean up any old state, then add it back
+        state.policies.remove(name);
         state.policies.insert(rbac);
         Ok(())
     }
@@ -385,7 +388,7 @@ impl Handler<XdsAuthorization> for ProxyStateUpdater {
         let handle = |res: XdsUpdate<XdsAuthorization>| {
             match res {
                 XdsUpdate::Update(w) => {
-                    self.updater.insert_authorization(&mut state, w.resource)?
+                    self.updater.insert_authorization(&mut state, w.name, w.resource)?
                 }
                 XdsUpdate::Remove(name) => self.updater.remove_authorization(&mut state, name),
             }
```

something like this

xocodergit:
> Apologies for the late review here, I didn't think it through enough on my initial review. How we avoid this in the rest of the code (workload/service) is to make "insert" do "remove + insert". WDYT of doing similar here?
> 
> ```diff
> diff --git a/src/xds.rs b/src/xds.rs
> index c9ebefe..d7a7e60 100644
> --- a/src/xds.rs
> +++ b/src/xds.rs
> @@ -273,12 +273,15 @@ impl ProxyStateUpdateMutator {
>      pub fn insert_authorization(
>          &self,
>          state: &mut ProxyState,
> +        name: Strng,
>          r: XdsAuthorization,
>      ) -> anyhow::Result<()> {
>          info!("handling RBAC update {}", r.name);
> 
>          let rbac = rbac::Authorization::try_from(r)?;
>          trace!("insert policy {}", serde_json::to_string(&rbac)?);
> +        // Remove the policy to clean up any old state, then add it back
> +        state.policies.remove(name);
>          state.policies.insert(rbac);
>          Ok(())
>      }
> @@ -385,7 +388,7 @@ impl Handler<XdsAuthorization> for ProxyStateUpdater {
>          let handle = |res: XdsUpdate<XdsAuthorization>| {
>              match res {
>                  XdsUpdate::Update(w) => {
> -                    self.updater.insert_authorization(&mut state, w.resource)?
> +                    self.updater.insert_authorization(&mut state, w.name, w.resource)?
>                  }
>                  XdsUpdate::Remove(name) => self.updater.remove_authorization(&mut state, name),
>              }
> ```
> 
> something like this

👍 i think this implementation(remove, then insert) is better, consistent, and much more simpler. but I prefer to consider that keeping policies in the right indexes is the responsibility of `PolicyStore`, what do you think if we do remove in `PolicyStore.insert`?

howardjohn:
Yeah I don't mind that, we just need to pass the XDS name further. Note the `XdsUpdate.name` and `Authorization.name` are not the same thing

howardjohn (approved):
Excellent, thanks!!

xocodergit:
> @xocodergit: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> test_ztunnel	[146c916](https://github.com/istio/ztunnel/commit/146c916b49c5ff8c64c8c1de4a7dbe8c8f70093c)	[link](https://prow.istio.io/view/gs/istio-prow/pr-logs/pull/istio_ztunnel/1248/test_ztunnel/1822813912383361024)	true	`/test test`

@istio-testing /retest-required

xocodergit:
@istio-testing /test test

xocodergit:
@istio-testing /test test

xocodergit:
@istio-testing /test test

xocodergit:
@istio-testing /test test

Issue #1196, opened 2024-07-05
Failure to do `apk update` when ambient captured from wolfi pods


howardjohn:
```
/ $apk update
fetch https://packages.wolfi.dev/os/x86_64/APKINDEX.tar.gz
WARNING: updating and opening https://packages.wolfi.dev/os: IO ERROR
2 unavailable, 0 stale; 27 distinct packages available

```

howardjohn:
ERROR 
```
2024-07-05T17:45:46.516241Z     error   copy:proxy{wl=default/shell-56bd5dbdbf-2h248}:outbound{id=c6f65b42f7665bfb56eea5a7ecfffb39}:receive     howardjohn: loop
2024-07-05T17:45:46.516258Z     error   copy:proxy{wl=default/shell-56bd5dbdbf-2h248}:outbound{id=c6f65b42f7665bfb56eea5a7ecfffb39}:receive     howardjohn: read err=false
2024-07-05T17:45:46.516277Z     error   copy:proxy{wl=default/shell-56bd5dbdbf-2h248}:outbound{id=c6f65b42f7665bfb56eea5a7ecfffb39}:receive     howardjohn: write 256
2024-07-05T17:45:46.516297Z     error   copy:proxy{wl=default/shell-56bd5dbdbf-2h248}:outbound{id=c6f65b42f7665bfb56eea5a7ecfffb39}:receive     howardjohn: wrote 256/256
2024-07-05T17:45:46.516309Z     error   copy:proxy{wl=default/shell-56bd5dbdbf-2h248}:outbound{id=c6f65b42f7665bfb56eea5a7ecfffb39}:receive     howardjohn: loop
2024-07-05T17:45:46.516323Z     error   copy:proxy{wl=default/shell-56bd5dbdbf-2h248}:outbound{id=c6f65b42f7665bfb56eea5a7ecfffb39}:receive     howardjohn: read err=false
2024-07-05T17:45:46.516340Z     error   copy:proxy{wl=default/shell-56bd5dbdbf-2h248}:outbound{id=c6f65b42f7665bfb56eea5a7ecfffb39}:receive     howardjohn: write 256
2024-07-05T17:45:46.516369Z     error   copy:proxy{wl=default/shell-56bd5dbdbf-2h248}:outbound{id=c6f65b42f7665bfb56eea5a7ecfffb39}:receive     howardjohn: wrote 256/256
[...]

PR #1249, opened 2024-08-02
copy: fix handling of blocked writers

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/1196

This adds a test which tests how the copy logic behaves in the face of
writers that do not always accept all the data we are writing. This
exposes two issues:
* If the write is Pending entirely, we dropped the saved buffer.
* If we did a partial write, our check was backwards causing us to not
  save the buffer.

### DIFF
--- a/src/copy.rs
+++ b/src/copy.rs
@@ -287,11 +287,17 @@ where
 
             // This is just a reference counter. Hold onto it in case the write() is not complete.
             let mut our_copy = buffer.clone();
-            let i = ready!(Pin::new(&mut *me.writer).poll_write_buf(cx, buffer))?;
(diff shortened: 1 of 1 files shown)


src/copy.rs
```diff
@@ -479,4 +486,91 @@ mod tests {
             assert_eq!(res.as_slice(), body);
         }
     }
+
+    #[tokio::test]
+    async fn copystress() {
```
bleggett, line 491: :+1:

bleggett (approved):
Nice. I apologize for blaming `apk` :D

howardjohn:
> Nice. I apologize for blaming `apk` :D

At least we didn't blame the compiler :slightly_smiling_face:

PR #1261, opened 2024-08-13
add json header to config dump, so the browser renders it nicely

yuval-k:
tested manually, and the browser now renders it as collapsible json


ilrudie (approved):
LGTM

PR #1265, opened 2024-08-16
dns: prefer the original family of the request

howardjohn:
This fixes DNS SE usage when in an IPv6 only env


src/state.rs
```diff
@@ -562,14 +562,20 @@ impl DemandProxyState {
             );
             return Err(Error::NoValidDestination(Box::new(dst_workload.clone())));
         }
-        let ip = Box::pin(self.resolve_workload_address(dst_workload, src_workload)).await?;
+        let ip = Box::pin(self.resolve_workload_address(
+            dst_workload,
+            src_workload,
+            original_target_address,
```
ilrudie, line 568: all we really care about is whether the original_target_address is ipv6 or not. Do we need to pass around a SocketAddr for that or would it be better to simply pass around a bool or enum to represent the preferred ip family?

src/config.rs
```diff
@@ -356,7 +356,12 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
     // Increase some defaults. Note these are NOT coming from /etc/resolv.conf (only some fields do, we don't override those),
     // but rather hickory's hardcoded defaults
     dns_resolver_opts.cache_size = 4096;
-    // TODO: should we override server_ordering_strategy based on our IP support?
+    dns_resolver_opts.ip_strategy = if ipv6_enabled {
+        // Lookup both in parallel. We will do filtering in a later stage to only appropriate IP families
+        LookupIpStrategy::Ipv4AndIpv6
```
ilrudie, line 361: Maybe a note that this is needed rather than `Ipv6thenIpv4` so that we can match the ip family of the original request when ipv6 is enabled rather than blanket using ipv6 whenever AAAA records exist. I did spend a sec wondering if we could optimize this using a better strategy.

src/state.rs
howardjohn: Technically no, but I was thinking it is a bit more flexible and conveys the meaning of the variable well. We change between whether we need the IP, the port, the family, etc a few times in this codepath (and just above we use the full IP for instance)

ilrudie (commented):
overall this looks good, couple of comments but nothing looks like a showstopper

ilrudie (approved):
added dnm to allow for addressing non-blocking comments. Drop when you're ready

src/config.rs
howardjohn: I expanded the comment

Issue #1268, opened 2024-08-20
Optimize endpoint management for large services


howardjohn:
Currently, things are pretty slow when there is a Service with many endpoints. Our update (add or remove) operations do a clone+modify approach, and we process updates one-by-one.

At scale, this approximates N^2 behavior. For example, to modify 1 service from50k endpoints to 1, we do `clone(50k) + clone(49999) + ...`.

In some testing, this can cause extreme performance degradation on the admin thread; the 50k removal was estimated to take ~40minutes to complete.

howardjohn:
Two main approaches I think, possibly all of them:
* Process XDS updates in batches, instead of one by one
  * Possibly even multiple XDS events at once, if we have a backlog?? Makes (N)ACK harder though.
* Make our Service have interior mutability on the Endpoints. This way add/remove is just a simple hashmap add/remove (though it will probably not be just a _simple_ hashmap, since we need some thread safety here)
* Optimize how we handle the updates so we do less remove+add, when we don't need to remove?

PR #1269, opened 2024-08-20
Improvements to Endpoint handling 

howardjohn:
Part of https://github.com/istio/ztunnel/issues/1268

This doesn't solve the problem -- we don't fix the primary performance issues. However, this sets us up for success by refactoring the data structures.

* A new EndpointSet is a collection of endpoints. For now this is just a wrapper around a hashmap, but the intent is we could swap it out with something that allows updates, or maybe add some things like load tracking for better LB.
* Drop address from Endpoint (and Service). Instead, an Endpoint just points to a workload (and its a part of a Service, so no need to point to the service). This also means for dual stack we only need 1 endpoint per Pod instead of 2.
* Trivial optimization to avoid cloning a Service when we always immediately drop it.

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -500,6 +500,7 @@ mod tests {
         sync::{Arc, RwLock},
     };
 
+    use crate::state::service::EndpointSet;
     use crate::{
         rbac::Connection,
         state::{
@@ -507,8 +508,7 @@ mod tests {
             service::{Endpoint, Service},
             workload::{
                 application_tunnel::Protocol as AppProtocol, gatewayaddress::Destination,
-                ApplicationTunnel, GatewayAddress, NamespacedHostname, NetworkAddress, Protocol,
-                Workload,
+                ApplicationTunnel, GatewayAddress, NetworkAddress, Protocol, Workload,
             },
             DemandProxyState,
         },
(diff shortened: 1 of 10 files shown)


src/proxy/inbound.rs
```diff
@@ -500,15 +500,15 @@ mod tests {
         sync::{Arc, RwLock},
     };
 
+    use crate::state::service::EndpointSet;
     use crate::{
         rbac::Connection,
         state::{
             self,
-            service::{endpoint_uid, Endpoint, Service},
+            service::{Endpoint, Service},
```
ilrudie, line 508: **nit** Add EndpointSet here rather than on its own

src/xds.rs
```diff
@@ -183,17 +183,7 @@ impl ProxyStateUpdateMutator {
...
-                state
-                    .services
-                    .remove_endpoint(&prev.uid, &endpoint_uid(&prev.uid, Some(prev_addr)));
-            }
-            if prev.workload_ips.is_empty() {
-                state
-                    .services
-                    .remove_endpoint(&prev.uid, &endpoint_uid(&prev.uid, None));
-            }
+            state.services.remove_endpoint(&prev.uid, &prev.uid);
```
ilrudie, line 186: Sending the exact same parameter twice looks odd. Can we just clean this up to only require 1 parameter now?

src/proxy/inbound.rs
howardjohn: `cargo format` not being opionated is tragic

src/xds.rs
howardjohn: nice, thanks

ilrudie (approved):
LGTM

PR #1274, opened 2024-08-22
Optimize XDS surrounding workload updates

howardjohn:
This PR adds a benchmark which adds N workloads (through the XDS path) and removes them.

Results of the benchmark before/after this PR:
```
xds/insert-remove       time:   [862.29 µs 886.71 µs 910.46 µs]
                        change: [-96.768% -96.599% -96.324%] (p = 0.00 < 0.05)
                        Performance has improved.
```

To keep the benchmark small, its only testing 1k workloads. I also tested with 50k:
```
Now: 329ms
Before: not patient enough to run. 5k takes 7s.
```

In real world tests, previously it was estimated to take ~40minutes of 100% CPU burning to get through 50k workloads. Now, with a real Istiod connection we see that Istiod takes roughly 30s to get through them all, and ztunnel is able to easily keep up with only spikes of ~10% CPU.

The old logic was pretty flawed. Basically, Service is an immutable object that contains a map of endpoints. For each workload update, we would find all relevant services, remove the workload, then add it back. As it is an immutable object, this is done by clone+add 1/clone+remove 1. The scaling here is O(N^2) (or, maybe even O(N^3) -- maybe we can simplify this to O(I failed my algorithms class))

This PR introduces a series of optimizations. All percentages are from the baseline with 500 objects (note: since scaling is polynomial, the improvements are far greater at large scales):
* Don't blindly `remove` the old service; its more expensive to remove+add an object to a map than it is to update it. `-57.107%`
* `Arc` the endpoints, so that when we clone the endpoint set we don't need to clone each endpoint: `-67.251%`
* Remove our `workload_to_services` index, and instead keep the list of services alongside `Workload`: `-92.323%`
* ~Use [im](https://github.com/bodil/im-rs) HashMap for endpoints, which is explicitly meant for this style of immutable data structures w/ efficient mutation: `-96.599%`~ Reverted due to licensing concerns.

Fixes https://github.com/istio/ztunnel/issues/1268

Note: a new benchmark harness is added to isolate from the `throughput` ones which do more advanced setup which is binary-wide. We should move a few others over in a followup

### DIFF
--- a/benches/basic.rs
+++ b/benches/basic.rs
@@ -1,42 +1,17 @@
-use std::cmp::Ordering::{Equal, Greater, Less};
-use std::future::Future;
-use std::io::Error;
-use std::net::{IpAddr, Ipv4Addr, SocketAddr};
-use std::sync::mpsc::{Receiver, SyncSender};
-use std::sync::{Arc, RwLock};
 use std::time::Duration;
-use std::{io, thread};
 
-use bytes::{BufMut, Bytes};
-use criterion::measurement::Measurement;
-use criterion::{
-    criterion_group, criterion_main, BenchmarkGroup, Criterion, SamplingMode, Throughput,
-};
-use hickory_resolver::config::{ResolverConfig, ResolverOpts};
+use bytes::Bytes;
+use criterion::{criterion_group, criterion_main, Criterion};
 use pprof::criterion::{Output, PProfProfiler};
-use prometheus_client::registry::Registry;
-use tokio::io::{AsyncReadExt, AsyncWriteExt};
-use tokio::net::TcpStream;
-use tokio::sync::Mutex;
-use tracing::info;
 
-use ztunnel::rbac::{Authorization, RbacMatch, StringMatch};
-use ztunnel::state::workload::{Protocol, Workload};
-use ztunnel::state::{DemandProxyState, ProxyRbacContext, ProxyState};
-use ztunnel::test_helpers::app::{DestinationAddr, TestApp};
-use ztunnel::test_helpers::linux::{TestMode, WorkloadManager};
-use ztunnel::test_helpers::tcp::Mode;
-use ztunnel::test_helpers::{helpers, tcp};
-use ztunnel::xds::{LocalWorkload, ProxyStateUpdateMutator, ProxyStateUpdater};
-use ztunnel::{app, identity, metrics, proxy, rbac, setup_netns_test, strng, test_helpers};
+use ztunnel::state::ProxyState;
+use ztunnel::xds::ProxyStateUpdateMutator;
 
 pub fn xds(c: &mut Criterion) {
-    use ztunnel::xds::istio::workload::address::Type as XdsAddressType;
-    use ztunnel::xds::istio::workload::TunnelProtocol as XdsProtocol;
+    use ztunnel::xds::istio::workload::Port;
+    use ztunnel::xds::istio::workload::Service as XdsService;
     use ztunnel::xds::istio::workload::Workload as XdsWorkload;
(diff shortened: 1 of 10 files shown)


benches/basic.rs
```diff
@@ -0,0 +1,100 @@
...
+    let updater = ProxyStateUpdateMutator::new_no_fetch();
+    let svc = XdsService {
+        hostname: "example.com".to_string(),
+        addresses: vec![XdsNetworkAddress {
+            network: "".to_string(),
+            address: vec![127, 0, 0, 3],
+        }],
+        ..Default::default()
+    };
+    updater.insert_service(&mut state, svc).unwrap();
```
ilrudie, line 42: all this is done in the bench function as well. was the intent to capture these as a closure and reuse them?

src/state/service.rs
```diff
@@ -80,37 +81,38 @@ impl<'de> serde::Deserialize<'de> for EndpointSet {
...
+        self.inner.insert(k, Arc::new(v));
     }
 
     pub fn contains(&self, key: &Strng) -> bool {
         self.inner.contains_key(key)
     }
 
     pub fn get(&self, key: &Strng) -> Option<&Endpoint> {
-        self.inner.get(key)
+        self.inner.get(key).map(|e| &**e)
```
ilrudie, line 107:
```suggestion
        self.inner.get(key).map(Arc::as_ref)
```

howardjohn: Nice, you are better than chatgpt :slightly_smiling_face:

benches/basic.rs
howardjohn: Thanks! forgot to clean this up

howardjohn:
I think we should probably backport this, but let it soak a bit in case there is some subtle issues introduced that we don't catch until later testing

src/state/service.rs
ilrudie: tbh, it looked sooooo crazy to me that I spent a likely shameful amount of time trying to understand how rustc doesn't alert the FBI when this tries to compile. After that it was "oh, isn't this the same as?" 😂

src/state/service.rs
```diff
@@ -461,56 +451,65 @@ impl ServiceStore {
...
                 let _ = self.by_host.insert(hostname.clone(), vec![service.clone()]);
             }
             Some(services) => {
-                services.push(service.clone());
+                if let Some((cur, _)) = services
+                    .iter()
+                    .find_position(|s| s.namespace == service.namespace)
+                {
+                    // Service already exists; replace the slot
+                    services[cur] = service.clone()
```
ilrudie, line 506: Does this implicitly drop 1 strong to the old Arc<Service> that was held by the vec? I suppose it must or it would leak

howardjohn: yeah it will Drop the old one.

src/state/service.rs
```diff
@@ -461,56 +451,65 @@ impl ServiceStore {
...
+        self.insert_internal(service, true)
+    }
+
+    fn insert_internal(&mut self, mut service: Service, endpoint_update_only: bool) {
         let namespaced_hostname = service.namespaced_hostname();
-        if let Some(endpoints) = self.staged_services.remove(&namespaced_hostname) {
-            for (wip, ep) in endpoints {
-                if service.should_include_endpoint(ep.status) {
-                    service.endpoints.insert(wip.clone(), ep);
+        // If we're replacing an existing service, remove the old one from all data structures.
```
ilrudie, line 471:
```suggestion
```

I think the comment is wrong

ilrudie (approved):
bit improvements for sure! I agree with letting is soak for a bit

PR #1275, opened 2024-08-22
Don't create V6 binds if global V6 flag is off

bleggett:
Existing code handles the case where V6 is globally enabled but not enabled in the local pod context.

It does not handle the case where V6 is globally _disabled_ but is enabled in the local pod context, which is a tad surprising.

Only the DNS handler is affected by this.

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -94,16 +94,10 @@ impl Server {
         // This is to ensure globally-enabled V6 support doesn't try to create V6 addresses in pods
         // that do not support it, and also ensure globally-disabled V6 support doesn't create V6 address
         // even if that's turned off.
-        let local_address = if let config::Address::Localhost(v6_global_enable, _) = address {
-            if v6_global_enable {
-                address.with_ipv6(socket_factory.ipv6_enabled_localhost().unwrap_or_else(|e| {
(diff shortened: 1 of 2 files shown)


src/dns/server.rs
```diff
@@ -89,10 +89,21 @@ impl Server {
...
-            true
-        }));
+        // if the address we got from config is supposed to be v6-enabled,
+        // actually check if the local pod context our socketfactory operates in supports V6.
+        // This is to ensure globally-enabled V6 support doesn't try to create V6 addresses in pods
+        // that do not support it, and also ensure globally-disabled V6 support doesn't create V6 address
+        // even if that's turned off.
+        let local_address = if let config::Address::Localhost(v6_global_enable, _) = address {
+            if v6_global_enable {
+                address.with_ipv6(socket_factory.ipv6_enabled_localhost().unwrap_or_else(|e| {
```
howardjohn, line 99: nit: maybe we should put this in with_ipv6 and maybe call it something like "maybe disable ipv6"?

howardjohn (approved):
Nice find.

src/dns/server.rs
bleggett: Yeah it's a bit too cute doing it out here, agreed.

PR #1278, opened 2024-08-23
Fix incorrect function naming

howardjohn:
The behavior is right, just the naming is off. Its a weird double
negative since we are checking its `!disabled`


bleggett (approved):

MorrisLaw (commented):
Still some reference to `ipv6_disabled_on_localhost` in [src/config.rs.](https://github.com/istio/ztunnel/blob/4de5e896fad8c5fb8744c2acba6b8f40f5abd194/src/config.rs#L291)

PR #1279, opened 2024-08-26
test: show how to enter user namespace as well

howardjohn:
To make debugging more realistic

### DIFF
--- a/src/test_helpers/namespaced.rs
+++ b/src/test_helpers/namespaced.rs
@@ -116,5 +116,5 @@ pub fn initialize_namespace_tests() {
     .expect("xtables bindmount");
 
     let pid = unsafe { getpid() };
-    eprintln!("Starting test in {tmp:?}. Debug with `sudo nsenter --mount --net -t {pid}`");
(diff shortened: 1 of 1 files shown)


src/test_helpers/namespaced.rs
```diff
@@ -116,5 +116,5 @@ pub fn initialize_namespace_tests() {
     .expect("xtables bindmount");
 
     let pid = unsafe { getpid() };
-    eprintln!("Starting test in {tmp:?}. Debug with `sudo nsenter --mount --net -t {pid}`");
+    eprintln!("Starting test in {tmp:?}. Debug with `sudo nsenter --mount --net --setuid=0 --preserve-credentials --user -t {pid}`");
```
bleggett, line 119: which wins? `--setuid=0` or `--preserve-credentials`? I thought the latter would ignore uid/gid changes?

howardjohn: We want to be uid 0 in the namespace (which is **mapped** to our original ID, but is not our original id), so we override that -- preserve-credentials is for the gid really

bleggett (approved):

PR #1280, opened 2024-08-26
dns: use new local workload fetch

howardjohn:
Followup on https://github.com/istio/ztunnel/pull/1218, applying the
same logic to DNS server

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -34,7 +34,7 @@ use std::sync::Arc;
 use std::time::Duration;
 use tracing::{debug, info, instrument, trace, warn};
 
-use crate::proxy::SocketFactory;
+use crate::proxy::{LocalWorkloadFetcher, SocketFactory};
 
 use crate::config::ProxyMode;
 use crate::dns::metrics::{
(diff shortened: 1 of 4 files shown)


src/dns/server.rs
```diff
@@ -74,20 +72,18 @@ impl Server {
...
+    pub async fn new(
         domain: String,
         address: config::Address,
-        network: S,
         state: DemandProxyState,
         forwarder: Arc<dyn Forwarder>,
         metrics: Arc<Metrics>,
         drain: DrainWatcher,
         socket_factory: &(dyn SocketFactory + Send + Sync),
-        allow_unknown_source: bool,
```
bleggett, line 90: oh ok, this _was_ just for testing. I was wondering why it was `false` everywhere.

howardjohn: Yeah I think we were just lazy in the tests basically/it was harder to get the source before

src/proxy.rs
```diff
@@ -190,6 +187,37 @@ impl LocalWorkloadInformation {
     pub fn workload_info(&self) -> Arc<WorkloadInfo> {
         self.wi.clone()
     }
+
+    pub fn as_fetcher(&self) -> Arc<LocalWorkloadFetcher> {
+        LocalWorkloadFetcher::new(self.wi.clone(), self.state.clone())
+    }
+}
+
+/// LocalWorkloadFetcher is essentially LocalWorkloadInformation without CA access.
```
bleggett, line 196: Eventually it might make sense to invert this a bit and do

```
pub struct LocalWorkloadInformation {
    fetcher: LocalWorkloadFetcher
    // full_cert_manager gives access to the full SecretManager. This MUST only be given restricted
    // access to the appropriate certificates
    full_cert_manager: Arc<SecretManager>,
}
```

but this makes sense as well.

bleggett (approved):

PR #1283, opened 2024-08-27
Test with coverage

Stevenjin8:


Makefile.core.mk
```diff
@@ -6,7 +6,7 @@ ifeq ($(TLS_MODE), boring)
 endif
 
 test:
-	RUST_BACKTRACE=1 cargo test --benches --tests --bins $(FEATURES)
+	FEATURES=$(FEATURES) ./scripts/test-with-coverage.sh
```
howardjohn, line 9: can we keep this as a special `coverage` target? Most ofthe time if I just want to test I don't want to waste all the effort and complexity of coverage.

And also not run that in presubmit.

scripts/test-with-coverage.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
```
bleggett, line 1:
```suggestion
#!/usr/bin/env sh
```

nit - can we stick to posix sh please, unless we really need a bashism here.

scripts/test-with-coverage.sh
```diff
@@ -0,0 +1,65 @@
...
+# Clean directory
+rm -rf "$profdata" "$profiles"
+mkdir -p "$profiles"
+
+echo "Running tests"
+cargo test --benches --tests --bins $FEATURES
+
+# Merge profraw data
+echo "Merging profraw files in $profiles to $profdata"
+llvm-profdata merge -sparse $(find "$profiles" -name '*.profraw') -o $profdata
```
bleggett, line 43: `llvm` is not a hard dep of  `rust`, is it? Can we add a check in the makefile that this binary exists in $PATH before running the script?

bleggett (commented):
Mostly nits, lgtm

scripts/test-with-coverage.sh
Stevenjin8: My understanding the rust very much depends on llvm. Though its not clear whether having rust means having the whole llvm toolchain.

bleggett: Oh my bad, it is. (Well, it doesn't have to be, but is in the default builds).

Nevermind then, probably safe to assume.

bleggett: Yeah on fedora for instance the binary tools appear to be separate/optional. But it's not a big deal.

Stevenjin8: Also, the test and coverage targets are different

scripts/test-with-coverage.sh
dhawton: Most scripts around Istio are bash, so probably for consistency sake it should be bash. Doesn't matter too much either way.

bleggett: Yep that's fair. I've encountered bugs using scripts that assume bash in environments that don't have bash. And if the script doesn't actually need bash, it's generally pointless and also less compatible to require it.

bleggett (approved):
TY!

dhawton (approved):

PR #1284, opened 2024-08-28
zds: fix retrying a bad netns

howardjohn:
Fixes https://github.com/istio/istio/issues/52858

The issue here is this sequence of events:

```
Pod update: no IP assigned
Get CNI plugin event, Netns=UID1
Partial add error no ztunnel connection

CNI event, Netns=UID2 (changed!)
Partial add error no ztunnel connection
Sending pod to ztunnel as part of snapshot
ACK error, cannot assigned

CNI event, Netns=UID3 (changed!)
No other log from the CNI plugin, which I think implies it did not fail
```

Basically, we **successfully start the proxy**, but have the same UID in the pending queue. When the pending retries, it kills the working on + fails to start (since it has its own netns which is bogus).

The fix is to remove a proxy from pending queue when its added

### DIFF
--- a/src/inpod/statemanager.rs
+++ b/src/inpod/statemanager.rs
@@ -231,6 +231,8 @@ impl WorkloadProxyManagerState {
             .await
         {
             Ok(()) => {
+                // If the workload is already pending, make sure we drop it, so we don't retry.
+                self.pending_workloads.remove(workload_uid);
                 self.update_proxy_count_metrics();
                 Ok(())
             }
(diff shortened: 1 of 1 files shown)


src/inpod/statemanager.rs
```diff
@@ -231,6 +231,8 @@ impl WorkloadProxyManagerState {
             .await
         {
             Ok(()) => {
+                // If the workload is already pending, make sure we drop it, so we don't retry.
+                self.pending_workloads.remove(workload_uid);
```
bleggett, line 235: I'm beginning to think we should probably key all of the workload stores with UID + NetNSID to avoid future goofiness like this.

bleggett (approved):
LGTM, nice find, ty!

src/inpod/statemanager.rs
howardjohn: one nice part about UID is it means we will never have multiple Proxy running for a given pod. I would worry if we have netns ID as a key we could end up with some leaky Proxy instances somehow maybe?

Might be possible, but needs care

PR #1290, opened 2024-09-04
Remove deadcode in waypoint_svc_ip_address

howardjohn:
Minor cleanup

### DIFF
--- a/src/state/workload.rs
+++ b/src/state/workload.rs
@@ -242,17 +242,6 @@ impl Workload {
             service_account: self.service_account.clone(),
         }
     }
-    pub fn waypoint_svc_ip_address(&self) -> Result<Option<IpAddr>, WaypointError> {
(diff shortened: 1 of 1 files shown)


src/state/workload.rs
```diff
@@ -242,17 +242,6 @@ impl Workload {
             service_account: self.service_account.clone(),
         }
     }
-    pub fn waypoint_svc_ip_address(&self) -> Result<Option<IpAddr>, WaypointError> {
```
zirain, line 245: is there any linter in rust like unsed function?

howardjohn: there is, i think maybe doesn't detect it due to it being public

zirain: oha, thinked it in golang-way, lower-case function means private.

keithmattix (approved):

ilrudie (approved):
lgtm

PR #1294, opened 2024-09-05
inbound: unconditionally log errors

howardjohn:
Previously, our error logging was built on an assumption that we would
always remember to log errors and return HTTP responses. This is not
great, we should enforce this in the type system -- we are only one `?`
away from broken error handling, which we do have one currently
(`local_workload_information`).

This builds into the type system enforcing we cannot accidentally mask
errors.

Instead of `serve_connect` returning an error (which goes to /dev/null),
we have it return nothing -- this should make it harder to insert any
new code in the wrong place that doesn't handle errors. Additionally, we
split up the function into 3 phases:
* Before we have full metrics context
* After we have full metrics context
* After we already sent a 200 back

This way we can use `?` ergonomically + safely, and consolidate where we
log/return error responses into a small set of places

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -12,22 +12,22 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use futures::stream::StreamExt;
+use futures_util::TryFutureExt;
+use http::{Method, Response, StatusCode};
 use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::Instant;
-
-use futures::stream::StreamExt;
-
-use http::{Method, Response, StatusCode};
 
 use tokio::sync::watch;
 
 use tracing::{debug, info, instrument, trace_span, Instrument};
 
-use super::{Error, LocalWorkloadInformation};
+use super::{ConnectionResult, Error, LocalWorkloadInformation, ResponseFlags};
 use crate::baggage::parse_baggage_header;
 use crate::identity::Identity;
 
+use crate::config::{Config, ProxyMode};
 use crate::drain::DrainWatcher;
 use crate::proxy::h2::server::H2Request;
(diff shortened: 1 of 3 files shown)


src/proxy/inbound.rs
```diff
@@ -95,7 +94,9 @@ impl Inbound {
                     let pi = self.pi.clone();
                     let (raw_socket, ssl) = tls.get_ref();
                     let src_identity: Option<Identity> = tls::identity_from_connection(ssl);
-                    let dst = crate::socket::orig_dst_addr_or_default(raw_socket);
+                    let dst = crate::socket::to_canonical(
```
ilrudie, line 97: **nit**: to_canonical has a direct in scope binding (ln 36)
```suggestion
                    let dst = to_canonical(
```

src/proxy/inbound.rs
```diff
@@ -155,91 +156,158 @@ impl Inbound {
...
+                InboundError(
+                    Error::ConnectAddress(req.uri().to_string()),
+                    StatusCode::BAD_REQUEST,
+                )
+            })?;
+
+        let destination_workload = pi
+            .local_workload_information
+            .get_workload()
+            .await
```
ilrudie, line 267: Would this be infallible? It seems like we shouldn't get to a point where the ztunnel is accepting connections of behalf of a workload it doesn't know about.

howardjohn: get_workload does a join with XDS, so that may not have landed yet

howardjohn: Added a comment about it, but i don't think its a safe enough assumption to `expect()` it

howardjohn: There could be future TLS providers that don't depend on the local workload, etc

ilrudie (approved):
LGTM

PR #1301, opened 2024-09-10
release.sh: support building alternative TLS providers

howardjohn:
Small helper to make these release scripts easy to build alternative TLS
modes

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -19,21 +19,25 @@ set -ex
 WD=$(dirname "$0")
 WD=$(cd "$WD" || exit; pwd)
 
-case $(uname -m) in
-    x86_64)
-      export ARCH=amd64;;
-    aarch64)
-      export ARCH=arm64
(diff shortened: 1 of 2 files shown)


scripts/release.sh
```diff
@@ -19,21 +19,25 @@ set -ex
...
-    aarch64)
-      export ARCH=arm64
-      # TODO(https://github.com/istio/ztunnel/issues/357) clean up this hack
-      sed -i 's/x86_64/arm64/g' .cargo/config.toml
-      ;;
-    *) echo "unsupported architecture"; exit 1 ;;
-esac
-
-cargo build --release
+if [[ "$TLS_MODE" == "boring" ]]; then
```
dhawton, line 22: Should we doc this in the README? https://github.com/istio/ztunnel?tab=readme-ov-file#tlscrypto-provider

howardjohn: good call, done

dhawton (approved):

PR #1304, opened 2024-09-10
Validate (and do not set in our test) invalid app_tunnnel

howardjohn:
We shouldn't accept this type of config, as it has no meaning. We also
don't need to test it


ilrudie (approved):
lgtm

PR #1305, opened 2024-09-10
Construct config only when we need it

howardjohn:
Avoid building it for `version` or `help`.

Just a minor cleanup, its not a huge deal.


ilrudie (approved):
lgtm

PR #1306, opened 2024-09-10
Allow referencing waypoints that are sent as workloads rather than se…

howardjohn:
…rvices

Currently, we are failing if the waypoint is a workload. This is fine for our typical cases, where we have a Gateway making a Service+Deployment, so we get a Service + Workloads.

But in other cases, we may use an external waypoint solution, where we only get an Address in the Gateway status, and we need to map it to something. In these cases, it may be desireable to support these as Workloads. For instance, we map a variety of things as Workloads -- not just Pods:
```
		PodWorkloads,
		WorkloadEntryWorkloads,
		ServiceEntryWorkloads,
		EndpointSliceWorkloads,
		NetworkGatewayWorkloads,
```

So this enables any of those to be used

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -240,13 +240,20 @@ impl ProxyState {
 
     /// Find either a workload or a service by hostname.
     pub fn find_hostname(&self, name: &NamespacedHostname) -> Option<Address> {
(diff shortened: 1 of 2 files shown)


ilrudie (approved):
lgtm

PR #1307, opened 2024-09-11
Support inline XDS config and auth

howardjohn:
Having to always use files is troublesome when we want to run in locked down environments that don't have file write access. Nice to be able to set these inline


keithmattix (approved):

ilrudie (approved):
LGTM

PR #1315, opened 2024-09-17
build: allow building off linux platforms

howardjohn:
Some small fixes to make things compile. Unfortunately I don't know a
great way to enforce this in CI, but for now better to fix then leave it
broken I suppose.


bleggett:
One of those times when Github Actions would be simpler than Prow :sweat_smile:

bleggett (approved):

PR #1316, opened 2024-09-17
tls: better error on TLS handshake failures

howardjohn:
Before: `error="io error: invalid peer certificate: ApplicationVerificationFailure"`
After: `error="io error: invalid peer certificate: Other(OtherError(SanError([], [Spiffe { trust_domain: \"\", namespace: \"default\", service_account: \"default\" }])))"`

Error still is a bit wonky, rustls doesn't give us much flexibility. But at least this lets a developer to understand what is going on

### DIFF
--- a/src/tls/lib.rs
+++ b/src/tls/lib.rs
@@ -15,7 +15,6 @@
 use super::Error;
 
 use crate::identity::{self, Identity};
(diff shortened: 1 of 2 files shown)


ilrudie (approved):
LGTM

PR #1328, opened 2024-10-01
logging: minor tweaks to debugging

howardjohn:
While debugging an issue (which was not a real issue, it turns out), I
was quite confused by the logs. This is mostly around not logging
service xds operations, and doing some confusing `remove` calls (which
handles both svc and workload) when we know its a workload. This just
cleans things upa bit


ilrudie (approved):
lgtm

PR #1330, opened 2024-10-01
Prevent `Error` enum from being huge

howardjohn:
This one case makes the whole enum big, which ends up bloating some call
stacks. Clippy has a lint for this I  think, but we must be under the
threshold.

Before
```
tracing::instrument::Instrumented<ztunnel::proxy::inbound_passthrough::InboundPassthrough::run::{{closure}}::{{closure}}::{{closure}}::{{closure}}>,2784
tracing::instrument::Instrumented<ztunnel::proxy::outbound::Outbound::run::{{closure}}::{{closure}}::{{closure}}::{{closure}}>,1648
tracing::instrument::Instrumented<ztunnel::proxy::socks5::Socks5::run::{{closure}}::{{closure}}::{{closure}}::{{closure}}>,1968
ztunnel::proxy::inbound::Inbound::run::{{closure}}::{{closure}}::{{closure}}::{{closure}},1400
```

After
```
tracing::instrument::Instrumented<ztunnel::proxy::inbound_passthrough::InboundPassthrough::run::{{closure}}::{{closure}}::{{closure}}::{{closure}}>,2464
tracing::instrument::Instrumented<ztunnel::proxy::outbound::Outbound::run::{{closure}}::{{closure}}::{{closure}}::{{closure}}>,1488
tracing::instrument::Instrumented<ztunnel::proxy::socks5::Socks5::run::{{closure}}::{{closure}}::{{closure}}::{{closure}}>,1808
ztunnel::proxy::inbound::Inbound::run::{{closure}}::{{closure}}::{{closure}}::{{closure}},1400
```

PR #1331, opened 2024-10-01
Reduce stack bloat on inbound

howardjohn:
Builds on https://github.com/istio/ztunnel/pull/1330
First, add new checks. Our inbound flow has 3 nested futures, we only
measured the outer one.

Next, two optimizations. Unfortunately, I dont really understand either
of them...  the stack sizing is a bit of black magic.

1. Drop #instrument for .instrument(). Result is the same but one uses
   500b less?
2. Inline handle_connection, which saves 1kb

Before
```
tracing::instrument::Instrumented<ztunnel::proxy::inbound_passthrough::InboundPassthrough::run::>,2464
tracing::instrument::Instrumented<ztunnel::proxy::outbound::Outbound::run::>,1488
tracing::instrument::Instrumented<ztunnel::proxy::socks5::Socks5::run::>,1808
ztunnel::proxy::h2::server::serve_connection<>>,6440
ztunnel::proxy::inbound::Inbound::run::,1400
ztunnel::proxy::inbound::Inbound::serve_connect::{{closure}},3576
```

After
```
tracing::instrument::Instrumented<ztunnel::proxy::inbound_passthrough::InboundPassthrough::run::>,1672
tracing::instrument::Instrumented<ztunnel::proxy::outbound::Outbound::run::>,1488
tracing::instrument::Instrumented<ztunnel::proxy::socks5::Socks5::run::>,1808
ztunnel::proxy::h2::server::serve_connection<>>6440
ztunnel::proxy::inbound::Inbound::run::,1400
tracing::instrument::Instrumented<ztunnel::proxy::inbound::Inbound::serve_connect::{{closure}}>,2128
```

### DIFF
--- a/src/proxy/connection_manager.rs
+++ b/src/proxy/connection_manager.rs
@@ -20,7 +20,6 @@ use serde::{Serialize, Serializer};
 use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
 use std::fmt::Formatter;
-use std::future::Future;
 use std::net::SocketAddr;
 
 use crate::drain;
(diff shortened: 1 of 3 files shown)


src/proxy/connection_manager.rs
```diff
@@ -82,19 +81,29 @@ pub struct ConnectionGuard {
...
-    pub async fn handle_connection(
-        mut self,
-        send: impl Future<Output = Result<(), Error>> + Sized,
-    ) -> Result<(), Error> {
-        let watch = self.watch.take().expect("watch cannot be taken twice");
+// For reasons that I don't fully understand, this uses an obscene amount of stack space when written as a normal function,
+// amounting to ~1kb overhead per connection.
+// Inlining it removes this entirely, and the macro ensures we do it consistently across the various areas we use it.
+#[macro_export]
+macro_rules! handle_connection {
```
ilrudie, line 88: What happens with `#[inline(always)]` rather than going all the way to a macro?

ilrudie:
This looks good. I do wonder if we can get most/all of the benefit by just asking the compiler to inline the call rather than using a macro though.

keithmattix:
+1 to avoiding a macro if we can

src/proxy/connection_manager.rs
howardjohn: :eyes:

ilrudie:
> Inlining has no impact. I can't say I understand why, but also feel the 1kb/connection benefit here is worth a small macro... would love to avoid it if anyone has a solution, though.

That is ultimately just a compiler suggestion. Things I could see impacting it are:

1. its a method and for some reason that form throws the compiler off, even though it's really just some syntactic sugar.
2. the method/function we asked to inline uses private methods/function and thus inline is not possible, essentially the same things you encountered when you made the macro

Anyhow, if you want to fiddle more go for it, but otherwise I think we did our due diligence on keeping ztunnel's syntax simpler for folks and came up lacking this time. I'll go ahead and approve but add a dnm in case you'd prefer to fiddle around with inlining more based on the suggestions I left.

ilrudie (approved):
Just drop the dnm label if you feel like you are done messing with non-macro inline stuff.

PR #1332, opened 2024-10-01
socks5: slightly improve error reporting

howardjohn:
This is the start of proper socks5 error reporting. There is some
followup needed, but its pretty invasive, so wanted to get at least the
base merged. The full thing is in
https://github.com/howardjohn/ztunnel/tree/socks5/report-error and I
will keep working on it and followup.

End result, using a Go client when we have a DNS lookup error:

Before: `socks connect tcp 127.0.0.1:15080->echo.ecs.local:8080: dial tcp 127.0.0.1:15080: connect: connection refused`
After: `socks connect tcp 127.0.0.1:15080->echo.ecs.local:8080: unknown error host unreachable`

### DIFF
--- a/src/proxy/socks5.rs
+++ b/src/proxy/socks5.rs
@@ -107,7 +107,7 @@ impl Socks5 {
                                 debug!(component="socks5", dur=?start.elapsed(), "connection completed");
                             }).instrument(span);
 
-                            assertions::size_between_ref(1000, 3000, &serve);
+                            assertions::size_between_ref(1000, 2000, &serve);
                             tokio::spawn(serve);
                         }
                         Err(e) => {
@@ -133,7 +133,6 @@ impl Socks5 {
 async fn handle_socks_connection(mut oc: OutboundConnection, mut stream: TcpStream) {
     match negotiate_socks_connection(&oc.pi, &mut stream).await {
         Ok(target) => {
-            // In many cases, when we are fail we don't have this.
             // TODO: ideally, we send the success after we connect. This allows us to actually give a
             // success only when we really succeeded, rather than if we completed the SOCKS handshake.
             // Additionally, it would allow us to get a proper address to send back/
@@ -152,10 +151,12 @@ async fn handle_socks_connection(mut oc: OutboundConnection, mut stream: TcpStre
         }
     }
 }
-// handle will process a SOCKS5 connection. This supports a minimal subset of the protocol,
-// sufficient to integrate with common clients:
+
+// negotiate_socks_connection will handle the negotiation of a SOCKS5 connection.
+// This ultimately outputs the target socket address, if the handshake is successful.
+// This supports a minimal subset of the protocol, sufficient to integrate with common clients:
 // - only unauthenticated requests
-// - only CONNECT, with IPv4 or IPv6
+// - only CONNECT, with IPv4/IPv6/Hostname
 async fn negotiate_socks_connection(
     pi: &ProxyInputs,
     stream: &mut TcpStream,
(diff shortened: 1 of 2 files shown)


src/proxy/socks5.rs
```diff
@@ -101,14 +100,14 @@ impl Socks5 {
...
-                                    _ = handle(oc, stream) => {}
+                                    _ = handle_socks_connection(oc, stream) => {}
                                 }
                                 // Mark we are done with the connection, so drain can complete
                                 drop(drain);
                                 debug!(component="socks5", dur=?start.elapsed(), "connection completed");
                             }).instrument(span);
 
-                            assertions::size_between_ref(1000, 2000, &serve);
+                            assertions::size_between_ref(1000, 3000, &serve);
```
ilrudie, line 110: Did things really grow this much? Probably OK as I think socks5 is a bit more niche.

src/proxy/socks5.rs
```diff
@@ -131,26 +130,56 @@ impl Socks5 {
         .await
     }
 }
-
+async fn handle_socks_connection(mut oc: OutboundConnection, mut stream: TcpStream) {
+    match negotiate_socks_connection(&oc.pi, &mut stream).await {
+        Ok(target) => {
+            // In many cases, when we are fail we don't have this.
```
ilrudie, line 136: What is _this_ in the comment? We don't have a target?

ilrudie: oh, I think this was meant to go refer to the dummy address and is part of a copy/paste from `send_error`

src/proxy/socks5.rs
```diff
@@ -185,40 +222,45 @@ async fn handle(mut oc: OutboundConnection, mut stream: TcpStream) -> Result<(),
...
             };
 
-            ip = dns_lookup(resolver.clone(), remote_addr, ds).await?;
-            // oc.pi.resolver.lookup()
-            // oc.pi.lookup_service_or_query(ds)
-            // return Err(anyhow::anyhow!("unsupported host {ds:?}"));
+            match dns_lookup(resolver.clone(), remote_addr, ds).await {
+                Ok(ip) => ip,
+                Err(e) => {
+                    return Err(SocksError::HostUnreachable(e));
```
ilrudie, line 256: Should this be like `UnknownHost` rather than unreachable?

src/proxy/socks5.rs
```diff
@@ -131,26 +130,56 @@ impl Socks5 {
...
+                socket::to_canonical(stream.peer_addr().expect("must receive peer addr"));
+            oc.proxy_to(stream, remote_addr, target).await
+        }
+        Err(e) => {
+            warn!("failed to negotiate socks connection: {e}");
+            send_error(&e, &mut stream).await;
+        }
+    }
+}
 // handle will process a SOCKS5 connection. This supports a minimal subset of the protocol,
```
ilrudie, line 155: This comment needs updating, took a crack at it but it can probably still be better worded
```suggestion
// negotiate_socks_connection will inspect a SOCKS5 connection to produce a socket address for supported connections. This supports a minimal subset of the protocol,
```

src/proxy/socks5.rs
howardjohn: The only errors in the socks protocol are:

```
             o  X'01' general SOCKS server failure
             o  X'02' connection not allowed by ruleset
             o  X'03' Network unreachable
             o  X'04' Host unreachable
             o  X'05' Connection refused
             o  X'06' TTL expired
             o  X'07' Command not supported
             o  X'08' Address type not supported
```

ilrudie: ah, yeah so the only real advantage would be around logging it more clearly maybe but at the end we would have to map this down to on of the above at which point we'd have this odd disconnect between what was logged and what was responded that would need mapping. Comment withdrawn

ilrudie (approved):
LGTM, thanks

PR #1337, opened 2024-10-07
tls: better error messages

howardjohn:
The Debug form of these messages is much harder to read than the
Display, so do some tricks to get these to work


ilrudie (approved):
lgtm

Issue #1356, opened 2024-11-03
metrics endpoint sends content-type: application/openmetrics-text 


craigbox:
Could ztunnel serve `/metrics` as `text/plain` if the user hasn't explicitly sent an `Accept:` header for `application/openmetrics-text`?  Without, the browser will download the file, rather than showing it as plain text on the screen.

https://github.com/spring-projects/spring-boot/issues/28446 is an example of this pattern being fixed in Spring Boot.

(A waypoint, OTOH, will return `text/plain` and work as expected)

howardjohn:
Envoy: `content-type: text/plain; charset=UTF-8`
Istio-agent: `Content-Type: text/plain; version=0.0.4; charset=utf-8`
Ztunnel: `content-type: application/openmetrics-text;charset=utf-8;version=1.0.0`

So yeah this looks right && there is a bug in istio-agent as well putting `version` when it shouldn't

ilrudie:
This definitely is fixable, just need to decide on the correct behavior. Based on the above from John, do we care to ever have the openmetrics-text content type or just always say its plain text?

howardjohn:
We need to say it openmetrics-text when prometheus requests it I believe

howardjohn:
Just to be clear, the envoy/istio-proxy ones were when i did NOT send the
'accept' header at all. Prometheus will - I didn't test how they behave
then, but its definitely some  non-plaintext content

ilrudie:
In testing, near as I can tell envoy ignores the accept header entirely. You have 2 different content-types being returned which are based on query params, not headers. If you include `?format=json` you get 'application/json' and for everything else you get 'text/plain'. This assumes you are querying the internal admin interface which has more options. If you use 15090/stats/prometheus then it always uses 'text/plain' even if you request json formatting.

howardjohn:
Well that might break with prometheus 3.0! https://github.com/prometheus/prometheus/pull/15136

Note we don't always hit envoy directly but rather proxy through istio-agent, so that one may matter less

ilrudie:
Interesting. I think envoy behavior is probably fine for this. They call out that text/plain is the content-type expected for Prom parser and Envoy specifically supports Prom format, but does not support OpenMetrics (which is similar but stricter from what I can gather).

howardjohn:
Ztunnel is more strictly openmetrics fwiw

ilrudie:
[...]

PR #1361, opened 2024-11-08
adjust metrics content-type

ilrudie:
mostly serve metrics with text/plain unless given an openmetrics Accept header

fixes https://github.com/istio/ztunnel/issues/1356

### DIFF
--- a/src/metrics/server.rs
+++ b/src/metrics/server.rs
@@ -73,25 +73,7 @@ async fn handle_metrics(
             .expect("builder with known status code should not fail");
     }
 
-    let response_content_type: &str = req
-        .headers()
-        .get_all(http::header::ACCEPT)
-        .iter()
-        .find_map(|v| {
-            match v
-                .to_str()
-                .unwrap_or_default()
-                .to_lowercase()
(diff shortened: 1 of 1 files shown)


src/metrics/server.rs
```diff
@@ -73,12 +73,26 @@ async fn handle_metrics(
...
+    let response_content_type = match req
+        .headers()
+        .get(http::header::ACCEPT)
+        .unwrap_or(&http::HeaderValue::from_str("").expect("static str must parse"))
+        .to_str()
+        .unwrap_or_default()
+        .to_lowercase()
+        .split(";")
+        .collect::<Vec<_>>()
+        .first()
```
howardjohn, line 85: Do we know that prometheus will always scrape with openmetrics as the first 'accept'

ilrudie: Good point, that is likely not a safe invariant

ilrudie: Finally got back around to this. Got it reworked to support multiple Accept headers and added testing.

howardjohn (approved):

PR #1366, opened 2024-11-18
Make loop protection more robust

howardjohn:
* Apply even for non-inpod
* Also block calls to localhost outbound
* Test ^


bleggett (approved):
> Apply even for non-inpod

I still worry a bit we may have a scenario where this needs to be tweaked for non-inpod, but we can always fix that when and if we need it.

PR #1377, opened 2024-12-05
Add default TCP keepalives

howardjohn:
Our goal with keepalives is to stop things from dropping our connection prematurely.
So we want this to be a short enough interval to achieve that goal, without causing
excessive pings.

Note that keepalives are not hop-by-hop. So without setting keepalives in Ztunnel,
an application may rely on keepalives which are now only going to the local Ztunnel.
This results in hitting timeout's and unexpected behavior. This only impacts TCP keepalives;
application level keepalives work fine.

Some popular services, like Google LBs have a 10 minute timeout, so we will aim to be
well below that.

Other systems' defaults:
* Go: 15s/15s, 9 retries
* Linux: 2hr delay, 75s interval, 9 retries (note: its off by default, though)
* Envoy: no default
* Linkerd2: 10s delay (then hitting Linux level settings for the rest)

Note that because keepalives are a property of the socket (which has two ends), not the connection,
we cannot somehow read what the peer set and forward it (which would be neat).

Below shows why Ztunnel breaks existing applications:
image

There have been >4 reports on slack that I think this will fix. One user tried out a pre-release and said it resolved their issues. Impacted users will see logs like `bytes_recv=7589 duration="1721673ms" error="receive: io error: Connection timed out (os error 110)"` (note the 'connection timed out' but >0 bytes).

This defaults to ON by default.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -39,6 +39,11 @@ const NODE_NAME: &str = "NODE_NAME";
 const PROXY_MODE: &str = "PROXY_MODE";
 const PROXY_WORKLOAD_INFO: &str = "PROXY_WORKLOAD_INFO";
 const PACKET_MARK: &str = "PACKET_MARK";
+const KEEPALIVE_TIME: &str = "KEEPALIVE_TIME";
+const KEEPALIVE_INTERVAL: &str = "KEEPALIVE_INTERVAL";
+const KEEPALIVE_RETRIES: &str = "KEEPALIVE_RETRIES";
+const KEEPALIVE_ENABLED: &str = "KEEPALIVE_ENABLED";
+const USER_TIMEOUT_ENABLED: &str = "USER_TIMEOUT_ENABLED";
 const INPOD_UDS: &str = "INPOD_UDS";
 const INPOD_PORT_REUSE: &str = "INPOD_PORT_REUSE";
 const CLUSTER_ID: &str = "CLUSTER_ID";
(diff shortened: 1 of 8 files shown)


src/config.rs
```diff
@@ -273,6 +303,16 @@ fn parse_default<T: FromStr>(env: &str, default: T) -> Result<T, Error> {
...
 }
 
+fn parse_duration(env: &str) -> Result<Option<Duration>, Error> {
+    parse::<String>(env)?
+        .map(|ds| duration_str::parse(&ds).map_err(|_| Error::EnvVar(env.to_string(), ds)))
+        .transpose()
+}
+
+fn parse_duration_default(env: &str, default: Duration) -> Result<Duration, Error> {
+    parse_duration(env).map(|v| v.unwrap_or(default))
```
ilrudie, line 313: why do we need to map this? seems like we could just
```suggestion
    parse_duration(env).unwrap_or(default)
```

src/config.rs
```diff
@@ -273,6 +303,16 @@ fn parse_default<T: FromStr>(env: &str, default: T) -> Result<T, Error> {
     parse(env).map(|v| v.unwrap_or(default))
 }
 
+fn parse_duration(env: &str) -> Result<Option<Duration>, Error> {
+    parse::<String>(env)?
+        .map(|ds| duration_str::parse(&ds).map_err(|_| Error::EnvVar(env.to_string(), ds)))
+        .transpose()
+}
+
+fn parse_duration_default(env: &str, default: Duration) -> Result<Duration, Error> {
```
ilrudie, line 312: This seems infallible? Why return a result?

howardjohn: parse_duration returns Result<Option>. It may fail to parse, for example if the var is `x` that is not a valid. We default if its unset, not if it is invalid

src/config.rs
howardjohn: parse_duration returns Result<Option<Duration>>, we want a Result<Duration>. We default if its unset, not if it is invalid

ilrudie: ah, yup. I missed the result wrapping the option. ty

src/config.rs
ilrudie: ty

src/proxy.rs
```diff
@@ -71,12 +71,32 @@ pub trait SocketFactory {
...
+            let cfg = self.0;
+            if cfg.keepalive_enabled {
+                let ka = TcpKeepalive::new()
+                    .with_time(cfg.keepalive_time)
+                    .with_retries(cfg.keepalive_retries)
+                    .with_interval(cfg.keepalive_interval);
+                tracing::trace!(
+                    "set keepalive: {:?}",
+                    socket2::SockRef::from(&s).set_tcp_keepalive(&ka)
+                );
```
ilrudie, line 89: Why would v6 not need keep alive?

howardjohn: It does, that is a bug. Nice catch!

ilrudie:
For clarity, it looks like this will introduce keep alive functionality for ~all zt connections:

1. client talking with ztunnel
2. ztunnel talking with server
3. pooled HBONE

Is that accurate?

howardjohn:
That is correct although now I am thinking we maybe don't want the HBONE ones since we already have the HTTP2 level ping.. although I suppose it probably doesn't hurt?

ilrudie:
yeah, I think it should be way lower overhead than sending the keepalives for all the non-pooled connections. It's probably fine 🤔

howardjohn:
Yeah I think we can go with this for now and optimize things as we get feedback if needed

ilrudie (approved):
lgtm

PR #1378, opened 2024-12-05
debugging: add some extra debug logs to introspect HBONE requests

howardjohn:
Right now we cannot see what is coming in on HBONE. I have >5 times done
a small local build to add this, so seems well worth having for
debugging.


ilrudie (approved):
lgtm

PR #1396, opened 2024-12-17
Expose configuration parsing errors to user

howardjohn:
Right now we say it failed... but not why. Just a simple QOL improvement


ilrudie (approved):
lgtm

PR #1397, opened 2024-12-18
Bump dependencies

howardjohn:
* Pull in various new deps
* Fixup deny.toml to remove exclusions we don't need anymore
* Move derivitive -> educe since its "unmaintained" and flags scanners


src/tls/control.rs
```diff
@@ -53,8 +54,16 @@ async fn root_to_store(root_cert: &RootCert) -> Result<rustls::RootCertStore, Er
...
-                .map_err(|e| Error::InvalidRootCert(e.to_string()))?;
+            let certs = {
+                let rustls_native_certs::CertificateResult { certs, errors, .. } =
+                    rustls_native_certs::load_native_certs();
+                if !errors.is_empty() {
+                    return Err(Error::InvalidRootCert(
+                        errors.into_iter().map(|e| e.to_string()).join(","),
+                    ));
+                }
+                certs
```
bleggett, line 65: Nit - can `roots.add_parsable_certificates(certs);` be moved into the `let` scope and the outer binding dropped?

bleggett (approved):

PR #1403, opened 2024-12-19
Tidy `DelWorkload` handling in a pre-snapshot context

bleggett:
Removes a `TODO` here (and also fixes a potential theoretical snapshot state bug, I think)

### DIFF
--- a/src/inpod/statemanager.rs
+++ b/src/inpod/statemanager.rs
@@ -178,9 +178,9 @@ impl WorkloadProxyManagerState {
                     // make sure we also opportunistically remove here before we
                     // get a snapshot
                     //
-                    // Note that unlike AddWorkload we do *not* need to remove the workload here,
-                    // as it should be auto-dropped subsequently during snapshot reconcile(), when
-                    // we actually get the `SnapshotSent` notification.
+                    // Note that even though AddWorkload starts the workload, we do *not* need
+                    // to stop it here, as it should be auto-dropped subsequently during snapshot
+                    // reconcile(), when we actually get the `SnapshotSent` notification.
                     self.snapshot_names.remove(&workload_uid);
                     // `reconcile()` will drop this workload later, but if the workload never successfully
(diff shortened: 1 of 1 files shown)


src/inpod/statemanager.rs
```diff
@@ -172,12 +173,12 @@ impl WorkloadProxyManagerState {
...
-                    // self.pending_workloads.remove(&workload_uid)
-                    return Err(Error::ProtocolError(
-                        "pod delete received before snapshot".into(),
-                    ));
+                    debug!("got workload delete before snapshot");
+                    // Since we insert here on AddWorkload before we get a snapshot,
+                    // make sure we also opportunistically remove here before we
+                    // get a snapshot
+                    self.snapshot_names.remove(&workload_uid);
+                    return Ok(());
```
howardjohn, line 181: the todo comment suggested `self.pending_workloads.remove(&workload_uid)`.. and reason we don't need to do that?

bleggett: Ah good catch, we do, I was originally letting this fall through to the `del_workload()`.

howardjohn (commented):
lgtm just want to check my understanding

src/inpod/statemanager.rs
bleggett: Oh actually - now I remember. When we get the `SnapshotDone` we would auto-drop anything that had previously been started but wasn't in the snapshot, so it _should_ get handled there.

bleggett: That being said, explicitly removing at-time-of-message is probably _simpler_  and safer so I'll just do that.

bleggett: Settled on a clarifying comment.

bleggett: ...but we can't do it in `reconcile()`, since `workload_states` only has successfully started proxies.

Updated and added comments to hopefully make this more clear.

howardjohn (approved):

PR #1409, opened 2025-01-03
xds: support overlapping IP addresses

howardjohn:
Fixes https://github.com/istio/istio/issues/54528

The original design was to dedupe IPs in the control plane. however,
this approach was flawed. While it may be possible to handle this in
the control plane properly, it would be quite challenging. Instead, we
will allow multiple in Ztunnel itself. In the unlikely event of a
conflciting IP, we will store all matches and pick the best one for the
situation.

Note we only lookup by IP in a few cases:
* On inbound, to get peer metadata *for telemetry*
* lookup_is_destination_this_waypoint for some niche cases
* Sending directly to a Pod IP, instead of to a service

Note that without this PR, if you have 2 objects with the same IP, the
last one updated will 'win'. If you remove one, it may remove any info
about the IP entirely.

### DIFF
--- a/src/state/workload.rs
+++ b/src/state/workload.rs
@@ -24,6 +24,7 @@ use serde::Deserialize;
 use serde::Deserializer;
 use serde::Serialize;
 use serde::Serializer;
+use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
 use std::convert::Into;
 use std::default::Default;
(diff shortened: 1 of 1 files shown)


src/state/workload.rs
```diff
@@ -632,14 +633,83 @@ pub struct WorkloadStore {
...
+                // Remove it if the UID matches, else do nothing
+                wl.uid == uid
+            }
+            WorkloadByAddr::Many(ws) => {
+                ws.retain(|w| w.uid != uid);
+                match ws.as_slice() {
+                    [] => true,
+                    [wl] => {
+                        // We now have one workload, transition to Single
+                        *self = WorkloadByAddr::Single(wl.clone());
```
keithmattix, line 678: Is this even worth it? It's more correct, but wonder if a high churn scenario could make this expensive. Maybe it's faster to just let a `::Many` stay `::Many`?

src/state/workload.rs
```diff
@@ -632,14 +633,83 @@ pub struct WorkloadStore {
...
+                        false
+                    }
+                    // We still have many. We removed already so no need to do anything
+                    _ => false,
+                }
+            }
+        }
+    }
+    pub fn get(&self) -> Arc<Workload> {
```
keithmattix, line 687: Does it make sense to add another criteria here for the network of the workload? Thinking of a scenario where the control plane is sending WDS to two clusters that have overlapping IP addresses. This isn't supported now, so the correct answer may be "modify this if/when we add support for it", but just throwing it out there

src/state/workload.rs
```diff
@@ -710,7 +788,7 @@ impl WorkloadStore {
 
     /// Finds the workload by address, as an arc.
     pub fn find_address(&self, addr: &NetworkAddress) -> Option<Arc<Workload>> {
-        self.by_addr.get(addr).cloned()
+        dbg!(dbg!(self.by_addr.get(addr)).map(WorkloadByAddr::get))
```
keithmattix, line 791: For my understanding: what does this macro do, and why do we call it twice?

src/state/workload.rs
keithmattix: Oh but on the other hand, large cluster with high churn could mean frequently overlapping ips...

howardjohn: Either case is probably expensive, the vec is probably going to be resized if we insert, so making a new many vs inserting 1 is probably likely to be expensive either way.

I wouldn't have done the optimization at all if it wasn't for the fact that you can only get into overlapping IP case with an invalid configuration. So the `many` is the "slow" path essential

src/state/workload.rs
howardjohn: The key into the map is already NetworkAddress so its already handled

src/state/workload.rs
howardjohn: its for debugging - shouldn't merge! but it basically dumps the value to stderr with some line info, etc

src/state/workload.rs
keithmattix: Ack - makes sense

src/state/workload.rs
keithmattix: Ah right `NetworkAddress` has the `network/` prefix. Makes sense

keithmattix (approved):

src/state/workload.rs
```diff
@@ -632,14 +633,83 @@ pub struct WorkloadStore {
...
+/// WorkloadByAddr is a small wrapper around a single or multiple Workloads
+/// We split these as in the vast majority of cases there is only a single one, so we save vec allocation.
+enum WorkloadByAddr {
+    Single(Arc<Workload>),
+    Many(Vec<Arc<Workload>>),
+}
+
+impl WorkloadByAddr {
+    // insert adds the workload
+    pub fn insert(&mut self, w: Arc<Workload>) {
```
bleggett, line 654: ooo, this is a cute pattern.

src/state/workload.rs
bleggett, line 687: Since this is, rather importantly, not really "just" a `get`, can we describe the behavior in a function comment (and maybe rename it to `get_by_heuristic` or something to make it more obvious what's happening in callers.

src/state/workload.rs
```diff
@@ -632,14 +633,83 @@ pub struct WorkloadStore {
...
+    pub fn get(&self) -> Arc<Workload> {
+        match self {
+            WorkloadByAddr::Single(workload) => workload.clone(),
+            WorkloadByAddr::Many(workloads) => workloads
+                .iter()
+                .max_by_key(|w| {
+                    // Setup a ranking criteria in the event of a conflict.
+                    // We prefer pod objects, as they are not (generally) spoof-able and is the most
+                    // likely to truthfully correspond to what is behind the service.
+                    let is_pod = w.uid.contains("//Pod/");
```
bleggett, line 696: Assuming that if two things share the same rank (which _shouldn't_ happen but is allowed by this data structure) - we just grab the first, which is more or less no worse than current behavior.

bleggett (approved):

PR #1414, opened 2025-01-07
tests: improve error messages

howardjohn:
Before:
```
thread '<unnamed>' panicked at tests/namespaced.rs:1411:43:
called `Result::unwrap()` on an `Err` value: Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'namespaced::local_captured_inpod' panicked at tests/namespaced.rs:1336:14:
called `Result::unwrap()` on an `Err` value: Any { .. }
```

After:
```
thread 'namespaced::local_captured_inpod' panicked at tests/namespaced.rs:1168:85:
called `Result::unwrap()` on an `Err` value: tcp client

Caused by:
    0: write failed
    1: Connection reset by peer (os error 104)
```

---

Access log change now looks like:

```
thread 'namespaced::local_captured_inpod' panicked at tests/namespaced.rs:1231:13:
Analyzed 40 logs but none matched our criteria. Closest 10 matches:

direction:"outboundx" != "outbound"

bytes_recv:"22" != "11"
bytes_sent:"11" != "22"
direction:"outboundx" != "inbound"

bytes_recv:missing
bytes_sent:missing
direction:missing
dst.identity:missing
dst.workload:missing
message:"connection complete" != "starting server"
scope:"access" != "ztunnel::test_helpers::inpod"
src.identity:missing
src.workload:missing
```

Before it just dumped all the logs

### DIFF
--- a/src/telemetry.rs
+++ b/src/telemetry.rs
@@ -417,6 +417,7 @@ pub mod testing {
     use once_cell::sync::Lazy;
     use serde_json::Value;
     use std::collections::HashMap;
+    use std::fmt::{Display, Formatter};
     use std::io;
     use std::sync::{Mutex, MutexGuard, OnceLock};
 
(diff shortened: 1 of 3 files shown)


bleggett:
> Before it just dumped all the logs

Should we do both and summarize at the top, since the filtering is fuzzy?

bleggett:
> > Should we do both and summarize at the top, since the filtering is fuzzy?
> 
> The logs are already logged as normal logs anyways so it was a bit redundant I feel

Ah fair enough.

bleggett (approved):

PR #1418, opened 2025-01-08
deps: drop http-types for specialized code

howardjohn:
This is a fairly large dependency for a small task, and it has not had
updates for a long time. This PR drops the dependency in favor of our
own parser + creation logic which handles only the bits we need. Unit
tests and a fuzzer are added for this function. The fuzzer has run on my
machine for >1hr without any findings.


fuzz/fuzz_targets/baggage.rs
```diff
@@ -21,6 +21,7 @@ use ztunnel::proxy::BAGGAGE_HEADER;
 
 fuzz_target!(|data: &[u8]| {
```
bleggett, line 22: Good enough if the fuzz tests are happy.

bleggett (approved):

PR #1420, opened 2025-01-10
Add inbound support for svc hostname in authority

jaellio:
Main changes:
- hbone_address supports hostname or ip when reading from the authority header
- when a hostname is included in the authority pseudo header in is included as header in the proxy protocol. The hbone address included in the address header is set to the destination service vip
- hostname is validated by looking up the respective service and verifying the destination pod ip is a member of that service. 

Todo:
- [x] refactor hbone address validation to improve testability
- [x] update error types
- [x] confirm if port should also be included with hostname in custom protocol type
- [x] intelligently select service vip for proxy protocol address header
- [ ] strictly validate hostname fqdn

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -493,14 +493,11 @@ pub enum Error {
 
 // Custom TLV for proxy protocol for the identity of the source
 const PROXY_PROTOCOL_AUTHORITY_TLV: u8 = 0xD0;
-// Custom TLV for including the svc hostname in the proxy protocol header
-const PROXY_PROTOCOL_SVC_HOSTNAME_TLV: u8 = 0xD1;
 
 pub async fn write_proxy_protocol<T>(
     stream: &mut TcpStream,
     addresses: T,
     src_id: Option<Identity>,
-    hostname_addr: Option<Strng>,
 ) -> io::Result<()>
 where
     T: Into<ppp::v2::Addresses> + std::fmt::Debug,
@@ -518,11 +515,6 @@ where
 
     if let Some(id) = src_id {
         builder = builder.write_tlv(PROXY_PROTOCOL_AUTHORITY_TLV, id.to_string().as_bytes())?;
-    }
-
-    // svc_hostname is None when the hbone_addr does not contain a svc hostname.
-    if let Some(svc_host) = hostname_addr {
-        builder = builder.write_tlv(PROXY_PROTOCOL_SVC_HOSTNAME_TLV, svc_host.as_bytes())?;
     }
 
     let header = builder.build()?;
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -851,28 +851,24 @@ impl From<(Strng, u16)> for HboneAddress {
     }
 }
 
-/*
 impl TryFrom<&http::Uri> for HboneAddress {
-    type Error = InboundError;
+    type Error = Error;
 
     fn try_from(value: &http::Uri) -> Result<Self, Self::Error> {
         match value.to_string().parse::<SocketAddr>() {
             Ok(addr) => Ok(HboneAddress::SocketAddr(addr)),
             Err(_) => {
-                let hbone_host = value.host().ok_or_else(|| InboundError(
-                    Error::NoValidAuthority(value.to_string()),
-                    StatusCode::BAD_REQUEST
-                ))?;
-                let hbone_port = value.port_u16().ok_or_else(|| InboundError(
-                    Error::NoValidAuthority(value.to_string()),
-                    StatusCode::BAD_REQUEST
-                ))?;
+                let hbone_host = value
+                    .host()
+                    .ok_or_else(|| Error::NoValidAuthority(value.to_string()))?;
+                let hbone_port = value
+                    .port_u16()
+                    .ok_or_else(|| Error::NoValidAuthority(value.to_string()))?;
                 Ok(HboneAddress::SvcHostname(hbone_host.into(), hbone_port))
             }
         }
     }
 }
-*/
 
 #[cfg(test)]
 mod tests {
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -485,12 +485,16 @@ pub enum Error {
     DnsEmpty,
 }
 
+// Custom TLV for proxy protocol for the identity of the source
 const PROXY_PROTOCOL_AUTHORITY_TLV: u8 = 0xD0;
+//TODO(jaellio) - Create our own TLV 0
+const PROXY_PROTOCOL_SVC_HOSTNAME_TLV: u8 = 0xD1;
 
 pub async fn write_proxy_protocol<T>(
     stream: &mut TcpStream,
     addresses: T,
     src_id: Option<Identity>,
+    svc_hostname: Option<String>,
 ) -> io::Result<()>
 where
     T: Into<ppp::v2::Addresses> + std::fmt::Debug,
@@ -504,6 +508,13 @@ where
 
     if let Some(id) = src_id {
         builder = builder.write_tlv(PROXY_PROTOCOL_AUTHORITY_TLV, id.to_string().as_bytes())?;
+    }
+
+    if let Some(svc_hostname) = svc_hostname {
+        builder = builder.write_tlv(
+            PROXY_PROTOCOL_SVC_HOSTNAME_TLV,
+            svc_hostname.as_bytes(),
+        )?;
     }
 
     let header = builder.build()?;
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -487,7 +487,7 @@ pub enum Error {
 
 // Custom TLV for proxy protocol for the identity of the source
 const PROXY_PROTOCOL_AUTHORITY_TLV: u8 = 0xD0;
-//TODO(jaellio) - Create our own TLV 0
+// Custom TLV for including the svc hostname in the proxy protocol header
 const PROXY_PROTOCOL_SVC_HOSTNAME_TLV: u8 = 0xD1;
 
 pub async fn write_proxy_protocol<T>(
@@ -502,6 +502,10 @@ where
     use ppp::v2::{Builder, Command, Protocol, Version};
     use tokio::io::AsyncWriteExt;
 
+    // When the hbone_addr populated from the authority header contains a svc hostname, the address included
+    // with respect to the hbone_addr is the SocketAddr <dst svc IP>:<original dst port>.
+    // This is done since addresses doesn't support hostnames.
+    // See ref https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt
     debug!("writing proxy protocol addresses: {:?}", addresses);
     let mut builder =
         Builder::with_addresses(Version::Two | Command::Proxy, Protocol::Stream, addresses);
@@ -510,10 +514,11 @@ where
         builder = builder.write_tlv(PROXY_PROTOCOL_AUTHORITY_TLV, id.to_string().as_bytes())?;
     }
 
-    if let Some(svc_hostname) = svc_hostname {
+    // svc_hostname is None when the hbone_addr does not contain a svc hostname.
+    if let Some(svc_host) = svc_hostname {
         builder = builder.write_tlv(
             PROXY_PROTOCOL_SVC_HOSTNAME_TLV,
-            svc_hostname.as_bytes(),
+            svc_host.as_bytes(),
         )?;
     }
 
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -516,10 +516,7 @@ where
 
     // svc_hostname is None when the hbone_addr does not contain a svc hostname.
     if let Some(svc_host) = svc_hostname {
-        builder = builder.write_tlv(
-            PROXY_PROTOCOL_SVC_HOSTNAME_TLV,
-            svc_host.as_bytes(),
-        )?;
+        builder = builder.write_tlv(PROXY_PROTOCOL_SVC_HOSTNAME_TLV, svc_host.as_bytes())?;
     }
 
     let header = builder.build()?;
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -494,7 +494,7 @@ pub async fn write_proxy_protocol<T>(
     stream: &mut TcpStream,
     addresses: T,
     src_id: Option<Identity>,
-    svc_hostname: Option<String>,
+    svc_hostname: Option<Strng>,
 ) -> io::Result<()>
 where
     T: Into<ppp::v2::Addresses> + std::fmt::Debug,
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -452,6 +452,9 @@ pub enum Error {
     #[error("no valid authority pseudo header: {0}")]
     NoValidAuthority(String),
 
+    #[error("no valid service port in authority header: {0}")]
+    NoValidSerivePort(String, u16),
+
     #[error("no valid routing destination for workload: {0}")]
     NoValidDestination(Box<Workload>),
 
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -500,7 +500,7 @@ pub async fn write_proxy_protocol<T>(
     stream: &mut TcpStream,
     addresses: T,
     src_id: Option<Identity>,
-    svc_hostname: Option<Strng>,
+    hostname_addr: Option<Strng>,
 ) -> io::Result<()>
 where
     T: Into<ppp::v2::Addresses> + std::fmt::Debug,
@@ -521,7 +521,7 @@ where
     }
 
     // svc_hostname is None when the hbone_addr does not contain a svc hostname.
-    if let Some(svc_host) = svc_hostname {
+    if let Some(svc_host) = hostname_addr {
         builder = builder.write_tlv(PROXY_PROTOCOL_SVC_HOSTNAME_TLV, svc_host.as_bytes())?;
     }
 
@@ -826,6 +826,13 @@ impl HboneAddress {
         match self {
             HboneAddress::SocketAddr(_) => None,
             HboneAddress::SvcHostname(s, _) => Some(s.into()),
+        }
+    }
+
+    pub fn hostname_addr(&self) -> Option<Strng> {
+        match self {
+            HboneAddress::SocketAddr(_) => None,
+            HboneAddress::SvcHostname(_, _) => Some(Strng::from(self.to_string())),
         }
     }
 }
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -449,11 +449,17 @@ pub enum Error {
     #[error("no service or workload for hostname: {0}")]
     NoHostname(String),
 
+    #[error("no endpoints for workload: {0}")]
+    NoWorkloadEndpoints(String),
+
     #[error("no valid authority pseudo header: {0}")]
     NoValidAuthority(String),
 
     #[error("no valid service port in authority header: {0}")]
     NoValidSerivePort(String, u16),
+
+    #[error("no valid target port for workload: {0}")]
+    NoValidTargetPort(String, u16),
 
     #[error("no valid routing destination for workload: {0}")]
     NoValidDestination(Box<Workload>),
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -446,6 +446,12 @@ pub enum Error {
     #[error("unknown waypoint: {0}")]
     UnknownWaypoint(String),
 
+    #[error("no service or workload for hostname: {0}")]
+    NoHostname(String),
+
+    #[error("no valid authority pseudo header: {0}")]
+    NoValidAuthority(String),
+
     #[error("no valid routing destination for workload: {0}")]
     NoValidDestination(Box<Workload>),
 
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -801,6 +801,79 @@ pub fn parse_forwarded_host(input: &str) -> Option<String> {
         .filter(|host| !host.is_empty())
 }
 
+#[derive(Debug, Clone)]
+pub enum HboneAddress {
+    SocketAddr(SocketAddr),
+    SvcHostname(Strng, u16),
+}
+
+impl HboneAddress {
+    pub fn port(&self) -> u16 {
+        match self {
+            HboneAddress::SocketAddr(s) => s.port(),
+            HboneAddress::SvcHostname(_, p) => *p,
+        }
+    }
+
+    pub fn ip(&self) -> Option<IpAddr> {
+        match self {
+            HboneAddress::SocketAddr(s) => Some(s.ip()),
+            HboneAddress::SvcHostname(_, _) => None,
+        }
+    }
+
+    pub fn svc_hostname(&self) -> Option<Strng> {
+        match self {
+            HboneAddress::SocketAddr(_) => None,
+            HboneAddress::SvcHostname(s, _) => Some(s.into()),
+        }
+    }
+}
+
+impl std::fmt::Display for HboneAddress {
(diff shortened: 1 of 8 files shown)


src/proxy/inbound.rs
```diff
@@ -375,22 +478,82 @@ impl Inbound {
...
         }
 
-        if conn.dst.ip() == hbone_addr.ip() {
-            // Normal case: both are aligned. This is allowed (we really only need the HBONE address for the port.)
-            return Ok(());
+        match hbone_addr {
+            // TODO(jaellio): Should we also validate the hbone_addr port?
+            HboneAddress::SvcHostname(hbone_host, hbone_port) => {
+                // Validate a service or workload exists for the hostname
+                let parts: Vec<&str> = hbone_host.split('.').collect();
```
jaellio, line 492: TODO - figure out if there is a better way to parse a hostname

src/proxy/inbound.rs
```diff
@@ -359,13 +465,15 @@ impl Inbound {
             },
             pi.metrics.clone(),
         ));
+        //let hbone_addr_clone = hbone_addr.clone();
```
Stevenjin8, line 468:
```suggestion
```

src/proxy/inbound.rs
```diff
@@ -228,16 +284,44 @@ impl Inbound {
         // At this point, we established the upstream connection and need to send a 200 back to the client.
         // we may still have failures at this point during the proxying, but we don't need to send these
         // at the HTTP layer.
+        // After sending a 200 back to the client, write the proxy protocol to the stream. This ensures
+        // that the server has all of the necessary information about the connection regardless of the protocol
+        // See https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt for more information about the
+        // proxy protocol.
+        //
+        // Why is the x-forwarded-for header not sufficient? Why do we need to write the proxy protocol?
```
keithmattix, line 292:
```suggestion

```

src/proxy/inbound.rs
```diff
@@ -262,38 +348,58 @@ impl Inbound {
...
+        })?;
+
+        let hbone_port = req.uri().port_u16().ok_or_else(|| {
+            InboundError(
+                Error::ConnectAddress(req.uri().to_string()),
+                StatusCode::BAD_REQUEST,
+            )
+        })?;
+
+        let hbone_authority = format!("{}:{}", hbone_host, hbone_port);
```
keithmattix, line 367: Might be able to be slightly more ergonomic here:
```rust
let hbone_addr = req
            .uri()
            .to_string()
            .as_str()
            .parse::<SocketAddr>()
            .map_err(|_| {
               let hbone_host = req.uri().host()
               let hbone_port = req.uri().port_u16()
               ....
            })?;
```

src/proxy/inbound.rs
```diff
@@ -262,38 +348,58 @@ impl Inbound {
...
             .get_workload()
             .await
             // At this point we already fetched the local workload for TLS, so it should be infallible.
             .map_err(InboundError::build(StatusCode::SERVICE_UNAVAILABLE))?;
 
-        // Check the request is allowed
-        Self::validate_destination(&pi.cfg, &pi.state, &conn, &destination_workload, hbone_addr)
-            .await
-            .map_err(InboundError::build(StatusCode::BAD_REQUEST))?;
+        // Check the request is allowed by verifying the connection IP and the hbone address IP are the same
```
keithmattix, line 382: What is the "connection IP"?

src/proxy/inbound.rs
```diff
@@ -262,38 +348,58 @@ impl Inbound {
...
+        let upstream_protocol_addr = Self::validate_destination(
```
keithmattix, line 383: Put another way: is there any logic that checks whether the destination pod is a part of the service being pointed to in the hostname?

src/proxy/inbound.rs
```diff
@@ -375,22 +483,82 @@ impl Inbound {
         state: &DemandProxyState,
         conn: &Connection,
         local_workload: &Workload,
-        hbone_addr: SocketAddr,
-    ) -> Result<(), Error> {
+        hbone_addr: HboneAddress,
+    ) -> Result<Option<SocketAddr>, Error> {
```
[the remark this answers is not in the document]
jaellio: Agree, just a temporary hack. Will fix

src/proxy/inbound.rs
```diff
@@ -375,22 +483,82 @@ impl Inbound {
...
 
-        if conn.dst.ip() == hbone_addr.ip() {
-            // Normal case: both are aligned. This is allowed (we really only need the HBONE address for the port.)
-            return Ok(());
+        match hbone_addr {
+            // TODO(jaellio): Should we also validate the hbone_addr port?
+            HboneAddress::SvcHostname(hbone_host, hbone_port) => {
+                // Validate a service or workload exists for the hostname
+                let parts: Vec<&str> = hbone_host.split('.').collect();
+                if parts.len() <= 1 {
```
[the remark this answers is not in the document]
jaellio: Agree. Kept it simple for the draft but this isn't sufficient atm.

src/proxy/inbound.rs
jaellio: This comment isn't correct. By connection IP I was referring to the `conn.dst.IP` and the conditional check [here](https://github.com/istio/ztunnel/pull/1420/files/72c87e0a118f47a30068d348d86096baef8a9d61#diff-0acf95d58db195dc1623fb8d2eface8e582459fdcf4a561c5fce509e46270facR550).

src/proxy/inbound.rs
```diff
@@ -262,38 +346,90 @@ impl Inbound {
...
+                match addr {
+                    Address::Service(svc) => {
+                        if !svc.contains_endpoint(&destination_workload) {
+                            // TODO(jaellio): Update error msg/type
+                            return Err(InboundError(
+                                Error::NoResolvedAddresses(destination_workload.to_string()),
+                                StatusCode::BAD_REQUEST,
+                            ));
+                        }
+                        // TODO(jaellio): Intelligently select the VIP
```
jaellio, line 404: @keithmattix with regards to https://github.com/istio/ztunnel/pull/1409 and intelligently selecting vip for a Address::Service, is it sufficient to rank vips by network comparison (higher rank if `vip.network` matches `conn.dst_network`) or should we also be comparing IpAddr ranges? Do we also want to implement ranking of Address::Services when finding hostnames (similar to how there would be duplicate workloads on lookup by IP)?

src/proxy/inbound.rs
```diff
@@ -262,38 +348,58 @@ impl Inbound {
...
+            &pi.state,
+            &conn,
+            &destination_workload,
+            hbone_addr.clone(),
+        )
+        .await
+        .map_err(InboundError::build(StatusCode::BAD_REQUEST))?;
 
         // Determine the next hop.
+        let hbone_addr_clone = hbone_addr.clone();
```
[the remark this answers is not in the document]
ilrudie: The part we care about in hbond_addr is a Strng, which is ArcStr being the covers. I don't recall the logic for why we did that exactly. Anyhow. Probably best to just pass a borrow of hbone_addr into find_inbound_upstream and then take a cheap clone internally when we need to.

howardjohn (commented):
A

src/proxy/inbound.rs
keithmattix: Sorry, missed this. Are you still blocked here?

src/proxy.rs
```diff
@@ -485,19 +485,27 @@ pub enum Error {
     DnsEmpty,
 }
 
+// Custom TLV for proxy protocol for the identity of the source
 const PROXY_PROTOCOL_AUTHORITY_TLV: u8 = 0xD0;
+// Custom TLV for including the svc hostname in the proxy protocol header
+const PROXY_PROTOCOL_SVC_HOSTNAME_TLV: u8 = 0xD1;
```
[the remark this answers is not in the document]
keithmattix: I mean, I imagine that if there's custom waypoint sandwiched with the ztunnel, it would want to use the hostname for routing (vs. an IP address we lookup) , right?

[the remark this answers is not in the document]
jaellio: @keithmattix correct me if I am wrong here. We were concerned about not preserving the hostname in the proxy protocol. Not positive about the exact implications of this for logging, but maybe it has more implications for security tools?

src/proxy/inbound.rs
jaellio: Based on [this comment](https://github.com/istio/ztunnel/pull/1420#discussion_r1917454377) we can't rely on the hostname to include the namespace. So right now we'll need to change some of the logic for finding a service by hostname and implement a ranking as suggested [here](https://github.com/istio/ztunnel/pull/1420#discussion_r1917455626)

jaellio: Not blocked, I'll implement the service ranking and vip ranking as suggested above. What just curious if you had any thoughts on the approach.

src/proxy/inbound.rs
```diff
@@ -262,38 +346,90 @@ impl Inbound {
         }
 
         let start = Instant::now();
+
+        // Extract the host or IP from the authority pseudo-header of the URI
         let hbone_addr = req
```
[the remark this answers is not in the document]
jaellio: Took a shot at it, let me know if it's what you had in mind

src/proxy/inbound.rs
```diff
@@ -262,38 +346,90 @@ impl Inbound {
...
-            .map_err(|_| {
-                InboundError(
-                    Error::ConnectAddress(req.uri().to_string()),
-                    StatusCode::BAD_REQUEST,
-                )
-            })?;
+            .map(HboneAddress::SocketAddr)
+            .unwrap_or_else(|_| {
+                let hbone_host = req.uri().host().unwrap();
+                let hbone_port = req.uri().port_u16().unwrap();
```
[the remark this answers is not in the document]
jaellio: Moved this logic to the TryFrom implementation for HboneAddress. Returns an error result if port or host are not set.

src/proxy/inbound.rs
jaellio: @keithmattix Implemented simple ranking for service (when searching without a namespace) and vip. There might be a simpler way to accomplish the ranking but took inspiration from https://github.com/istio/ztunnel/pull/1409

src/proxy/inbound.rs
jaellio: Took a shot at this with ranking - might be a simpler way to do it in rust but I am not at all familiar :) Let me know if it's similar to what you had in mind.

src/proxy.rs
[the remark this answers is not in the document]
keithmattix: Yeah logging is one thing, but I'm really more thinking about fidelity. The sandwiched proxy should use the data on the wire as much as possible rather than an IP that ztunnel provides based on state that it may have on hand

src/proxy/inbound.rs
```diff
@@ -227,16 +235,42 @@ impl Inbound {
...
+                                Some(addr) => addr,
+                                None => {
+                                    // This should never happen, as we should have already validated the destination
+                                    // and resolved the hostname to an IP address.
+                                    return Err(Error::ConnectAddress(ri.hbone_addr.to_string()));
+                                }
+                            }
+                        }
+                    };
+                    // TODO(jaellio): Also include port?
```
keithmattix, line 264: Yeah this should also include port

src/proxy/inbound.rs
```diff
@@ -261,38 +297,82 @@ impl Inbound {
...
+            if !svc.contains_endpoint(&destination_workload) {
+                // TODO(jaellio): Update error msg/type
+                return Err(InboundError(
+                    Error::NoHostname(destination_workload.to_string()),
+                    StatusCode::BAD_REQUEST,
+                ));
+            }
+            // Select the VIP by network match
+            let svc_address = svc
+                .vips
```
keithmattix, line 351: How does this impact dualstack services that have both an IPv4 and an IPv6 VIP? Do we want to defer to v4 in some way? This looks right otherwise

src/proxy/inbound.rs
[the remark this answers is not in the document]
keithmattix: +1 to having it take a borrow or copying the port if that's really all we need

src/proxy/inbound.rs
```diff
@@ -365,26 +445,65 @@ impl Inbound {
...
+        hbone_host: Strng,
+    ) -> Result<Service, Error> {
+        // Validate a service exists for the hostname
+        // TODO(jaellio): Currently returns an error is no service is found. Workload lookup is not supported.
+        let services = state.read().find_service_by_hostname(&hbone_host)?;
+
+        services
+            .iter()
+            .max_by_key(|s| {
+                let is_local_namespace = s.namespace == local_workload.namespace;
```
keithmattix, line 466: 👍🏾

src/proxy/inbound.rs
jaellio: Updated to include port. Fmt is <hostname>:<port>. Should this value instead be taken directly from the authority value  of the uri rather than manually formatted?

src/proxy/inbound.rs
jaellio: Good callout. Not sure. I'll look into this

src/proxy/inbound.rs
```diff
@@ -250,8 +283,10 @@ impl Inbound {
         ri.result_tracker.record(res);
     }
 
+    // build_inbound_request builds up the context for an inbound request.
     async fn build_inbound_request(
```
jaellio, line 287: @howardjohn Any suggestions for testing `build_inbound_request` or it best to break out the parts we want to test into separate methods? Is testing for methods like these typically captured in the istio tests?

howardjohn: The hard part of testing this is likely making a `H2Request`? Since that would pretty much need to come from a real connection which makes it more e2e test than a small unit test.

We only actually need properties of the request (uri/method/headers) though - in theory we could make the function take in `Parts` (not currently exposed) only which would make it pretty simple to have just simple unit tests. Might be worth doing so...

howardjohn (commented):
Having some form of E2E test would probably be good here. https://github.com/istio/ztunnel/blob/5e6234f3ceb0d5583139c2d5d8f8688a721af6bb/tests/namespaced.rs#L830 shows an example test sending a raw hbone request, we could likely do one for a hostname request (the test is also setting the authroity header, just in that test to a bogus IP)

src/proxy/inbound.rs
jaellio: Updated build_inbound_request to accept  `Parts` instead of `H2Request`. To simplify testing even further, I could define a trait and implement it for `Parts` and a mock `Parts` since I can't create a Parts instance directly (I'm currently creating a `Request` and accessing the `Parts` in the test. Both options are relatively simple/minimal code

jaellio: Never mind, added a `RequestParts` trait

src/proxy/inbound.rs
jaellio: Updated logic prefer ipv4

keithmattix (approved):
This looks right to me; leaving it to @istio/wg-networking-maintainers-ztunnel for final approval

src/proxy.rs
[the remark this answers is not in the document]
howardjohn: Its kind of awkward though since you are _only_ going to get this info for cross-network. So I am not sure what a usage can really do meaningfully..

src/proxy.rs
```diff
@@ -485,19 +491,27 @@ pub enum Error {
...
+    hostname_addr: Option<Strng>,
 ) -> io::Result<()>
 where
     T: Into<ppp::v2::Addresses> + std::fmt::Debug,
 {
     use ppp::v2::{Builder, Command, Protocol, Version};
     use tokio::io::AsyncWriteExt;
 
+    // When the hbone_addr populated from the authority header contains a svc hostname, the address included
+    // with respect to the hbone_addr is the SocketAddr <dst svc IP>:<original dst port>.
```
howardjohn, line 512: The addresses is the source address, not the destination address

tests/namespaced.rs
```diff
@@ -885,6 +885,76 @@ mod namespaced {
...
+
+        run_tcp_server(server)?;
+        let srv = resolve_target(manager.resolver(), "server");
+        client
+            .run(move || async move {
+                let builder =
+                    hyper::client::conn::http2::Builder::new(ztunnel::hyper_util::TokioExecutor);
+
+                let request = hyper::Request::builder()
+                    .uri(format!("{SERVER_HOSTNAME}:{SERVER_PORT}"))
```
howardjohn, line 924: This isn't quite right. The service port is `80`, but we are using `8080` here (SERVER_PORT is 8080).

We will need to perform the service port -> target port translation in inbound.rs I think

src/proxy.rs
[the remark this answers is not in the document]
jaellio: @keithmattix thoughts on including this in the future if needed but remove it from this PR?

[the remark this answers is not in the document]
keithmattix: Yeah works for me

src/proxy.rs
jaellio: Could you clarify what you mean? Previously, `addresses` was always the authority header value. My understanding was that this was the <dest ip>:<org dest port>

howardjohn: Ah my bad - I didn't realize its actually `(src IP, destination authority header)`. I thought it was just the source

tests/namespaced.rs
jaellio: Okay, want to make sure I am on the same page. For the test, should the connection be to the dest service IP or the dest pod IP?

> We will need to perform the service port -> target port translation in inbound.rs I think

Is this specific to when we're using a svc hostname in the authority or is the translation required when building an inbound request to a service? When the authority header is configured on the client side, would the port be the service port or the target port?

jaellio: Updated - I think the connection is using the correct IP and port, but please double check

src/proxy.rs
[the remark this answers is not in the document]
jaellio: Removed

src/proxy/inbound.rs
```diff
@@ -614,22 +786,127 @@ mod tests {
...
             .unwrap();
-        let hbone_addr = format!("{hbone_dst}:{TARGET_PORT}").parse().unwrap();
-        let res = Inbound::validate_destination(&cfg, &state, &conn, &local_wl, hbone_addr)
-            .await
-            .map(|_| Inbound::find_inbound_upstream(&state, &conn, &local_wl, hbone_addr));
+        let hbone_addr =
+            if let Ok(addr) = format!("{hbone_dst}:{TARGET_PORT}").parse::<SocketAddr>() {
+                HboneAddress::SocketAddr(addr)
+            } else {
+                HboneAddress::SvcHostname(hbone_dst.into(), 80)
```
keithmattix, line 793:
```suggestion
                HboneAddress::SvcHostname(hbone_dst.into(), SERVER_PORT)
```
?

jaellio: Updated

src/proxy/inbound.rs
```diff
@@ -365,26 +460,65 @@ impl Inbound {
...
+        services
+            .iter()
+            .max_by_key(|s| {
+                let is_local_namespace = s.namespace == local_workload.namespace;
+                match is_local_namespace {
+                    true => 1,
+                    false => 0,
+                }
+            })
+            .cloned()
```
howardjohn, line 487: We should not clone. https://github.com/istio/ztunnel/pull/1475 fixes this for the last existing codepath but your PR uses the existing one I am fixing there. Should be a simple improvement once that PR merges to rebase on it

src/proxy/inbound.rs
```diff
@@ -365,26 +460,65 @@ impl Inbound {
...
+            .ok_or_else(|| Error::NoResolvedAddresses(hbone_host.to_string()))
```
howardjohn, line 488: `no ip addresses were resolved for workload: {0}` is the error message, but this is a service name which seems a bit awkward. May be worth a new enum for this case?

src/proxy/inbound.rs
howardjohn: We don't need to block on that BTW, I can rebase 1475 on this if this merges first

tests/namespaced.rs
```diff
@@ -885,6 +885,76 @@ mod namespaced {
...
+    async fn test_svc_hostname() -> anyhow::Result<()> {
+        let mut manager = setup_netns_test!(Shared);
+        let zt = manager.deploy_ztunnel(DEFAULT_NODE).await?;
+        manager
+            .service_builder("server")
+            .addresses(vec![NetworkAddress {
+                network: strng::EMPTY,
+                address: TEST_VIP.parse::<IpAddr>()?,
+            }])
+            .ports(HashMap::from([(80u16, 8080u16)]))
```
howardjohn, line 898: A service can have two things here: a numeric target port and a named target port (mapping from Kubernetes API). In the named target port case, this would be `(80, 0)`. The intent is the worklaod then has a port map like `80 -> 8080` and we can use that. I think we still need to handle that case here.

Which I guess implies two test cases may be useful here, though I don't mind making one of them a unit test in inbound.rs if preferred (either is fine with me)

jaellio: @howardjohn I don't think we currently have access to the workload port mapping. Prior to this change we'd only care about the target port and could get this directly from the authority header. I think beyond a test case this might require some WDS changes. Could support for named target ports be a TODO or with a quick follow up or is this a blocker?

src/proxy/inbound.rs
jaellio: Updated

tests/namespaced.rs
jaellio: Update - we do have access through the service endpoints. We can lookup endpoint by workload uid

src/proxy/inbound.rs
```diff
@@ -363,7 +363,33 @@ impl Inbound {
...
+                                .ok_or_else(|| {
+                                    InboundError(
+                                        Error::NoWorkloadEndpoints(
+                                            destination_workload.to_string(),
+                                        ),
+                                        StatusCode::BAD_REQUEST,
+                                    )
+                                })?;
+                        // Get the target port from the endpoint
+                        dest_port = *endpoint.port.get(&hbone_addr.port()).ok_or_else(|| {
```
jaellio, line 381: lmk if there is a better way to do this in Rust. Looking into it - just looks stinky

tests/namespaced.rs
jaellio: @howardjohn Added an additional e2e and the logic to support named target ports

tests/namespaced.rs
```diff
@@ -886,7 +886,16 @@ mod namespaced {
     }
 
     #[tokio::test]
-    async fn test_svc_hostname() -> anyhow::Result<()> {
+    async fn test_svc_hostname_port() -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
jaellio: Resolved by modifying the `setup_netns_test` macro.

src/proxy/inbound.rs
```diff
@@ -227,14 +235,33 @@ impl Inbound {
         // At this point, we established the upstream connection and need to send a 200 back to the client.
         // we may still have failures at this point during the proxying, but we don't need to send these
         // at the HTTP layer.
+        // After sending a 200 back to the client, write the proxy protocol to the stream. This ensures
```
howardjohn, line 238: nit: this implies we always send PROXY but we only do in a few (uncommon) cases

howardjohn (approved):
:+1: nice work!

Issue #1423, opened 2025-01-13
Implement double-hbone client in ztunnel


keithmattix:

keithmattix:
Blocked on control plane doc approval

PR #1429, opened 2025-01-17
WIP: Feature/double hbone

Stevenjin8:
# Initial double HBONE implementation

Right now, inner HBONE will only hold one connect tunnel. Once the inner tunnel terminates, so will the outer tunnel (but not the outer HBONE). So when ztunnel receives its first connection to a double HBONE host (E/W gateway), it will perform two TLS handshakes. Subsequent connections to the same host will perform one TLS handshake.

This behavior is not great, but if we put the inner HBONE in the connection pool, then we pin ourselves to a pod in the remote cluster since ztunnel performs connection pooling, but is not aware of the E/W gateway's routing decision.

That being said, I think this is a good place to stop and think about control plane implementation and get some feedback on how I'm approaching this.

Tasks:
- [x] Implement double HBONE 
- [x] Fix TLS code (I don't think I can do this without more control plane info since SANs are set for services)
- [x] metrics
- [ ] Implement proper inner HBONE connection pooling
- [x] Tests

Some open questions:
- Do I need to make any changes to metrics?
- How do should we do inner HBONE pooling? My suggestion is to have up to `N` inner HBONE connections per E/W or per remote cluster.
- Good ways to test for race conditions in connection terminations? Right now, it seems that connections end gracefully without race conditions, but that's just on my machine.

References:

- [Multi-network Ambient Protocol](https://docs.google.com/document/d/13c_r1doDCoRoEHuxk_eV8rnPi7X2pQgUWay9KMSSUrs/edit?tab=t.0)
- [Multi-network Ambient: Control Plane and ServiceDiscovery](https://docs.google.com/document/d/1Wg6sx9ZUJL4AsHj5wM1kMx3E436s5wg2qoMqoI-bqbQ/edit?tab=t.0)


src/proxy/pool.rs
```diff
@@ -83,6 +83,26 @@ struct ConnSpawner {
 
 // Does nothing but spawn new conns when asked
 impl ConnSpawner {
+    async fn new_unpooled_conn(
```
Stevenjin8, line 86: Anything here we can do higher up I think, but things might change if we decide to implement pooling in this PR

src/proxy/outbound.rs
```diff
@@ -107,7 +112,7 @@ impl Outbound {
                                 debug!(component="outbound", dur=?start.elapsed(), "connection completed");
                             }).instrument(span);
 
-                            assertions::size_between_ref(1000, 1750, &serve_outbound_connection);
```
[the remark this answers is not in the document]
bleggett: by looking a the current size and adding a small amount of buffer.

src/proxy/pool.rs
bleggett: Yeah if we want double-hbone conns to be unpooled and thus need ~none of this surrounding machinery, then I'd be inclined to just start `proxy/double-hbone.rs` and use that directly.

src/proxy/outbound.rs
```diff
@@ -202,10 +204,58 @@ impl OutboundConnection {
...
+        // This always drops ungracefully
+        // drop(conn_client);
+        // tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+        // drain_tx.send(true).unwrap();
+        // tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+        drain_tx.send(true).unwrap();
+        let _ = driver_task.await;
+        // this sleep is important, so we have a race condition somewhere
+        // tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+        res
```
Stevenjin8, line 256: Does anybody have any info on how to properly drop/terminate H2 connections over stream with nontrivial drops (e.g. shutting down TLS over HTTP2 CONNECT). Right now, I'm just dropping things/aborting tasks randomly until something works

src/proxy/outbound.rs
```diff
@@ -217,12 +267,12 @@ impl OutboundConnection {
         copy::copy_bidirectional(copy::TcpStreamSplitter(stream), upgraded, connection_stats).await
     }
 
-    async fn send_hbone_request(
+    fn create_hbone_request(
```
Stevenjin8, line 270: Git merge is getting confused here

src/proxy/outbound.rs
keithmattix: Are you asking about how to cleanup after, for example, a RST_STREAM to the inner tunnel? Or something else

Stevenjin8: Kinda. I mostly mean the outer TLS stream because that's what I've looked at. It seems like if I drop `conn_client` before termination `driver_task` the TCP connection will close **without** sending close notifies. So yes, I'm asking if there is a way to explicitly do cleanup rather than relying on implicit drops.

src/config.rs
```diff
@@ -70,7 +70,7 @@ const IPV6_ENABLED: &str = "IPV6_ENABLED";
 
 const UNSTABLE_ENABLE_SOCKS5: &str = "UNSTABLE_ENABLE_SOCKS5";
 
-const DEFAULT_WORKER_THREADS: u16 = 2;
+const DEFAULT_WORKER_THREADS: u16 = 40;
```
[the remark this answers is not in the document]
Stevenjin8: I was hoping it would making debugging async rust easier (it didn't)

src/proxy/outbound.rs
```diff
@@ -219,40 +219,56 @@ impl OutboundConnection {
         req: &Request,
         connection_stats: &ConnectionResult,
     ) -> Result<(), Error> {
-        // Outer HBONE
-        let upgraded = Box::pin(self.send_hbone_request(remote_addr, req)).await?;
-
-        // Inner HBONE
-        let upgraded = TokioH2Stream::new(upgraded);
+        // TODO: dst should take a hostname? and upstream_sans currently contains E/W Gateway certs
         let inner_workload = pool::WorkloadKey {
```
Stevenjin8, line 223: Will reorganize later.

src/proxy/outbound.rs
```diff
@@ -397,23 +471,29 @@ impl OutboundConnection {
...
+                SocketAddr::from((us.selected_workload_ip, self.hbone_port))
+            }
             Protocol::TCP => us.workload_socket_addr(),
         };
         let hbone_target_destination = match us.workload.protocol {
-            Protocol::HBONE => Some(us.workload_socket_addr()),
+            Protocol::HBONE | Protocol::DOUBLEHBONE => Some(us.workload_socket_addr()),
             Protocol::TCP => None,
         };
+        let (upstream_sans, final_sans) = match us.workload.protocol {
```
Stevenjin8, line 491: My understanding from talking to @keithmattix is that `Workload.subjectAlternativeNames` will be repurposed to contain the identities of remote pods/waypoints, so I should change the logic of the other protocols to only use `us.workload.identity` instead of `us.workload_and_services_san`.

src/proxy/pool.rs
```diff
@@ -511,10 +578,10 @@ impl WorkloadHBONEPool {
 #[derive(Debug, Clone)]
 // A sort of faux-client, that represents a single checked-out 'request sender' which might
 // send requests over some underlying stream using some underlying http/2 client
-struct ConnClient {
-    sender: H2ConnectClient,
+pub struct ConnClient {
```
Stevenjin8, line 581: fixme

src/proxy/outbound.rs
keithmattix: I see the code changed; do you still need help figuring this out?

src/proxy/outbound.rs
keithmattix: Yes, I think this is correct; only the double hbone codepath needs to be added/changed because there are two sans being considered: the e/w gateway SAN and the SANs of the backends. So what you have looks right to me

src/proxy/outbound.rs
Stevenjin8: Im still not confident in it. It works (on my machine), but I couldn't find any docs on proper connection termination/dropping.

src/config.rs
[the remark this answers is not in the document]
bleggett: (if you haven't already found it, `tokio-console` can be helpful)

keithmattix:
Not stale

src/proxy/h2.rs
```diff
@@ -138,6 +140,53 @@ impl Drop for DropCounter {
...
+}
+
+impl tokio::io::AsyncRead for TokioH2Stream {
+    fn poll_read(
+        mut self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+        buf: &mut tokio::io::ReadBuf<'_>,
+    ) -> Poll<std::io::Result<()>> {
+        let pinned = std::pin::Pin::new(&mut self.0.read);
+        copy::ResizeBufRead::poll_bytes(pinned, cx).map_ok(|bytes| buf.put(bytes))
```
howardjohn, line 158: This seems like there is no protection against ReadBuf's rules: `Panics if self.remaining() is less than buf.len().`

howardjohn: Can we plausible write a test around this like we have in copy.rs wrapping in `WeirdIO`? It detects these types of bugs

src/proxy/outbound.rs
```diff
@@ -195,19 +198,89 @@ impl OutboundConnection {
...
+        let res = copy::copy_bidirectional(
+            copy::TcpStreamSplitter(stream),
+            inner_upgraded,
+            connection_stats,
+        )
+        .await;
+
+        let _ = drain_tx.send(true);
+        let _ = driver_task.await;
+        // Here, there is an implicit, drop(conn_client).
```
howardjohn, line 277: there is no `conn_client` in scope so this is a bit confusing

src/proxy/outbound.rs
```diff
@@ -195,19 +198,89 @@ impl OutboundConnection {
...
 
-        let res = match req.protocol {
-            Protocol::HBONE => {
+        let res = match (
+            req.protocol,
+            req.actual_destination_workload
+                .as_ref()
+                .and_then(|wl| wl.network_gateway.as_ref()),
+        ) {
+            (_, Some(_)) => {
```
howardjohn, line 207: I don't think there is a guarantee that something that has a network_gateway supports HBONE. Or that something that has a network_gateway is even in a different network

examples/inpodserver.rs
```diff
@@ -50,6 +52,7 @@ async fn main() {
         .await
         .unwrap();
     sender.wait_forever().await.unwrap();
+    tokio::time::sleep(Duration::from_secs(99999999)).await;
```
Stevenjin8, line 55: remove me

src/proxy/socks5.rs
```diff
@@ -107,7 +107,7 @@ impl Socks5 {
                                 debug!(component="socks5", dur=?start.elapsed(), "connection completed");
                             }).instrument(span);
 
-                            assertions::size_between_ref(1000, 2000, &serve);
+                            assertions::size_between_ref(1000, 10000, &serve);
```
Stevenjin8, line 110: was erroring at at 5600

src/proxy/outbound.rs
```diff
@@ -195,19 +198,89 @@ impl OutboundConnection {
...
+        let cert = self
+            .pi
+            .local_workload_information
+            .fetch_certificate()
+            .await?;
+        let connector = cert.outbound_connector(wl_key.dst_id.clone())?;
+        let tls_stream = connector
+            .connect(
+                upgraded,
+                rustls::pki_types::ServerName::IpAddress(wl_key.dst.ip().into()),
```
howardjohn, line 255: The servername we pass has no meaning in the underlying library so I think its confusing to do the effort to pick a specific IP here vs always hardcoding a dummy IP and hiding it in the `OutboundConnector` like we do today

src/state.rs
```diff
@@ -70,9 +70,30 @@ pub struct Upstream {
...
     pub destination_service: Option<ServiceDescription>,
+    /// Use this port instead of the default one for the proxy protocol.
+    pub proxy_protocol_port_override: Option<u16>,
 }
 
 impl Upstream {
+    /// If there is a network gateway, use <service hostname>:<port>. Otherwise, use
+    /// <ip address>:<port>. Fortunately, for double-hbone, the authority/host is the same
+    /// for inner and outer CONNECT request, so we can reuse this for inner double hbone,
+    /// outer double hbone, and normal hbone.
```
Stevenjin8, line 81:
```suggestion
    /// The authority header value for hbone requests. It is the same for normal hbone, and the inner/outer
    /// requests of double hbone
```

src/proxy/outbound.rs
```diff
@@ -195,19 +198,89 @@ impl OutboundConnection {
...
+
+        // Proxy
+        let res = copy::copy_bidirectional(
+            copy::TcpStreamSplitter(stream),
+            inner_upgraded,
+            connection_stats,
+        )
+        .await;
+
+        let _ = drain_tx.send(true);
```
howardjohn, line 275: Do we need this? Doesn't dropping all references to the connection already make it close?

src/proxy/h2.rs
Stevenjin8: Yes, I'm holding back on unit tests until there is some agreement on the implementation

src/proxy/outbound.rs
Stevenjin8: It closes it, but in my testing, it was not graceful (e.g. the proper close notifies weren't sent). For example, if the outer connection http2 CONNECT stream closes before the inner TLS stream has a chance to send the close notify.

Stevenjin8: But since there have been some refactors, ill take another look.

src/proxy/outbound.rs
Stevenjin8: Right now, we use the peer address of the tcp stream, but if it has no meaning I agree that its better to just have a dummy variable and get rid of the argument.

src/proxy/outbound.rs
Stevenjin8: So we should check that the protocol is HBONE to do double hbone? what about the inner tunnel? I guess my question is how should we explicitly encode protocol information.? Maybe have it be a field in `network_gateway`?

src/proxy/socks5.rs
howardjohn: If we went from 2000 to 5600 we have a problem we need to address. The size here is the per-connection memory overhead, so increases here directly lead to increased memory utilization

src/proxy/outbound.rs
Stevenjin8: @howardjohn I took another look at it seems like connections aren't being closed particularly gracefully. Namely, I'm seeing FIN being sent before close notifies, But I do have a janky setup. How do we generally test for this?

src/proxy/socks5.rs
Stevenjin8: I looked into this a bit more and "solved" it by boxing the futures. However, I don't know if this makes a difference the resulting binary because the size of futures can change based on compiler optimizations. For example, if I leave `proxy_to_double_hbone` unboxed, and restore the limit to 2000, the tests will pass if I build them with `--release`, but fail otherwise. So the size of `serve` seems more like a red herring.

src/proxy/outbound.rs
Stevenjin8: @howardjohn does WDS/network gateway need a new `protocol` field? or are we good just checking that the protocol is HBONE and returning an Error in the rest of the cases? cc @keithmattix

keithmattix:
Fixes #1423 1423

src/proxy/outbound.rs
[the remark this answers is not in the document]
howardjohn: I think we covered this live on the WG call, but just to close the loop here - this must not grow (beyond a trivial amount) else it meas every connection will use that much additional memory. Typically the fix here is to Box::pin the futures.

src/proxy/outbound.rs
```diff
@@ -195,19 +199,98 @@ impl OutboundConnection {
             metrics,
         ));
 
-        let res = match req.protocol {
-            Protocol::HBONE => {
+        let res = match (
+            req.protocol,
+            req.actual_destination_workload
```
howardjohn, line 204: The presence of a network gateway does not imply we need to reach it through the network gateway. For example, a local workload in a cluster will have a network gateway defined.

The WDS protocol is not contextual. A workload w/ a network gateway does not mean "To reach this workload, always go through this network gateway [and the control plane will dynamically set or unset this for you]", it means "if you need to traverse a network boundary, here is how to do it".

howardjohn: More generally, I wonder if the code might be simplified to have a new Protocol::DoubleHbone (maybe we make a new enum like RequestProtocol though)

src/proxy/outbound.rs
```diff
@@ -453,7 +551,7 @@ fn baggage(r: &Request, cluster: String) -> String {
     )
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
howardjohn, line 554: Not needed

src/proxy/outbound.rs
howardjohn: Sorry github hid this comment so end up saying something similar but with slightly more context in https://github.com/istio/ztunnel/pull/1429#discussion_r2008209563.

I think the WDS protocol has enough information, we just need to use the context of (src,dst) to determine whether a cross network traversal is required

src/proxy/outbound.rs
Stevenjin8: Agreed.

src/proxy/outbound.rs
Stevenjin8: So just looking at the source/ dest networks should give enough info on whether to use the network gateway , and I can put this logic next to the waypoint logic in ‘build_request’ (I think that’s what the function is called).

src/proxy/socks5.rs
Stevenjin8: Ok, I take this back. Not sure if the size of futures changes based on optimization level, but I do find it a little silly that we have async functions at always box a futures certain futures for example

https://github.com/istio/ztunnel/blob/dc54d850a774e033cc0aa36357f623b8141e0926/src/proxy/outbound.rs#L170-L175

this always runs, so what's the point of boxing it? the memory usage will always be the same, except that we are moving it from the stack to the heap. The only team I think this makes sense is that if a future is really big and is not in the happy path. @howardjohn

Stevenjin8:
@howardjohn @bleggett I think this is ready for another review. Here are some of the changes

* Set dummy ServerName in tls connect
* Handle case where the network gateway is a Hostname
* write integration tests 

I still need to 

* Write unit tests
* Understand the metrics situation.

src/proxy/h2.rs
```diff
@@ -240,3 +302,4 @@ fn h2_to_io_error(e: h2::Error) -> std::io::Error {
         std::io::Error::new(std::io::ErrorKind::Other, e)
     }
 }
+
```
Stevenjin8, line 305: TODO write tests

src/proxy/h2/client.rs
```diff
@@ -147,10 +145,10 @@ impl H2ConnectClient {
 
 pub async fn spawn_connection(
     cfg: Arc<config::Config>,
-    s: TlsStream<TcpStream>,
+    s: impl AsyncRead + AsyncWrite + Unpin + Send + 'static,
     driver_drain: Receiver<bool>,
     wl_key: WorkloadKey,
-) -> Result<H2ConnectClient, Error> {
+) -> Result<(H2ConnectClient, tokio::task::JoinHandle<()>), Error> {
```
Stevenjin8, line 151: yeah I don't think we need this, doesn't seem like we care too much about gracefule shutdowns

src/proxy/outbound.rs
```diff
@@ -721,43 +1007,43 @@ mod tests {
...
+                protocol: OutboundProtocol::HBONE,
                 hbone_destination: "127.0.0.2:80",
                 destination: "127.0.0.2:15008",
             }),
         )
         .await;
     }
 
-    #[tokio::test]
-    async fn build_request_source_waypoint() {
```
Stevenjin8, line 733: fixme

src/state.rs
```diff
@@ -732,26 +771,52 @@ impl DemandProxyState {
...
         &self,
         network: Strng,
         source_workload: &Workload,
         addr: SocketAddr,
         resolution_mode: ServiceResolutionMode,
     ) -> Result<Option<Upstream>, Error> {
+        // update cache?
         self.fetch_address(&network_addr(network.clone(), addr.ip()))
             .await;
+        // find upstream and select destination workload
```
Stevenjin8, line 807: TODO: get rid of extra comments

tests/namespaced.rs
```diff
@@ -800,7 +1034,9 @@ mod namespaced {
                 let connector = cert.outbound_connector(vec![dst_id]).unwrap();
                 let hbone = SocketAddr::new(srv.ip(), 15008);
                 let tcp_stream = TcpStream::connect(hbone).await.unwrap();
+
```
Stevenjin8, line 1037: get rid of extra stuff

src/proxy/h2.rs
Stevenjin8: tests are in poo.rs

src/tls/workload.rs
```diff
@@ -163,19 +171,12 @@ pub struct OutboundConnector {
...
-                .ip()
-                .into(),
-        );
+    pub async fn connect<IO>(self, stream: IO) -> Result<client::TlsStream<IO>, io::Error>
+    where
+        IO: AsyncRead + AsyncWrite + Unpin,
+    {
         let c = tokio_rustls::TlsConnector::from(self.client_config);
-        c.connect(dest, stream).await
+        c.connect(DUMMY_DOMAIN.clone(), stream).await
```
howardjohn, line 179: Can you please revert this. We should NOT be sending an SNI.

rustls makes this kind of annoying, but if you just set an IP it will not send anything. We can put 0.0.0.0 if you want.

Putting a dummy IP != putting a dummy domain

src/proxy/outbound.rs
```diff
@@ -357,16 +434,69 @@ impl OutboundConnection {
...
 
+        // Check whether we are using an E/W gateway and sending cross network traffic
+        if let Some(ew_gtw) = &us.workload.network_gateway {
+            if us.workload.network != source_workload.network {
+                let svc = us
+                    .destination_service
+                    .as_ref()
+                    .expect("Workloads with network gateways must be service addressed.");
+
+                let (actual_destination, upstream_sans, final_sans) = match &ew_gtw.destination {
```
howardjohn, line 456: nit: can we just put this all in state.rs like fetch_waypoint but fetch_network_gateway?

src/proxy/outbound.rs
```diff
@@ -357,16 +434,69 @@ impl OutboundConnection {
...
                 intended_destination_service: None,
                 actual_destination: target,
                 upstream_sans: vec![],
+                final_sans: vec![],
             });
         };
 
+        // Check whether we are using an E/W gateway and sending cross network traffic
+        if let Some(ew_gtw) = &us.workload.network_gateway {
+            if us.workload.network != source_workload.network {
```
howardjohn, line 450: IIUC this means that if we find a workload in another network but it does NOT have a gateway, we will just send to the IP, which will not work (or worse, may go to some actual destination unrelated to the target). Do we have something protecting against this

src/proxy/outbound.rs
```diff
@@ -357,16 +434,69 @@ impl OutboundConnection {
...
+                    Destination::Address(address) => (
+                        SocketAddr::from((address.address, ew_gtw.hbone_mtls_port)),
+                        us.workload_and_services_san(),
+                        us.service_sans(),
+                    ),
+                    Destination::Hostname(host) => {
+                        let us_gtw = self
+                            .pi
+                            .state
+                            .fetch_upstream_by_host(
```
howardjohn, line 466: If we follow the above comment about following fetch_watchpoint this would probably end up calling `find_hostname --> find_upstream_from_service`

howardjohn (commented):
Overall looks good, mostly some minor comments, only blocker is the dummy SNI

src/proxy/outbound.rs
Stevenjin8: Seems like we ignore shutdown errors so I'm not going to worry about this

src/proxy/outbound.rs
Stevenjin8: I can add it, but I don't think we ever did this check

src/state.rs
```diff
@@ -835,6 +801,56 @@ impl DemandProxyState {
...
+    pub async fn fetch_network_gateway(
+        &self,
+        gtw: &GatewayAddress,
+        source_workload: &Workload,
+        original_destination_address: SocketAddr,
+        original_us: &Upstream,
+    ) -> Result<(SocketAddr, Vec<Identity>, Vec<Identity>), Error> {
+        match &gtw.destination {
+            Destination::Address(address) => Ok((
+                SocketAddr::from((address.address, gtw.hbone_mtls_port)),
```
howardjohn, line 817: The intent of these APIs is the address are not raw IPs we send to, but lookup keys into workloads/services. We should use the same logic as fetch_waypoint which calls find_upstream

Stevenjin8: ok, so I can assume that the address has a Workload object behind it

howardjohn (approved):

zirain:
the PR title is terrible.

keithmattix:
That's my fault; I removed the hold before he could change it

PR #1430, opened 2025-01-17
Serialize SA match as camelCase

howardjohn:
Not sure a good way to make sure we don't miss these. Obviously we could
add a test for the individual field but that only helps after we
notice...

### DIFF
--- a/src/rbac.rs
+++ b/src/rbac.rs
@@ -283,6 +283,7 @@ impl StringMatch {
 }
 
 #[derive(Debug, Hash, Eq, PartialEq, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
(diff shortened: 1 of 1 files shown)


keithmattix:
How does this fail today?

howardjohn:
> How does this fail today?

It doesn't fail per-se, its just the only place we use this casing. Noticed while fixing istioctl to support this field

bleggett:
Sorry - why are we only using camel casing here?

howardjohn:
Why _here_ - because every other part of the config dump does.

Why the config dump.... pretty arbitrary really, most of the k8s ecosystem is camelCase so it seems appropriate to do the same. As is envoy's

bleggett:
Ah, ok, the `from` impl.

bleggett (approved):

Issue #1431, opened 2025-01-17
ztunnel CSR inconsistency with sidecar


zliu-rh:
Tracking two issues we observed with CSRs from ztunnel:

1. Sidecar CSRs have a `Subject` of `O=` (organization blank), ztunnels CSRs do not have a Subject
2. When using the `tls-ring` feature, the CSR does not set the SAN extension as critical

---

1. is causing an issue with Istio RA's verification, [discussion here](https://github.com/istio/istio/pull/54431#discussion_r1913799777)
2. when combined with 1., causes issue with specific issuers (e.g. AWS PCA returns error `MalformedCSRException: CSR must mark the SAN extension critical when it has an empty subject.`)

kuberkaul:
https://github.com/istio/ztunnel/pull/1550 resolved it !

dhawton:
Bringing this back up.. we'll see this eventually... rcgen 0.13.3 got yanked (https://crates.io/crates/rcgen/0.13.3 and https://github.com/rustls/rcgen/issues/324), so anything doing `cargo update` will revert rcgen back to 0.13.2.

cc: @yuval-k

PR #1432, opened 2025-01-17
Make csr test stricter and more correct

howardjohn:
Part of https://github.com/istio/ztunnel/issues/1431

Fails without https://github.com/rustls/rcgen/pull/311


bleggett (approved):

yuval-k:
see #1550

PR #1441, opened 2025-01-28
metrics: add src/dst locality information to labels

howardjohn:
Understanding cross-zone/cross-region traffic is one of the most
critical pieces of observing a cluster, for a cost, reliability, and
latency standpoint. Lots of work has been done in this space from
standalone tools (https://github.com/polarsignals/kubezonnet/) to
enterprise upsells (I *think* Buoyant enterprise has this but not in
OSS?); we can provide this exact data trivially.


bleggett (approved):

bleggett:
@howardjohn might be worth backporting and adding to the 1.25 release highlights?

howardjohn:
+1 to a backport

PR #1452, opened 2025-02-11
Add xDS and CA headers/metadata support

Winbobob:
### Summary
This PR make it possible to add additional xDS and CA headers/metadata (fetched from env vars). We could use these headers/metadata for different purposes, such as routing.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -88,6 +88,9 @@ const ISTIO_META_PREFIX: &str = "ISTIO_META_";
 const DNS_CAPTURE_METADATA: &str = "DNS_CAPTURE";
 const DNS_PROXY_ADDR_METADATA: &str = "DNS_PROXY_ADDR";
 
+const ISTIO_XDS_HEADER_PREFIX: &str = "XDS_HEADER_";
+const ISTIO_CA_HEADER_PREFIX: &str = "CA_HEADER_";
+
 /// Fetch the XDS/CA root cert file path based on below constants
 const XDS_ROOT_CA_ENV: &str = "XDS_ROOT_CA";
 const CA_ROOT_CA_ENV: &str = "CA_ROOT_CA";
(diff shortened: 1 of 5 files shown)


Winbobob:
cc @howardjohn mind taking a look? Thanks a lot

src/xds/client.rs
```diff
@@ -628,9 +630,17 @@ impl AdsClient {
                 .await?,
         )?;
 
+        let mut req = tonic::Request::new(outbound);
+        self.config.xds_headers.iter().for_each(|(k, v)| {
+            let key: tonic::metadata::MetadataKey<_> = k.as_str().parse().unwrap();
```
howardjohn, line 635: It would probably be good to not panic on error. Either return an error here, or maybe we can make the `xds_headers` object a `Vec<(MetadataKey, MetadataValue)>` or similar upfront so we fail at config construction time instead of down the line

howardjohn (commented):
Overall LGTM, just one small comment.

Also, welcome to Istio :slightly_smiling_face:

src/xds/client.rs
Winbobob: Thank you for the comment!

I tried to use `Vec<(AsciiMetadataKey, AsciiMetadataValue)>` but both `AsciiMetadataKey` and `AsciiMetadataValue` do not implement the `serde::Serialize` trait.

So at the end I created a wrapper struct called `MetadataVector`, which implements serde::Serialize` trait.

src/config.rs
```diff
@@ -326,6 +372,40 @@ fn parse_args() -> String {
...
+
+    for (key, value) in env::vars() {
+        let stripped_key: Option<&str> = key.strip_prefix(prefix);
+        match stripped_key {
+            Some(stripped_key) => {
+                // attempt to parse the stripped key
+                match AsciiMetadataKey::from_str(stripped_key) {
+                    Ok(metadata_key) => {
+                        // attempt to parse the value
+                        match AsciiMetadataValue::from_str(&value) {
```
howardjohn, line 386: nit: more idiomatic would probably be to do `AsciiMetadataValue::from_str(&value).ok_or_else(||Error::InvalidHeaderValue(value))?`

howardjohn (approved):

mikemorris:
Does this feel reasonable to include in 1.25? Didn't see a backport label and we're about to start enforcing a code freeze...

src/config.rs
Winbobob: https://github.com/istio/ztunnel/pull/1460

mikemorris:
Adding backport label for 1.25 per discussion and approval in February 26th combined working group meeting.

PR #1461, opened 2025-02-21
Experimental Support for ambient on Windows

keithmattix:
Part of istio/istio#27893. Istio CNI piece is at istio/istio#55216

More information to follow in an [upcoming IstioDay talk ](https://colocatedeventseu2025.sched.com/event/1u5ht/a-new-frontier-sidecarless-service-mesh-on-windows-with-istios-ambient-mesh-keith-mattix-mitch-connors-microsoft) as well as a blog post, but in summary: we've made substantial progress towards support Istio on Windows, and we decided the time is right to share where we are with the community! After discussing with the rest of Istio TOC, we have agreed that a long-lived experimental branch is the best place for this code to live for now as we work towards productionizing it and getting CI set up. 

**What works**
- in pod traffic redirection
- HBONE upgrade
- waypoint forwarding
- ZDS communication with ztunnel

**What doesn't work**
- DNS proxying (haven't tested, redirection in CNI isn't implemented)
- Communicating with istiod via DNS (windows host process pods can't resolve via kube-dns) 
- Tests that rely on creating a netns programmatically (pre-WS2025, I don't think this is possible because of how compartments and namespaces work)
- Probably more stuff I'm forgetting

I'm opening this PR as a draft for folks to take a look and comment on the diff/approach before merging it into the experimental branch. Feel free to reach out with any questions or concerns. Welcome to Istio, Windows!


proto/zds.proto
```diff
@@ -23,11 +37,18 @@ message WorkloadInfo {
...
+  string guid = 2;
+}
+
 // Add a workload to the ztunnel. this will be accompanied by ancillary data contianing
 // the workload's netns file descriptor.
 message AddWorkload {
   string uid = 1;
   WorkloadInfo workload_info = 2;
+  // The namespace for the workload. Windows only.
+  WindowsNamespace windows_namespace = 3;
```
bleggett, line 51: Why not make this an `optional` in WorkloadInfo?

keithmattix: Honestly I think it was in an earlier draft of this, but I can't remember why the shift. Agreed that an optional is much cleaner

howardjohn (commented):
What happened to a "small 60 line change" :stuck_out_tongue: 

Good stuff. Will take a look soon

proto/zds.proto
keithmattix: Ah, I remember now, it was because I wanted to model it like the auxiliary data (i.e. the FD) from the unix domain socket communication. I don't really have any preference, and went ahead and just stuck it on workload_info

build.rs
```diff
@@ -74,7 +74,19 @@ fn main() -> Result<(), anyhow::Error> {
         .nth_back(3)
         .unwrap();
 
-    match Command::new("common/scripts/report_build_info.sh").output() {
+    
+    cfg_if::cfg_if! {
+        if #[cfg(target_os = "windows")] {
+            output = Command::new("powershell.exe")
```
grnmeira, line 80:
```suggestion
           let  output = Command::new("powershell.exe")
```

build.rs
```diff
@@ -74,7 +74,19 @@ fn main() -> Result<(), anyhow::Error> {
...
+            .arg("common/scripts/report_build_info.ps1")
+            .output();
+        } else {
+            let output: std::io::Result<std::process::Output> = Command::new("common/scripts/report_build_info.sh").output();
+
+        }
+    }
+
+    match output {
         Ok(output) => {
```
grnmeira, line 90: Minor issue here where it doesn't print anything when powershell fails to execute the script (eg. because of permission policies).
```
       Ok(output) => {
            if output.status.success() {
                for line in String::from_utf8(output.stdout).unwrap().lines() {
                    // Each line looks like `istio.io/pkg/version.buildGitRevision=abc`
                    if let Some((key, value)) = line.split_once('=') {
                        let key = key.split('.').last().unwrap();
                        println!("cargo:rustc-env=ZTUNNEL_BUILD_{key}={value}");
                    } else {
                        println!("cargo:warning=invalid build output {line}");
                    }
                }
            } else {
                println!(
                    "cargo:warning=build info script failed (stderr): {}",
                    String::from_utf8(output.stderr)
                        .unwrap()
                        .replace("\r\n", "")
                );
            }
        }
```

keithmattix:
not stale

keithmattix:
Not stale

grnmeira:
@howardjohn @craigbox @therealmitchconnor can we get the Ztunnel experimental branch in as well? Thank you!

craigbox:
Doesn't look like it needs a docs maintainer.

PR #1436, opened 2025-01-22
Add support for OpenSSL

tedjpoole:
Add support for using OpenSSL as an alternative crypto provider, implemented using the [rustls-openssl](https://crates.io/crates/rustls-openssl) and [openssl](https://crates.io/crates/openssl) crates, and guarded by a new feature called `tls-openssl`.

fixes #149

Issue #1323, opened 2024-09-20
tls-rustls-aws-ls-sys-fips build option


xnox:
Hi,

I see that there is tls-boring and tls-ring build flags available.

would you consider adding tls-aws-ls ? specifically to use rustls, with aws-ls-sys, with/without the fips feature flag of the latter one.

If such addition would be welcomed, let me know if you need help implementing it. The primary benefit of getting access to https://csrc.nist.gov/projects/cryptographic-module-validation-program/certificate/4759 CVMP cert under FIPS 140-3.

howardjohn:
fwiw the boring crate is also has a cvmp 140-3 certification

howardjohn:
Anyhow, from a technical standpoint it's harder to add providers than other projects since we need to do CSRs. This is the "rcgen" stuff. Might be simple but need to make sure we have a solution. 

from a project standpoint I think the main thing is we want to make sure any we are are actively maintained by someone so we don't end up with a bunch of providers causing issues.

xnox:
Basically the aws-lc is now the default rustls provider (instead of ring), and more stuff is likely standertise on it.

And the fips version of it the 2.0.0 lifecycle policy is much kinder, thus easier to deploy compliance with.

howardjohn:
@xnox do you have a specific desire to have the AWS module or to have a FIPS validated module? Id the former, would be curious to know why

Currently the plan has been to offer boringssl as the FIPS validated module primarily, as is the same module used in Go and Envoy

howardjohn:
If we do this one other thing I am reminded of - boringssl and aws-lc add some onerous build steps. With boring we are able to use a pre compiled build. It would be great to do this with aws as well if we adopt it. 

If it's the 3rd most recommended choice I suppose build ergonomics may not be as important, though.

xnox:
> If we do this one other thing I am reminded of - boringssl and aws-lc add some onerous build steps. With boring we are able to use a pre compiled build. It would be great to do this with aws as well if we adopt it.
> 
> If it's the 3rd most recommended choice I suppose build ergonomics may not be as important, though.

I expect most distributions to provide a prebuilt aws-lc for usage in aws-lc fips crate. And the build instructions are actually fairly relaxed for asw-lc. At least they appear easier to meet than boringssl ones.

xnox:
[...]

PR #1466, opened 2025-02-25
Add support for AWS-LC-RS TLS provider

howardjohn:
Related PR: https://github.com/istio/ztunnel/pull/1436/

Ref https://github.com/istio/ztunnel/issues/1323

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ edition = "2024"
 rust-version = "1.85"
 
 [features]
-default = ["tls-aws-lc"]
+default = ["tls-ring"]
 jemalloc = ["dep:tikv-jemallocator", "dep:jemalloc_pprof"]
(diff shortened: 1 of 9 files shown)


Cargo.toml
```diff
@@ -5,10 +5,11 @@ edition = "2021"
 rust-version = "1.77"
 
 [features]
-default = ["tls-ring"]
+default = ["tls-aws-lc"]
```
bleggett, line 8: LGTM, and I think this is a better default anyway (there's nothing very special about `ring` and upstream rust-tls already moved to `aws-lc`), but if we make it default README needs another bump to reflect.

bleggett (approved):

Cargo.toml
howardjohn: Oh I actually meant to make this !default in this PR and split it out

jacob-delgado (approved):

PR #1470, opened 2025-02-27
Add benchmarks for load balancing

howardjohn:


bleggett (approved):
LGTM -> is this expected when I run the benches, or just due to local-machine variability?

> Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 5.8s, enable flat sampling, or reduce sample count to 60.

howardjohn:
happens for me to, I think it just means the test is not super fast so its complaining it takes slightly too long?

PR #1472, opened 2025-02-27
Implement capacity

howardjohn:
Ztunnel side impl of https://github.com/istio/istio/pull/55284

This has a small perf impact. I have some ideas on improving it though.

```
load_balance/basic-10   time:   [510.99 ns 511.45 ns 511.93 ns]
                        thrpt:  [1.9534 Melem/s 1.9552 Melem/s 1.9570 Melem/s]
                 change:
                        time:   [+25.568% +26.059% +26.629%] (p = 0.00 < 0.05)
                        thrpt:  [-21.029% -20.672% -20.362%]
                        Performance has regressed.
load_balance/basic-1000 time:   [30.696 µs 30.757 µs 30.822 µs]
                        thrpt:  [32.444 Kelem/s 32.513 Kelem/s 32.577 Kelem/s]
                 change:
                        time:   [+18.856% +19.240% +19.608%] (p = 0.00 < 0.05)
                        thrpt:  [-16.393% -16.136% -15.865%]
                        Performance has regressed.
load_balance/basic-10000
                        time:   [439.77 µs 441.01 µs 442.05 µs]
                        thrpt:  [2.2622 Kelem/s 2.2675 Kelem/s 2.2739 Kelem/s]
                 change:
                        time:   [+8.0823% +10.660% +13.024%] (p = 0.00 < 0.05)
                        thrpt:  [-11.523% -9.6330% -7.4779%]
                        Performance has regressed.
load_balance/locality-10
                        time:   [579.62 ns 580.77 ns 581.93 ns]
                        thrpt:  [1.7184 Melem/s 1.7218 Melem/s 1.7253 Melem/s]
                 change:
                        time:   [+4.6631% +4.9633% +5.2462%] (p = 0.00 < 0.05)
                        thrpt:  [-4.9847% -4.7286% -4.4553%]
                        Performance has regressed.
load_balance/locality-1000
                        time:   [36.301 µs 36.417 µs 36.564 µs]
                        thrpt:  [27.350 Kelem/s 27.459 Kelem/s 27.548 Kelem/s]
                 change:
                        time:   [-3.0653% -2.8098% -2.5421%] (p = 0.00 < 0.05)
                        thrpt:  [+2.6084% +2.8910% +3.1622%]
                        Performance has improved.
load_balance/locality-10000
                        time:   [491.06 µs 492.14 µs 493.39 µs]
                        thrpt:  [2.0268 Kelem/s 2.0319 Kelem/s 2.0364 Kelem/s]
                 change:
                        time:   [-20.042% -13.052% -8.0796%] (p = 0.00 < 0.05)
                        thrpt:  [+8.7898% +15.011% +25.065%]
                        Performance has improved.
```

### DIFF
--- a/proto/workload.proto
+++ b/proto/workload.proto
@@ -16,6 +16,7 @@ syntax = "proto3";
 
 package istio.workload;
 
+import "google/protobuf/wrappers.proto";
 import "google/protobuf/any.proto";
 
 option go_package="pkg/workloadapi";
@@ -266,6 +267,11 @@ message Workload {
 
   // Extension provides a mechanism to attach arbitrary additional configuration to an object.
   repeated Extension extensions = 26;
+
+  // Capacity for this workload.
+  // This represents the amount of traffic the workload can handle, relative to other workloads
+  // If unset, the capacity is default to 1.
+  google.protobuf.UInt32Value capacity = 27;
 
   // Reservations for deleted fields.
   reserved 15;
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -659,6 +659,7 @@ mod tests {
                 subzone: "subezone".to_string(),
             }),
             extensions: Default::default(),
+            capacity: Default::default(),
             // ..Default::default() // intentionally don't default. we want all fields populated
         };
 
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -897,6 +897,7 @@ mod tests {
             application_tunnel: None,
             locality: Default::default(),
             services: Default::default(),
+            capacity: Default::default(),
         }
     }
 
@@ -927,6 +928,7 @@ mod tests {
             application_tunnel: None,
             locality: Default::default(),
             services: Default::default(),
+            capacity: Default::default(),
         }
     }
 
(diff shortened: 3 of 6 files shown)


bleggett (approved):
> This has a small perf impact. I have some ideas on improving it though.

Oh it's literally just the stdlib `choose_weighted` impl causing the change here, ok. Even if we can't do better than that impl we kinda need to do something like this so yeah doesn't leave us  with much choice.

howardjohn:
> > This has a small perf impact. I have some ideas on improving it though.
> 
> Oh it's literally just the stdlib `choose_weighted` impl causing the change here, ok. Even if we can't do better than that impl we kinda need to do something like this so yeah doesn't leave us with much choice.

Yeah basically with `choose` you can do it on an iterator but with `choose_weighted` you need to collect first. 

There is an iterator choose_weighted algo (not in stdlib) that works on iterators but it generates a rand num for every element (vs 1 rand number) so I suspect its worse but not tested

Most of the cost of the LB function (unchanged in this PR) are around looking up the Workload for each Endpoint, so I think just storing those together would be a bigger win than any optimization in this area anyways

howardjohn:
Eventually we will probably do least_conn LB algorithm which will rewrite this anyways

PR #1490, opened 2025-03-21
misc refactors around inbound code

howardjohn:
Some of the initial choses we made have not really held up over the
tweaks we have made over the year. Clean things up a bit. This should
have no meaningful user impact, mostly a refactoring.

* check_from_network_gateway is not right; it is assuming we get the
  identity from the EW GW; in the modern multi-network scheme that is
never true as its forwarded e2e from the client (double hbone). Instead,
change the heuristic to be "if hostname, it is multi-network".
* Instead of trying to create the PROXY protocol handling in the main
  forwarding logic, refactor it into find_inbound_upstream.

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -22,7 +22,7 @@ use tokio::sync::watch;
 
 use tracing::{Instrument, debug, info, info_span, trace_span};
 
-use super::{ConnectionResult, Error, LocalWorkloadInformation, ResponseFlags};
+use super::{ConnectionResult, Error, HboneAddress, LocalWorkloadInformation, ResponseFlags};
 use crate::baggage::parse_baggage_header;
 use crate::identity::Identity;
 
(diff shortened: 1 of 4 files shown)


src/proxy/inbound.rs
```diff
@@ -606,51 +485,126 @@ impl Inbound {
...
-                // connection to
-                (new_target, protocol)
+                // connection to.
+
+                // Which address we will send in the tunnel
+                let tunnel_target = match hbone_addr {
+                    HboneAddress::SvcHostname(h, port) => {
+                        // PROXY cannot currently send to hostnames, so we will need to select an IP to
+                        // use instead
+                        let vip = services[0]
```
ilrudie, line 563: This could panic if we happen to have no services, yeah? I think we want to use first() instead.

src/proxy/inbound.rs
```diff
@@ -294,128 +282,36 @@ impl Inbound {
         // Extract the host or IP from the authority pseudo-header of the URI
         let hbone_addr: HboneAddress = req
             .uri()
-            .clone()
             .try_into()
-            .map_err(|e: Error| InboundError(e, StatusCode::BAD_REQUEST))?;
+            .map_err(InboundError::build(StatusCode::BAD_REQUEST))?;
```
keithmattix, line 286: Is a lambda less performant or something? Or is this just readability?

howardjohn: Just readability, exact same compilation

src/proxy/inbound.rs
ilrudie: Sounds good. I would prefer the expect personally.

ilrudie (approved):
lgtm, added a hold so you can decide whether or not to make the couple suggestions before it goes in

PR #1492, opened 2025-03-25
[1.25] Add support for OpenSSL

tedjpoole:
Back port of #1436 

Add support for using OpenSSL as an alternative crypto provider, implemented using the [rustls-openssl](https://crates.io/crates/rustls-openssl) and [openssl](https://crates.io/crates/openssl) crates, and guarded by a new feature called `tls-openssl`.


howardjohn:
Hmm, I am not sure this makes sense to backport as it is purely a new feature. Note we also didn't backport the AWS-LC provider

tedjpoole:
Sorry for the noise

PR #1500, opened 2025-04-01
display protocol in accesslog

zirain:
display src/dst protocol in accesslog, which help understand the connection better.

```console
2025-04-01T14:01:02.476412Z	info	access	connection complete	src.addr=10.244.3.9:52462 src.workload="simple-http-waypoint-547d6d7b8b-4xckr" src.protocol=TCP src.namespace="default" src.identity="spiffe://cluster.local/ns/default/sa/simple-http-waypoint" dst.addr=10.244.3.8:15008 dst.hbone_addr=10.244.3.8:8080 dst.service="httpbin.default.svc.cluster.local" dst.workload="httpbin-54845dfb75-4tsvd" dst.protocol=HBONE dst.namespace="default" dst.identity="spiffe://cluster.local/ns/default/sa/httpbin" direction="inbound" bytes_sent=564 bytes_recv=156 duration="5013ms"
```

### DIFF
--- a/src/proxy/metrics.rs
+++ b/src/proxy/metrics.rs
@@ -33,7 +33,7 @@ use crate::metrics::DefaultedUnknown;
 use crate::proxy::{self, HboneAddress};
 
 use crate::state::service::ServiceDescription;
-use crate::state::workload::Workload;
+use crate::state::workload::{Protocol, Workload};
 use crate::strng::{RichStrng, Strng};
 
 #[derive(Debug)]
@@ -344,9 +344,9 @@ impl Metrics {
 /// ConnectionResult abstracts recording a metric and emitting an access log upon a connection completion
 pub struct ConnectionResult {
     // Src address and name
-    src: (SocketAddr, Option<RichStrng>),
+    src: (SocketAddr, Option<RichStrng>, Option<Protocol>),
     // Dst address and name
-    dst: (SocketAddr, Option<RichStrng>),
+    dst: (SocketAddr, Option<RichStrng>, Option<Protocol>),
     hbone_target: Option<HboneAddress>,
     start: Instant,
 
@@ -435,10 +435,15 @@ impl ConnectionResult {
         metrics: Arc<Metrics>,
     ) -> Self {
         // for src and dest, try to get pod name but fall back to "canonical service"
-        let mut src = (src, conn.source.as_ref().map(|wl| wl.name.clone().into()));
+        let mut src = (
+            src,
+            conn.source.as_ref().map(|wl| wl.name.clone().into()),
+            conn.source.as_ref().map(|wl| wl.protocol),
+        );
         let mut dst = (
             dst,
             conn.destination.as_ref().map(|wl| wl.name.clone().into()),
+            conn.destination.as_ref().map(|wl| wl.protocol),
         );
         let tl = CommonTrafficLabels::from(conn);
         metrics.connection_opens.get_or_create(&tl).inc();
@@ -454,6 +459,7 @@ impl ConnectionResult {
 
             src.addr = %src.0,
             src.workload = src.1.as_deref().map(to_value),
+            src.protocol = src.2.as_ref().map(display),
             src.namespace = tl.source_workload_namespace.to_value(),
             src.identity = tl.source_principal.as_ref().filter(|_| mtls).map(to_value_owned),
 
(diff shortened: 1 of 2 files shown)


howardjohn (changes requested):
A "src.protocol=HBONE" does not make sense for an outbound connection. That seems to imply there is mTLS between the app and ztunnel which is not the case. 

Using the protocol from the workload object in general is broken, as just because they *can* support mTLS doesn't mean they actually sent it. We may be incorrectly logging it's mTLS when it's not, which is a security concern. 

The access logs already securely indicate the protocol (https://ambientmesh.io/docs/security/verify-mtls/#confirm-with-ztunnel-logs) as well so I don't think this is necessary

zirain:
this's not for confirming mTLS, it's a quick insight for sandwitch or not.

howardjohn:
I don't understand what this has to do with sandwich.

For outbound, the src.protocol is **always** HBONE since otherwise ztunnel wouldn't see it. the same for inbound (but for dst.protocol). So those seem confusing to report.

I don't understand why I would care if the peer is supporting HBONE...

Stevenjin8:
I do think this  might be useful if we have double hbone, but I agree that we shoudn't fish this field out of the workload.

zirain:
```
2025-04-03T02:49:23.224282Z     info    access  connection complete     src.addr=10.244.1.21:46182 src.workload="sleep-64d59dd8d9-4gl56" src.protocol=HBONE src.namespace="default" src.identity="spiffe://cluster.local/ns/default/sa/sleep" dst.addr=10.244.1.22:15008 dst.hbone_addr=10.96.186.42:8000 dst.service="httpbin.default.svc.cluster.local" dst.workload="simple-http-waypoint-7b8b74dfb9-xnwq4" dst.protocol=TCP dst.namespace="default" dst.identity="spiffe://cluster.local/ns/default/sa/simple-http-waypoint" direction="outbound" bytes_sent=79 bytes_recv=691 duration="17ms"
2025-04-03T02:49:28.228792Z     info    access  connection complete     src.addr=10.244.1.22:45224 src.workload="simple-http-waypoint-7b8b74dfb9-xnwq4" src.protocol=TCP src.namespace="default" src.identity="spiffe://cluster.local/ns/default/sa/simple-http-waypoint" dst.addr=10.244.1.20:15008 dst.hbone_addr=10.244.1.20:8080 dst.service="httpbin.default.svc.cluster.local" dst.workload="httpbin-68f6749f8-ttnzq" dst.protocol=HBONE dst.namespace="default" dst.identity="spiffe://cluster.local/ns/default/sa/httpbin" direction="inbound" bytes_sent=565 bytes_recv=156 duration="5009ms"
```

the dst.protocol is `TCP` when using waypoint, and it's `HBONE` when using sandwich

howardjohn:
Which doesn't even make sense at all to say "dst.protocol=TCP" when we are connecting to a waypoint *which is over hbone*...

zirain:
there's actually a difference, right?
I counld find out information from zc config, but worth it to show in `accesslog` IMO.

I'm fine to close this PR, if you think this bring more confuse than benifit.

howardjohn:
I don't even get what you want to show tbh. The fact a waypoint shows TCP vs HBONE for sandwiches vs normal is really a quirk of the implementation that has no meaning and certainly shouldn't be user facing. We could change it in the control plane and there would be no behavior change - waypoints *always* use hbone so the protocol field is ignored.

zirain:
the traffic flow is `sleep -1-> waypoint/nginx-sandwich-waypoint -2-> httpbin`
with native waypoint, ztunnel show TCP protocol in outbound
with sandwich(e.g. nginx-ingress) ,  ztunnel show HBONE protocol in outbound

howardjohn:
Yes, I understand the field has 2 different values based on these two deployment types. However, the `protocol={TCP/HBONE}` has literally no meaning to the user because **its not the protocol**. Its the value of a field **that has no meaning and is not used**.

zirain:
> Yes, I understand the field has 2 different values based on these two deployment types. However, the `protocol={TCP/HBONE}` has literally no meaning to the user because **its not the protocol**. Its the value of a field **that has no meaning and is not used**.

IMO, this's useful some case for debuging as it's the state just return the current.
we'd better make it configable or give it a better name instead of `protocol` which is confused, WDYT?

howardjohn:
you could say the same for the other 20 fields in the workload.. this seems incredibly niche if you need it it can just be looked up from the workload name...?

howardjohn:
This change 100% will cause widespread confusion for users so I am very against hurting users to help develops debug niche configurations

zirain:
fair enough

PR #1501, opened 2025-04-03
PROXY Protocol: send to localhost

stevenctl:
For "sandwich" it's probably desirable to have the sandwiched proxy bind listeners to localhost to prevent things other than the zTunnel from reaching it. 

This PR make it so that any inbound request that would use PROXY will send to localhost. There is an env var to turn this behavior on; long term it should be on by default, forcing sandwich implementations to follow this pattern. 

* In this mode we also bind the `src` address coming from zTunnel to be localhost
* I'm currently guessing between `::1` and `127.0.0.1` based on the IP family of the actual destination address. Maybe this can be the `::ffff:127.0.0.1` or the guess should be based on the ip family of `orig_src`? 
* Currently I don't see the need for a per-proxy/workload setting here but we can revisit that in the future.  

I've tested this alongside an envoy bound like

```
      "listener": {
       "@type": "type.googleapis.com/envoy.config.listener.v3.Listener",
       "name": "proxy_protocol_inbound",
       "address": {
        "socket_address": {
         "address": "::ffff:127.0.0.1",
         "port_value": 15088,
         "ipv4_compat": true
        }
       },
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -106,6 +106,8 @@ const CERT_SYSTEM: &str = "SYSTEM";
 const PROXY_MODE_DEDICATED: &str = "dedicated";
 const PROXY_MODE_SHARED: &str = "shared";
 
+const LOCALHOST_APP_TUNNEL: &str = "LOCALHOST_APP_TUNNEL";
+
 #[derive(serde::Serialize, Clone, Debug, PartialEq, Eq)]
 pub enum RootCert {
     File(PathBuf),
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -746,6 +751,8 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
         fake_self_inbound: false,
         xds_headers: parse_headers(ISTIO_XDS_HEADER_PREFIX)?,
         ca_headers: parse_headers(ISTIO_CA_HEADER_PREFIX)?,
+
+        localhost_app_tunnel: parse_default(LOCALHOST_APP_TUNNEL, false)?,
```
howardjohn, line 755: probably ok to default to true?

howardjohn (approved):

PR #1505, opened 2025-04-03
Regenerate certs with a valid CA

howardjohn:
The current root we use doesn't declare itself as supporting
keyCertSigning, which is rejected in some more strict usages. Regenerate
it with the constraint


ilrudie (approved):
LGTM

PR #1507, opened 2025-04-04
Support TLS for metrics endpoint

harsh-ps-2003:
Based on the [RFC](https://docs.google.com/document/d/19vMr22inhLAPjWcVJkZWiG_GY6ZG3PLxG9QqAmIlnIc/edit?usp=sharing) :

Adds support for mTLS via HBONE tunnelling along with already existing HTTP metrics endpoint via leveraging k8s network policy for enforcement.


src/proxy/inbound.rs
```diff
@@ -525,6 +530,28 @@ impl Inbound {
...
+                    "ztunnel".to_string(),
+                    hbone_addr.port(),
+                )),
+            };
+            
+            let target = SocketAddr::new(target_ip, metrics_port);
+            tracing::debug!("ztunnel metrics request forwarded to {}", target);
+            return Ok((target, None, Vec::new()));
+        }
+
```
ilrudie, line 554: Can likely remove, see comment on test asserting this behavior.

src/identity/manager.rs
```diff
@@ -125,6 +126,77 @@ impl Identity {
             Identity::Spiffe { trust_domain, .. } => trust_domain.clone(),
         }
     }
+
+    /// Creates a new proxy Identity using the Kubernetes Downward API environment variables.
+    pub fn from_downward_api() -> Result<Self, Error> {
+        let trust_domain = std::env::var("TRUST_DOMAIN")
+            .or_else(|_| std::env::var("CLUSTER_DOMAIN"))
+            .unwrap_or_else(|_| "cluster.local".to_string());
```
ilrudie, line 134: We should probably take a TODO here to figure out if we should have a hard coded default like this. It seems like probably not.

src/identity/manager.rs
```diff
@@ -125,6 +126,77 @@ impl Identity {
             Identity::Spiffe { trust_domain, .. } => trust_domain.clone(),
         }
     }
+
+    /// Creates a new proxy Identity using the Kubernetes Downward API environment variables.
+    pub fn from_downward_api() -> Result<Self, Error> {
+        let trust_domain = std::env::var("TRUST_DOMAIN")
```
howardjohn, line 132: Nit: this is not really the 'downward API' since its just a hardcoded env var someone will set.

src/identity.rs
```diff
@@ -34,6 +34,23 @@ pub mod mock {
     };
 }
 
+/// Gets the Identity for ztunnel itself using the Downward API
+pub fn get_ztunnel_identity() -> Result<Identity, Error> {
```
howardjohn, line 38: rather than read this each time we want to fetch it, we should just put it in config.rs and read once at startup. Then invalid config will fail fast instead of later at runtime.

src/proxy/inbound.rs
```diff
@@ -525,6 +530,28 @@ impl Inbound {
     ) -> Result<(SocketAddr, Option<TunnelRequest>, Vec<Arc<Service>>), Error> {
         // We always target the local workload IP as the destination. But we need to determine the port to send to.
         let target_ip = conn.dst.ip();
+        
+        // When ztunnel is serving its own metrics over mTLS - the target is the ztunnel itself
+        if local_workload.name.starts_with("ztunnel") {
```
howardjohn, line 535: This is not going to be safe or reliable mechanism to detect "ourselves".

src/proxy/inbound.rs
```diff
@@ -525,6 +530,28 @@ impl Inbound {
...
+            let metrics_port = match hbone_addr {
+                HboneAddress::SocketAddr(addr) if matches!(cfg.stats_addr, ConfigAddress::SocketAddr(s) if addr.port() == s.port())
+                    || matches!(cfg.stats_addr, ConfigAddress::Localhost(_, p) if addr.port() == p) => {
+                    match cfg.stats_addr {
+                        ConfigAddress::Localhost(_, port) => port,
```
howardjohn, line 540: We should not support sending to localhost. If they set stats address to localhosty, that means they don't want it accessible externally. We should not bypass that. 

Also, it makes this much simpler since we can just remove this code

src/identity/manager.rs
harsh-ps-2003: `cluster.local` is the default domain name used for internal DNS resolution within a Kubernetes cluster, that's why I thought of using it as default value. Referred - https://komodor.com/learn/kubernetes-dns-a-beginners-guide/

src/app.rs
```diff
@@ -128,7 +128,7 @@ pub async fn build_with_cert(
         config.clone(),
         state.clone(),
         cert_manager.clone(),
-        proxy_metrics,
+        proxy_metrics.clone(),
```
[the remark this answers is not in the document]
harsh-ps-2003: Oh yeah, it's already reference counted! Nit from my side, sorry.

src/identity/manager.rs
howardjohn: trust domain and cluster domain are not correlated in any way other than having the same default value. We should probably default to cluster.local since we do it all over in istio, but we should not read `CLUSTER_DOMAIN`

src/identity/manager.rs
harsh-ps-2003: A bit confused here. 

Referring this - https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/

Dont we simply use the env variables to get the values from pod metadata during runtime, they are populated by Kubernetes' Downward API mechanism?

When we specify in pod manifest :

```
env:
  - name: POD_NAMESPACE
    valueFrom:
      fieldRef:
        fieldPath: metadata.namespace
  - name: POD_NAME
    valueFrom:
      fieldRef:
        fieldPath: metadata.name
  - name: SERVICE_ACCOUNT
    valueFrom:
      fieldRef:
        fieldPath: spec.serviceAccountName
```

the function consumes the values populated by Downward API. Isnt this downward API :(

ilrudie: Downward API isn't the same as any environment variable. TRUST_DOMAIN is a normal variable you just set and read. Could probably just be `from_env` or something like that. IDK

harsh-ps-2003: Yes, TRUST_DOMAIN is a regular config env var that can be set directly. 
But POD_NAMESPACE, SERVICE_ACCOUNT, and POD_NAME would normally come from Kubernetes Downward API in a real pod.

src/config.rs
```diff
@@ -293,6 +293,10 @@ pub struct Config {
 
     // If true, when AppTunnel is set for
     pub localhost_app_tunnel: bool,
+
+    pub ztunnel_identity: identity::Identity,
+
+    pub ztunnel_workload: state::WorkloadInfo,
```
ilrudie, line 299: These should probably be options. If someone doesn't set the variables ztunnel shouldn't fail, it just won't be able to start its own inbound listener and proxy to itself.

src/proxy/inbound.rs
```diff
@@ -514,9 +516,12 @@ impl Inbound {
...
-    /// find_inbound_upstream determines the next hop for an inbound request.
-    #[expect(clippy::type_complexity)]
-    fn find_inbound_upstream(
+    /// Resolves the target address for an inbound request.
+    /// This handles both regular workload traffic and ztunnel's own services.
+    /// 
+    /// When ztunnel is the target workload (identified by workload name starting with "ztunnel"),
+    /// this function will handle routing to local services like metrics or metadata endpoints.
+    pub(super) fn resolve_inbound_target(
         cfg: &Config,
```
ilrudie, line 525: All the other changes were backed out, we should probably also go ahead and back out the rename and these comments which no longer apply.

src/config.rs
howardjohn: or at the very least, it should be optional in Dedicated mode, which would never do this and should not be required to set the env vars (they typically are run outside of k8s in this mode)

src/config.rs
```diff
@@ -1048,6 +1101,50 @@ pub mod tests {
         validate_metadata_vector(&cfg.ca_headers, expected_ca_headers.clone());
     }
 
+    #[test]
+    fn test_ztunnel_identity_workload() {
```
ilrudie, line 1105: https://docs.rs/envtestkit/latest/envtestkit/, wonder if we should use something like this for the testing 🤔

howardjohn: its basically broken to set env vars in tests. unless you use the same library in ALL tests (so they share a global mutex to serialize tests) which sucks.

howardjohn: Better to structure in a way such that you do not depend on the env var.

ToxiWatch:
Hi @howardjohn, we're researchers from Stockholm University.

Your recent [comment](https://github.com/istio/ztunnel/pull/1507#discussion_r2035910381) was flagged by [Detoxify](https://github.com/unitaryai/detoxify) (an automated tool) as potentially toxic.

> its basically broken to set env vars in tests. unless you use the same library in all tests (so...

While detection isn't perfect, [research](https://ieeexplore.ieee.org/document/9397528) shows toxic interactions on GitHub can discourage developers, leading to stress and burnout. If you'd like, here are a few ways to improve clarity or tone:
- Rephrase with a more constructive style
- Add context or soften strong language
- Try tools like ChatGPT for a neutral rewrite

You can share feedback in our [1-minute survey](http://otree-extra-url-d0bb646a8437.herokuapp.com/?hash_id=20d546f8a0bfd0681cdba0417eaf5ea9197633c4d113cc48bcdf36d06764b9af) (opt-out available at the end).

Thanks for helping make open source a more welcoming place — click our profile to learn more about us!

React to this comment to let us know you saw it!

src/app.rs
```diff
@@ -128,14 +129,56 @@ pub async fn build_with_cert(
...
         drain_rx.clone(),
     )
     .map_err(|e| anyhow::anyhow!("failed to start proxy factory {:?}", e))?;
 
     if config.proxy_mode == config::ProxyMode::Shared {
         tracing::info!("shared proxy mode - in-pod mode enabled");
+        
+        // Create ztunnel inbound listener only if both identity and workload info are set up
+        if let (Some(ztunnel_identity), Some(ztunnel_workload)) = (&config.ztunnel_identity, &config.ztunnel_workload) {
+            if ztunnel_workload.service_account == "ztunnel" {
```
ilrudie, line 143: probably don't need this check

src/proxy/inbound.rs
```diff
@@ -536,15 +537,33 @@ impl Inbound {
...
+                    } else {
+                        IpAddr::V6(Ipv6Addr::LOCALHOST)
+                    };
+                    let target = SocketAddr::new(loopback_ip, 15020);
+                    debug!(%target, "Redirecting ztunnel metrics request to loopback");
+                    return Ok((target, None, vec![]));
+                }
+            }
+        }
+
```
ilrudie, line 563: We shouldn't require a special case to handle this traffic

src/config.rs
ilrudie: Not sure how to restructure a test which boils down to "we correctly read env vars". Maybe we just remove this test and the functionality will be stressed at a higher level

howardjohn (commented):
Nice work! looks pretty good, just a few small comments. An e2e test under `tests/` would be great as well

src/proxy/inbound.rs
```diff
@@ -536,15 +537,33 @@ impl Inbound {
...
-    fn find_inbound_upstream(
+    pub(super) fn find_inbound_upstream(
         cfg: &Config,
         state: &DemandProxyState,
         conn: &Connection,
         local_workload: &Workload,
         hbone_addr: &HboneAddress,
     ) -> Result<(SocketAddr, Option<TunnelRequest>, Vec<Arc<Service>>), Error> {
+        // Special case for ztunnel metrics endpoint
+        if local_workload.service_account == "ztunnel" {
```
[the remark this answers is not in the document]
harsh-ps-2003: When the special case is removed, I see no access log entry for the Prometheus -> ztunnel metrics connection, indicating the TCP connection attempt likely failed at the network level before the HBONE proxy logic could even log the attempt. You are right, I shouldn't use loopback.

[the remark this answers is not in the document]
ilrudie: Without special casing he's seeing a timeout from `freebind_connect`. In his testing he found this worked so pushed while it was working.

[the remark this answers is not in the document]
howardjohn: Ah the freebind requires iptables. We should just disable it for this flow specifically

tests/namespaced.rs
```diff
@@ -1389,6 +1393,86 @@ mod namespaced {
...
+        let dst_addr = format!("{zt_ip}:15020");
+        let want = HashMap::from([
+            ("scope", "access"),
+            ("src.workload", "prometheus"),
+            ("dst.workload", "ztunnel-node"), // Actual workload name from logs
+            // ("dst.hbone_addr", ...), // Not present in logs 
+            ("dst.addr", &dst_addr),
+            ("bytes_sent", &sent),
+            // ("bytes_recv", &recv), // Removed due to variability
+            ("direction", "outbound"),
```
ilrudie, line 1465: We should validate the inbound direction. Inbound is the new bit here.

tests/namespaced.rs
```diff
@@ -1389,6 +1393,86 @@ mod namespaced {
...
+        run_hbone_metrics_scrape_client(prometheus, manager.resolver(), &target).await?;
+
+        // BYTES_RECV = bytes sent by client = size of GET request (observed as 41 with hyper client)
+        // BYTES_SENT = bytes received by client = size of metrics response (variable, so I removed assertion)
+        const EXPECTED_METRICS_REQ_SIZE: u64 = 41; // Update expected size based on observation
+        let metrics = [
+            (CONNECTIONS_OPENED, 1),
+            (CONNECTIONS_CLOSED, 1),
+            (BYTES_RECV, EXPECTED_METRICS_REQ_SIZE), 
+            // (BYTES_SENT, ???),
```
ilrudie, line 1449: Seems ok to leave it out for test reliability. The client does validation of the contents. It can do more if we deem the existing state insufficient.

tests/namespaced.rs
```diff
@@ -1740,12 +1872,13 @@ mod namespaced {
...
     }
 
-    async fn hbone_read_write_stream(stream: &mut TcpStream) {
+    async fn hbone_read_write_stream(stream: &mut TcpStream) -> anyhow::Result<()> {
         const BODY: &[u8] = b"hello world";
-        stream.write_all(BODY).await.unwrap();
+        stream.write_all(BODY).await?;
         let mut buf = [0; BODY.len() + WAYPOINT_MESSAGE.len()];
-        stream.read_exact(&mut buf).await.unwrap();
+        stream.read_exact(&mut buf).await?;
```
ilrudie, line 1879: why? We want these to panic on errors so we assure tests fail if the r/w fail. Arguably updating to an expect could be helpful for locating where a test failure occurred.

tests/namespaced.rs
```diff
@@ -1389,6 +1393,86 @@ mod namespaced {
...
+            ("dst.workload", "ztunnel-node"), // Actual workload name from logs
+            // ("dst.hbone_addr", ...), // Not present in logs 
+            ("dst.addr", &dst_addr),
+            ("bytes_sent", &sent),
+            // ("bytes_recv", &recv), // Removed due to variability
+            ("direction", "outbound"),
+            ("message", "connection complete"),
+            // Identities are not logged for direct metrics scrape connection
+            // ("src.identity", ...),
+            // ("dst.identity", ...),
```
ilrudie, line 1469: This is the HBONE test. Identities should be logged.

tests/namespaced.rs
```diff
@@ -1423,66 +1423,145 @@ mod namespaced {
...
-        // Ztunnel will intercept and handle the connection via HBONE
-        let target_addr = SocketAddr::new(server_ip, SERVER_PORT);
-        info!(target=%target_addr, "Client attempting connection");
-        run_tcp_client(client, manager.resolver(), &target_addr.to_string())?;
-        info!(target=%target_addr, "Client connection complete");
+        let zt_identity_str = zt.ztunnel_identity.as_ref().unwrap().to_string();
+        let cert_manager_clone = zt.cert_manager.clone();
+        let zt_identity_clone = zt.ztunnel_identity.as_ref().expect("Ztunnel identity missing").clone();
+
+        // Client connects to ztunnel's HBONE port (15008), sends CONNECT to metrics port, then GET /metrics
```
ilrudie, line 1445: The client doesn't speak HBONE so it should not connect directly to 15008. It's trying to reach metrics on 15020 (or where ever we setup the metrics listener) but ztunnel gets that connection and tunnels it over HBONE (in this case from outbound on the client side to its own inbound hbone listener).

src/config.rs
ilrudie: @harsh-ps-2003 , we should probably clean this up. We don't want to be setting env vars in tests but now there logic uses battle tested parse rather than a custom env var reading anyway.

src/identity.rs
```diff
@@ -54,6 +54,8 @@ pub enum Error {
     Forgotten,
     #[error("BUG: identity requested {0}, but only allowed {1:?}")]
     BugInvalidIdentityRequest(Identity, Arc<WorkloadInfo>),
+    #[error("No ztunnel identity configured")]
+    NoIdentity,
```
ilrudie, line 58: I think this is no longer used now that we don't error out if we can start the zt inbound

src/proxy.rs
```diff
@@ -259,6 +260,8 @@ pub(super) struct ProxyInputs {
     socket_factory: Arc<dyn SocketFactory + Send + Sync>,
     local_workload_information: Arc<LocalWorkloadInformation>,
     resolver: Option<Arc<dyn Resolver + Send + Sync>>,
+    // If true, inbound connections created with these inputs will not attempt to preserve the original source IP.
+    pub disable_inbound_freebind: bool,
```
stevenctl, line 264: Is this used? I don't see anywhere besides `outbound.rs` which has it `false`.

src/proxy/inbound.rs
```diff
@@ -244,10 +245,12 @@ impl Inbound {
                     SocketAddr::new(loopback, ri.upstream_addr.port()),
                 )
             } else {
-                (
-                    enable_original_source.then_some(ri.rbac_ctx.conn.src.ip()),
-                    ri.upstream_addr,
-                )
+                let upstream_src_ip = if pi.disable_inbound_freebind {
```
stevenctl, line 248: If we don't need this logic we should delete it. Also a comment somewhere should explain _why_ we want to opt-out of original src for the new usecase

src/proxy/inbound.rs
```diff
@@ -1018,7 +1023,7 @@ mod tests {
             };
             Some(GatewayAddress {
                 destination: Destination::Address(NetworkAddress {
-                    network: strng::EMPTY,
+                    network: strng::new(""),
```
stevenctl, line 1026: Also confused why this change is needed

src/proxyfactory.rs
```diff
@@ -130,13 +129,57 @@ impl ProxyFactory {
...
+                false,
             );
             result.connection_manager = Some(cm);
             result.proxy = Some(Proxy::from_inputs(pi, drain).await?);
         }
 
         Ok(result)
     }
+
+    pub async fn create_ztunnel_inbound_listener(
```
stevenctl, line 141: Can we add a comment here explaining that this is a listener that proxies traffic for the ztunnels own servers and not things that the ztunnel is proxying on behalf of? The "ztunnel inbound listener" seems a little overloaded/non-specific. Same with the logs. Maybe something like `ztunnel management listener` or something

src/test_helpers/linux.rs
```diff
@@ -124,38 +124,90 @@ impl WorkloadManager {
...
+        let proxy_mode = match current_mode {
+            Shared => ProxyMode::Shared,
+            Dedicated => ProxyMode::Dedicated,
         };
         let (mut tx_cfg, rx_cfg) = mpsc_ack(1);
         tx_cfg.send(initial_config).await?;
         let local_xds_config = Some(ConfigSource::Dynamic(Arc::new(Mutex::new(rx_cfg))));
+
+        // These are ONLY populated for Shared mode runtime config
+        let (ztunnel_identity_config, ztunnel_workload_config) = match proxy_mode {
```
stevenctl, line 182: nit: can we dedupe this with the above?

stevenctl (commented):
Mostly LGTM just left some nits

src/proxy/inbound.rs
ilrudie: Our CNI plugin configures captured pods so that ztunnel may freebind in their netns. It doesn't configure capture for the ztunnel pods though. For ztunnel itself, freebinds are blocked blocked by the kernel. The connection just times out if we try to freebind_connect to ourself.

src/app.rs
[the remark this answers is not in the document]
stevenctl: looks like this still needs to be cleaned up

src/proxy/inbound.rs
```diff
@@ -466,7 +475,7 @@ impl Inbound {
             return Ok(());
         };
         if conn.dst.ip() == hbone_addr.ip() {
-            // Normal case: both are aligned. This is allowed (we really only need the HBONE address for the port.)
+            // Normal case: both are aligned. This is allowed
```
stevenctl, line 478:
```suggestion
            // Normal case: both are aligned. This is allowed (we really only need the HBONE address for the port.)
```

src/test_helpers/linux.rs
```diff
@@ -512,16 +582,18 @@ impl<'a> TestWorkloadBuilder<'a> {
...
         let node = self.w.workload.node.clone();
-        let network_namespace = if self.manager.mode == Dedicated && zt.is_some() {
-            // This is a bit of hack. For dedicated mode, we run the app and ztunnel in the same namespace
-            // We probably should express this more natively in the framework, but for now we just detect it
-            // and re-use the namespace.
-            tracing::info!("node already has ztunnel and dedicate mode, sharing");
-            zt.as_ref().unwrap().namespace.clone()
-        } else {
-            self.manager
+        let network_namespace = match (self.manager.mode, zt.is_some()) {
```
stevenctl, line 585: i'm not sure changing this to a match is necessary or makes the code clearer

stevenctl (commented):
LGTM, just some very small ones left

src/test_helpers/linux.rs
harsh-ps-2003: A match is just more idiomatic Rust.

stevenctl (approved):

PR #1510, opened 2025-04-07
Better support for root certificates

howardjohn:
* Split the last element of the returned response from the CA. This
  allows support for multiple roots. Note this requires a change to
Istio (which I am making) to rever the regression in 1.25.1
(https://github.com/istio/istio/issues/55793).
* Split root from chain in the config_dump. Without this, it is
  impossible to know if something is a root or intermediate.


ilrudie (commented):
The changes look reasonable but it seems like we should have some test coverage for the new multi-root cert parsing.

src/tls/certificate.rs
```diff
@@ -49,12 +50,14 @@ pub struct Expiration {
...
+    /// chain is the entire trust chain, excluding the leaf and root
     pub chain: Vec<Certificate>,
-    pub private_key: PrivateKeyDer<'static>,
+    pub(in crate::tls) private_key: PrivateKeyDer<'static>,
 
-    /// precomputed roots
-    pub roots: Arc<RootCertStore>,
+    /// precomputed roots. This is used for verification
+    root_store: Arc<RootCertStore>,
+    /// original roots, used or debugging
```
Stevenjin8, line 59:
```suggestion
    /// original roots, used for debugging
```

src/tls/certificate.rs
```diff
@@ -218,32 +240,57 @@ fn parse_key(mut key: &[u8]) -> Result<PrivateKeyDer<'static>, Error> {
...
+            .iter()
             .map(|x| x.to_vec())
             .map(parse_cert)
             .collect::<Result<Vec<_>, _>>()?;
         let key: PrivateKeyDer = parse_key(key)?;
 
-        let mut roots = RootCertStore::empty();
-        roots.add_parsable_certificates(chain.iter().last().map(|c| c.der.clone()));
+        let mut roots_store = RootCertStore::empty();
+        roots_store.add_parsable_certificates(roots.iter().map(|c| c.der.clone()));
```
Stevenjin8, line 261: Should we check the return status of `add_parsable_certificates`? Feels like we should at least log an error if we find unparseable ones.

https://docs.rs/rustls/latest/rustls/struct.RootCertStore.html#method.add_parsable_certificates
> Returns the number of certificates added, and the number that were ignored.

howardjohn:
@ilrudie added tests and addressed comments

ilrudie (approved):
LGTM

mikemorris:
@howardjohn Is the fix in https://github.com/istio/istio/pull/55835 complete (resolving the regression) without these cherrypicks - is this an additive improvement or part of the fix?

PR #1513, opened 2025-04-08
chore: fix make failed on mac

zirain:


zirain:
error message:
```console
docker: Error response from daemon: Mounts denied:
The path /fake/path/does/not/exist is not shared from the host and is not known to Docker.
You can configure shared paths from Docker -> Preferences... -> Resources -> File Sharing.
See https://docs.docker.com/desktop/settings/mac/#file-sharing for more info.
make: *** [Makefile:44: build] Error 125
```

howardjohn (approved):

Issue #1531, opened 2025-04-18
istio_build metric doesn't have build tag value


namm2:
I've install ztunnel v1.25.2 and the `istio_build` metric doesn't have correct `tag` label:

> {
  __name__="istio_build",
  app="ztunnel",
  app_kubernetes_io_instance="ztunnel",
  app_kubernetes_io_managed_by="Helm",
  app_kubernetes_io_name="ztunnel",
  app_kubernetes_io_part_of="istio",
  app_kubernetes_io_version="1.25.2",
  component="ztunnel",
  controller_revision_hash="74c4cd5f7b",
  helm_sh_chart="ztunnel-1.25.2",
...
  tag="unknown"
}

I noticed this because the built-in Istio dashboards are using this label and ztunnel pods show "unknown" version.

PR #1532, opened 2025-04-18
fix istio_build metric

zirain:
fixes: https://github.com/istio/ztunnel/issues/1531

xref: https://github.com/istio/istio/blob/186a232f97549f766d0f852408a9982ff0376d38/pilot/docker/Dockerfile.ztunnel#L27


ilrudie:
Thanks! We should cherry-pick to 1.26 for sure. Should we also try to fix in 1.25 and 1.24?

ilrudie (approved):
LGTM, Thanks!

PR #1536, opened 2025-04-21
rbac: add counters to config dump

howardjohn:
```
rbac                    time:   [16.109 ms 16.223 ms 16.345 ms]
                        change: [-7.9751% -1.4256% +3.8734%] (p = 0.71 > 0.05)
                        No change in performance detected.
```

No perf impact as expected

This gives us 
```
NAMESPACE    POLICY NAME                   ACTION SCOPE            HITS MISSES
default      httpbin                       Allow  Namespace        7    3
istio-system istio_converted_static_strict Deny   WorkloadSelector 2    10
```

### DIFF
--- a/src/rbac.rs
+++ b/src/rbac.rs
@@ -17,6 +17,9 @@ use ipnet::IpNet;
 use std::fmt;
 use std::fmt::{Display, Formatter};
 use std::net::SocketAddr;
+use std::sync::Arc;
+use std::sync::atomic::AtomicU64;
+use std::sync::atomic::Ordering::Relaxed;
 use tracing::{instrument, trace};
 use xds::istio::security::Address as XdsAddress;
 use xds::istio::security::Authorization as XdsRbac;
@@ -31,7 +34,7 @@ use crate::state::workload::{WorkloadError, byte_to_ip};
 use crate::strng::Strng;
 use crate::{strng, xds};
 
-#[derive(Debug, Hash, Eq, PartialEq, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 #[serde(rename_all = "camelCase", deny_unknown_fields)]
 pub struct Authorization {
     pub name: Strng,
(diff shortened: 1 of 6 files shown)


src/rbac.rs
```diff
@@ -31,14 +34,32 @@ use crate::state::workload::{WorkloadError, byte_to_ip};
...
+    #[serde(default)]
+    pub stats: Arc<AuthzCounter>,
+}
+
+#[derive(Debug, Default, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
+pub struct AuthzCounter {
+    hits: AtomicU64,
+    misses: AtomicU64,
+}
```
ilrudie, line 54: hit and miss are relative to the kind of policy. hit a deny == denied. hit an allow == allowed. Would users be better served with "allows" and "denies" which inform them of a concrete result taken in association with this policy?

howardjohn: I worry an allow policy with `deny: 123` looks alarming. Its not really a deny, its more like it failed to allow

Stevenjin8: So if a policy doesn't match, `miss`/`deny` will be incremented, even if another policy later on allow or denies it?

Stevenjin8: I also wonder if the order of operations might be confusing cause the stats to be confusing. For example, if there are multiple deny policies, but only one of them matches, then the order in which the deny policies are evaluated will affect the stats (assuming there is some short circuiting), which seems weird.

Stevenjin8: Maybe a better say of saying this is, why do we need `miss` when we can just do `total_requests - hit`

howardjohn: there is no total request counter for a policy

Stevenjin8: I mean total request counter across for ztunnel in general

howardjohn: That doesn't exist is the first issue. We could add it but it doesn't help the use case. 
Miss means "policy applies to the request but does not match"

Stevenjin8: Ok then only my second comment then: This doesn't seem to take into account short circuiting. Just because a policy could apply, short circuiting might mean that a counter doesn't get incremented, which might be the intended behavior

PR #1545, opened 2025-04-29
show crypto provider in ztunnel version

Stevenjin8:
It would be nice if ztunnel showed the crypto provider in version.

Envoy does a similar thing, where the output looks like

```
envoy  version: 739644f84930a8c0d416319aea97f58c2222f7ef/1.32.2-dev/Modified/RELEASE/BoringSSL
```

or

```
envoy  version: 739644f84930a8c0d416319aea97f58c2222f7ef/1.32.2-dev/Modified/RELEASE/BoringSSL-FIPS
```

### DIFF
--- a/src/version.rs
+++ b/src/version.rs
@@ -34,10 +34,16 @@ pub struct BuildInfo {
     build_status: String,
     git_tag: String,
     pub istio_version: String,
+    crypto_provider: String,
 }
 
 impl BuildInfo {
     pub fn new() -> Self {
+        #[cfg(feature = "tls-ring")]
+        let crypto_provider = "tls-ring".to_string();
+        #[cfg(feature = "tls-boring")]
+        let crypto_provider = "tls-boring".to_string();
+
         BuildInfo {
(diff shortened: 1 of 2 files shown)


src/version.rs
```diff
@@ -34,10 +34,20 @@ pub struct BuildInfo {
...
     git_tag: String,
     pub istio_version: String,
+    crypto_provider: String,
 }
 
 impl BuildInfo {
     pub fn new() -> Self {
+        #[cfg(feature = "tls-aws-lc")]
+        let crypto_provider = "tls-aws-lc".to_string();
+        #[cfg(feature = "tls-ring")]
```
ilrudie, line 44: For organization, would it make sense to have the provider be static in tls/lib.rs rather than defined here? Like TLS_VERSIONS

https://github.com/istio/ztunnel/blob/903cf079de4a81925fa1debcb1c48ea20f364169/src/tls/lib.rs#L41

ilrudie (approved):
LGTM, Thanks

PR #1561, opened 2025-05-16
tls: add PQC compliance policy

### LINKED ISSUE #1560: Feature request: PQC for mesh-internal traffic
This feature was described in detail here: https://github.com/istio/istio/issues/56330.

Status of the support for `X25519MLKEM768` in ztunnel crypto providers:
1. `aws-lc-rs` - has support.
2. `ring` - no support.
3. `boring` - no support, and will not support, because ML-KEM would break FIPS-compliance.
4. `openssl` - not yet, but it's already possible with [rustls-liboqs](https://github.com/tofay/rustls-liboqs).

jewertow:
Fixes #1560

### DIFF
--- a/src/tls/lib.rs
+++ b/src/tls/lib.rs
@@ -68,14 +68,24 @@ pub(super) fn provider() -> Arc<CryptoProvider> {
 
 #[cfg(feature = "tls-aws-lc")]
 pub(super) fn provider() -> Arc<CryptoProvider> {
-    Arc::new(CryptoProvider {
+    let mut provider = CryptoProvider {
         // Limit to only the subset of ciphers that are FIPS compatible
         cipher_suites: vec![
             rustls::crypto::aws_lc_rs::cipher_suite::TLS13_AES_256_GCM_SHA384,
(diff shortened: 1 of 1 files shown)


src/tls/lib.rs
```diff
@@ -68,14 +68,24 @@ pub(super) fn provider() -> Arc<CryptoProvider> {
...
         // Limit to only the subset of ciphers that are FIPS compatible
         cipher_suites: vec![
             rustls::crypto::aws_lc_rs::cipher_suite::TLS13_AES_256_GCM_SHA384,
             rustls::crypto::aws_lc_rs::cipher_suite::TLS13_AES_128_GCM_SHA256,
         ],
         ..rustls::crypto::aws_lc_rs::default_provider()
-    })
+    };
+
+    let kx_groups = match std::env::var("COMPLIANCE_POLICY").as_deref() {
```
howardjohn: In every other part of the codebase we do not read raw env vars and instead pass them into config::Config.

For this function Ican see how passing the config around may be tedious. So if we are going to do it can we at least read it once and store it in a global var (once_cell::sync::Lazy can be used)?

jewertow: Done

howardjohn (approved):

PR #1562, opened 2025-05-19
WIP: initial idea for improved resolution

ilrudie:
TODO:

- [ ] Tests
- [ ] Read rootNamespace or parse from XDS address as a fallback instead of just hardcoding `istio-system`


ilrudie:
WDS should probably be updated to include the creation-time of the resource which defines the service. ztunnel can then sort services so that or_first() will be a stable return of the oldest definition of the hostname.

Likely, sorted insert should be performed rather than sorting during name resolution.

ilrudie:
not stale

howardjohn (approved):

Issue #1555, opened 2025-05-12
multicluster: ztunnel load balancing across waypoints based on combo of endpoint count and waypoint count


keithmattix:

keithmattix:
@krinkinmu will take a stab at this

krinkinmu:
@keithmattix just to make sure I'm understanding correctly the task here let me restate the request as I undesrtand it.

Background: For multiucluster scenario in ambient mode the client ztunnel will have to first make a decision whether to route the request locally (and therefore, when L7 is involved, to waypoints in the local cluster) or remotely (via E/W gateway, which in turn will send to remote waypoints if necessary).

That's based on https://docs.google.com/document/d/1Wg6sx9ZUJL4AsHj5wM1kMx3E436s5wg2qoMqoI-bqbQ/edit?usp=sharing), specifically once traffic reaches waypoint it stays within the cluster of that waypoint, therefore we need to decide which cluster to send data to, before we pick waypoint or E/W gateway.

With that background in mind, my understanding of the task here is to implement the load balancing in ztunnel that will look at how many backends we have in local and remote clusters, use that number to affect the weight of the clusters and based on that (and any other input that controls the load balancing policy) change load balancing algorthim to pick the cluster to send traffic to.

I'm not entirely sure  where the waypoint count comes into play here, but I'd imagine we want load balancing to distribute the load between waypoints sensibly enough to avoid a single waypoint getting all the load (and within the limits of what load balancing policy allows).

keithmattix:
@krinkinmu That's generally correct; where I envision the waypoint count coming in is as a part of the capacity heuristic. Consider a crazy scenario:
```
1M QPS
cluster1: 1 waypoint, 1k backends
clusterr2: 1000 waypoints, 1 backend
```

Now people absolutely should never do this, but the general takeaway is that there may be only 1 waypoint serving a bunch of backends or a bunch of waypoints serving 1 backend, so we don't necessarily want to lean on _just_ backend endpoint count or _just_ waypoint count. We probably want both to factor in some way

krinkinmu:
Here is what I'm trying to try and implement this week and play around with it.

In the [build_request](https://github.com/istio/ztunnel/blob/master/src/proxy/outbound.rs#L374) method there is a short circuit logic that goes as follows:
[...]

PR #1565, opened 2025-05-27
Multinetwork/Support remote networks for services with waypoints

krinkinmu:
Currently `build_request` when it sees a service with a waypoint resolves the waypoint backend and routes request there using regular HBONE.

In multi network scenario though the waypoint may have workload on a remote network and to reach it we have to go through E/W gateway and use double HBONE.

This change enables handling of services with waypoint on a remote network.

Some of the assumptions that were used when I prepared this change:

1. We assume uniformity of configuration (e.g., if service X in local cluster has a waypoint, then service X in remote network also has a waypoint, if waypoint is service addressable, then it's using service to address waypoint both locally and on remote network) 
2. Split-horizon representation of waypoint workloads, just like with
   any regular workloads and services (e.g., in the local cluster
   instead of an actual waypoint workload pointing to a pod on another
   network we will have a "proxy" representation that just has network
   gateway).

Both of those can be in hanled by the controlplane (e.g., controlplane can generate split-horizon workloads and when configuration is non-uniform, just filter out remote configs for remote networks), though we don't yet have a complete implementation.

Tracking issue #1555


keithmattix:
We should create a follow-up issue to address this scenario if the config isn't uniform (for beta). Such a scenario will probably require changes to WDS

howardjohn (commented):
So is the net effect in traffic just the scenario: local service has "use waypoint" but no local waypoint pods and remote has waypoint pods now sends remote?

Since it seems like most scenarios will behave the same, seeing as waypoints always have "prefer close" so they will never pick a remote waypoint if there is a local one?

krinkinmu:
> So is the net effect in traffic just the scenario: local service has "use waypoint" but no local waypoint pods and remote has waypoint pods now sends remote?
> 
> Since it seems like most scenarios will behave the same, seeing as waypoints always have "prefer close" so they will never pick a remote waypoint if there is a local one?

Pretty much, if there is a healthy local waypoint it will be picked unless the user overrides it.

We discussed it internally with @keithmattix and @Stevenjin8 and basically for the initial version the scenario we want to support is failover and not load-balancing between clusters. And current behavior kind of makes sense from this point of view (aside from not using double HBONE).

The reason we didn't go for more is because with L7 policies we just can't know for sure where the requests are going to get routed and because of that it's not easy to consider capacity accurately. We might come up with some heuristic to overcome this, but it's not clear how it would look like and if we want to rely on a heuristic for load balancing between networks.


A side note, unrelated to the above, there is another change in behavior that this PR introduces - it changes hbone destination for double hbone to use the original target port instead of translating it. The desirable behavior for E/W gateway is similar to what waypoint does - we give it the original target we tried to connect, and it decides where to route the request exactly.

keithmattix:
@howardjohn @ilrudie Can we get some eyes on this? It's pretty small and has been sitting for a week

src/proxy/outbound.rs
```diff
@@ -446,37 +532,16 @@ impl OutboundConnection {
...
-                // Do not try to send cross-network traffic without network gateway.
-                return Err(Error::NoValidDestination(Box::new((*us.workload).clone())));
-            }
+            return self
+                .build_request_through_gateway(
+                    source_workload.clone(),
+                    us,
+                    service
+                        .as_ref()
+                        .expect("Workloads with network gateways must be service addressed."),
```
ilrudie: It seems like we should return an error rather than panic.

src/proxy/outbound.rs
```diff
@@ -369,6 +369,81 @@ impl OutboundConnection {
...
+        upstream: Upstream,
+        // This is a target service we wanted to reach in the first place.
+        //
+        // NOTE: Crossing network boundaries is only supported for services
+        // at the moment, so we should always have a service we could use.
+        service: &Service,
+        target: SocketAddr,
+    ) -> Result<Request, Error> {
+        // This function is called when the next hop we selected is on a
+        // different network, so we expect the upstream workload to have
```
ilrudie: If this is intended to describe what the function does it should precede the function in our codebase.

src/proxy/outbound.rs
```diff
@@ -369,6 +369,81 @@ impl OutboundConnection {
...
+                // The outer tunnel of double HBONE is terminated by the E/W
+                // gateway and so for the credentials of the next hop
+                // (upstream_sans) we use gateway credentials.
+                //
+                // The inner HBONE tunnel is terminated by either the server
+                // we want to reach or a Waypoint in front of it, depending on
+                // the configuration. So for the final destination credentials
+                // (final_sans) we use the upstream workload credentials.
+                upstream_sans: gateway_upstream.workload_and_services_san(),
+                final_sans: upstream.service_sans(),
```
ilrudie: **nit:** comment about a field should be with the field
```suggestion
                // The outer tunnel of double HBONE is terminated by the E/W
                // gateway and so for the credentials of the next hop
                // (upstream_sans) we use gateway credentials.
                upstream_sans: gateway_upstream.workload_and_services_san(),
                // The inner HBONE tunnel is terminated by either the server
                // we want to reach or a Waypoint in front of it, depending on
                // the configuration. So for the final destination credentials
                // (final_sans) we use the upstream workload credentials.
                final_sans: upstream.service_sans(),
```

src/proxy/outbound.rs
krinkinmu: I don't have a strong opinion and updated the code to return an error.

I could not really think of a case where this error would ever happen other than map lookups starting to return completely wrong keys. Here is how I think about it:

1. This error triggers when the target address refers to a workload instead of a service
2. This workload has to be on a remote network, otherwise we will not execute this code
3. If it's on a remote network, then find_upstream will never find it, because it does not match the local network used in all lookups.

I don't think that there is any harm in returning an error for a condition that should never trigger, but I also don't think that panicking here is bad either.

src/proxy/outbound.rs
krinkinmu: Reworked the comments a bit, PTAL.

src/proxy/outbound.rs
ilrudie: https://doc.rust-lang.org/std/macro.debug_assert.html seems like a good middle ground here. We can panic in non-optimized builds to find bugs but have a more resilient optimized runtime

krinkinmu: Makes sense, I added the debug_assert in addition to returning the error. It should make it easier to diagnose the test failures if we ever hit this issue.

src/proxy/outbound.rs
```diff
@@ -393,6 +469,17 @@ impl OutboundConnection {
                 .fetch_service_waypoint(&target_service, &source_workload, target)
                 .await?
             {
+                if waypoint.workload.network != source_workload.network {
+                    return self
```
ilrudie: I think it would be good to have a `debug!` here noting we're building a request to a waypoint on another network.

src/proxy/outbound.rs
```diff
@@ -446,37 +533,25 @@ impl OutboundConnection {
...
+            // This situation "should never happen" because for workloads fetch_upstream
+            // above only checks the workloads on the same network as this ztunnel
+            // instance and therefore it should not be able to find a workload on a
+            // different network.
+            debug_assert!(
+                service.is_some(),
+                "workload on remote network is not service addressed"
+            );
+            let service = service.as_ref().ok_or(Error::NoService(target))?;
+            return self
```
ilrudie, line 555: `debug!` message would likely be helpful here so we can know we tried to send to another network.

ilrudie (commented):
mostly LGTM, I think we should toss in a handful of debug logs just to help clarify what happened when we are troubleshooting.

src/proxy/outbound.rs
krinkinmu: Added a few debug entries to note when we pick a waypoint or an upstream workload on a remote network and also when we successfully build a request through an E/W gateway. With this we should be able to tell if we are trying to talk to a remote destination and if something happens that prevented us from successfully creating a request to E/W gateway.

src/proxy/outbound.rs
krinkinmu: See above.

Issue #1537, opened 2025-04-22
Dynamic thread count


howardjohn:
Today, we can pick a fixed thread count. This is problematic if you have mixed nodes, say some 2 node  with some 128 core nodes. 

We should introduce a ratio, so you can say something like '0.25 threads per core'.

PR #1566, opened 2025-06-03
Allow dynamic configuration of thread count

howardjohn:
Fixes https://github.com/istio/ztunnel/issues/1537

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -266,7 +266,8 @@ fn new_data_plane_pool(num_worker_threads: usize) -> mpsc::Sender<DataPlaneTask>
             .thread_name_fn(|| {
                 static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);
                 let id = ATOMIC_ID.fetch_add(1, Ordering::SeqCst);
-                format!("ztunnel-proxy-{id}")
+                // Thread name can only be 16 chars so keep it short
+                format!("ztunnel-{id}")
             })
             .enable_all()
             .build()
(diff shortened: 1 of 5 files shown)


src/app.rs
```diff
@@ -266,7 +266,7 @@ fn new_data_plane_pool(num_worker_threads: usize) -> mpsc::Sender<DataPlaneTask>
             .thread_name_fn(|| {
                 static ATOMIC_ID: AtomicUsize = AtomicUsize::new(0);
                 let id = ATOMIC_ID.fetch_add(1, Ordering::SeqCst);
-                format!("ztunnel-proxy-{id}")
+                format!("ztunnel-{id}")
```
howardjohn, line 270: thread name has a limit so this caused any 2-digit numbers to be truncated

keithmattix: nit: consider adding this as a code comment and document the thread name char limit

howardjohn: thanks, done

src/config.rs
```diff
@@ -1094,4 +1148,61 @@ pub mod tests {
             assert!(metadata.vec.contains(&(key, value)));
         }
     }
+
+    #[test]
+    fn test_parse_worker_threads() {
```
ilrudie, line 1153: Will we flake if `test_parse_worker_threads` and `test_get_cpu_count` run in parallel since they are both setting/removing the same env vars and making assertions based on the contents?

ilrudie (commented):
mostly LGTM, a little concerned because it seems like the new tests can't be run in parallel safely though

src/config.rs
howardjohn: yes it does. will need to fix.

We should probably have some form of 'env var provider' abstraction

ilrudie: Short term we could probably combine them into a single test to work around it. Little wonky but probably ok

howardjohn: good idea done

ilrudie (approved):
LGTM aside from the extra whitespace tests are complaining about.

PR #1570, opened 2025-06-05
don't send to empty address

stevenctl:
https://github.com/istio/istio/issues/56531

If we have a Workload with no addresses, we should only send to it if it's a cross-network workload and we have a gateway.

This is the behavior we already get for sidecar and waypoint.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -1584,6 +1584,22 @@ mod tests {
             },
             ..test_helpers::test_default_workload()
         };
+        let wl_empty_ip = Workload {
+            uid: "cluster1//v1/Pod/default/wl_empty_ip".into(),
+            name: "wl_empty_ip".into(),
+            namespace: "default".into(),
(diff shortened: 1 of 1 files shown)


src/state.rs
```diff
@@ -1708,5 +1752,12 @@ mod tests {
             vec!["cluster1//v1/Pod/default/wl_match"],
             "failover full match selects closest match",
         );
+        assert_not_endpoint(
+            &wl_no_locality,
```
ilrudie, line 1756:
```suggestion
            &wl_empty_ip,
```

ilrudie (commented):
mostly lgtm, typo in the test assertion though I think

src/state.rs
stevenctl: This is testing sending from `wl_no_locality` as the src

ilrudie: oh, right

PR #1580, opened 2025-06-20
Steven/buffer inner h2 streams

Stevenjin8:
`TokioH2Stream` is a wrapper to implement `tokio::io::AsyncRead` for `H2StreamReadHalf`.
I originally implemented `tokio::io::AsyncRead::poll_read(self, buf)` for `TokioH2Stream` by calling `H2StreamReadHalf::poll_bytes` and writing the result into `buf` and errored if `buf` wasn't big enough. (Note that we cannot control how many bytes  `H2StreamReadHalf::poll_bytes` returns).

I thought `buf` would be pretty big. Turns out its usually 4096 bytes. This is set internally by tokio-rustls and is not configurable. A Server Hello + Certificate Request + Encrypted Extensions can easily reach 5000+ bytes causing an error. Not to mention that a large or jumbo frame will easily surpass that. This implements some buffering for `TokioH2Stream` to make sure that we never error because `buf` is not big enough.


src/proxy/h2.rs
```diff
@@ -154,18 +160,26 @@ impl tokio::io::AsyncRead for TokioH2Stream {
         cx: &mut Context<'_>,
         buf: &mut tokio::io::ReadBuf<'_>,
     ) -> Poll<std::io::Result<()>> {
-        let pinned = std::pin::Pin::new(&mut self.0.read);
+        // Just return the bytes we have left over and don't poll the stream because
```
howardjohn, line 163: I think this can be simplified (but give equivilent) like so:

```rust
        let this = &mut self;
        if this.buf.is_empty() {
            let res = ready!(Pin::new(&mut this.stream.read).poll_bytes(cx))?;
            this.buf = res;
        }
        let cnt = std::cmp::min(this.buf.len(), read_buf.remaining());
        read_buf.put_slice(&this.buf[..cnt]);
        this.buf = this.buf.split_off(cnt);
        Poll::Ready(Ok(()))
```

Stevenjin8: Mostly  agree. What's the point of `let this = &mut self;`

howardjohn: Probably not needed and was just copy and pasted from others and not cleaned up. Its common to need to do things _like_ that for projections

src/proxy/pool.rs
```diff
@@ -612,13 +613,28 @@ mod test {
...
         c.write_all(b"abcde").await.unwrap();
-        let mut b = [0u8; 0];
-        // Crucially, this should error rather than panic.
-        if let Err(e) = c.read(&mut b).await {
-            assert_eq!(e.kind(), io::ErrorKind::Other);
-        } else {
-            panic!("Should have errored");
-        }
+        let mut b = [0u8; 100];
+        // Properly buffer reads and don't error
```
howardjohn, line 617: `WeirdIO` may be useful here

howardjohn (approved):

PR #1581, opened 2025-06-23
Double hbone/use intended port

Stevenjin8:
Bug in double hbone implementation. The authority in double hbone should have the service port (intended port), not the destination port. That is, we do not consider port mappings in the src ztunnel.


ilrudie:
https://github.com/istio/ztunnel/pull/1565 also has this I think

Stevenjin8:
@ilrudie yep, just saw it after trying to merge... should update the test for it, but will put in draft for now

Issue #1585, opened 2025-07-03
ZTunnel runs out of File Descriptors on heavily loaded nodes


oddur:
When running ztunnel on nodes that are heavily loaded, it runs out of file descriptors and randomly starts dropping connections. 

The logs show errors like 

```
b57f79f-46cr9}	TLS handshake error: Too many open files (os error 24)
2025-07-02T15:25:52.617977Z	error	proxy::outbound:proxy{wl=datadog/datadog-agent-lxh24}	Failed TCP handshake Too many open files (os error 24)
```

This is on a GKE node with 110 pods(  currently the default max pods per node out of the box on gke ) 

We worked around this by manually patching the Daemonset of the ztunnel to run 

```
     Args:
      ulimit -n 65536 && exec /usr/local/bin/ztunnel proxy ztunnel
```

To up the soft limit of file handles to 65536 (an arbitrary value ). This resolves the issue for us, but is not a permanent solution, since it involves patching the CRD manually. 

I want to be able to have the ztunnel set its own soft limit at startup. 

Taking inspiration from [here](https://github.com/istio/istio/pull/56507/files#diff-192c798ff0002ea4b84c1587abde1169fdd7cbce8de60a004b109bfc6bd592adL22-L33)  — it could set it's own softlimit to the maximum allowed value on startup. 

```
func raiseFileLimits() (uint64, error) {
	rlimit := unix.Rlimit{}
	if err := unix.Getrlimit(unix.RLIMIT_NOFILE, &rlimit); err != nil {
		return 0, err
	}
	rlimit.Cur = rlimit.Max
	if err := unix.Setrlimit(unix.RLIMIT_NOFILE, &rlimit); err != nil {
		return 0, err
	}

	return rlimit.Cur, nil
}

```

This would allow ztunnel to run on highly loaded nodes out of the box.   Any other solution we should consider ?

Relevant Slack Conversation: https://istio.slack.com/archives/C041EQL1XMY/p1751474124761639

keithmattix:
Looks like there was only a 1.27 cherry-pick

PR #1586, opened 2025-07-03
increasing limit for open files

grnmeira:
This PR helps checking and increasing the max number of opened files (rlimit). This should help with #1585

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -31,7 +31,8 @@ pub static malloc_conf: &[u8] = b"prof:true,prof_active:true,lg_prof_sample:19\0
 
 // We use this on Unix systems to increase the number of open file descriptors
 // if possible. This is useful for high-load scenarios where the default limit
-// is too low, which can lead to droopped connections and other issues (#1585).
+// is too low, which can lead to droopped connections and other issues:
+// see: https://github.com/istio/ztunnel/issues/1585
 fn increase_open_files_limit() {
     #[cfg(unix)]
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -43,6 +44,19 @@ fn main() -> anyhow::Result<()> {
         }
     };
 
+    if let Ok((soft_limit, hard_limit)) = getrlimit(Resource::RLIMIT_NOFILE) {
```
howardjohn: Thanks! can we put this into a split out function, guard behind `#[cfg(unix)]` and call it under the `proxy` function?

howardjohn: And add a comment about why we do this

plucked:
Thank you for jumping on this issue so quickly. 🤟🏻

src/main.rs
```diff
@@ -28,6 +29,25 @@ static ALLOC: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 #[unsafe(export_name = "malloc_conf")]
 pub static malloc_conf: &[u8] = b"prof:true,prof_active:true,lg_prof_sample:19\0";
 
+// We use this on Unix systems to increase the number of open file descriptors
+// if possible. This is useful for high-load scenarios where the default limit
+// is too low, which can lead to droopped connections and other issues (#1585).
```
dhawton:
```suggestion
// is too low, which can lead to droopped connections and other issues:
// see: https://github.com/istio/ztunnel/issues/1585
```

Best to be specific

PR #1594, opened 2025-07-08
build: support compilation on Windows using MSVC

yskopets:
## Summary

* add support for `ztunnel` compilation on Windows using MSVC
* fix compilation errors in main sources
* statically link `ztunnel.exe`


keithmattix:
Are you aware of #1461? We're using mingw and gnu targets, but probably better to keep all of the windows stuff togetehr

PR #1601, opened 2025-07-24
respect IPV6 setting for DNS server

stevenctl:
Currently the IPV6_ENABLED flag only affects zTunnel -> upstream DNS instead of making it so we don't include IPv6 addresses in our internal DNS server's responses. This patch respects that flag in our DNS server.


howardjohn (approved):

keithmattix:
Do we want this in 1.27

stevenctl:
yes, thanks for the reminder

PR #1608, opened 2025-08-07
Retrying compartmentless containers

grnmeira:
### Problem
We have an issue with containerd and Windows' HCS where namespaces for pods don't yet have a compartment ID when the CNI signals ztunnel for a new workload that's just been created. Without that ID we can't create ztunnel proxy sockets inside the pod's network compartment. The compartment ID will only be available after ztunnel signals the CNI that the workload has been assimilated by ztunnel (which can't happen without a compartment ID 🫤).
### What this PR does
This PR mitigates the stated problem with ztunnel replying an ACK to the CNI during the ADD operation, even though the workload proxies haven't been yet instantiated inside ztunnel. After a timeout, ztunnel tries again to add the workflow, and now the HCS API returns a valid compartment ID, which allows the creation of a proxy without any problems.

A more detailed flow looks like:
1. CNI sends an ADD command to ztunnel
2. ztunnel queries HCS and checks if a comapartment ID is available for that pod
2.1 If the compartment ID is available, we create a proxy for the workload and ACK the CNI
2.2 If the compartment ID is not available, we mark the workload as pending inside ztunnel and ACK the CNI.
3. The CNI receives an ACK and the pod creation proceeds.
4. After a timeout, ztunnel tries to add the pending workload again.
### How the PR does it
It introduces a queue of events that runs in the same thread where ZDS commands are processed. These events are managed in a quite simple way at the moment, but can be expanded in the future for more advanced handling of future "internal events", including the current retries of pending workloads.
### Caveats
We'll keep retrying compartmentless workloads even if they're failing in an unretriable way (to be fixed in a different PR). There's no current way to signal CNI after sending an ACK for the ADD command that the workload actually failed to be assimilated by ztunnel.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -112,6 +112,7 @@ tracing-core = "0.1"
 tracing-appender = "0.2"
 tokio-util = { version = "0.7", features = ["io-util"] }
 educe = "0.6"
+uuid = "1.17.0"
 
 [target.'cfg(target_os = "linux")'.dependencies]
 netns-rs = "0.1"
--- a/src/inpod/windows.rs
+++ b/src/inpod/windows.rs
@@ -34,8 +34,7 @@ mod workloadmanager;
 #[cfg(any(test, feature = "testing"))]
 pub mod test_helpers;
 
-
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct WorkloadData {
     workload_uid: WorkloadUid,
     workload_info: Option<WorkloadInfo>,
--- a/src/inpod/windows/admin.rs
+++ b/src/inpod/windows/admin.rs
@@ -24,6 +24,7 @@ use std::sync::RwLock;
 #[derive(Debug, Serialize, Deserialize, Clone, Copy)]
 pub enum State {
     Pending,
+    WaitingCompartment,
     Up,
 }
 
@@ -61,11 +62,7 @@ pub struct WorkloadManagerAdminHandler {
 }
 
 impl WorkloadManagerAdminHandler {
-    pub fn proxy_pending(
-        &self,
-        uid: &crate::inpod::WorkloadUid,
-        workload_info: &WorkloadInfo,
-    ) {
+    pub fn proxy_pending(&self, uid: &crate::inpod::WorkloadUid, workload_info: &WorkloadInfo) {
         let mut state = self.state.write().unwrap();
 
         // don't increment count here, as it is only for up and down. see comment in count.
@@ -78,6 +75,30 @@ impl WorkloadManagerAdminHandler {
                     uid.clone(),
                     ProxyState {
                         state: State::Pending,
+                        connections: None,
+                        count: 0,
+                        info: workload_info.clone(),
+                    },
+                );
+            }
+        }
+    }
+    pub fn proxy_waiting_compartment(
+        &self,
+        uid: &crate::inpod::WorkloadUid,
+        workload_info: &WorkloadInfo,
+    ) {
+        let mut state = self.state.write().unwrap();
+
+        match state.get_mut(uid) {
+            Some(key) => {
+                key.state = State::WaitingCompartment;
+            }
+            None => {
+                state.insert(
+                    uid.clone(),
+                    ProxyState {
+                        state: State::WaitingCompartment,
                         connections: None,
                         count: 0,
                         info: workload_info.clone(),
--- a/src/inpod/windows/config.rs
+++ b/src/inpod/windows/config.rs
@@ -12,9 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-
-
-use crate::inpod::windows::namespace::InpodNamespace;
+use crate::inpod::windows::namespace::NetworkNamespace;
 use crate::proxy::DefaultSocketFactory;
 use crate::{config, socket};
 
@@ -36,14 +34,14 @@ impl InPodConfig {
             ..cfg.socket_config
         };
         Ok(InPodConfig {
-            cur_namespace: InpodNamespace::current()?,
+            cur_namespace: NetworkNamespace::current()?,
             reuse_port: cfg.inpod_port_reuse,
             socket_config,
         })
     }
     pub fn socket_factory(
         &self,
-        netns: InpodNamespace,
+        netns: NetworkNamespace,
     ) -> Box<dyn crate::proxy::SocketFactory + Send + Sync> {
         let base = crate::proxy::DefaultSocketFactory(self.socket_config);
         let sf = InPodSocketFactory::from_cfg(base, self, netns);
@@ -62,14 +60,14 @@ impl InPodConfig {
 
 struct InPodSocketFactory {
     inner: DefaultSocketFactory,
-    netns: InpodNamespace,
+    netns: NetworkNamespace,
 }
 impl InPodSocketFactory {
-    fn from_cfg(inner: DefaultSocketFactory, _: &InPodConfig, netns: InpodNamespace) -> Self {
+    fn from_cfg(inner: DefaultSocketFactory, _: &InPodConfig, netns: NetworkNamespace) -> Self {
         Self::new(inner, netns)
     }
-    fn new(inner: DefaultSocketFactory,netns: InpodNamespace) -> Self {
-        Self {inner, netns }
+    fn new(inner: DefaultSocketFactory, netns: NetworkNamespace) -> Self {
+        Self { inner, netns }
     }
 
     fn run_in_ns<S, F: FnOnce() -> std::io::Result<S>>(&self, f: F) -> std::io::Result<S> {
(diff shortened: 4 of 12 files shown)


src/inpod/windows/statemanager.rs
```diff
@@ -339,6 +362,98 @@ impl WorkloadProxyManagerState {
         }
     }
 
+    pub async fn retry_comparmentless(&mut self, poddata: &WorkloadData) -> Result<(), Error> {
```
MikeZappa87: compartmentless*

src/inpod/windows.rs
```diff
@@ -34,8 +34,7 @@ mod workloadmanager;
 #[cfg(any(test, feature = "testing"))]
 pub mod test_helpers;
 
-
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
keithmattix, line 37: Are we sure we want to clone this?

src/inpod/windows/admin.rs
```diff
@@ -61,23 +62,43 @@ pub struct WorkloadManagerAdminHandler {
 }
 
 impl WorkloadManagerAdminHandler {
-    pub fn proxy_pending(
+    pub fn proxy_pending(&self, uid: &crate::inpod::WorkloadUid, workload_info: &WorkloadInfo) {
+        let mut state = self.state.write().unwrap();
```
keithmattix: Does it make sense to merge these?

src/inpod/windows/config.rs
```diff
@@ -36,14 +34,14 @@ impl InPodConfig {
             ..cfg.socket_config
         };
         Ok(InPodConfig {
-            cur_namespace: InpodNamespace::current()?,
+            cur_namespace: NetworkNamespace::current()?,
```
keithmattix: Why the rename?

MikeZappa87: +1, the current naming makes no distiction between the specific namespace. Does this rename add value?

src/inpod/windows/admin.rs
grnmeira: Good point, they're pretty much the same.

src/inpod/windows/config.rs
grnmeira: I think it was more because I removed one layer there, and this was the original "inner" struct. I'll change it back to `InpodNamespace` 👍

src/inpod/windows.rs
grnmeira: Reason here is just to facilitate persisting this information across requests in our message loop (as now we can retry workloads later). We can make an effort to simplify that struct, or use an Rc<WorkloadData> if you think that's desirable.

keithmattix: Ack, let's optimize later if we feel like we need it. This doesn't affect Linux so I feel good coming  back to it

src/inpod/windows/statemanager.rs
```diff
@@ -339,6 +366,98 @@ impl WorkloadProxyManagerState {
...
+            namespace: info.namespace.clone(),
+            service_account: info.service_account.clone(),
+        };
+
+        if compartment_id == 0 {
+            info!(
+                "network compartment ID not yet available for namespace {}",
+                netns.namespace_guid
+            );
+            self.compartmentless_workloads
```
keithmattix: Do we ever pop this off the list? Could this create duplicates?

src/inpod/windows/statemanager.rs
```diff
@@ -339,6 +366,98 @@ impl WorkloadProxyManagerState {
...
+                .push((uid.clone(), info.clone(), netns.clone()));
+        }
+
+        debug!(
+            "compartment id {} found for network namespace {}",
+            compartment_id, netns.namespace_guid
+        );
+        let new_netns = InpodNamespace::new(netns.namespace_guid.clone()).map_err(|e| {
+            self.compartmentless_workloads
```
keithmattix: Why potentially do this again?

src/inpod/windows/statemanager.rs
```diff
@@ -339,6 +366,98 @@ impl WorkloadProxyManagerState {
...
+                .push((uid.clone(), info.clone(), netns.clone()));
+            Error::NamespaceError(format!("unable to create new network namespace: {}", e))
+        })?;
+        match self.add_workload(&uid, info.clone(), new_netns).await {
+            Ok(()) => {}
+            Err(e) => {
+                self.compartmentless_workloads
```
keithmattix: I might be missing something, but isn't there a possibility that we add this guid to the list 3 times?

MikeZappa87: For compartmentless_workloads? Its a Vec<(WorkloadUid, WorkloadInfo, InpodNamespace)> it has nothing would prevent it from having duplicate entries.

grnmeira: That vector shouldn't be there anymore. It should've been removed after the queue of events was created.

grnmeira:
pinging @Stevenjin8 and @howardjohn for a closer look

Cargo.toml
```diff
@@ -112,6 +112,7 @@ tracing-core = "0.1"
 tracing-appender = "0.2"
 tokio-util = { version = "0.7", features = ["io-util"] }
 educe = "0.6"
+uuid = "1.17.0"
```
Stevenjin8: this is windows only right?

grnmeira: Actually I think we don't need that anymore, as we're using windows::core::UUID now.

src/inpod/windows/workloadmanager.rs
```diff
@@ -272,6 +268,65 @@ impl<'a> WorkloadProxyManagerProcessor<'a> {
...
+
+    fn enqueue_local_event(&mut self, delay: Duration, event_type: EventType) {
+        let expiration = tokio::time::Instant::now() + delay;
+        let new_event = InternalEvent {
+            event_type,
+            expiration,
+        };
+        match self
+            .event_queue
+            .binary_search_by_key(&expiration, |event| event.expiration)
```
Stevenjin8: we can use `partition_point` instead: https://doc.rust-lang.org/std/primitive.slice.html#method.partition_point

src/inpod/windows/workloadmanager.rs
```diff
@@ -272,6 +268,65 @@ impl<'a> WorkloadProxyManagerProcessor<'a> {
...
+                continue;
+            }
+            match &event.event_type {
+                EventType::RetryAllPendingWorkloads => self.retry_proxies().await,
+                EventType::RetryWorkload(previous_timeout, poddata) => {
+                    if let Err(e) = self.state.retry_compartmentless(poddata).await {
+                        // We can't tell the CNI that a node needs to be removed due
+                        // to an unretriable error. So at the moment we always retry,
+                        // always increasing the the timeout by a factor on each attempt.
+                        warn!("error while retyring workload: {}", e);
```
Stevenjin8:
```suggestion
                        warn!("error while retrying workload: {}", e);
```

src/inpod/windows/workloadmanager.rs
```diff
@@ -272,6 +268,65 @@ impl<'a> WorkloadProxyManagerProcessor<'a> {
...
+                        let new_timeout = previous_timeout.mul(2);
+                        debug!(
+                            "retyring workload {:?} in {} seconds",
+                            poddata,
+                            new_timeout.as_secs()
+                        );
+                        self.enqueue_local_event(
```
Stevenjin8, line 299: After some thought, this doesn't matter, but it is a bit sketchy to modify the event queue as we are iterating over it

src/inpod/windows/workloadmanager.rs
```diff
@@ -272,6 +268,65 @@ impl<'a> WorkloadProxyManagerProcessor<'a> {
...
+            }
+            match &event.event_type {
+                EventType::RetryAllPendingWorkloads => self.retry_proxies().await,
+                EventType::RetryWorkload(previous_timeout, poddata) => {
+                    if let Err(e) = self.state.retry_compartmentless(poddata).await {
+                        // We can't tell the CNI that a node needs to be removed due
+                        // to an unretriable error. So at the moment we always retry,
+                        // always increasing the the timeout by a factor on each attempt.
+                        warn!("error while retyring workload: {}", e);
+                        let new_timeout = previous_timeout.mul(2);
```
Stevenjin8, line 293: we should put a cap on this.

Stevenjin8: and add jitter

src/inpod/windows/workloadmanager.rs
Stevenjin8: I think it works out, but feels a bit fragile (like the fact that enqueue_local_event will always enqueue an event that has a timeout that has an expiration > now (otherwise the retry event would get immediately deleted). It would be nice to at least have some comments for this.

grnmeira: I get ya, agreed. We're also making async calls. I don't see much issues for the short term, but in the future it might generate some pain points. Let me see if we can make it a bit better.

src/inpod/windows/workloadmanager.rs
grnmeira: This is just a work around at the moment so people can test it the Windows implementation. In order to fix this properly we need a change in ZDS or a fix in containerd/Windows' HCS. Let me add some extra comments and let me know if it's clearer. There's an issue open for that (#1609), and I've added your points about jitter and a cap.

But the underlining reason is we need to always ACK to the CNI even before the retries start. So if at some point we reach that cap/timeout, we can't really do anything. Kubernetes will still think there's a container hanging around, that's why we keep retrying. If you think this behaviour is detrimental to the experimental branch or think of any alternatives let me know.

Stevenjin8 (approved):

PR #1614, opened 2025-08-13
Update rust version and dependencies

howardjohn:


src/socket.rs
```diff
@@ -32,12 +32,12 @@ pub fn set_freebind_and_transparent(socket: &TcpSocket) -> io::Result<()> {
     let socket = SockRef::from(socket);
     match socket.domain()? {
         Domain::IPV4 => {
-            socket.set_ip_transparent(true)?;
-            socket.set_freebind(true)?;
+            socket.set_ip_transparent_v4(true)?;
+            socket.set_freebind_v6(true)?;
```
ilrudie:
```suggestion
            socket.set_freebind_v4(true)?;
```
Is using v6 here intentional? It seems like we want to use the v4 variant here.

ilrudie (commented):
One suspicious bit, otherwise lgtm

PR #1626, opened 2025-09-03
Add open file metrics

Stevenjin8:
Add the `process_max_fds` `process_open_fds` to ztunnel metrics.

Resolves https://github.com/istio/istio/issues/57477

### DIFF
--- a/src/metrics/process.rs
+++ b/src/metrics/process.rs
@@ -1,0 +1,88 @@
+// Copyright Istio Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use nix::sys::resource::{Resource, getrlimit};
+use prometheus_client::collector::Collector;
+use prometheus_client::encoding::{DescriptorEncoder, EncodeMetric};
+use prometheus_client::metrics;
+use tracing::error;
+
+// Track open fds
+#[derive(Debug)]
+pub struct ProcessMetrics {}
+const FD_PATH: &str = "/proc/self/fd";
+
+impl ProcessMetrics {
+    pub fn new() -> Self {
+        Self {}
+    }
+
+    fn encode_open_fds(&self, encoder: &mut DescriptorEncoder) -> Result<(), std::fmt::Error> {
+        // Count open fds by listing /proc/self/fd
+        let open_fds = match std::fs::read_dir(FD_PATH) {
+            Ok(entries) => entries.count() as u64,
+            Err(e) => {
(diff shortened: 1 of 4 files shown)


src/metrics/process.rs
```diff
@@ -0,0 +1,94 @@
...
+use nix::sys::resource::{Resource, getrlimit};
+use prometheus_client::collector::Collector;
+use prometheus_client::encoding::{DescriptorEncoder, EncodeMetric};
+use prometheus_client::metrics;
+use tracing::error;
+
+// Track open fds
+#[derive(Debug)]
+pub struct ProcessMetrics {}
+const FD_PATH: &str = "/proc/self/fd";
```
howardjohn: From prom go code:

```
	// Alternately, the undocumented proc_pidinfo(PROC_PIDLISTFDS) can be used to
	// return a list of open fds, but that requires a way to call C APIs.  The
	// benefits, however, include fewer system calls and not failing when at the
	// open file soft limit.

	if dir, err := os.Open("/dev/fd"); err != nil {
		return 0.0, err
	} else {
		defer dir.Close()

		// Avoid ReadDir(), as it calls stat(2) on each descriptor.  Not only is
		// that info not used, but KQUEUE descriptors fail stat(2), which causes
		// the whole method to fail.
		if names, err := dir.Readdirnames(0); err != nil {
			return 0.0, err
		} else {
			// Subtract 1 to ignore the open /dev/fd descriptor above.
			return float64(len(names) - 1), nil
		}
	}
```

* /proc/self/fd vs /dev/fd -- is it relevant?
* Does the PROC_PIDLISTFDS make more sense for us in rust?
* Does read_dir do stat like it does in Go?

Stevenjin8: @howardjohn 

* `proc/self/fd` vs `/dev/fd` seem to be symlinks to in linux. I don't feel strongly, so we can default to whatever Go does.
* isn't `PROC_PIDLISTFDS` a mac thing? Same with any KQUEUE concerns.
* Pretty sure not. It uses a constant number of syscalls. here is the strace from on request to `/metrics`
```
recvfrom(24, "GET /metrics HTTP/1.1\r\nHost: loc"..., 8192, 0, NULL, NULL) = 86
openat(AT_FDCWD, "/dev/fd", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = 25
newfstatat(25, "", {st_mode=S_IFDIR|0500, st_size=26, ...}, AT_EMPTY_PATH) = 0
getdents64(25, 0x605ee1969360 /* 28 entries */, 32768) = 672
getdents64(25, 0x605ee1969360 /* 0 entries */, 32768) = 0
close(25)                               = 0
prlimit64(0, RLIMIT_NOFILE, NULL, {rlim_cur=1024*1024, rlim_max=1024*1024}) = 0
writev(24, [{iov_base="HTTP/1.1 200 OK\r\ncontent-type: t"..., iov_len=119}, {iov_base="# HELP process_open_fds Number o"..., iov_len=2062}], 2) = 2181
```

howardjohn (approved):

d-mankowski-synerise:
Thanks for implementing this! One question, will it be backported to 1.27, or do we have to wait for 1.28?

Stevenjin8:
> Thanks for implementing this! One question, will it be backported to 1.27, or do we have to wait for 1.28?

1.28, we don’t backport new features.

PR #1634, opened 2025-09-22
Add complete wildcard support

Stevenjin8:
So we can have host be "*".


Stevenjin8:
cc @jaellio @grnmeira

howardjohn (commented):
What does this change in terms of user facing behavior?

The test is very unit-level so hard to grok the meaning

PR #1641, opened 2025-09-30
Set timeouts and keepalive

Stevenjin8:
To stop sockets from leaking when client completely disconnects.

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -34,7 +34,6 @@ pub use metrics::*;
 
 use crate::identity::{Identity, SecretManager};
 
-use crate::config::SocketConfig;
 use crate::dns::resolver::Resolver;
 use crate::drain::DrainWatcher;
 use crate::proxy::connection_manager::{ConnectionManager, PolicyWatcher};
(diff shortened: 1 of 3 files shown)


src/proxy.rs
```diff
@@ -133,6 +134,29 @@ impl DefaultSocketFactory {
     }
 }
 
+fn setup_steam(cfg: &SocketConfig, s: &TcpStream) -> io::Result<()> {
```
keithmattix: Should we make inbound use a factory like the outbound calls?

howardjohn: the `self.listener` is already a wrapper around TcpListener that adds `nodelay`.

Perhaps it would make sense to make that take in a SocketConfig and automatically apply it

howardjohn: Note we only want this on inbound_plaintext though (??), but use Listener in other places, so should be optional and only used there.

Stevenjin8: adding this is kinda awkward as is. We can:
1. have `accept` take socket options
2. apply it to all sockets. (We already do so for outbound)
3. have multiple socket option (e.g. outbound_cfg, inbound_cfg, inbound_plaintext_cfg).

Option 2 seems best to me. I don't see a great reason to set these socket options for outbound, but not inbound.

howardjohn: keepalive over localhost is probably not really needed.

But a 4th option seems to just be `fn set_socket_options(&mut self, socket_options: SocketConfig)` on `socket::Listener`? And set it only in `InboundPassthrough::new`

jaellio: `setup_stream` instead of `setup_steam`

src/socket.rs
```diff
@@ -145,29 +144,57 @@ mod linux {
 }
 
 /// Listener is a wrapper For TCPListener with sane defaults. Notably, setting NODELAY
```
jaellio, line 146: Update comment to include keepalive?

howardjohn (approved):

PR #1646, opened 2025-10-02
Make certificate DER fields public

mantasmatelis:
By making these fields public, bits of ztunnel (mainly `identity::caclient`) become broadly useful for other software that wants to retrieve certificates as ztunnel would.


louiscryan:
hey Mantas, no obvious technical objection per se but we generally don't think of ztunnel as a 'library' and want to be careful about scope creep on this front. 

Can you elaborate on the need / use-case a bit?

keithmattix:
This seems reasonable to me since it's so small. We should have a larger conversation if your needs expand though

PR #1659, opened 2025-10-28
dry run auth

Stevenjin8:
might be a slow rollout

Here are the semantics:
* first check for dry-run deny, and log all that match
* then check for real deny. short circuit if any match
* then check for dry run accept and log all that match
* then check for real accept.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -1471,7 +1471,7 @@ mod tests {
 
     fn create_dry_run_wildcard_rbac_policy(action: rbac::RbacAction) -> rbac::Authorization {
         rbac::Authorization {
-            action: action,
+            action,
             namespace: "ns1".into(),
             name: "wildcard".into(),
             rules: vec![vec![]],


src/state.rs
```diff
@@ -570,6 +575,11 @@ impl DemandProxyState {
 
         // Allow and deny logic follows https://istio.io/latest/docs/reference/config/security/authorization-policy/
 
+        for pol in deny_dry_run.iter() {
+            if pol.matches(conn) {
+                debug!(policy = pol.to_key().as_str(), "dry-run deny policy match");
```
Stevenjin8, line 580: I've always found it curious that these are debug logs in envoy.

ilrudie:
Was there a document around this or was it just discussion in WG?

Stevenjin8:
> Was there a document around this or was it just discussion in WG?

@ilrudie sorry for the ghost. Yes, we did discuss this in a WG group meeting (https://docs.google.com/document/d/1wsa06GGiq1LEGwhkiPP0FKIZJqdAiue-VeBonWAzAyk/edit?tab=t.0#heading=h.mlb4al6widfz). We agreed on the behavior described in the PR IIRC.

ilrudie:
Awesome. Thanks for the notes @Stevenjin8 . This is on my list to take a look at.

src/state.rs
ilrudie: makes them kind of ugly too
```
2025-12-31T14:09:03.793034Z     debug   state:proxy{wl=server/httpbin-654df84766-hwj4m}:inbound{id=1f022ae2187b6d519923ca493b1dbaec peer=10.244.1.4:46190}      dry-run deny policy match       policy="server/dry-run-example"
```

PR #1660, opened 2025-10-30
feat: implement crl support in ztunnel

nilekhc:
This PR extends the work of [CRL support](https://github.com/istio/istio/pull/56308) by consuming the ca-crl config map in zTunnel and validates the request accordingly.

Fixes https://github.com/istio/ztunnel/issues/1678

### DIFF
--- a/src/tls/certificate.rs
+++ b/src/tls/certificate.rs
@@ -320,6 +320,7 @@ impl WorkloadCertificate {
             }
         }
 
+        // TODO: check if our own certificate is revoked in the CRL and log warning
         let raw_client_cert_verifier = builder.build()?;
 
         let client_cert_verifier =
@@ -336,17 +337,11 @@ impl WorkloadCertificate {
         Ok(sc)
     }
 
-    pub fn client_config(
-        &self,
-        identity: Vec<Identity>,
-        crl_manager: Option<Arc<crate::tls::crl::CrlManager>>,
-    ) -> Result<ClientConfig, rustls::Error> {
+    // TODO: add CRL support for outbound connections (client verifying server certs)
+    // this requires a separate design due to complexity - deferred for follow-up
+    pub fn client_config(&self, identity: Vec<Identity>) -> Result<ClientConfig, rustls::Error> {
         let roots = self.root_store.clone();
-        let verifier = IdentityVerifier {
-            roots,
-            identity,
-            crl_manager,
-        };
+        let verifier = IdentityVerifier { roots, identity };
         let mut cc = ClientConfig::builder_with_provider(crate::tls::lib::provider())
             .with_protocol_versions(tls::TLS_VERSIONS)
             .expect("client config must be valid")
@@ -362,12 +357,8 @@ impl WorkloadCertificate {
         Ok(cc)
     }
 
-    pub fn outbound_connector(
-        &self,
-        identity: Vec<Identity>,
-        crl_manager: Option<Arc<crate::tls::crl::CrlManager>>,
-    ) -> Result<OutboundConnector, Error> {
-        let cc = self.client_config(identity, crl_manager)?;
+    pub fn outbound_connector(&self, identity: Vec<Identity>) -> Result<OutboundConnector, Error> {
+        let cc = self.client_config(identity)?;
         Ok(OutboundConnector {
             client_config: Arc::new(cc),
         })
@@ -486,7 +477,7 @@ mod test {
         });
 
         let stream = TcpStream::connect(addr).await.unwrap();
-        let client = cert2.outbound_connector(vec![id], None).unwrap();
+        let client = cert2.outbound_connector(vec![id]).unwrap();
         let mut tls = client.connect(stream).await.unwrap();
 
         let _ = tls.write(b"hi").await.unwrap();
--- a/src/socket.rs
+++ b/src/socket.rs
@@ -178,7 +178,6 @@ impl Listener {
                 let res = SockRef::from(&stream).set_tcp_keepalive(&ka);
                 tracing::trace!("set keepalive: {:?}", res);
             }
-            #[cfg(target_os = "linux")]
             if cfg.user_timeout_enabled {
                 let ut = cfg.keepalive_time + cfg.keepalive_retries * cfg.keepalive_interval;
                 let res = SockRef::from(&stream).set_tcp_user_timeout(Some(ut));


howardjohn:
I'm a bit confused. It seems there was a design doc that was approved for this (not sure by who, as I was out of office and there are no meeting notes or listed approvers) but the design is completely different than this

src/proxy.rs
```diff
@@ -120,6 +120,7 @@ impl DefaultSocketFactory {
                 socket2::SockRef::from(&s).set_tcp_keepalive(&ka)
             );
         }
+        #[cfg(target_os = "linux")]
```
nilekhc, line 121: I was getting an error on macos due to cross compilation. I have to add this to some places to fix it. Please let me know if there is a better way to handle this.

src/proxy/connection_manager.rs
```diff
@@ -260,6 +260,39 @@ impl ConnectionManager {
         // potentially large copy under read lock, could require optimization
         self.drains.read().expect("mutex").keys().cloned().collect()
     }
+
+    /// Close all inbound and outbound connections
+    /// This is called when certificate revocations are detected
+    pub fn close_all_connections(&self) {
```
howardjohn: This seems really dangerous and bad behavior. How strongly do you feel about keeping this?

Does any other application do this?

keithmattix: +1 I don't think we should be closing every single connection just because of one revocation. What does Envoy do?

nilekhc: The only reason I addd this since http connections are long lived and does not check updated crl. Is there a better way to handle this? I know envoy has some timeout we can configure. Not comepletly sure how zTunnel handles it. Happy to tweak this.

howardjohn: My intuition is to do nothing. But at the very least, only close ones that are revoked

keithmattix: Doing nothing feels wrong IMO; the purpose of a CRL is to be able to break a particular trust chain, and you don't want existing connections to continue. Only closing the ones that are revoked is the right solution to me

howardjohn: Do we need it on outbound as well though? or is only inbound sufficient.

The outbound part adds substantial complexity and there is not currently really a concept of outbound authz in istio

nilekhc: I have updated the PR to only drain the connection that has revoked cert. 


> The outbound part adds substantial complexity and there is not currently really a concept of outbound authz in istio

should we just validate inbound connections only then?

@howardjohn @keithmattix

ilrudie: Complexity aside, is there a strong argument we can make for preserving outbound connections which wouldn't presently be allowed due to revoked certs?

howardjohn: Is there any other software that does that?

keithmattix: Looks like [OpenVPN](https://openvpn.net/community-docs/revoking-certificates.html) forces a reboot:

>"...and have the new CRL take effect immediately for newly connecting clients. If the client whose certificate you are revoking is already connected, you can restart the server via a signal (SIGUSR1 or SIGHUP) and flush all clients, or you can telnet to the management interface and explicitly kill the specific client..."

ilrudie: Is restart your clients if you want the CRL to take effect quickly the UX we want then?

nilekhc: imo, zTunnel should terminate that connection selectively and avoid restarting the workload. Since the user provides/updates the CRL file, they’re already aware of potentially revoked certificates, so why ask them to take action twice?
(1) Create or update the CRL, and
(2) Restart the workload pod.

ilrudie: Playing advocate aside, I suspect that most folks would accept a limitation where existing connections are not re-evaluated on a CRL refresh. This seems to be a fairly common behavior. I think I would lean towards breaking support for re-evaluating existing connections into a separate PR (assuming we do want to explore that functionality).

keithmattix: Separate PR seems reasonable to me so we can evaluate the complexity on its own

nilekhc: I have reverted the outbound related changes and now it only selectively terminates inbound connections. PTAL. Thanks

/cc @howardjohn @keithmattix @ilrudie

nilekhc:
@howardjohn @ilrudie @keithmattix Could you PTAL when you get a chance? Thanks.

src/proxy/connection_manager.rs
```diff
@@ -260,6 +321,120 @@ impl ConnectionManager {
...
+
+        // Log all revoked serials we're checking against
+        for (idx, serial) in revoked_serials.iter().enumerate() {
+            tracing::debug!(
+                "revoked serial {}: {} bytes (hex: {})",
+                idx + 1,
+                serial.len(),
+                serial
+                    .iter()
+                    .map(|b| format!("{:02x}", b))
```
howardjohn: why are we converting to string here and on L372? it will be quite expensive, we can just compare the bytes directly?

src/proxy/inbound.rs
```diff
@@ -120,13 +121,92 @@ impl Inbound {
...
+                        conn: conn.clone(),
+                        dest_workload: destination_workload,
+                    };
+                    let mut tls_guard = pi
+                        .connection_manager
+                        .register_tls_connection(&rbac_ctx, client_cert_serials.clone());
+                    debug!("TLS connection registered for revocation tracking");
+
+                    // Get watcher BEFORE serving to listen for revocation-based termination
+                    let tls_drain_watch = tls_guard.watcher();
```
ilrudie: serv_connect will already create a connection guard which will end the connection when asked. I don't think we need a second connection guard

src/proxy/inbound.rs
```diff
@@ -153,11 +234,29 @@ impl Inbound {
...
+                            debug!("TLS connection terminated due to certificate revocation");
+                            tls_guard.release();
+                            // Return an error to signal abnormal termination
+                            // This helps with proper cleanup and error reporting
+                            Err(std::io::Error::new(
+                                std::io::ErrorKind::ConnectionAborted,
+                                "connection terminated due to certificate revocation"
+                            ).into())
+                        }
+                    }
```
ilrudie: similar to above, we already take a guard that closes the connection, I don't think we need

src/proxy/connection_manager.rs
nilekhc: Good point. updated

src/proxy/inbound.rs
nilekhc: refactored

src/proxy/inbound.rs
nilekhc: refactored

nilekhc:
> It seems to me this bundles functionality that belongs in something like PolicyWatcher into the ConnectionManager. Please take a look at that, specifically how run works and is invoked and works for an example of how similar functionality was implemented.

Thanks for the pointers @ilrudie. I refactored the code with CRLWatcher much like PolicyWatcher. PTAL Thanks.

src/proxy/inbound.rs
```diff
@@ -120,13 +121,63 @@ impl Inbound {
...
+                    let rbac_ctx = ProxyRbacContext {
+                        conn: conn.clone(),
+                        dest_workload,
+                    };
+
+                    let mut tls_guard = pi
+                        .connection_manager
+                        .register_tls_connection(&rbac_ctx, client_cert_serials.clone());
+                    let tls_drain = tls_guard.watcher();
+                    debug!("registered TLS connection for CRL monitoring");
```
ilrudie: Why do we need a second connection guard at a different level? The guard created inside the request_handler seems sufficient. It'll already include the certificates involved in the connection and can  be closed if the connection needs to terminate due to revoked certificates.

src/proxy/inbound.rs
ilrudie: Maybe a commit didn't get pushed. We still have a second connection guard being created

src/proxy/inbound.rs
nilekhc: It was not closing exiting connections properly. Worked for new connections so I had to add this. But let me take a look at it again and validate.

nilekhc: @ilrudie, so this is mainly to keep track of idle connections. In my teseting I found that, if we remove TLS guard and rely only on per-request guards:
 - for idle TLS connection (no active requests), no guards exist
 - so when CRL update happens -> CrlWatcher finds no connections to close
 - and, revoked certificate remains connected until next request
 
wdyt?
 
 No TLS gurd works for new connections since it pretty straight forward check.

ilrudie: Is this a correctness issue? A long-lived connection would still be closed.

What remains is only an idle HBONE which can't be used by the application?

nilekhc: > Is this a correctness issue? 

Maybe. My goal is to terminate ongoing connection as well. But open to only rejecting new connections since ongoing connections will eventually timeout.

src/proxy/connection_manager.rs
```diff
@@ -185,12 +187,14 @@ impl ConnectionManager {
         state: &DemandProxyState,
         ctx: &ProxyRbacContext,
         dest_service: Option<String>,
+        client_cert_serials: Option<Vec<Vec<u8>>>,
     ) -> Result<ConnectionGuard, Error> {
```
ilrudie, line 188: **note:** I'm not sure where to leave this. It's not a comment about this line of code per se. More a semi-relevant anchor for my thought process.

This effectively ties a property of the HBONE transport being used (certificate serials) to the application connection using it. It's not incorrect to know details about the underlying transport mechanism I think. It means we can deny application level traffic based on the transport which would be used for it.

The rub is, I think we can get caught in a bad state by doing this and not also invalidating the transport underneath. Even if we get updated certificates, the underlying transport is already established and would continue to be used causing denies.

ilrudie (commented):
Can we clarify the behaviors expected in this PR after the WG discussion. From a look through it seems like:

- net-new connections (no previous HBONE established) trying to use revoked certs: DENY
- new application connection (open HBONE) trying to use revoked certs: DENY
- open application connection (application actively sending traffic) presently using a revoked certs: CLOSE
- no application connection (open HBONE): nothing, HBONE connection lives until timeout

Is there any path to an application actually sending traffic with a revoked cert?

I bring this up because my understanding of the WG discussion is that we would move the entirety of this checking to the HBONE layer. New HBONE connections with revoked certs are not accepted, but existing ones continue to be functional. I worry there are edges, I tried to describe one in a comment tied to a line of code in the connection manager, which are tricky to foresee when mixing the HBONE and application properties.

src/proxy/connection_manager.rs
nilekhc: Fixed

src/tls/crl.rs
```diff
@@ -0,0 +1,466 @@
...
+                }
+            }
+        }
+
+        Ok(manager)
+    }
+
+    /// Load or reload the CRL from disk
+    pub fn load_crl(&self) -> Result<bool, CrlError> {
+        let mut inner = self.inner.write().unwrap();
```
ilrudie, line 95: Avoid unwrap. If you truly believe this is infallible, prefer expect. If it can fail, we should handle the error gracefully.

src/proxyfactory.rs
```diff
@@ -55,13 +57,35 @@ impl ProxyFactory {
...
+        let crl_manager = if config.enable_crl {
+            match tls::crl::CrlManager::new(config.crl_path.clone(), config.allow_expired_crl) {
+                Ok(manager) => {
+                    let manager_arc = Arc::new(manager);
+
+                    if let Err(e) = manager_arc.start_file_watcher() {
+                        tracing::error!("failed to start CRL file watcher: {}", e);
+                    }
+
+                    Some(manager_arc)
```
ilrudie, line 75: Seems odd. Is there a reason to store a manager that failed to start? I think this puts us in a loop where no certs can validate.

src/tls/crl.rs
```diff
@@ -0,0 +1,466 @@
...
+            return Err(CrlError::ParseError(
+                "no valid CRL blocks found in PEM data".to_string(),
+            ));
+        }
+
+        Ok(crls)
+    }
+
+    /// Verify CRL validity period
+    fn verify_crl_validity(
```
ilrudie: nit: `validate_crl` maybe?

src/tls/crl.rs
```diff
@@ -0,0 +1,466 @@
...
+                return Err(CrlError::ExpiredCrl);
+            }
+            warn!("CRL is expired but allow_expired_crl is enabled");
+        }
+
+        Ok(())
+    }
+
+    /// Check if any certificate in the chain is revoked
+    pub fn is_revoked_chain(
```
ilrudie: Is there a reason we didn't use something like https://docs.rs/rustls-webpki/latest/webpki/enum.CertRevocationList.html#method.find_serial?

ilrudie (commented):
I had to time box looking at this so here's a partial review. Overall impression, this seems like a better direction. Curious about why we choose not to use something like rustls webpki for checking revocations and instead implement our own.

src/tls/crl.rs
nilekhc: Good point. fixed it.

src/tls/crl.rs
nilekhc: done

src/tls/crl.rs
nilekhc: Ah. This is better. validating CRL with rustls-webpki. Thanks for the pointer.

src/proxyfactory.rs
nilekhc: This is just a check for crl filewatch. The CRL manager is already correctly initialized during new(), and the file watcher is only for detecting updates, so validation still works even if the watcher fails (you’d just need a restart zTunnel to pick up new CRLs). There’s also no risk, if the CRL list is empty, we call load_crl() once and fail fast if it doesn’t load.

nilekhc: I did change it to warn and then added some more log statements to provide additonal details.

ilrudie: Gotcha, If we have a crl manager and the crl file doesn't exist yet or is empty then we'll keep trying to load it for every connection attempt and deny all traffic until the first successful load that validates. This allows recovery even if the watch failed.

On the other path, we could already have loaded a valid CRL into the manager, then have the watch fail and in that case you can run without it and you just can't get updates.

ilrudie:
@istio/wg-networking-maintainers-ztunnel, anyone else have cycles to dive into this? I'm not sure I have time anymore. My present concern is that match certs to our crls seems pretty custom and I'm not convinced it's correct. I don't have the knowledge of correct CRL impl to say for sure, but it seems like we need to be checking revocation against the instant that a cert is used. This isn't exhaustive per se, just an issue I think the impl could have which needs digging into.

src/test_helpers/app.rs
```diff
@@ -113,6 +115,7 @@ impl TestApp {
             Ok(client.request(req).await?)
         };
 
+        #[cfg(target_os = "linux")]
```
Stevenjin8, line 118: I'd really rather not have this littered all over the place. The only other platform that we support (and barely do so) is windows.

src/tls/crl.rs
```diff
@@ -0,0 +1,412 @@
...
+                    debug!("  next update: {:?}", next_update);
+                }
+                let revoked_count = crl_info.tbs_cert_list.revoked_certificates.len();
+                debug!("  revoked certificates: {}", revoked_count);
+            }
+
+            parsed_crls.push(crl);
+        }
+
+        let has_new_revocations = parsed_crls.len() != inner.crl_list.len();
```
Stevenjin8: I don't think this is true. Expired certs can get removed from CRLs:

> An entry MUST NOT be removed
>  from the CRL until it appears on one regularly scheduled CRL issued
 > beyond the revoked certificate's validity period.

https://datatracker.ietf.org/doc/html/rfc5280#section-3.3

src/tls/crl.rs
```diff
@@ -0,0 +1,412 @@
...
+                }
+                in_pem = false;
+                continue;
+            }
+            if in_pem {
+                base64_data.push_str(line.trim());
+            }
+        }
+
+        if crls.is_empty() {
```
Stevenjin8: Should read the design doc, but whats wrong with having an empty crl file?

src/tls/crl.rs
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    /// Check if a certificate is revoked using webpki's native CRL API
+    pub fn is_cert_revoked(&self, cert: &CertificateDer) -> Result<bool, CrlError> {
+        use x509_parser::prelude::*;
+
+        let inner = self
+            .inner
+            .read()
+            .map_err(|e| CrlError::LockError(format!("failed to acquire read lock: {}", e)))?;
```
Stevenjin8: tbh we can unwrap here. we do it in other places too. The only way this errors if the mutex is poisoned

https://github.com/istio/ztunnel/blob/cd100cd3ad7484d56baf5de5c8ac49f33239c1ca/src/state.rs#L532-L534

src/tls/crl.rs
```diff
@@ -0,0 +1,412 @@
...
+            })?;
+
+            let crl = CertRevocationList::from(owned_crl);
+
+            // use x509-parser for detail logging
+            use x509_parser::prelude::*;
+            if let Ok((_, crl_info)) = CertificateRevocationList::from_der(der_data) {
+                debug!("CRL {}:", idx + 1);
+                debug!("  issuer: {}", crl_info.tbs_cert_list.issuer);
+                debug!("  this update: {:?}", crl_info.tbs_cert_list.this_update);
```
Stevenjin8: Something like 

```rs
     debug!(
         crl = idx + 1,
         issuer = %tbs.issuer,
         this_update = ?tbs.this_update,
         next_update = ?tbs.next_update,
         revoked = tbs.revoked_certificates.len(),
         "loaded CRL"
     );
```

src/tls/crl.rs
```diff
@@ -0,0 +1,412 @@
...
+        let cert_serial = parsed_cert.serial.to_bytes_be();
+        debug!("checking certificate serial: {:?}", cert_serial);
+
+        // check the certificate against ALL CRLs using webpki's native API
+        for (idx, crl) in inner.crl_list.iter().enumerate() {
+            debug!("checking against CRL {}", idx + 1);
+
+            match crl.find_serial(&cert_serial) {
+                Ok(Some(revoked_cert)) => {
+                    error!(
```
Stevenjin8: can we also use logging with key values

Stevenjin8 (changes requested):
Some minor comments, will take it for a run tmr. Generally there are *a lot* of debug statements. Would be nice to cut down a bit, espcially on happy paths.

src/tls/workload.rs
```diff
@@ -112,6 +122,29 @@ impl ClientCertVerifier for TrustDomainVerifier {
             .base
             .verify_client_cert(end_entity, intermediates, now)?;
         self.verify_trust_domain(end_entity)?;
+
+        // check CRL if enabled
+        if let Some(crl_manager) = &self.crl_manager {
+            debug!("crl checking enabled for client certificate");
```
Stevenjin8:
```suggestion
```

src/tls/workload.rs
```diff
@@ -112,6 +122,29 @@ impl ClientCertVerifier for TrustDomainVerifier {
...
+            let is_revoked = crl_manager
+                .is_revoked_chain(end_entity, intermediates)
+                .map_err(|e| {
+                    error!("crl validation failed for client certificate: {}", e);
+                    rustls::Error::General(format!("certificate revocation check failed: {}", e))
+                })?;
+
+            if is_revoked {
+                error!("client certificate is REVOKED - rejecting connection");
```
Stevenjin8: These are all at most debug. See rbac code:

https://github.com/istio/ztunnel/blob/cd100cd3ad7484d56baf5de5c8ac49f33239c1ca/src/state.rs#L591-L601

src/tls/crl.rs
```diff
@@ -0,0 +1,412 @@
...
+            debug!("CRL is in DER format");
+            // Single DER-encoded CRL
+            vec![data]
+        };
+
+        debug!("found {} CRL block(s) in file", der_crls.len());
+
+        let mut parsed_crls = Vec::new();
+
+        for (idx, der_data) in der_crls.iter().enumerate() {
```
Stevenjin8: Do we ever verify that the CRL is signed by our root cert?

Stevenjin8: we should also verify that there are no critical extensions that we do not understand.

> If a CRL contains a critical extension that the application cannot process, then the application MUST NOT use that CRL to determine the status of certificates

src/tls/crl.rs
```diff
@@ -0,0 +1,412 @@
...
+        // check all intermediate certificates
+        for (idx, intermediate) in intermediates.iter().enumerate() {
+            debug!("checking intermediate certificate {} in chain", idx);
+            if self.is_cert_revoked(intermediate)? {
+                error!("intermediate CA certificate at position {} is REVOKED", idx);
+                return Ok(true);
+            }
+        }
+
+        debug!("certificate chain validation passed - no revoked certificates found");
```
Stevenjin8:
```suggestion
```

src/tls/crl.rs
Stevenjin8: and that the current time is within `thisUpdate` and `nextUpdate`

EDIT: and everything listed here
https://datatracker.ietf.org/doc/html/rfc5280#section-6.3.3

src/tls/crl.rs
```diff
@@ -0,0 +1,412 @@
...
+
+            match crl.find_serial(&cert_serial) {
+                Ok(Some(revoked_cert)) => {
+                    error!(
+                        "certificate with serial {:?} is REVOKED in CRL {}",
+                        cert_serial,
+                        idx + 1
+                    );
+                    error!("revocation date: {:?}", revoked_cert.revocation_date);
+                    if let Some(reason) = revoked_cert.reason_code {
```
Stevenjin8: No changes required here, but just pointing out that `reason` could be `RemoveFromCrl` which would mean to not revoke it, but according to the webpki docs:

>     /// RemoveFromCrl only appears in delta CRLs that are unsupported.

src/tls/crl.rs
```diff
@@ -0,0 +1,412 @@
...
+            return Err(CrlError::ParseError(
+                "no valid CRL blocks found in PEM data".to_string(),
+            ));
+        }
+
+        Ok(crls)
+    }
+
+    /// Check if a certificate is revoked using webpki's native CRL API
+    pub fn is_cert_revoked(&self, cert: &CertificateDer) -> Result<bool, CrlError> {
```
Stevenjin8: Hmm, I think this should return a `bool` rather than a `Result`. If we cannot get CRLs for any reason, its up to us what the failure policy should be, and I think that behavior should be defined in the CrlManager. Here, the only reason for failure is if we have invalid PEM/DER. `find_serial` is infallible.

src/tls/crl.rs
```diff
@@ -0,0 +1,412 @@
...
+        intermediates: &[CertificateDer],
+    ) -> Result<bool, CrlError> {
+        debug!(
+            "checking certificate chain against CRL (chain length: {})",
+            1 + intermediates.len()
+        );
+
+        // check leaf certificate
+        debug!("checking leaf certificate");
+        if self.is_cert_revoked(end_entity)? {
```
Stevenjin8: can we used `is_revoked_chain` here?

src/config.rs
```diff
@@ -78,6 +78,9 @@ const HTTP2_FRAME_SIZE: &str = "HTTP2_FRAME_SIZE";
 
 const UNSTABLE_ENABLE_SOCKS5: &str = "UNSTABLE_ENABLE_SOCKS5";
 
+const ENABLE_CRL: &str = "ENABLE_CRL";
+const CRL_PATH: &str = "CRL_PATH";
```
Stevenjin8: can these be like `XDS_ADDRESS` where we just have one environment variable, and empty means disable.

src/test_helpers/app.rs
nilekhc: I'll removed this once I have an approval, I can't cross compile to validate changes on my local.

src/tls/crl.rs
nilekhc: You're right, thanks for catching this. I've added a module-level comment clarifying RFC 5280 Section 3.3 behavior: "per RFC 5280 section 3.3, entries may be removed from CRLs after the certificate's validity period expires". 

Also renamed the misleading variable `has_new_revocations` to `crl_count_changed`.

src/tls/crl.rs
nilekhc: Good point. Fixed, empty CRL files are now treated as "no revocations" instead of returning an error. This makes sense since an empty CRL legitimately means no certificates are currently revoked.

src/tls/crl.rs
nilekhc: Done. Replaced all map_err(...)? on RwLock operations with .unwrap() to match the pattern used elsewhere in the codebase. Also removed the now-unused LockError variant from CrlError.

src/tls/crl.rs
nilekhc: Done. Consolidated into a single structured log with key-value pairs:

```rust 
debug!(
      crl = idx + 1,
      issuer = %tbs.issuer,
      this_update = ?tbs.this_update,
      next_update = ?tbs.next_update,
      revoked = tbs.revoked_certificates.len(),
      "loaded crl"
  );
```

src/tls/crl.rs
nilekhc: Done. Changed to structured logging:
```rust  
debug!(
      serial = ?cert_serial,
      crl = idx + 1,
      revocation_date = ?revoked_cert.revocation_date,
      reason = ?revoked_cert.reason_code,
      "certificate is revoked"
  );
```

src/tls/workload.rs
nilekhc: Done

src/tls/workload.rs
nilekhc: Done. Changed all CRL-related `error!` logs to `debug!` to match the RBAC pattern.

src/tls/crl.rs
nilekhc: Done

src/tls/crl.rs
nilekhc: Thanks for the context. Added an inline comment noting this

src/tls/crl.rs
nilekhc: Done. Changed `is_cert_revoked()` and `is_revoked_chain()` to return bool. Errors are now handled internally with debug logging and fail-open policy. If CRL check fails for any reason, we allow the connection and log the error.

src/tls/crl.rs
nilekhc: `is_revoked_chain` is already the only method called externally (in workload.rs). The is_cert_revoked call here is internal to is_revoked_chain itself, used to check each certificate in the chain.

src/config.rs
nilekhc: Done. Removed `ENABLE_CRL` environment variable. `CRL_PATH` is now the single control:
- Empty/unset -> CRL disabled
- Non-empty -> CRL enabled with that path

Changed `Config.crl_path` from `PathBuf` to `Option<PathBuf>`

src/tls/crl.rs
nilekhc: Implemented time bounds validation and investigated webpki's other validations:
  - Signature verification: Not performed during loading. This implementation relies on the CRL being from a trusted source (e.g., mounted ConfigMap).
  - Critical extensions: Handled by webpki's from_der() - it rejects delta CRLs, indirect CRLs, and validates known extensions.
  - Time bounds: Added `is_crl_time_valid()` function that validates `thisUpdate <= now <= nextUpdate` using x509-parser's ASN1Time. Time bounds are checked:
    a. At CRL load time: expired/not-yet-valid CRLs are skipped with a warning
    b. At revocation lookup time: CRLs that expired since load are skipped

If a CRL has no nextUpdate field, we accept it (per RFC 5280, nextUpdate is optional).

Stevenjin8: Are you sure webpki's from_der will reject anything with unkonwn critical extensions? Regardless, we should cryptographically verify anything that we can, including the crl. Regardless, it would probably be better to use 

https://docs.rs/rustls-webpki/0.103.8/webpki/struct.EndEntityCert.html#method.verify_for_usage

src/tls/crl.rs
```diff
@@ -92,80 +106,136 @@ impl CrlManager {
...
 
-        let has_new_revocations = parsed_crls.len() != inner.crl_list.len();
+        let crl_count_changed = parsed_crls.len() != inner.crl_list.len();
 
-        if has_new_revocations {
-            warn!(
-                "CRL file changed - reloaded with {} CRL(s)",
-                parsed_crls.len()
-            );
+        if crl_count_changed {
```
Stevenjin8: remove

nilekhc: Done

src/tls/crl.rs
```diff
@@ -0,0 +1,343 @@
...
+impl std::fmt::Debug for CrlManager {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("CrlManager").finish_non_exhaustive()
+    }
+}
+
+// stores the webpki CRL for use with verify_for_usage
+struct CrlEntry {
+    crl: OwnedCertRevocationList,
+}
```
Stevenjin8: We dont need this anymore right?

src/tls/crl.rs
```diff
@@ -0,0 +1,343 @@
...
+                debug!("crl not loaded, attempting to load now");
+                if let Err(e) = self.load_crl() {
+                    debug!(error = %e, "failed to load crl");
+                    return Vec::new();
+                }
+            }
+        }
+
+        let inner = self.inner.read().unwrap();
+        inner.crl_list.iter().map(|e| e.crl.clone()).collect()
```
Stevenjin8: somewhat unfortunate that we clone here, but not a blocker for me

src/tls/workload.rs
```diff
@@ -93,13 +103,131 @@ impl TrustDomainVerifier {
...
 }
 
 // Implement our custom ClientCertVerifier logic. We only want to add an extra check, but
 // need a decent amount of boilerplate to do so.
 impl ClientCertVerifier for TrustDomainVerifier {
     fn root_hint_subjects(&self) -> &[DistinguishedName] {
-        self.base.root_hint_subjects()
+        // return distinguished names from root store for client cert hints
+        static EMPTY: &[DistinguishedName] = &[];
+        EMPTY
```
Stevenjin8: This doesn't seem right.

src/tls/workload.rs
```diff
@@ -108,11 +236,45 @@ impl ClientCertVerifier for TrustDomainVerifier {
         intermediates: &[CertificateDer<'_>],
         now: UnixTime,
     ) -> Result<ClientCertVerified, rustls::Error> {
-        let res = self
-            .base
-            .verify_client_cert(end_entity, intermediates, now)?;
+        // get CRLs if CRL manager is enabled
+        let crls = self.crl_manager.as_ref().map(|m| m.get_crls());
+
+        // use verify_for_usage for all certificate verification
```
Stevenjin8: this comment is kinda confusing because the verify_for_usage is called in verify_cert_chain

src/tls/crl.rs
nilekhc: removed

src/tls/crl.rs
nilekhc: removed

src/tls/workload.rs
nilekhc: removed

src/tls/workload.rs
nilekhc: updaed the implementation based on our discussion

src/tls/certificate.rs
```diff
@@ -305,11 +306,24 @@ impl WorkloadCertificate {
...
+        let mut builder = WebPkiClientVerifier::builder_with_provider(
             self.root_store.clone(),
+            crate::tls::lib::provider(),
         );
+
+        // add CRLs if available
+        if let Some(ref mgr) = crl_manager {
+            let crls = mgr.get_crl_ders();
+            if !crls.is_empty() {
+                builder = builder.with_crls(crls).allow_unknown_revocation_status(); // fail-open for unknown status
```
Stevenjin8, line 319: what? Isn't this flipped?

src/tls/crl.rs
```diff
@@ -0,0 +1,325 @@
...
+// CrlManager handles certificate revocation list (CRL) loading.
+//
+// validation notes:
+// - webpki's from_der() validates ASN.1 structure, CRL version (v2), and rejects
+//   delta CRLs and indirect CRLs with unknown critical extensions
+// - CRL signature verification is performed by rustls's WebPkiClientVerifier
+// - time bounds (thisUpdate/nextUpdate) handling depends on whether
+//   enforce_revocation_expiration() is called on the verifier builder
+// - per RFC 5280 section 3.3, entries may be removed from CRLs after the
+//   certificate's validity period expires
```
Stevenjin8: A lot of this comment is wrong now and in the wrong place. I woudl just remove it

nilekhc: I agree. removed.

src/tls/certificate.rs
nilekhc: Its fail open.

Stevenjin8: nope. I just can't read.

src/proxy/inbound.rs
nilekhc: fixed

src/proxy/inbound.rs
nilekhc: Now handleing via WebPkiClientVerifier

src/tls/crl.rs
nilekhc: refactored with WebPkiClientVerifier

src/tls/crl.rs
nilekhc: removed verify_for_usage and handleing everything with WebPkiClientVerifier

src/tls/crl.rs
nilekhc: refactored

src/tls/crl.rs
```diff
@@ -24,16 +24,6 @@ use std::time::Duration;
 use tracing::{debug, info, warn};
 
 // CrlManager handles certificate revocation list (CRL) loading.
```
Stevenjin8: This comment is still in the wrong place.

nilekhc: done

src/tls/certificate.rs
```diff
@@ -298,15 +298,29 @@ impl WorkloadCertificate {
...
+        // build the base client cert verifier with optional CRL support
+        let mut builder = WebPkiClientVerifier::builder_with_provider(
             self.root_store.clone(),
             crate::tls::lib::provider(),
-        )
-        .build()?;
+        );
+
+        // add CRLs if available
+        if let Some(ref mgr) = crl_manager {
```
howardjohn, line 316: is there a reason we are not doing the client config, and only the server config?

src/tls/certificate.rs
```diff
@@ -298,15 +298,29 @@ impl WorkloadCertificate {
...
+            let crls = mgr.get_crl_ders();
+            if !crls.is_empty() {
+                builder = builder.with_crls(crls).allow_unknown_revocation_status(); // fail-open for unknown status
+            }
+        }
+
+        let raw_client_cert_verifier = builder.build()?;
```
howardjohn, line 324: not blocking but might be nice to look at the CRL and *our own cert* (`cert_and_intermediates_der`) and check if its revoked and warn

src/tls/crl.rs
```diff
@@ -0,0 +1,314 @@
...
+                            debug!(event_count = events.len(), "crl directory events detected");
+
+                            // reload CRL for any changes in the watched directory
+                            // this handles Kubernetes ConfigMap updates (..data symlink changes)
+                            // as well as direct file writes and text editor saves
+                            debug!("crl directory changed, reloading");
+                            match manager.load_crl() {
+                                Ok(crl_count_changed) => {
+                                    debug!("crl reloaded successfully after file change");
+                                    if crl_count_changed {
```
howardjohn: nit: just because the count did not change does not mean the content did not change?

src/tls/crl.rs
```diff
@@ -0,0 +1,314 @@
...
+        // store validated DER bytes
+        inner.crl_ders = validated_ders;
+
+        debug!(count = inner.crl_ders.len(), "crl loaded successfully");
+        Ok(crl_count_changed)
+    }
+
+    /// parses PEM-encoded CRL data that may contain multiple CRL blocks
+    /// returns a Vec of DER-encoded CRLs (empty vec if no blocks found)
+    fn parse_pem_crls(pem_data: &[u8]) -> Result<Vec<Vec<u8>>, CrlError> {
```
howardjohn, line 147: can you use rustls_pemfile instead of rolling our own, as we do in the rest of the TLS code?

src/tls/crl.rs
```diff
@@ -0,0 +1,314 @@
...
+        Ok(crls)
+    }
+
+    /// returns CRLs as DER bytes for rustls's with_crls().
+    /// if no CRLs are loaded, attempts to load them first.
+    pub fn get_crl_ders(&self) -> Vec<CertificateRevocationListDer<'static>> {
+        // try to load if not already loaded
+        {
+            let inner = self.inner.read().unwrap();
+            if inner.crl_ders.is_empty() {
```
howardjohn: this doesn't look right. is_empty does not imply its not loaded, since it could be loaded and found to be empty. Then we will always load?

src/tls/crl.rs
```diff
@@ -0,0 +1,314 @@
...
+                drop(inner);
+                debug!("crl not loaded, attempting to load now");
+                if let Err(e) = self.load_crl() {
+                    debug!(error = %e, "failed to load crl");
+                    return Vec::new();
+                }
+            }
+        }
+
+        let inner = self.inner.read().unwrap();
```
howardjohn: we should not lock twice, just use the lock from L190. If you put it in an `else` it should be fine

src/tls/crl.rs
```diff
@@ -0,0 +1,314 @@
...
+            move |result: DebounceEventResult| {
+                match result {
+                    Ok(events) => {
+                        if !events.is_empty() {
+                            debug!(event_count = events.len(), "crl directory events detected");
+
+                            // reload CRL for any changes in the watched directory
+                            // this handles Kubernetes ConfigMap updates (..data symlink changes)
+                            // as well as direct file writes and text editor saves
+                            debug!("crl directory changed, reloading");
```
howardjohn, line 229: So this will reload only for *new* server_configs. So this will take ~12hrs to rollout

howardjohn: rustls supports dynamic certs with with_cert_resolver but doesn't seem to with CRLs which is unfortunate

src/tls/certificate.rs
Stevenjin8: I think somewhere we agreed to verify crls only in the on inbound connections

Stevenjin8: ok maybe i made that up

src/tls/certificate.rs
Stevenjin8: I had this thought too. Would be also nice if istioctl zc certificates showed revoked certs as invalide

src/tls/certificate.rs
nilekhc: @howardjohn, During KubeCon, when we discussed this, we decided to implement CRLs only for inbound connections, since outbound scenarios are more complex and require a follow-up design. I’m thinking of adding outbound scenarios separately.

src/tls/crl.rs
Stevenjin8: Don't we create a new server_config for each new incoming hbone connection?

src/tls/certificate.rs
Stevenjin8: https://github.com/istio/ztunnel/pull/1660#discussion_r2479148136

nilekhc: Added a TODO comment to track this for follow-up.

src/tls/certificate.rs
nilekhc: Good idea! Added a TODO to track this. It would require adding `is_serial_revoked()` to CrlManager. I'll address in a follow-up.

src/tls/crl.rs
nilekhc: Good catch! Fixed it. I'm now using hash comparison to detect actual content changes

src/tls/crl.rs
nilekhc: done

src/tls/crl.rs
nilekhc: Makes sense. Added a `loaded: bool` flag to distinguish _not loaded yet_ from _loaded but empty_. I'm now checking `!inner.loaded` instead of _inner.crl_ders.is_empty()_.

src/tls/crl.rs
nilekhc: Yeah. I added a commnet on CRLManager to document it.

src/tls/crl.rs
nilekhc: Good catch! Refactored.

src/tls/crl.rs
Stevenjin8: can we use an Option type instead?

src/tls/certificate.rs
Stevenjin8: https://github.com/istio/ztunnel/pull/1660#discussion_r2478858479 This is where we talk about not including it in this PR.

src/tls/crl.rs
```diff
@@ -135,75 +146,69 @@ impl CrlManager {
             validated_ders.push(der_data);
         }
 
-        let crl_count_changed = validated_ders.len() != inner.crl_ders.len();
+        // use hash comparison to detect actual content changes, not just count changes
+        let crl_changed =
+            Self::compute_crl_hash(&validated_ders) != Self::compute_crl_hash(&inner.crl_ders);
```
Stevenjin8: This doesn't make much sense. If we hash we're going to have to iterate through all the contents each time at which point we might as well just compare the contents

src/tls/crl.rs
```diff
@@ -49,6 +56,7 @@ impl std::fmt::Debug for CrlManager {
 struct CrlManagerInner {
     crl_ders: Vec<Vec<u8>>,
```
Stevenjin8:
```suggestion
    crl_ders: Option<Vec<Vec<u8>>>,
```

src/tls/crl.rs
Stevenjin8: TBH I don't really think we have to look for updates at all.

Stevenjin8: we have a debouncer anyways

nilekhc: done

src/tls/crl.rs
nilekhc: done

src/tls/crl.rs
```diff
@@ -0,0 +1,303 @@
...
+
+            debug!(crl = idx + 1, "loaded crl");
+
+            validated_ders.push(der_data);
+        }
+
+        // store validated DER bytes
+        inner.crl_ders = Some(validated_ders);
+
+        debug!(
```
stevenctl, line 136: super nit: the logs could probably be combined into something like

```rust
 debug!(
      path = ?inner.crl_path,
      format = if is_pem { "PEM" } else { "DER" },
      count = validated_ders.len(),
      "crl loaded successfully"
  );
```

src/tls/crl.rs
```diff
@@ -0,0 +1,303 @@
...
+            inner._debouncer = Some(debouncer);
+        }
+
+        debug!("crl file watcher started successfully");
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
stevenctl, line 266: should we test the happy path here too?

src/proxyfactory.rs
```diff
@@ -55,13 +57,40 @@ impl ProxyFactory {
...
+                            crl validation will continue with current file, but \
+                            crl updates will require restarting ztunnel.",
+                            e
+                        );
+                    }
+
+                    Some(manager_arc)
+                }
+                Err(e) => {
+                    tracing::debug!("failed to initialize crl manager: {}", e);
```
stevenctl: this probably warrants higher than `debug` scope, someone set CURL_PATH and then the file was unreadable?

src/test_helpers/app.rs
```diff
@@ -65,6 +66,7 @@ impl From<(&Bound, Arc<SecretManager>)> for TestApp {
             tcp_dns_proxy_address: app.tcp_dns_proxy_address,
             udp_dns_proxy_address: app.udp_dns_proxy_address,
             cert_manager,
+            #[cfg(target_os = "linux")]
```
stevenctl, line 69: i think someone else already noted that we should probably clean these up before merge

src/tls/crl.rs
stevenctl: something like

```rust
// or pre-generated one for testdata
let crl_params = rcgen::CertificateRevocationListParams{ ... } ;
let crl = crl_param.signed_by(...);
let crl_pem = crl.pem().unwrap();

// write to file or wrap constructor for testability

let manager = CrlManager::new(...).expect("we can successfully parse CRL");

// check that it loaded one CRL

```

stevenctl (commented):
LGTM but we should have at least one happy path unit test for loading CRLs. An integration test would be nice. 


Something like:

  1. Generate client cert 
  2. Generate CRL revoking it 
  3. Create server with server_config(Some(crl_manager))
  4. Attempt TLS handshake -> assert failure

src/tls/crl.rs
```diff
@@ -311,4 +293,58 @@ mod tests {
...
+            rcgen::KeyUsagePurpose::KeyCertSign,
+            rcgen::KeyUsagePurpose::CrlSign,
+        ];
+
+        // create issuer from CA params and key
+        let issuer = Issuer::from_params(&ca_params, &ca_key_pair);
+
+        // create CRL with one revoked certificate
+        let crl_params = CertificateRevocationListParams {
+            this_update: time::OffsetDateTime::now_utc(),
```
Stevenjin8, line 321: Can we use `std::time::Instant::now()` instead? that way we don't add a whole crate just for a unit test

nilekhc: rcgen requires `time::OffsetDateTime` and it's not compatible with `std::time::Instant` or `std::time::SystemTime`. Also `time` is already a transitive dependency.

src/tls/crl.rs
nilekhc: done. Also removed some additional debug statements.

src/tls/crl.rs
nilekhc: added a test.

src/proxyfactory.rs
nilekhc: updated to warn.

src/test_helpers/app.rs
nilekhc: Yes. I'm going to remove all these isntances once I have the final approval.

src/tls/crl.rs
howardjohn: > Don't we create a new server_config for each new incoming hbone connection?

Ah! I stand corrected. thanks

Stevenjin8 (approved):

stevenctl (approved):
FWIW we could add a integration test directly in this repo (one that actually sets up the proxy and tries to send traffic)

PR #1664, opened 2025-11-11
removing CNAME record answer for wildcards

### LINKED ISSUE #1663: Ztunnel returning CNAME with wildcard domains
When ztunnel captures DNS requests that match wildcarded hosts, it returns a CNAME record with the wildcarded domain name and a canonical name also containing a wildcard. That shouldn't be the case according to RFCs 1034 and 1035. That causes issues with Ubuntu's standard resolver, where it doesn't go through the following A records when the CNAME mapping doesn't contain FQDNs.

grnmeira:
This PR removes special treating of wildcards for DNS answers with CNAME record.

When resolving a domain that matches a wildcarded service, for example, resolving www.example.com against service *.example.com, ztunnel would create CNAME records mapping a wildcarded host name to a wildcarded canonical name, which can cause applications to reject the answer. Curl on Ubuntu is an example of application that rejects an authoritative answer with CNAME and A records when the CNAME contains mappings that are not FQDNs.

This should fix #1663

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -672,7 +672,9 @@ impl Resolver for Store {
 
         // If the service was found by stripping off one of the search domains, create a
         // CNAME record to map to the appropriate canonical name.
-        if let Some(stripped) = service_match.alias.stripped && !service_match.name.is_wildcard(){
+        if let Some(stripped) = service_match.alias.stripped
+            && !service_match.name.is_wildcard()
+        {
             // Create a CNAME record to map from the requested name -> stripped name.
             records.push(cname_record(requested_name.clone(), stripped.clone()));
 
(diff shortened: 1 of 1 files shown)


src/dns/server.rs
```diff
@@ -677,32 +677,15 @@ impl Resolver for Store {
...
-                let cname_record_name = service_match
-                    .name
-                    .clone()
-                    .append_domain(&stripped.search_domain)
-                    .unwrap();
-                let canonical_name = service_match.name;
-                records.push(cname_record(cname_record_name, canonical_name));
-
-                // For wildcards, continue using the original requested hostname for IP records.
-            } else {
```
grnmeira, line 696: @ilrudie would have any insights of why we would do this? It looks like we used to map a wildcard name to another wildcard name. Is there any scenarios where that may be needed?

Issue #1694, opened 2025-12-15
Socket leak on inbound connection


Stevenjin8:
@dkulchinsky. To summarize, you have your falco daemonset export logs to an otel collector. When both are in the mesh and `keep-alive` is set to false, thre are socket leaks. correct?

dkulchinsky:
@Stevenjin8 wanted to add a couple of data points here:

1. The issue subject says `Socket leak on outbound connection`, but wanted to mention that the leaks are observed on the destination ztunnel instances (where otel collector is running), so I'd assuming this is happening on `inbound`?

2. We have expanded Istio ambient deployment into few additional clusters that are based on GKE, they have a largely similar setup (identical application stack), however we are not observing this issue on these clusters.

Stevenjin8:
@dkulchinsky yep thats correct. Sorry we couldn't get it in. It was a security release so we wanted to move relatively fast.

Stevenjin8:
@dkulchinsky BTW we just released 1.28.3

PR #1695, opened 2025-12-16
enhance ztunnel metrics

lcopi:
Enhances ztunnel metrics for debugging:
~- Adds direction (inbound/outbound) to the connection metrics~
- adds a connection failed metric
- adds socket open/closed metrics 

The socket metrics could potentially help debug issues like https://github.com/istio/ztunnel/issues/1694

Edit: I removed the direction label for the metrics because `reporter[source|destination]` is sufficient

### DIFF
--- a/src/proxy/metrics.rs
+++ b/src/proxy/metrics.rs
@@ -578,7 +578,7 @@ impl ConnectionResult {
     }
 
     // Record our final result, with more details as a response flag.
-    pub fn record_with_flag<E: std::error::Error>(
+    pub fn record_with_flag<E: std::error::Error + 'static>(
         mut self,
         res: Result<(), E>,
         flag: ResponseFlags,
@@ -588,7 +588,7 @@ impl ConnectionResult {
     }
 
     // Record our final result.
-    pub fn record<E: std::error::Error>(mut self, res: Result<(), E>) {
+    pub fn record<E: std::error::Error + 'static>(mut self, res: Result<(), E>) {
         // If no specific flag was set and we have an error, try to infer the failure reason
         if self.tl.response_flags == ResponseFlags::None
             && let Err(ref err) = res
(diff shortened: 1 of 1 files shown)


src/proxy/inbound_passthrough.rs
```diff
@@ -189,6 +201,7 @@ impl InboundPassthrough {
                 destination: Some(upstream_workload),
                 connection_security_policy: metrics::SecurityPolicy::unknown,
                 destination_service: ds,
+                direction: Direction::proxy,
```
stevenctl: It's a little confusing or unclear what `proxy` means as a direction. This will always have the socket metric marked inbound but the direction may be proxy?

src/proxy/inbound.rs
```diff
@@ -108,7 +108,20 @@ impl Inbound {
...
+
                 let serve_client = async move {
+                    let _socket_guard = metrics::SocketCloseGuard::new(
+                        metrics_for_socket_close,
+                        Reporter::destination,
+                        metrics::Direction::inbound,
+                    );
                     let tls = match acceptor.accept(raw_socket).await {
                         Ok(tls) => tls,
                         Err(e) => {
```
stevenctl, line 125: we'll drop the guard here so we track close, but we could record something that indicates why it closed in error cases

src/proxy/inbound_passthrough.rs
```diff
@@ -76,7 +76,19 @@ impl InboundPassthrough {
                 let pi = self.pi.clone();
                 match socket {
                     Ok((stream, remote)) => {
+                        let socket_labels = metrics::CommonTrafficLabels::for_socket(
+                            Reporter::destination,
+                            metrics::Direction::inbound,
```
stevenctl: super nit: you have `Direction` imported, no need to quality with `metrics::`

src/proxy/metrics.rs
```diff
@@ -534,6 +621,10 @@ impl ConnectionResult {
         // Unconditionally record the connection was closed
         self.metrics.connection_close.get_or_create(tl).inc();
 
+        if tl.response_flags == ResponseFlags::ConnectionFailure {
```
stevenctl: should things like TLS failures (We already `log_early_deny` or RBAC failures be included here?

src/proxy/metrics.rs
```diff
@@ -330,12 +345,84 @@ impl Metrics {
...
         );
 
+        let connection_failures = Family::default();
+        registry.register(
+            "tcp_connections_failed",
+            "The total number of TCP connections that failed to establish (unstable)",
+            connection_failures.clone(),
+        );
+
+        let socket_opens = Family::default();
```
stevenctl: rather than total opens/closes, would a gauge of tcp_sockets_active make sense?

stevenctl: depending on what exactly we want to debug using these metrics

Stevenjin8: We currently have open file descriptors. Out of curiousity, is there a case where these two metrics would be substantially difference?

lcopi: In the steady state healthy case the socket count should be roughly equivalent to the connection count. In the case where a socket leak exists, it will trend somewhere between the open file descriptor metric and the connection metric. Probably the only benefit from the socket metric is in having the reporter label to indicate direction. 

I adjusted the socket metric to be a gauge in https://github.com/istio/ztunnel/pull/1695/commits/9c2f9012319b7a3cd03ef70750c58764104373aa but I can remove it completely if we don't think it is providing enough value

src/proxy/inbound_passthrough.rs
lcopi: fixed in https://github.com/istio/ztunnel/pull/1695/commits/bf871581908b379e4dd3570f1b5ecc72b2fba761

src/proxy/inbound_passthrough.rs
lcopi: changed to inbound in https://github.com/istio/ztunnel/pull/1695/commits/bf871581908b379e4dd3570f1b5ecc72b2fba761

src/proxy/metrics.rs
lcopi: expanded the applicable failures in https://github.com/istio/ztunnel/pull/1695/commits/ce9038a5c635589f90b7385dec22f00322e6c27b

src/proxy/inbound.rs
lcopi: I adjusted the socket metric to be a gauge so I don't know if it makes sense to include any error codes as metric labels now

src/proxy/inbound_passthrough.rs
lcopi: I ended up removing direction completely since reporter covers it

src/proxy/inbound_passthrough.rs
lcopi: I ended up removing direction completely since it is covered by reporter

src/proxy/metrics.rs
```diff
@@ -519,9 +589,41 @@ impl ConnectionResult {
...
+        if self.tl.response_flags == ResponseFlags::None
+            && let Err(ref err) = res
+        {
+            self.tl.response_flags = Self::extract_failure_reason(err);
+        }
         self.record_internal(res)
     }
 
+    // Extract failure reason from error type
+    fn extract_failure_reason(err: &dyn std::error::Error) -> ResponseFlags {
```
stevenctl: can we downcast the error? the string matching approach seems potentially brittle

lcopi: changed to error downcasting in https://github.com/istio/ztunnel/pull/1695/commits/be79f9aaa21387e45e4488197f0297f28c8cdcc8

PR #1704, opened 2025-12-29
canonical wds service

ilrudie:
implement the ztunnel side of https://github.com/istio/istio/pull/58576

This improves how names are resolved in ambient mode. It allows a wds service to be marked as canonical, at which point it will be used whenever there is not a matching hostname defined in the same namespace as the client..

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -1435,6 +1435,12 @@ mod tests {
                 expect_records: vec![a(n("canonical.svc."), ipv4("10.10.10.141"))],
                 ..Default::default()
             },
+            Case {
+                name: "success: namespace-local service should be preferred over canonical",
+                host: "canonical.with.local",
+                expect_records: vec![a(n("canonical.with.local."), ipv4("10.10.10.150"))],
+                ..Default::default()
+            },
         ];
 
         // Create and start the proxy.
@@ -1760,6 +1766,17 @@ mod tests {
                 "canonical.svc",
                 NS3,
                 &[na(NW1, "10.10.10.141")],
+            ),
+            // Client in NS1 should prefer local over canonical
+            xds_namespaced_external_service(
+                "canonical.with.local",
+                NS1,
+                &[na(NW1, "10.10.10.150")],
+            ),
+            xds_namespaced_external_canonical_service(
+                "canonical.with.local",
+                NS2,
+                &[na(NW1, "10.10.10.151")],
             ),
             with_fqdn(
                 "details.ns2.svc.cluster.remote",


src/dns/server.rs
```diff
@@ -390,21 +410,34 @@ impl Store {
...
-                    Some(service) => Some(service),
-                    None => match self.prefered_service_namespace.as_ref() {
-                        Some(prefered_namespace) => services.iter().find_or_first(|service| {
-                            service.namespace == prefered_namespace.as_str()
-                        }),
-                        None => services.first(),
-                    },
-                };
+                    .fold_while(MatchReason::None, |r, s| {
+                        if s.namespace == client.namespace {
```
Stevenjin8, line 416: namespace takes precedence over canonical, so we want to check it first

proto/workload.proto
```diff
@@ -89,6 +89,8 @@ message Service {
 
   // Extension provides a mechanism to attach arbitrary additional configuration to an object.
   repeated Extension extensions = 10;
+
+  bool canonical = 11;
```
keithmattix, line 95: Can/should we document what this means

ilrudie: https://github.com/ilrudie/istio-istio/blob/7cfedf51998d0286eaf98f428680230b17be32f7/pkg/workloadapi/workload.proto#L95 has an attempt. I can move it over if we like that.

src/dns/server.rs
```diff
@@ -187,6 +187,26 @@ impl Server {
     }
 }
 
+enum MatchReason<'a> {
+    Canonical(&'a Arc<Service>),
+    First(&'a Arc<Service>),
+    Namespace(&'a Arc<Service>),
+    PreferredNamespace(&'a Arc<Service>),
+    None,
+}
```
Stevenjin8, line 196: This is definitely not a blocker, but more of a neat idea I had. If you implemement `Ord` (and `PartialOrd`) for this:

```rs
impl<'a> Ord for MatchReason<'a> {
    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
        use MatchReason::*;
        let self_rank: i32 = match self {
            Canonical(_) => 3,
            PreferredNamespace(_) => 2,
            Namespace(_) => 1,
            First(_) => 0,
        };
        let other_rank: i32 = match other {
            Canonical(_) => 3,
            PreferredNamespace(_) => 2,
            Namespace(_) => 1,
            First(_) => 0,
        };
        self_rank.cmp(&other_rank)
    }
}
```

then you can do

```rs
let service: Option<&Arc<Service>> = services Vec<Arc<Service>>
    .iter() Iter<'_, Arc<Service>>
    .map(|s: &Arc<Service>| {
        if s.namespace == client.namespace {
            MatchReason::Namespace(s)
        } else if s.canonical {
            MatchReason::Canonical(s)
        } else if let Some(preferred_ns: &String) = &self.prefered_service_namespace
            && s.namespace == *preferred_ns
        {
            MatchReason::PreferredNamespace(s)
        } else {
            MatchReason::First(s)
        }
    }) impl Iterator<Item = MatchReason<'_>>
    .max() Option<MatchReason<'_>>
    .map(|mr: MatchReason<'_>| mr.into());

```

It also lets you get rid of the `None` variant in the enum. The downside is I'm not sure if `max` will short circuit.

proto/workload.proto
stevenctl: I think it would be good to have that copied over as this is where the behavior is actually implemented. 

The comment could be slightly more explicit: 

> canonical marks this Service as taking priority during hostname lookups, when there is not a match in the namespace of the client.

src/dns/server.rs
```diff
@@ -390,21 +410,34 @@ impl Store {
...
-                        None => services.first(),
-                    },
-                };
+                    .fold_while(MatchReason::None, |r, s| {
+                        if s.namespace == client.namespace {
+                            itertools::FoldWhile::Done(MatchReason::Namespace(s))
+                        } else if s.canonical {
+                            itertools::FoldWhile::Continue(MatchReason::Canonical(s))
+                        } else {
+                            // TODO: I vote we deprecate prefered_service_namespace
```
stevenctl: +1, but can you put your name on this TODO or link an issue to it?

src/dns/server.rs
```diff
@@ -390,21 +410,34 @@ impl Store {
...
-                        Some(prefered_namespace) => services.iter().find_or_first(|service| {
-                            service.namespace == prefered_namespace.as_str()
-                        }),
-                        None => services.first(),
-                    },
-                };
+                    .fold_while(MatchReason::None, |r, s| {
+                        if s.namespace == client.namespace {
+                            itertools::FoldWhile::Done(MatchReason::Namespace(s))
+                        } else if s.canonical {
```
stevenctl, line 418: having multiple marked canonical is UB right? do we have sorting or anything on the list so it would be consistent? 

A good control plane would probably not send multiple marked canonical of course.

src/state/service.rs
```diff
@@ -55,6 +55,9 @@ pub struct Service {
 
     #[serde(default, skip_serializing_if = "is_default")]
     pub ip_families: Option<IpFamily>,
+
+    #[serde(default)]
```
stevenctl, line 59: should this also have skip serializing on default?

src/dns/server.rs
```diff
@@ -1394,6 +1428,12 @@ mod tests {
                 expect_records: vec![a(n("everywhere.io."), ipv4("10.10.10.112"))],
                 ..Default::default()
             },
+            Case {
```
stevenctl, line 1432: we should have a test case that NS local can win out over one marked canonical

src/dns/server.rs
ilrudie: Yeah, UB and a good control plane should not send this.

We don't have consistent sorting so IF you have two services marked as canonical (control plane bug) and IF you have two ztunnels who's wds input ordering is differed enough that their data structures are ordered differently you could have inconsistent behaviors node-to-node. This isn't actually any worse than how it was when we just did "match namespace or first" since "first" was essentially the same UB. I think it's probably ok but maybe we can add a debug_assertion? Would our integ test builds run debug assertions or are they optimized builds already?

src/state/service.rs
ilrudie: To my mind it would be nice to serialize `canonical: false` here instead of it being omitted. I can change if we feel strongly.

proto/workload.proto
ilrudie: Added the comment, also updated i/i PR so they match

src/dns/server.rs
ilrudie: issue opened and linked into the TODO, TY

PR #1728, opened 2026-01-17
Add x forwarded network header

### LINKED ISSUE #1725: ztunnel should send an x-origin-source header in inner connect requests
This header will contain the network ztunnel is running in. This is to help destination envoys know not to use metadata discovery

keithmattix:
Successor of #1727 in light of our forthcoming AI policy. Feel free to refer to that PR to see the prompt and iterations. 

Fixes #1725

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -31,7 +31,8 @@ use crate::drain::DrainWatcher;
 use crate::proxy::h2::server::{H2Request, RequestParts};
 use crate::proxy::metrics::{ConnectionOpen, Reporter};
 use crate::proxy::{
-    BAGGAGE_HEADER, ProxyInputs, TRACEPARENT_HEADER, TraceParent, X_FORWARDED_NETWORK_HEADER, metrics,
+    BAGGAGE_HEADER, ProxyInputs, TRACEPARENT_HEADER, TraceParent, X_FORWARDED_NETWORK_HEADER,
+    metrics,
 };
 use crate::rbac::Connection;
 use crate::socket::to_canonical;


src/proxy/inbound.rs
```diff
@@ -393,11 +395,13 @@ impl Inbound {
         // We may need a more explicit indicator in the future.
         // Note: previously this attempted to check that the src identity was equal to the Gateway;
         // this check is broken as the gateway only forwards an HBONE request, it doesn't initiate it itself.
-        let from_gateway = matches!(hbone_addr, HboneAddress::SvcHostname(_, _));
+        let from_gateway = req.headers().get(X_ORIGIN_SOURCE_HEADER).is_some();
```
Stevenjin8: Should we also check the value here? Check that its in a different network

src/proxy/outbound.rs
```diff
@@ -312,7 +316,14 @@ impl OutboundConnection {
                 FORWARDED,
                 build_forwarded(remote_addr, &req.intended_destination_service),
             )
-            .header(TRACEPARENT_HEADER, self.id.header())
+            .header(TRACEPARENT_HEADER, self.id.header());
+
+        // Add x-istio-origin-network header for inner CONNECT requests in double HBONE
+        if let Some(network) = origin_network {
+            builder = builder.header(X_ORIGIN_SOURCE_HEADER, network.as_str());
+        }
```
Stevenjin8, line 324: I don't understand why we need to pass `network` as an argument. Can't we grab use `self.pi.cfg.network` here?

Stevenjin8: oh I see, its to control whether to add it or not in the single vs double hbone path

Stevenjin8: I still think its better to just always add it and make our `from_gateway` determination check if the x-istio-origin-network header is different than the local one.

Stevenjin8: This are minor nits. Non blocking

keithmattix: The idea was that single-cluster would have zero overhead increase with this. Realistically, it's never more than a handful of bytes, but being able to say the approach is zero cost for single cluster/network is less controversial IMO

src/proxy/inbound.rs
```diff
@@ -393,11 +395,18 @@ impl Inbound {
         // We may need a more explicit indicator in the future.
         // Note: previously this attempted to check that the src identity was equal to the Gateway;
         // this check is broken as the gateway only forwards an HBONE request, it doesn't initiate it itself.
-        let from_gateway = matches!(hbone_addr, HboneAddress::SvcHostname(_, _));
+        let from_gateway = req
```
ilrudie, line 399: Idea here, more than request for immediate change. If we don't have the header, we're assuming that we are in the same network and it's OK to perform a lookup. We may actually just be in a version mismatch where the source cluster doesn't know it's supposed to send this.
Should we consider always adding src network header so that we know that not seeing this header is a version mismatch (which means we should not perform a lookup, because all local ztunnels are likely the same version modulo a temporary upgrade?)

Stevenjin8: https://github.com/istio/ztunnel/pull/1728#discussion_r2712822135 :)

src/proxy/inbound.rs
```diff
@@ -393,11 +395,18 @@ impl Inbound {
...
+            .get(X_ORIGIN_SOURCE_HEADER)
+            .and_then(|h| h.to_str().ok())
+            .map(|s| s.eq_ignore_ascii_case(&pi.cfg.network))
+            .unwrap_or(false);
+
         if from_gateway {
             debug!("request from gateway");
         }
         let source = match from_gateway {
+            // TODO: use baggage instead of looking up by IP
```
ilrudie: Sorry, why do this? We query baggage for derived_source. Then when building labels we apply derived first and then override with _this_ source so we'd be overriding baggage with baggage.

src/proxy/inbound.rs
keithmattix: Once we get ambient multicluster to beta (hopefully this release 🤞🏾), I think moving to always sending the headers makes sense. As Steven pointed out, the goal was to make this zero-cost for single-cluster at this juncture. I don't know how I feel about the absence of the header being a signal to not use workload discovery. It feels a bit too implicit to me (specifically the part about assuming all ztunnels in the cluster are at the same version).

src/proxy/inbound.rs
keithmattix: Oh thanks for clarifying; I didn't see the end to end flow. I'll remove.

src/proxy/inbound.rs
ilrudie: Seems a fair point. Optimize around single network and accept the risk in multi-network that you may get some mangled telemetry if there is an address collision and the source zt is not new enough to implement this behavior.

src/proxy.rs
```diff
@@ -558,6 +558,7 @@ pub struct TraceParent {
 
 pub const BAGGAGE_HEADER: &str = "baggage";
 pub const TRACEPARENT_HEADER: &str = "traceparent";
+pub const X_ORIGIN_SOURCE_HEADER: &str = "x-istio-origin-network";
```
ilrudie: bikeshed
```suggestion
pub const X_ORIGIN_NETWORK_HEADER: &str = "x-istio-origin-network";
```

keithmattix: Good call; missed this in a refactor

src/proxy/inbound.rs
```diff
@@ -393,7 +395,13 @@ impl Inbound {
         // We may need a more explicit indicator in the future.
         // Note: previously this attempted to check that the src identity was equal to the Gateway;
         // this check is broken as the gateway only forwards an HBONE request, it doesn't initiate it itself.
-        let from_gateway = matches!(hbone_addr, HboneAddress::SvcHostname(_, _));
+        let from_gateway = req
+            .headers()
+            .get(X_ORIGIN_NETWORK_HEADER)
```
ilrudie: I think the logic is inverted. We should be "from_gateway" when header value != our network, yeah?

Related, we can get away with not comparing the values. The way this is setup if the header exists at all, we're from a remote network

keithmattix: Good catch; this logic is definitely inverted. I feel like we should actually compare the values though since I think it's likely that we get to a point where we always send the header

Issue #1625, opened 2025-09-01
Opentelemetry support


JorTurFer:
Hello

I'm checking the configuration docs and I can't see any information about sending logs via opentelemetry (and the code doesn't use otel either). 
Currently we have to use file reader exporters to expose the logs from ztunnel (we find them quite useful) and it'd be nice if sending them via opentelemetry (like other istio components do) is an option. I have no idea about rust so maybe this is a stupid thing, but I'd say that there is stable support for rust SDK and logs -> https://github.com/open-telemetry/opentelemetry-rust 

Of course, the most telemetry the best, and if adding traces and metrics via otel is possible, why no, but at least logs would be a nice addition

JorTurFer:
Would you accept a contribution to configure ztunnel for opentelemetry protocol?

mzacho:
We may be able to use the `opentelemetry-appender-tracing` crate, to preserve the current plain/json logging to stdout via `tracing_appender`, as well as format `tracing` events as OTLP LogRecords and ship to otel collector.

From the docs:

> Integrates as a [Layer](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/trait.Layer.html) from [tracing-subscriber](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/), allowing to be used alongside other tracing layers, such as fmt

https://docs.rs/opentelemetry-appender-tracing/0.31.1/opentelemetry_appender_tracing/

PR #1733, opened 2026-01-26
OpenTelemetry logging

mzacho:
This PR adds as the ability to send tracing log Events to an OpenTelemetry collector via gRPC. When `OTLP_LOGGING=true` it uses [opentelemetry-tracing-appender](https://docs.rs/opentelemetry-appender-tracing/latest/opentelemetry_appender_tracing/) to add a OTEL tracing bridge as a layer to the current tracing subscriber registry. The current plain/json logging to stdout is not modified, so with this change ztunnel can either log:

1. to stdout using the plain format
2. to stdout using the plain format AND to otel collector in OTLP format
3. to stdout using the json format
4. to stdout using the json format AND to otel collector in OTLP format

The collector endpoint can be configured with `OTEL_COLLECTOR_ENDPOINT`.

cc @ilrudie, you mentioned in #1625 you'd be curios to see how intrusive OpenTelemetry support is.


src/telemetry.rs
```diff
@@ -76,12 +85,69 @@ fn fmt_layer(writer: NonBlocking) -> Box<dyn Layer<Registry> + Send + Sync + 'st
...
+    Box::new(otel_layer.with_filter(filter_otel))
+}
+
+#[cfg(feature = "otel")]
+fn otlp_logger_provider() -> opentelemetry_sdk::logs::SdkLoggerProvider {
+    use opentelemetry_otlp::{LogExporter, WithExportConfig};
+    use opentelemetry_sdk::logs::SdkLoggerProvider;
+
+    let mut builder = LogExporter::builder()
+        .with_tonic()
```
howardjohn: A very common use case is going to be wanting to connect to an in-cluster collector endpoint using mesh mtls. using the out-of-the-box client will not support this case.

src/telemetry.rs
```diff
@@ -76,12 +85,69 @@ fn fmt_layer(writer: NonBlocking) -> Box<dyn Layer<Registry> + Send + Sync + 'st
...
+    // (https://github.com/open-telemetry/opentelemetry-rust/issues/2877),
+    // filtering like this is the best way to suppress such logs.
+    //
+    // The filter levels are set as follows:
+    // - Allow `trace` level and above by default.
+    // - Completely restrict logs from `hyper`, `tonic`, `h2`, and `reqwest`.
+    //
+    // Note: This filtering will also drop logs from these components even when
+    // they are used outside of the OTLP Exporter.
+    let filter_otel = EnvFilter::new("trace")
```
howardjohn: generally I don't think that we want to just blindly send (even a subset) logs to otel. We probably only want access logs, and we want those to follow otel semantic conventions. This requires specialized calls specifically for otel. I wouldn't be surprised if it ends up not even using `tracing` at all.

src/telemetry.rs
mzacho: That makes sense. I don't know how to implement it though.

mzacho: Any pointers/ ideas?

src/telemetry.rs
rbkrabbe: I would want the same logs that are currently logged to stdout to be available in OTEL. It seems weird to require that people run a file log exporter in order to get logs.

But I agree with the OTEL semantic conventions.

PR #1734, opened 2026-01-26
Ambient Multicluster Telemetry

keithmattix:
Ztunnel side of ambient multicluster telemetry. Sends baggage in response headers and uses incoming baggage as a source for metrics.

Please reference the [design doc](https://docs.google.com/document/d/1gyDsViXTHlozTbHFuggNmYFrL6P3om4ZvTMdy_-Ul74/edit?tab=t.0#heading=h.xw1gqgyqs5b) for a refresher on the design

### DIFF
--- a/src/baggage.rs
+++ b/src/baggage.rs
@@ -31,13 +31,34 @@ pub struct Baggage {
 
 pub fn baggage_header_val(baggage: &Baggage, workload_type: &str) -> String {
     [
-        baggage.cluster_id.as_ref().map(|cluster| format!("k8s.cluster.name={cluster}")),
-        baggage.namespace.as_ref().map(|namespace| format!("k8s.namespace.name={namespace}")),
-        baggage.workload_name.as_ref().map(|workload| format!("k8s.{workload_type}.name={workload}")),
-        baggage.service_name.as_ref().map(|service| format!("service.name={service}")),
-        baggage.revision.as_ref().map(|revision| format!("service.version={revision}")),
-        baggage.region.as_ref().map(|region| format!("cloud.region={region}")),
-        baggage.zone.as_ref().map(|zone| format!("cloud.availability_zone={zone}")),
+        baggage
+            .cluster_id
+            .as_ref()
+            .map(|cluster| format!("k8s.cluster.name={cluster}")),
+        baggage
+            .namespace
+            .as_ref()
+            .map(|namespace| format!("k8s.namespace.name={namespace}")),
+        baggage
+            .workload_name
+            .as_ref()
+            .map(|workload| format!("k8s.{workload_type}.name={workload}")),
+        baggage
+            .service_name
+            .as_ref()
+            .map(|service| format!("service.name={service}")),
+        baggage
+            .revision
+            .as_ref()
+            .map(|revision| format!("service.version={revision}")),
+        baggage
+            .region
+            .as_ref()
+            .map(|region| format!("cloud.region={region}")),
+        baggage
+            .zone
+            .as_ref()
+            .map(|zone| format!("cloud.availability_zone={zone}")),
     ]
     .into_iter()
     .flatten()
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -220,7 +220,9 @@ impl Inbound {
                 // At this point in processing, we never built up full context to log a complete access log.
                 // Instead, just log a minimal error line.
                 metrics::log_early_deny(src, dst, Reporter::destination, e);
-                if let Err(err) = req.send_error(build_response(code, None, pi.cfg.enable_enhanced_baggage)) {
+                if let Err(err) =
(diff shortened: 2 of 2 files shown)


src/baggage.rs
```diff
@@ -29,6 +29,22 @@ pub struct Baggage {
     pub zone: Option<Strng>,
 }
 
+#[allow(clippy::too_many_arguments)]
```
stevenctl: could avoid the `allow` and add an easy test by making this accept a `Baggage` and assert that 


```rust
let baggage = Baggage{...};
let formatted = baggage_header_val(baggage);
let parsed = parse_baggage_header(formatted);
assert(formatted == baggage);
```

src/baggage.rs
stevenctl, line 103: not part of your PR, but I think `baggage_parser_empty_values` is weird... why do we call `parse_baggage_header` before we append the to `hm`?

src/proxy/inbound.rs
```diff
@@ -445,19 +453,22 @@ impl Inbound {
             ConnectionOpen {
                 reporter: Reporter::destination,
                 source,
-                derived_source: Some(derived_source),
-                destination: Some(destination_workload),
+                derived_source: Some(derived_source.clone()),
```
stevenctl: i don't think we need to clone derived source

src/baggage.rs
krinkinmu: Yes, that was a mistake. Fixed now.

src/proxy/inbound.rs
krinkinmu: Indeed, we can move here - fixed in the latest revision.

src/baggage.rs
krinkinmu: Re-worked the code a bit and added a test to check that we can parse the values we generate here.

src/baggage.rs
```diff
@@ -29,20 +29,37 @@ pub struct Baggage {
...
-    format!(
-        "k8s.cluster.name={cluster},k8s.namespace.name={namespace},k8s.{workload_type}.name={workload_name},service.name={name},service.version={version},cloud.region={region},cloud.availability_zone={zone}",
-    )
+pub fn baggage_header_val(baggage: &Baggage, workload_type: &str) -> String {
+    let mut items = Vec::new();
+    baggage
+        .cluster_id
+        .as_ref()
+        .inspect(|cluster| items.push(format!("k8s.cluster.name={cluster}")));
+    baggage
```
stevenctl: super nit: you could do something like

```rust
[
  baggage.property.as_ref().map(|v| format!("k8s.property.key={v}")),
  ...
]
.into_iter()
.flatten()
.collect::<Vec<_>>()
.join(",")
```

src/proxy/outbound.rs
```diff
@@ -209,91 +211,131 @@ impl OutboundConnection {
...
+                revision: baggage.revision,
+            });
+            Result::<_, Error>::Ok((derived_workload, drain_tx, inner_upgraded))
+        })
         .await;
 
-        let _ = drain_tx.send(true);
+        match res {
+            Err(e) => {
+                let connection_stats = Box::new(connection_stats_builder.build());
```
stevenctl: I don't think we need `Box` (The `Ok` branch doesn't have it)

src/proxy/outbound.rs
```diff
@@ -209,91 +211,131 @@ impl OutboundConnection {
...
         req: &Request,
-        connection_stats: &ConnectionResult,
-    ) -> Result<(), Error> {
-        let upgraded = Box::pin(self.send_hbone_request(remote_addr, req)).await?;
-        copy::copy_bidirectional(copy::TcpStreamSplitter(stream), upgraded, connection_stats).await
+        connection_stats_builder: Box<ConnectionResultBuilder>,
+    ) {
+        let connection_stats = Box::new(connection_stats_builder.build());
+        let res = (async {
+            let (upgraded, _) = Box::pin(self.send_hbone_request(remote_addr, req)).await?;
```
stevenctl, line 333: So here, we intentionally discard baggage from the response and don't use it to enrich any metrics, with the assumption that when we send (inner/single) hbone that we know everything already as a client, correct?

krinkinmu: That's pretty much correct.

In double hbone the actual destination workload is an E/W gateway and that's not the right workload to use for the metric labels, so we have to rely on the baggage in case of double HBONE. For plain HBONE, we could use baggage or could use info we already have in WDS.

We want to limit impact of these changes to single-cluster deployments, so given that it makes sense to keep the HBONE connections as is and only use baggage for double-HBONE for now.

Hypothetically, in the future we could offer baggage-based peer metadata discovery as an alternative to xDS-based, but even if we do that, we still will have to push WDS to ztunnels, so in ztunnels specifically there is little to be gained from that.

stevenctl (approved):

PR #1738, opened 2026-01-28
Address review comments

krinkinmu:
The fixes include:

1. Rework of the baggage_header_val to take Baggage structure and workload type as input
2. Adding some tests to make sure that the baggage value we generate can be parsed by our code correctly
3. Fix one of the tests for baggage parsing function

Other than that, I reworked the code to not include items with empty values into baggage string.

And finally, in a few places we took cluster from the proxy config when we already had workload available - I removed that and now cluster always comes from the workload structure.

+cc @keithmattix @grnmeira @Stevenjin8

### DIFF
--- a/src/baggage.rs
+++ b/src/baggage.rs
@@ -18,7 +18,7 @@ use hyper::{
     http::HeaderValue,
 };
 
-#[derive(Default)]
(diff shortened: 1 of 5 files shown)


keithmattix:
Pushing through since this is a part of a PR review for #1734

PR #1745, opened 2026-01-29
dry-run no allow policies match

ilrudie:
try out handling the dry-run case where allows exist, but none matched so if they became enforced we would deny this conn

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -119,7 +119,9 @@ const ENABLE_ENHANCED_BAGGAGE: &str = "ENABLE_RESPONSE_BAGGAGE";
 
 /// When true, authorization policy logs are emitted at INFO level instead of DEBUG.
 pub static AUTHZ_POLICY_INFO_LOGGING: once_cell::sync::Lazy<bool> =
-    once_cell::sync::Lazy::new(|| env::var("AUTHZ_POLICY_INFO_LOGGING").unwrap_or_default() == "true");
+    once_cell::sync::Lazy::new(|| {
+        env::var("AUTHZ_POLICY_INFO_LOGGING").unwrap_or_default() == "true"
+    });
 
 #[derive(serde::Serialize, Clone, Debug, PartialEq, Eq)]
 pub enum RootCert {


src/state.rs
```diff
@@ -592,11 +592,18 @@ impl DemandProxyState {
...
         for pol in allow_dry_run.iter() {
             if pol.matches(conn) {
+                dry_run_allow_matched = true;
                 debug!(policy = pol.to_key().as_str(), "dry-run allow policy match");
             }
         }
+        if allow.is_empty() && !allow_dry_run.is_empty() && !dry_run_allow_matched {
+            // this is going to be an allow, but the conn would be denied if dry-run policies
+            // became enforced because none matched
+            debug!("dry-run no allow policies match");
```
Stevenjin8:
```suggestion
            debug!("no dry-run allow policies match");
```

We should mention that is is because there are no allow policies, so request would have been denied. That said, I can't think of a concise way of saying this.

ilrudie: I put "dry-run" first intentionally to follow the form of the other ones ("dry-run " + verbiage-from-not-dry-run-message). It's grammatically, not good though. I agree. Could go either way on it TBH but the way it is was intentional.

find-arka: Would this be the message when we have a _allow nothing empty spec_  AuthorizationPolicy with dry-run specific annotation?
```yaml
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  name: allow-nothing-httpbin-dry-run
  namespace: httpbin
  annotations:
    "istio.io/dry-run": "true"
spec:
  {}
```

ilrudie: Yes, but only if there are no enforced allow policies. My logic here is: if there are already enforced allows, then the only behavior change would potentially come from a dry-run which does match (assuming that no existing allow policies match). The existence of 1 or more enforced allows already put us in a deny if no matches state, so enforcing new policies which also don't match is a non-event.

find-arka: What about: `dry-run: no allow policy matched, enforcing would block this connection` ?

ilrudie: Do we want to also change the format of the others to be `dry-run:` then?

find-arka: That would be good for consistency I think.

find-arka: Is there any disadvantage of setting the log level to `info`?
The only use case I can imagine for dry run AuthorizationPolicy is user checking dynamically if a policy is going to impact in the desired manner or not. If the log level is `info` user can quickly verify that by looking at the logs without changing the log level on ztunnel.

Stevenjin8: TBH i agree with this. The only reason I had it as debug is because in envoy, we emit them as debug logs.

ilrudie: Yeah, I was thinking about this too. Debug is an annoyance we inherited from the envoy impl. Since this missed the boat for 1.29 I was kind of thinking about a small reworking to allow configuring dry-run log level. That way we can have debug for backwards compatibility but allow the flexibility to move them to another log level that is more convenient.

I'll work something up. Thanks for the feedback

src/proxy/metrics.rs
```diff
@@ -487,6 +487,19 @@ pub fn log_early_deny<E: std::error::Error>(
     );
 }
 
+/// Logs a message at INFO level if AUTHZ_POLICY_INFO_LOGGING is set to "true",
+/// otherwise logs at DEBUG level.
+#[macro_export]
+macro_rules! authpol_log {
```
ilrudie, line 493: todo: is this the best spot?

close to access_log is a plus

named metrics.rs is a minus

Stevenjin8: > named metrics.rs is a minus

same for access_log, so its kinda whatever.

src/config.rs
```diff
@@ -117,6 +117,12 @@ const PROXY_MODE_SHARED: &str = "shared";
 const LOCALHOST_APP_TUNNEL: &str = "LOCALHOST_APP_TUNNEL";
 const ENABLE_ENHANCED_BAGGAGE: &str = "ENABLE_RESPONSE_BAGGAGE";
 
+/// When true, authorization policy logs are emitted at INFO level instead of DEBUG.
+pub static AUTHZ_POLICY_INFO_LOGGING: once_cell::sync::Lazy<bool> =
+    once_cell::sync::Lazy::new(|| {
+        env::var("AUTHZ_POLICY_INFO_LOGGING").unwrap_or_default() == "true"
+    });
```
Stevenjin8, line 124: But TBH, dry run logs is alpha, so don't think we really have to worry about this at all as long as we are clear about it in our upgrade notes.

ilrudie: I thought it was kind of rough that only dry-run logs would be fixed, so this impacts all authz policy logging and hence, is configurable with a default to using debug.

We could of course leave the old logs at debug and only allow config for dry-runs (kind of ugly IMO) or we could allow them to be configured independently (kind of complex, although sometimes Istio is complex)

Stevenjin8: oh that makes sense

PR #1746, opened 2026-01-30
prioritize canonical services on inbound

stevenctl:
We sometimes do hostname lookups on the inbound side. These had only a simple "is the service in the client's namespace" check. This expands it to match the behavior of the DNS server. 

I also refactor that prioritization code from the DNS server to live in the ServiceStore

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -12,15 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use hickory_proto::ProtoErrorKind;
 use hickory_proto::op::ResponseCode;
 use hickory_proto::rr::rdata::{A, AAAA, CNAME};
 use hickory_proto::rr::{Name, RData, Record, RecordType};
-use hickory_proto::ProtoErrorKind;
 use hickory_resolver::config::{NameServerConfig, ResolverConfig, ResolverOpts};
 use hickory_resolver::system_conf::read_system_conf;
+use hickory_server::ServerFuture;
 use hickory_server::authority::LookupError;
 use hickory_server::server::Request;
-use hickory_server::ServerFuture;
 use once_cell::sync::Lazy;
 use rand::rng;
 use rand::seq::SliceRandom;
@@ -45,10 +45,10 @@ use crate::dns::resolver::{Answer, Resolver};
 use crate::drain::{DrainMode, DrainWatcher};
 use crate::metrics::{DeferRecorder, IncrementRecorder, Recorder};
 use crate::proxy::Error;
-use crate::state::service::{IpFamily, Service, ServiceMatch};
-use crate::state::workload::address::Address;
-use crate::state::workload::Workload;
 use crate::state::DemandProxyState;
+use crate::state::service::{IpFamily, Service, ServiceMatch};
+use crate::state::workload::Workload;
+use crate::state::workload::address::Address;
 use crate::strng::Strng;
 use crate::{config, dns};
 
--- a/src/state/service.rs
+++ b/src/state/service.rs
@@ -23,10 +23,10 @@ use tracing::trace;
 use xds::istio::workload::Service as XdsService;
 
 use crate::state::workload::{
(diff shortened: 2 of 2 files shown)


src/state/service.rs
```diff
@@ -371,6 +371,11 @@ impl ServiceStore {
         self.by_host.get(hostname).map(|v| v.to_vec())
     }
 
+    pub fn get_best_by_host(&self, hostname: &Strng, ns: Option<&Strng>) -> Option<Arc<Service>> {
+        let services = self.get_by_host(hostname)?;
+        Some(ServiceMatch::find_best_match(services.iter(), ns, None)?.clone())
```
ilrudie, line 379: Should get_best_by_host optionally take preferred_namespace to match the inner API and provide more options?

src/state.rs
```diff
@@ -281,10 +281,14 @@ impl ProxyState {
     }
 
     /// Find services by hostname.
-    pub fn find_service_by_hostname(&self, hostname: &Strng) -> Result<Vec<Arc<Service>>, Error> {
+    pub fn find_service_by_hostname(
+        &self,
+        hostname: &Strng,
+        namespace: &Strng,
```
ilrudie, line 287: If we expose this as Option<&Strng> (any maybe preferred_namespace) to match the inner APIs does it give us more flexibility? It seems a bit odd to unify the approach but then toss out some of the extra fidelity.

src/dns/server.rs
```diff
@@ -383,11 +363,11 @@ impl Store {
                 let search_name_str = search_name.to_string().into();
                 search_name.set_fqdn(true);
 
-                let services: Vec<Arc<Service>> = state
-                    .services
-                    .get_by_host(&search_name_str)
-                    .iter()
-                    .flatten()
+                let Some(services) = state.services.get_by_host(&search_name_str) else {
```
ilrudie: **nit** seems messy, match is probably a cleaner way to get iter of sercives

src/state/service.rs
stevenctl: I was thinking we could expose that if we actually want the functionality, but for now we don't plumb that value through to many places.

src/state.rs
stevenctl: If anything I'd argue for making the `get_best_by_host` have namespace be _non_-optional, as we don't have any codepaths where it can be `None`.

stevenctl: "YAGNI"

src/dns/server.rs
stevenctl: switched from `let else` to `into_iter.flatten`

ilrudie: I don't think that commit got pushed

ilrudie: this was the suggestion by the way, not sure it's actually that much clearer in the end
```rust
                let services: Vec<Arc<Service>> = match state.services.get_by_host(&search_name_str)
                {
                    Some(svc) => {
                        svc
                            .into_iter()
                            // Remove things without a VIP, unless they are Kubernetes headless services.
                            // This will trigger us to forward upstream.
                            // TODO: we should have a reliable way to distinguish these. In sidecars, we use
                            // `svc.Attributes.ServiceRegistry`, but we don't pass anything similar over WDS.
                            // For now, checking the domain is good enough.
                            // This does mean a `.svc.cluster.local` ServiceEntry will use these semantics, but
                            // its better than *ALL* ServiceEntry doing this
                            .filter(|s| {
                                // Domain will be like `.svc.cluster.local.` (trailing .), so ignore the last character.
                                let domain = ".".to_string() + &self.svc_domain.to_utf8();

                                let domain = domain
                                    .strip_suffix('.')
                                    .expect("the svc domain must have a trailing '.'");
                                !s.vips.is_empty() || s.hostname.ends_with(domain)
                            })
                            // Get the service matching the client namespace. If no match exists, just
                            // return the first service.
                            .collect()
                    }
                    None => continue, //not found, that's ok just try the next search name
                };

```

stevenctl: I think the `continue` is ~implicit since the list is empty

PR #1749, opened 2026-02-04
Add ENABLE_OUTBOUND_ORIG_SRC config for outbound source IP preservation

bekerr1:
For our multi-NIC pods, we use policy based routing (PBR) to influence the outbound interface. The routing looks something like the following

```
$ ip rule show
0:      from all lookup local
32763:  from <ETH0_IP_ADDR> lookup 200
32766:  from all lookup main
32767:  from all lookup default

# Table 200
$ ip route show table 200
default via 169.254.1.1 dev eth0 

# Main table
$ ip route show
default via 169.254.2.1 dev eth1 proto static 
<NETWORK_ADDRESS_RANGE> via 169.254.1.1 dev eth0 proto static 
<POD_SUBNET_ADDRESS_RANGE> via 169.254.1.1 dev eth0 proto static 
<SERVICE_CIDR_RANGE> via 169.254.1.1 dev eth0 proto static 
```

The intention of this change is to enable Ambient to honor the outbound source address it uses (obtained from the peer addr) in cases where it could receive redirected packets from multiple interfaces.

More on our specific use-case here for those interested:
https://github.com/istio/istio/discussions/58681

I wasn't sure if creating a separate env variable and making the distinction between ENABLE_ORIG_SRC, which is currently only used for inbound, was preferred over just re purposing the existing env. Open to suggestions here.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -69,6 +69,7 @@ const CONNECTION_TERMINATION_DEADLINE: &str = "CONNECTION_TERMINATION_DEADLINE";
 // (Our forceful shutdown is more graceful than a SIGKILL, as we can close connections cleanly).
 const TERMINATION_GRACE_PERIOD_SECONDS: &str = "TERMINATION_GRACE_PERIOD_SECONDS";
 const ENABLE_ORIG_SRC: &str = "ENABLE_ORIG_SRC";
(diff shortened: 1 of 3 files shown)


Stevenjin8 (commented):
lgtm implementation wise, but want to give other folks a chance to take a look.

PR #1756, opened 2026-02-11
feat: connection strategies to support racing clients

### LINKED ISSUE #1456: SQL Server Multi-Subnet Cluster connection failures
Summary
=======

We observe connection failures to `SQL Server Multi-Subnet Clusters` when the sql client application is running inside an ambient mesh and connecting to SQL Servers outside of the mesh.

I _think_ the cause is due to ztunnel's TCP proxy "optimistically" completing TCP handshakes between the downstream client (i.e. dotnet sql client) and ztunnel _prior_ to completing & validating the handshake between the upstream (i.e. sql server multi subnet cluster) and ztunnel.

Details
=====

My understanding of how SQL Server Multi-Subnet Clusters operate, is that they actively coordinate amongst each other so that only a single instance is listening for connections at any given time. SQL Server clients discover the cluster's servers via a DNS record, of which returns a set of IP addresses bound to the cluster (example response in a screenshot below).

After the client obtains the list of cluster IPs via DNS, it then, in parallel, attempts to open a connection to each of these IPs. In a normal connection flow, once the client successfully completes a TCP handshake to the single, active IP, it then uses the connection for the sql client session and discards the other connection attempts.

Now, when mixing in the ambient mesh, ztunnel's pass-through TCP proxy "optimistically" & "immediately" (10s of microseconds) completes the TCP handshake for _all_ connection attempts to _all_ cluster IPs, regardless if the upstream connection is possible.

Due to race conditions, the downstream side of the proxy (between the sql client & ztunnel) often completes its TCP handshake for the IP of the _inactive_ server.  At this point, the sql client believes it has successfully established a connection to the inactive server, and drops any other connection to the active server. Subsequent writes to this otherwise invalid connection fails as the upstream side of the connection cannot be established.

Random Thoughts on Possible Solutions
-----------------------------------------

There's probably better ideas, but to capture various options and references, here's some thoughts on how this might be resolved:

This could be solved if support is added for `excludeOutboundPorts (ref: https://github.com/istio/istio/issues/49829). We'd exclude the sql server port and force all other traffic through the mesh.

Another possible solution is to verify the upstream connection handshake prior completing to the downstream handshake... though I'd assume this would have other impacts and downsides.

Given that our workloads connection both to sql server and other services inside the mesh, we'd prefer not to opt these workloads out of the mesh entirely via the `istio.io/dataplane-mode: none` option, but worth a call out.

Packet Captures
============

DNS
----

DNS response for the multi subnet cluster.

IP 10.x.x.72 is _inactive_
IP 10.x.x.7 is _active_

Image

Success Scenario
------------------

Here's a typical connection sequence when the mesh is inactive, generated by a test app that opens a connection every ~two seconds.

Successful connection to active server on 10.x.x.7:

Image

Failed connections to inactive server on 10.x.x.72:

Image

Failure Scenario
----------------

Here's a connection attempt to the _invalid_ address, 10.x.x.72, for the downstream side of the connection between the sql client and ztunnel.
This shows the TCP handshake completing between the client and ztunne, the subsequent data packet, and finally connection termination due to no valid listener on the upstream side.

Image

And corresponding upstream packets between ztunnel and the inactive sql server on 10.x.x.72, which all fail:

Image

keithmattix:
Fixes #1456 

I put copilot on the design @jaellio and @grnmeira proposed [here](https://docs.google.com/document/d/1cNA376XwRSsmB2NXNee30iZTw8gaAj8t8I4QDEFK99s/edit?tab=t.0#heading=h.xw1gqgyqs5b), and did a rough sanity check of the approach. This PR creates the idea of a ConnectionStrategy which allows racing connections within ztunnel to solve for cases where the application client does racing on its own (e.g. .NET SQL Server client)

### DIFF
--- a/src/proxy/outbound.rs
+++ b/src/proxy/outbound.rs
@@ -872,9 +872,7 @@ impl OutboundConnection {
                     .iter()
                     .map(|ip| SocketAddr::new(*ip, us.port))
                     .collect();
-                UpstreamTarget::Race {
-                    candidates,
-                }
+                UpstreamTarget::Race { candidates }
             } else {
                 UpstreamTarget::Single(actual_destination)
             };
@@ -2357,7 +2355,7 @@ mod tests {
                         proxy_stream,
                         &req,
                         source_addr,
-                        None, // no hbone_target
+                        None,         // no hbone_target
                         refused_addr, // orig_dst: use first candidate as the original destination
                         start,
                         metrics,


src/proxy/outbound.rs
```diff
@@ -422,6 +428,67 @@ impl OutboundConnection {
...
+            ?candidates,
+            "racing upstream connections (FIRST_HEALTHY_RACE)"
+        );
+
+        let res = (async {
+            let socket_factory = self.pi.socket_factory.as_ref();
+            let mut futs = futures_util::stream::FuturesUnordered::new();
+            for &addr in candidates {
+                futs.push(super::freebind_connect(None, addr, socket_factory));
+            }
```
Stevenjin8: If we do timeouts, then we need to cancel the other futures

src/proxy/outbound.rs
Stevenjin8: only one of `actual_destination` and `race_candidates` can ever be set right? in that case we should make it an enum

src/proxy/outbound.rs
```diff
@@ -211,6 +211,12 @@ impl OutboundConnection {
             metrics,
         ));
 
+        if !req.race_candidates.is_empty() {
+            self.proxy_to_race(source_stream, &req, connection_result_builder)
```
Stevenjin8: what if the destination is in the mesh?

keithmattix: The only time that would happen in the cluster is for a headless service right (based on the kubernetes DNS spec). If it's an external meshed destination, I'm less sure. Are you saying we should consider doing a racing hbone implementation?

Stevenjin8: yeah, or at least make sure the the destination isn't in the mesh

Stevenjin8:
> 1. I'm definitely not an envoy/C++ expert, but don't they disable the timer here: https://github.com/Stevenjin8/envoy/blob/77a4c7abbbe698e78f8be9c16e80dee189b4fea3/source/common/network/multi_connection_base_impl.cc#L471? Which would mean an immediate attempt?

It only disables if the connection fails. Suppose a tcp handshake takes 500 ms to complete and our dns query replies with 10 entries. As implemented, we would always create 10 tcp connections every time, which is excessive, but not the worst thing. If we add a 300ms timer, then we would do do 2 tcp handshakes (assuming they all succeed). If we get an immediate error, then we can cancel the timer and move on immediately since there are no packets in flight.

src/proxy/outbound.rs
keithmattix: According to copilot, they're automatically canceled when they go out of scope, but I added manual drops just in case

src/proxy/outbound.rs
keithmattix: Done

src/proxy/outbound.rs
keithmattix: Gated it so that we only race if the protocol is TCP

src/proxy/outbound.rs
Stevenjin8: or maybe not

keithmattix:
@Stevenjin8 Good call; implemented a more gradual race approach

src/proxy/outbound.rs
```diff
@@ -196,21 +197,27 @@ impl OutboundConnection {
...
+            req.upstream_target.addr(),
             req.protocol,
         );
 
         let metrics = self.pi.metrics.clone();
         let hbone_target = req.hbone_target_destination.clone();
         let connection_result_builder = Box::new(ConnectionResultBuilder::new(
             source_addr,
-            req.actual_destination,
+            req.upstream_target.addr(),
```
Stevenjin8, line 225: hmm, i didn't realize we would have to change code here.

Stevenjin8: Maybe we could have `Request` be an enum. I feel like I'm making suggestions that I don't fully understand

src/proxy/outbound.rs
```diff
@@ -422,6 +429,124 @@ impl OutboundConnection {
...
+                        biased;
+                        result = futures_util::StreamExt::next(&mut futs) => {
+                            match result {
+                                Some(Ok(tcp)) => {
+                                    winner = Some(tcp);
+                                    break;
+                                }
+                                Some(Err(err)) => {
+                                    debug!(?err, "race candidate connection failed");
+                                    // Immediate failure: start next candidate right away
```
Stevenjin8: This isn't an immediate failure, just a failure.

src/proxy/outbound.rs
keithmattix: I do actually kind of like the enum approach; it reduces potential confusion if both somehow get set. I don't think this diff is too bad honestly; we're just switching to use the wrapped enum type

src/proxy/outbound.rs
keithmattix:
```suggestion
                                    // Race candidate ailure: start next candidate right away
```

src/proxy/outbound.rs
Stevenjin8: Yeah I've been prototyping it a bit. I'm growing less fond of the `Request` struct since it has a lot of semantics that aren't part of the type system. For example, `hbone_target_destination` only makes sense if the protocol is `HBONE`. My biggest concern is around telemetry reporting the wrong address with the "representative address" stuff.

Stevenjin8: @keithmattix sorry I've been kinda all over the place with this PR.  Here are my thoughts. The semantics of `Request` are kinda a mess and if we want to simplify them, we can leave that to another PR. Thus, I like what we have in d747592b9ec4b2974b40af0a7b7c7fe2d6e4022f with two exceptions:

1. We should cherry-pick in 0e7aba6 
2. Report correct metrics and logs:

```diff
+        if !req.race_candidates.is_empty() {
+            // Put connection tracking logic in proxy_to_race since we don't know the 
+            // actual destination.
+            self.proxy_to_race(source_stream, &req, connection_result_builder)
+                .await;
+            return;
+        }

        let _conn_guard = self.pi.connection_manager.track_outbound(
            source_addr,
            dest_addr,
            req.actual_destination,
            req.protocol,
        );
        let metrics = self.pi.metrics.clone();
        let hbone_target = req.hbone_target_destination.clone();
        let connection_result_builder = Box::new(ConnectionResultBuilder::new(
            source_addr,
            req.actual_destination,
            hbone_target,
            start,
            Self::conn_metrics_from_request(&req),
            metrics,
        ));

-        if !req.race_candidates.is_empty() {
-            self.proxy_to_race(source_stream, &req, connection_result_builder)
-                .await;
-            return;
-        }

        match req.protocol {
            OutboundProtocol::DOUBLEHBONE => {
 ```

Comment (keithmattix) [reply src/proxy/outbound.rs:225]: Talked with steven; we need to move some of the telemetrycreporting into `proxy_to_race` to ensure the correct IP is logged.

Comment (keithmattix) [reply src/proxy/outbound.rs:225]: I might keep the request stuff as is if it's all the same to you? I think the enum is cleaner. If so, then I think the reporting is already done correctly because we use the representative IP (i.e. the one we picked).

Comment (Stevenjin8) [reply src/proxy/outbound.rs:225]: The reporting is not correct

```
# /etc/hosts
...
10.0.0.5 example.com # no host with 10.0.0.5
10.0.0.4 example.com # eth0
```

```yaml
# examples/localhost.yaml
workloads:
- uid: cluster1//v1/Pod/default/local
  name: local
  namespace: default
  serviceAccount: default
  workloadIps: ["127.0.0.1"]
  protocol: HBONE
  node: local
  network: ""
  services: {}
- uid: cluster1//v1/Pod/default/local-tcp
  name: dns
  namespace: default
  serviceAccount: default
  workloadIps: []
  hostname: example.com
  protocol: TCP
  node: local
  network: ""
  services:
    "default/example.com":
      8000: 8000
services:
- name: local
  namespace: default
  hostname: example.com
  vips:
    - /127.10.0.1
  ports:
    8000: 8000
  canonical: true
  loadBalancer:
    routingPreferences: []
    healthPolicy: AllowAll
    mode: Standard
    connectStrategy: FirstHealthyRace
```

In one terminal, I run

```bash
FAKE_CA="true" \
XDS_ADDRESS="" \
LOCAL_XDS_PATH=./examples/localhost.yaml  \
PROXY_MODE=dedicated \
PROXY_WORKLOAD_INFO=default/local/default \
UNSTABLE_ENABLE_SOCKS5="true" \
RUST_LOG="debug" \
cargo run --features testing
```

and in another one, I curl 

```bash
curl -x socks5://localhost:15080  127.10.0.1:8000
```

In the first terminal, i see

```
...
2026-03-02T19:57:30.339880Z     debug   proxy::socks5:socks5{id=65fa56c883649c4ab68b4a2ca09040bd}       connection started        component="socks5"
2026-03-02T19:57:30.340138Z     debug   state:socks5{id=65fa56c883649c4ab68b4a2ca09040bd}       wait for workload wl=default.default (local)
2026-03-02T19:57:30.340179Z     debug   state:socks5{id=65fa56c883649c4ab68b4a2ca09040bd}       got sub, waiting for workload     wl=default.default (local)
2026-03-02T19:57:30.340225Z     debug   state:socks5{id=65fa56c883649c4ab68b4a2ca09040bd}       fetch address     network_addr.address=127.10.0.1
2026-03-02T19:57:30.340255Z     debug   state:socks5{id=65fa56c883649c4ab68b4a2ca09040bd}       fetch address     network_addr.address=127.10.0.1
2026-03-02T19:57:30.340295Z     debug   state:socks5{id=65fa56c883649c4ab68b4a2ca09040bd}       service example.com does not have port 800
2026-03-02T19:57:30.340305Z     debug   state:socks5{id=65fa56c883649c4ab68b4a2ca09040bd}       Service example.com has no healthy endpoints
2026-03-02T19:57:30.340337Z     warn    access  connection failed       src.addr=127.0.0.1:51914 dst.addr=127.10.0.1:800 direction="outbound" error="no healthy upstream: 127.10.0.1:800"
2026-03-02T19:57:30.340411Z     debug   proxy::socks5:socks5{id=65fa56c883649c4ab68b4a2ca09040bd}       connection completed      component="socks5" dur=759.175µs
2026-03-02T19:57:51.256575Z     debug   proxy::socks5:socks5{id=f7cf2ada92f3031467956619d4b88a4c}       connection started        component="socks5"
2026-03-02T19:57:51.256857Z     debug   state:socks5{id=f7cf2ada92f3031467956619d4b88a4c}       wait for workload wl=default.default (local)
2026-03-02T19:57:51.256897Z     debug   state:socks5{id=f7cf2ada92f3031467956619d4b88a4c}       got sub, waiting for workload     wl=default.default (local)
2026-03-02T19:57:51.256948Z     debug   state:socks5{id=f7cf2ada92f3031467956619d4b88a4c}       fetch address     network_addr.address=127.10.0.1
2026-03-02T19:57:51.256989Z     debug   state:socks5{id=f7cf2ada92f3031467956619d4b88a4c}       fetch address     network_addr.address=127.10.0.1
2026-03-02T19:57:51.257385Z     debug   hickory_proto::xfer::dns_handle:socks5{id=f7cf2ada92f3031467956619d4b88a4c}       querying: example.com. A
2026-03-02T19:57:51.257495Z     debug   hickory_resolver::name_server::name_server_pool:socks5{id=f7cf2ada92f3031467956619d4b88a4c}       sending request: [Query { name: Name("example.com."), query_type: A, query_class: IN }]
2026-03-02T19:57:51.257586Z     debug   hickory_resolver::name_server::name_server:socks5{id=f7cf2ada92f3031467956619d4b88a4c}    reconnecting: NameServerConfig { socket_addr: 127.0.0.53:53, protocol: Udp, tls_dns_name: None, http_endpoint: None, trust_negative_responses: false, bind_addr: None }
2026-03-02T19:57:51.257676Z     debug   hickory_proto::xfer:socks5{id=f7cf2ada92f3031467956619d4b88a4c} enqueueing message:QUERY:[Query { name: Name("example.com."), query_type: A, query_class: IN }]
2026-03-02T19:57:51.257762Z     debug   hickory_proto::xfer::dns_handle:socks5{id=f7cf2ada92f3031467956619d4b88a4c}       querying: example.com. AAAA
2026-03-02T19:57:51.257803Z     debug   hickory_resolver::name_server::name_server_pool:socks5{id=f7cf2ada92f3031467956619d4b88a4c}       sending request: [Query { name: Name("example.com."), query_type: AAAA, query_class: IN }]
2026-03-02T19:57:51.257853Z     debug   hickory_resolver::name_server::name_server:socks5{id=f7cf2ada92f3031467956619d4b88a4c}    existing connection: NameServerConfig { socket_addr: 127.0.0.53:53, protocol: Udp, tls_dns_name: None, http_endpoint: None, trust_negative_responses: false, bind_addr: None }
2026-03-02T19:57:51.257882Z     debug   hickory_proto::xfer:socks5{id=f7cf2ada92f3031467956619d4b88a4c} enqueueing message:QUERY:[Query { name: Name("example.com."), query_type: AAAA, query_class: IN }]
2026-03-02T19:57:51.258010Z     debug   hickory_proto::udp::udp_client_stream   final message: ; header 19890:QUERY:RD:NoError:QUERY:0/0/1
; edns version: 0 dnssec_ok: false z_flags: 0 max_payload: 1232 opts: 0
; query
;; example.com. IN A

2026-03-02T19:57:51.258098Z     debug   hickory_proto::udp::udp_client_stream   final message: ; header 49161:QUERY:RD:NoError:QUERY:0/0/1
; edns version: 0 dnssec_ok: false z_flags: 0 max_payload: 1232 opts: 0
; query
;; example.com. IN AAAA

2026-03-02T19:57:51.258204Z     debug   hickory_proto::udp::udp_stream:socks5{id=f7cf2ada92f3031467956619d4b88a4c}        created socket successfully
2026-03-02T19:57:51.258394Z     debug   hickory_proto::udp::udp_stream:socks5{id=f7cf2ada92f3031467956619d4b88a4c}        created socket successfully
2026-03-02T19:57:51.258627Z     debug   hickory_proto::udp::udp_client_stream:socks5{id=f7cf2ada92f3031467956619d4b88a4c} received message id: 19890
2026-03-02T19:57:51.258685Z     debug   hickory_proto::error:socks5{id=f7cf2ada92f3031467956619d4b88a4c} response: ; header 19890:RESPONSE:RD,AA,RA:NoError:QUERY:2/0/1
; edns version: 0 dnssec_ok: false z_flags: 0 max_payload: 65494 opts: 0
; query
;; example.com. IN A
; answers 2
example.com. 0 IN A 10.0.0.5
example.com. 0 IN A 10.0.0.4
; nameservers 0
; additionals 1

2026-03-02T19:57:51.258748Z     debug   hickory_proto::error:socks5{id=f7cf2ada92f3031467956619d4b88a4c} response: ; header 19890:RESPONSE:RD,AA,RA:NoError:QUERY:2/0/1
; edns version: 0 dnssec_ok: false z_flags: 0 max_payload: 65494 opts: 0
; query
;; example.com. IN A
; answers 2
example.com. 0 IN A 10.0.0.5
example.com. 0 IN A 10.0.0.4
; nameservers 0
; additionals 1

2026-03-02T19:57:51.258887Z     debug   hickory_proto::udp::udp_client_stream:socks5{id=f7cf2ada92f3031467956619d4b88a4c} received message id: 49161
2026-03-02T19:57:51.258936Z     debug   hickory_proto::error:socks5{id=f7cf2ada92f3031467956619d4b88a4c} response: ; header 49161:RESPONSE:RD,AA,RA:NoError:QUERY:0/0/1
; edns version: 0 dnssec_ok: false z_flags: 0 max_payload: 65494 opts: 0
; query
;; example.com. IN AAAA
; answers 0
; nameservers 0
; additionals 1

2026-03-02T19:57:51.259001Z     debug   hickory_resolver::lookup_ip:socks5{id=f7cf2ada92f3031467956619d4b88a4c}   one of ipv4 or ipv6 lookup failed in ipv4_and_ipv6 strategy: proto error: no records found for Query { name: Name("example.com."), query_type: AAAA, query_class: IN }
2026-03-02T19:57:51.259069Z     debug   proxy::outbound:socks5{id=f7cf2ada92f3031467956619d4b88a4c}     built request to workload
2026-03-02T19:57:51.259145Z     debug   access  connection opened       src.addr=127.0.0.1:50572 src.workload="local" src.namespace="default" dst.addr=10.0.0.5:8000 dst.service="example.com" dst.workload="dns" dst.namespace="default" direction="outbound"
2026-03-02T19:57:51.259174Z     debug   proxy::outbound:socks5{id=f7cf2ada92f3031467956619d4b88a4c}     racing upstream connections (FIRST_HEALTHY_RACE)  count=2 candidates=[10.0.0.5:8000, 10.0.0.4:8000]
2026-03-02T19:57:51.264584Z     info    access  connection complete     src.addr=127.0.0.1:50572 src.workload="local" src.namespace="default" dst.addr=10.0.0.5:8000 dst.service="example.com" dst.workload="dns" dst.namespace="default" direction="outbound" bytes_sent=79 bytes_recv=5935 duration="7ms"
2026-03-02T19:57:51.264639Z     debug   proxy::socks5:socks5{id=f7cf2ada92f3031467956619d4b88a4c}       connection completed      component="socks5" dur=8.308918ms
```

Specifically, the line

```
2026-03-02T19:57:51.264584Z     info    access  connection complete     src.addr=127.0.0.1:50572 src.workload="local" src.namespace="default" dst.addr=10.0.0.5:8000 dst.service="example.com" dst.workload="dns" dst.namespace="default" direction="outbound" bytes_sent=79 bytes_recv=5935 duration="7ms"
```

is incorrect

keithmattix: Good catch; it had to do with how we were passing in connection_result_builder. I had the AI fix it and add a test that seems correct at first glance.

src/proxy/outbound.rs
```diff
@@ -206,15 +197,30 @@ impl OutboundConnection {
         let _conn_guard = self.pi.connection_manager.track_outbound(
             source_addr,
             dest_addr,
-            req.actual_destination,
+            req.upstream_target.addr(),
             req.protocol,
         );
```
Stevenjin8: Ok, this is  more of what I imaged, but this still isn't quite right. I haven't tested this, but if I did the same setup as before (with `example.com` resolving to `10.0.0.4` and `10.0.0.5`), `istioctl zc connections` would give incorrect info.

Stevenjin8: so then you can plumb this into `proxy_to_race` by having a match earlier, but then there is kinda no point in having a `representative` field in `UpstreamTarget`

keithmattix: Oh nice catch; I didn't understand the importance of the conn_guard. Moved that and removed the representative field

src/proxy/outbound.rs
```diff
@@ -742,6 +914,36 @@ fn baggage(r: &Request) -> String {
...
+
+impl UpstreamTarget {
+    /// Returns a representative address for this target.
+    /// For Single, this is the actual destination.
+    /// For Race, this is the first candidate (used only as a fallback;
+    /// the real winning address is determined after racing).
+    fn addr(&self) -> SocketAddr {
+        match self {
+            UpstreamTarget::Single(addr) => *addr,
+            UpstreamTarget::Race { candidates, .. } => candidates[0],
```
Stevenjin8, line 942: but then this branch never gets used... but I guess that's ok. Maybe we emit a log? I really don't know.

src/proxy/outbound.rs
```diff
@@ -755,10 +957,8 @@ struct Request {
     // service, but rather the original intended service.
     // May be unset in case of non-service traffic
     intended_destination_service: Option<ServiceDescription>,
-    // The address we should actually request to. This is the "next hop" address; could be a waypoint, network gateway,
-    // etc.
-    // When using HBONE, the `hbone_target_destination` is the inner :authority and `actual_destination` is the TCP destination.
-    actual_destination: SocketAddr,
+    // The upstream target to connect to. Either a single address or multiple candidates to race.
+    upstream_target: UpstreamTarget,
```
Stevenjin8, line 961: Could we keep the spirit of the delete comment

src/proxy/outbound.rs
keithmattix: Hmm good point. I can add an assertion here because we should generally never get to this branch

src/proxy/outbound.rs
```diff
@@ -432,6 +440,144 @@ impl OutboundConnection {
...
+            Some(tcp) => tcp,
+            None => {
+                // All candidates failed; RST-close the downstream socket
+                rst_close(stream);
+                return;
+            }
+        };
+
+        // Now that we know the winning destination, build the connection stats
+        // with the actual peer address rather than the representative.
```
Stevenjin8, line 549: There is no concept of representative anymore

src/proxy/outbound.rs
Stevenjin8: Ok here is my issue with UpstreamTarget being an enum (though I wouldn't block on this because this seems more like a symptom of `Request` than anything else). We check for the variant of upstream target

```rs
# https://github.com/keithmattix/ztunnel/blob/b27e47e82b99e29ae8d2fa4847c9037846143c03/src/proxy/outbound.rs#L210-L224
        if let UpstreamTarget::Race { .. } = &req.upstream_target {
            // Box::pin to avoid inflating the parent future's size with the racing state machine.
            Box::pin(self.proxy_to_race(
                source_stream,
                &req,
                source_addr,
                hbone_target,
                dest_addr,
                start,
                metrics,
            ))
            .await;
            return;
        }
```

and then we check that again!

```rs
# https://github.com/keithmattix/ztunnel/blob/b27e47e82b99e29ae8d2fa4847c9037846143c03/src/proxy/outbound.rs#L470-L477 
         let candidates = match &req.upstream_target {
            UpstreamTarget::Race { candidates, .. } => candidates,
            UpstreamTarget::Single(_) => {
                error!("BUG: proxy_to_race called with Single upstream target");
                return;
            }
        };
```

Its not a huge deal as long as we clearly label the paths that we should never reach  with some logs, and I can refactor it later.

PR #1766, opened 2026-02-25
 warn when own certificate appears in a configured CRL

jlojosnegros:
When CRL checking is enabled, ztunnel validates peer certificates against the configured CRLs but never checks whether its own workload certificate appears in any of those lists. If the local certificate is revoked, peers will start rejecting inbound HBONE connections without ztunnel having any indication of why.

This adds a self-revocation check inside WorkloadCertificate::build_tls_config:
after fetching the CRL DERs from the manager, `find_cert_revocation` scans each
CRL for the workload certificate's serial number. If a match is found, a `warn!`
log is emitted with the certificate serial, SPIFFE identity, and revocation
reason (when present in the CRL entry). The check is purely diagnostic — the
TLS config is still built normally so ztunnel continues to operate fail-open,
consistent with the existing allow_unknown_revocation_status policy.

`find_cert_revocation` is intentionally kept narrow: it only compares serial
numbers within CRLs that can be successfully parsed; malformed CRL entries
are logged at debug level and skipped.

A unit test covers both the positive case (own serial listed in CRL returns
Some) and the negative case (unrelated serial returns None).

Closes #1767

### DIFF
--- a/src/tls/certificate.rs
+++ b/src/tls/certificate.rs
@@ -33,6 +33,8 @@ use tracing::warn;
 
 use crate::tls;
 use x509_parser::certificate::X509Certificate;
+use x509_parser::prelude::FromDer;
+use x509_parser::revocation_list::CertificateRevocationList;
 
 #[derive(Clone, Debug)]
 pub struct Certificate {
@@ -237,6 +239,37 @@ fn parse_key(mut key: &[u8]) -> Result<PrivateKeyDer<'static>, Error> {
     }
 }
 
+/// Searches the provided CRL DER blobs for `cert`'s serial number.
+/// Returns `Some(reason)` if the certificate is listed as revoked,
+///     where `reason`(`Option<String>`) is `None` when the CRL entry carries no reason extension.
+/// Returns `None` if the certificate is not found in any CRL.
+fn find_cert_revocation(
+    cert: &Certificate,
+    crls: &[rustls::pki_types::CertificateRevocationListDer<'_>],
+) -> Option<Option<String>> {
+    let our_serial = &cert.parsed().serial;
+
+    for crl_der in crls {
+        // CertificateRevocationList::from_der returns (remaining_bytes, parsed_crl).
+        // we only care about the parsed_crl
+        match CertificateRevocationList::from_der(crl_der.as_ref()) {
+            Ok((_, crl)) => {
+                for revoked in crl.iter_revoked_certificates() {
+                    if revoked.serial() == our_serial {
+                        //reason_code() retruns Option<(critical: bool, ReasonCode)>
+                        let reason = revoked.reason_code().map(|(_, code)| code.to_string());
+                        return Some(reason);
+                    }
+                }
+            }
+            Err(e) => {
+                tracing::debug!("failed to parse CRL for self-revocation check: {e:?}");
+            }
+        }
+    }
+    None
+}
+
 impl WorkloadCertificate {
     pub fn new(key: &[u8], cert: &[u8], chain: Vec<&[u8]>) -> Result<WorkloadCertificate, Error> {
         let cert = parse_cert(cert.to_vec())?;
@@ -316,11 +349,15 @@ impl WorkloadCertificate {
         if let Some(ref mgr) = crl_manager {
             let crls = mgr.get_crl_ders();
             if !crls.is_empty() {
+                // Warn if ztunnel's own certificate is in the CRL.
+                if let Some(reason) = find_cert_revocation(&self.cert, &crls) {
+                    warn!(serial = %self.cert.serial(), identity = ?self.cert.identity(), reason = reason.as_deref().unwrap_or("unexpecified"), "own certificate is listed in the CRL; peers may reject connections until certificate is renewed");
+                }
+
                 builder = builder.with_crls(crls).allow_unknown_revocation_status(); // fail-open for unknown status
             }
         }
 
-        // TODO: check if our own certificate is revoked in the CRL and log warning
         let raw_client_cert_verifier = builder.build()?;
 
         let client_cert_verifier =
@@ -424,6 +461,11 @@ mod test {
     use crate::tls::WorkloadCertificate;
     use crate::tls::mock::{TEST_ROOT, TEST_ROOT_KEY, TEST_ROOT2, TEST_ROOT2_KEY, TestIdentity};
 
+    use rcgen::{
+        CertificateRevocationListParams, Issuer, KeyIdMethod, KeyPair, RevocationReason,
+        RevokedCertParams, SerialNumber,
+    };
+    use rustls::pki_types::CertificateRevocationListDer;
     use std::str::FromStr;
     use std::sync::Arc;
     use std::time::Duration;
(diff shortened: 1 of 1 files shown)


jlojosnegros:
Hi again, 
bumping this as it's been a while. 
CI is green and the change is focused. 
Happy to make any adjustments if needed. 

Is there anything blocking the review?


cc: @nilekhc

src/tls/certificate.rs
```diff
@@ -316,11 +349,15 @@ impl WorkloadCertificate {
         if let Some(ref mgr) = crl_manager {
             let crls = mgr.get_crl_ders();
             if !crls.is_empty() {
+                // Warn if ztunnel's own certificate is in the CRL.
+                if let Some(reason) = find_cert_revocation(&self.cert, &crls) {
+                    warn!(serial = %self.cert.serial(), identity = ?self.cert.identity(), reason = reason.as_deref().unwrap_or("unexpecified"), "own certificate is listed in the CRL; peers may reject connections until certificate is renewed");
```
ilrudie, line 354: typo
```suggestion
                    warn!(serial = %self.cert.serial(), identity = ?self.cert.identity(), reason = reason.as_deref().unwrap_or("unspecified"), "own certificate is listed in the CRL; peers may reject connections until certificate is renewed");
```

src/tls/certificate.rs
```diff
@@ -237,6 +239,37 @@ fn parse_key(mut key: &[u8]) -> Result<PrivateKeyDer<'static>, Error> {
...
+    let our_serial = &cert.parsed().serial;
+
+    for crl_der in crls {
+        // CertificateRevocationList::from_der returns (remaining_bytes, parsed_crl).
+        // we only care about the parsed_crl
+        match CertificateRevocationList::from_der(crl_der.as_ref()) {
+            Ok((_, crl)) => {
+                for revoked in crl.iter_revoked_certificates() {
+                    if revoked.serial() == our_serial {
+                        //reason_code() retruns Option<(critical: bool, ReasonCode)>
```
ilrudie, line 259: typo
```suggestion
                        //reason_code() returns Option<(critical: bool, ReasonCode)>
```

src/tls/certificate.rs
```diff
@@ -237,6 +239,37 @@ fn parse_key(mut key: &[u8]) -> Result<PrivateKeyDer<'static>, Error> {
...
 }
 
+/// Searches the provided CRL DER blobs for `cert`'s serial number.
+/// Returns `Some(reason)` if the certificate is listed as revoked,
+///     where `reason`(`Option<String>`) is `None` when the CRL entry carries no reason extension.
+/// Returns `None` if the certificate is not found in any CRL.
+fn find_cert_revocation(
+    cert: &Certificate,
+    crls: &[rustls::pki_types::CertificateRevocationListDer<'_>],
+) -> Option<Option<String>> {
```
ilrudie, line 249: Option<Option<...>> is difficult to understand. Consider putting the `unwrap_or("unspecified")` in this function to clean up the signature and make it a bit easier to parse.

src/tls/certificate.rs
```diff
@@ -237,6 +239,37 @@ fn parse_key(mut key: &[u8]) -> Result<PrivateKeyDer<'static>, Error> {
...
+    let our_serial = &cert.parsed().serial;
+
+    for crl_der in crls {
+        // CertificateRevocationList::from_der returns (remaining_bytes, parsed_crl).
+        // we only care about the parsed_crl
+        match CertificateRevocationList::from_der(crl_der.as_ref()) {
+            Ok((_, crl)) => {
+                for revoked in crl.iter_revoked_certificates() {
+                    if revoked.serial() == our_serial {
```
ilrudie, line 258: Technically, this isn't correct and can produce a false positive. Serial numbers are only required to be unique within the scope of the CA which issues them, not globally. I'm not sure what the odds of a conflict are, likely miniscule but it would be very concerning if it happened to you.

src/tls/certificate.rs
```diff
@@ -237,6 +239,37 @@ fn parse_key(mut key: &[u8]) -> Result<PrivateKeyDer<'static>, Error> {
...
 
+/// Searches the provided CRL DER blobs for `cert`'s serial number.
+/// Returns `Some(reason)` if the certificate is listed as revoked,
+///     where `reason`(`Option<String>`) is `None` when the CRL entry carries no reason extension.
+/// Returns `None` if the certificate is not found in any CRL.
+fn find_cert_revocation(
+    cert: &Certificate,
+    crls: &[rustls::pki_types::CertificateRevocationListDer<'_>],
+) -> Option<Option<String>> {
+    let our_serial = &cert.parsed().serial;
```
ilrudie, line 250: Only checking the cert for revocation seems incorrect. Another likely use would be to revoke the IA in which case the cert would be invalid but we would not catch it. For the specific use you have in mind here, that's probably not as likely but it makes this checking less useful.

src/tls/certificate.rs
```diff
@@ -237,6 +239,37 @@ fn parse_key(mut key: &[u8]) -> Result<PrivateKeyDer<'static>, Error> {
     }
 }
 
+/// Searches the provided CRL DER blobs for `cert`'s serial number.
+/// Returns `Some(reason)` if the certificate is listed as revoked,
+///     where `reason`(`Option<String>`) is `None` when the CRL entry carries no reason extension.
+/// Returns `None` if the certificate is not found in any CRL.
+fn find_cert_revocation(
```
ilrudie, line 246: organization, should checking for revocation be functionality of the crl manager itself?

src/tls/certificate.rs
```diff
@@ -316,11 +349,15 @@ impl WorkloadCertificate {
         if let Some(ref mgr) = crl_manager {
             let crls = mgr.get_crl_ders();
             if !crls.is_empty() {
+                // Warn if ztunnel's own certificate is in the CRL.
```
ilrudie, line 352: I think the comment is misleading, see other comment about verbiage

src/tls/certificate.rs
ilrudie, line 354: The verbiage here seems misleading. It suggest that ztunnel's certificate is revoked where in reality it's whatever certificate I'm trying to serve.

src/tls/certificate.rs
ilrudie, line 334: Putting this warning logic under `server_config` means it only triggers when we're going to serve inbound HBONE. It would likely be better to invoke this during initial handling of a certificate. Consider somewhere near
```rust
                        Ok(certs) => {
                             tracing::debug!(%id, "certificate fetch succeeded");
```
in manager.rs

ilrudie (changes requested):

src/tls/certificate.rs
conradhanson: +1 this should likely be handled by the crl manager itself

conradhanson, line 246: is there a particular reason why we are not leveraging webpki for verifying the ztunnel cert serial against the current CRL? I'm thinking about the drift between this and our use of webpki in general for handling CRL.

On the client side, we leverage `WebPkiClientVerifier` for CRL checks on new inbound connections. I'll be opening a PR soon for CRL checks on new outbound connection's cert chain which also leverages webpki.

conradhanson: with webpki, an impl could look something like
```
use webpki::end_entity
let ee = EndEntityCert::try_from(<ztunnel's cert>);
// build revocation options with RevocationOptionsBuilder to load CRL
ee.verify_for_usage(...)
```

conradhanson: building the revocation options to include `RevocationCheckDepth::Chain` will make it so that both ztunnel's leaf cert as well as its intermediate certs are checked.

src/tls/certificate.rs
conradhanson, line 334: Alternatives to placing the check here (since here will cause the check to execute for every inbound connection) could possibly be in the crl manager itself upon startup/everytime CRL is updated, or possibly in `WorkloadCertificate` instantiation when the cert is ztunnel's. not entirely sure which one makes more sense at the moment.

PR #1801, opened 2026-03-20
[Backport/v1.28] identity: reload CA root cert channel on file change

jlojosnegros:
Backport of #1775 to release-1.28

Everything is almost directly backported.
Note: some dependencies added.

Introduces a modification over the master PR, a boolean env var CA_CERT_WATCHER (default: true) that allows operators to disable the CA root cert file watcher at runtime.
When set to false, no watcher thread is started and the gRPC channel is never rebuilt on cert rotation — the startup-time cert is retained permanently.

The flag is only effective when ca_root_cert is a file path; Static and Default certs never start a watcher regardless.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -912,6 +912,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37909eebbb50d72f9059c3b6d82c0463f2ff062c9e95845c43a6c9c0355411be"
 
 [[package]]
+name = "file-id"
+version = "0.2.3"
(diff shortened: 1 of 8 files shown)


ilrudie:
I already did a 1.29 cherry-pick [here](https://github.com/istio/ztunnel/pull/1824). We can flip that flag there too if folks want.

The flag was requested [here](https://istio.slack.com/archives/C0A3G1LTUTZ/p1774015372206399) in slack for 1.29 where I had brought up the topic of cherry-picking more generally.

fjglira (approved):

PR #1861, opened 2026-04-22
Add runtime TLS configuration via xDS MeshSettings

zmiklank:
Implement runtime mesh-wide TLS configuration that can be pushed from
istiod to ztunnel via xDS, enabling dynamic updates without pod restart.

MeshSettings provides:
- trustDomain: SPIFFE identity trust domain for mTLS validation
- trustDomainAliases: additional trust domains for multi-cluster scenarios
  - this is a new feature in ztunnel, let me know if it is better to be removed.
- meshMTLS: TLS protocol configuration (min version, cipher suites, ECDH curves)

Configuration priority: xDS > environment variables (if exist) > defaults.

This aligns ztunnel with Envoy sidecar configuration model, allowing
operators to use existing MeshConfig fields to configure ztunnel's
TLS behavior mesh-wide.

Feature was tested locally, and needs istio complementary part to work. I will create a PR to Istio in following days.
Feature is backwards compatible, so if user decides to not use the XDS meshConfig in the mesh, everything will work as before.
In hope of easier review this was implemented in 2 steps (in 2 commits), the first one delivering minimal changes, the second one centralising the mesh_config and its operations into one file.
Fixes: #1817


src/proxy/inbound.rs
```diff
@@ -175,7 +176,7 @@ impl Inbound {
                 };
                 // This is small since it only handles the TLS layer -- the HTTP2 layer is boxed
                 // and measured above.
-                assertions::size_between_ref(1000, 1600, &serve_client);
+                assertions::size_between_ref(1000, 2200, &serve_client);
```
howardjohn: this seems like quite the jump?

zmiklank: hmm, yes. It is because the whole DemandProxyState was added to the `serve_client`. 
I pushed the commit which instead of the whole DemandProxyState only uses newly introduced MeshConfigHandle. I also used it in other places where it made sense.

PR #1874, opened 2026-04-23
update rustls-openssl to include hkdf_info fix from upstream

zmiklank:
This is needed for running ztunnel with openssl crypto in a FIPS mode without TLS12 enabled.
More info: https://github.com/tofay/rustls-openssl/commit/a21035c0ff42c78cfbce4ba15f8c1e0dd3c09660.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -261,7 +261,7 @@ dependencies = [
  "bitflags 2.10.0",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
+ "itertools 0.10.5",
  "log",
  "prettyplease",
  "proc-macro2",
@@ -1806,15 +1806,6 @@ dependencies = [
 
 [[package]]
 name = "itertools"
-version = "0.13.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "413ee7dfc52ee1a4949ceeb7dbc8a33f2d6c088194d9f922fb8318faf1f01186"
-dependencies = [
- "either",
-]
-
-[[package]]
-name = "itertools"
 version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2b192c782037fadd9cfa75548310488aabdbf3d2da73885b31bd0abd03351285"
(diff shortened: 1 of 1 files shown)


Stevenjin8:
@zmiklank I'd much rather wait for a new release rather than pinning to a commit. Also, the CI seems to fail for openssl 3.

zmiklank:
Hi @Stevenjin8, thanks for looking at this. 

> I'd much rather wait for a new release rather than pinning to a commit. 

Yes, that was my plan too, but the commit that I am referring to in this PR is from January, so not sure what is the appropriate time for waiting. However, if you would prefer it, I can try to ask on rustls-openssl upstream about planned releases.

> Also, the CI seems to fail for openssl 3.

Where do you please see the CI results? I see only that CI is waiting for /ok-to-test label.

Stevenjin8:
> Hi @Stevenjin8, thanks for looking at this.
> 
> > I'd much rather wait for a new release rather than pinning to a commit.
> 
> Yes, that was my plan too, but the commit that I am referring to in this PR is from January, so not sure what is the appropriate time for waiting. However, if you would prefer it, I can try to ask on rustls-openssl upstream about planned releases.
> 
> > Also, the CI seems to fail for openssl 3.
> 
> Where do you please see the CI results? I see only that CI is waiting for /ok-to-test label.

Yeah, if you could ask them to make a release. Here is the failing CI: https://github.com/tofay/rustls-openssl/actions/runs/21175489488/job/60903077443

zmiklank:
Hi @Stevenjin8, rustls-openssl upstream released new release including the fix. Also the CI is green now. I cannot really tell why it was failing then, as the logs are gone now.

Cargo.toml
```diff
@@ -40,7 +40,7 @@ boring-sys = { version = "4", optional = true }
 ring = { version = "0.17", optional = true }
 
 # Enabled with 'tls-openssl'
-rustls-openssl = { version = "0.3", optional = true, features = ["tls12"]}
+rustls-openssl = { version = "0.3.1", optional = true, features = ["tls12"]}
```
Stevenjin8: I don't think this diff is necessary anymore. You can just run `cargo update -p rustls-openssl` on master

zmiklank: Done.

zmiklank:
Hi @Stevenjin8. I just checked and the [update is in the master branch](https://github.com/istio/ztunnel/blob/a3592e6821cc261fe9a610dcb245939bf5e07865/Cargo.lock#L3192) already. It must have gotten there by some automation. We can close this PR I think. Thank you for your review :).

PR #1904, opened 2026-05-13
[release-1.29] bump cargo deps incl fuzz lock

PetrMc:
Same dep bumps as #1903 with fuzz/Cargo.lock updated so check-clean-repo passes. Closes #1903


PetrMc:
@fjglira - is it ok for you to approve to unblock dependebot? Thanks!

fjglira (approved):
lgtm

PR #1905, opened 2026-05-18
Upgrade criterion

fraenkel:
Break the dependency between pprof and criterion.
Reimplement the pprof/criterion support with only protobuf support.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -122,7 +122,7 @@ tempfile = { version = "3.27", optional = true}
 
 [target.'cfg(target_os = "linux")'.dependencies]
 netns-rs = "0.2"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -153,7 +153,7 @@ incremental = true
 ztunnel = { version = "0.0.0", path = ".", default-features = false, features = ["testing"] }
 
 # presently pprof-rs is pinned to this old version of criterion
-criterion = { version = "0.5.1", features = ["async_tokio", "html_reports"] }
+criterion = { version = "0.8", features = ["async_tokio", "html_reports"] }
```
ilrudie, line 155: we can probably lose the comment on 155 since this PR is addressing that

ilrudie (approved):
Nice! LGTM. I added DNM because I think we should clean up the one comment first

PR #1910, opened 2026-05-18
feat: track xDS config freshness without re-arming readiness

iandvt:
## Summary

This keeps ztunnel readiness as a startup gate and exposes xDS freshness
through metrics instead of changing `/healthz/ready` after startup.

After the initial xDS sync completes, `/healthz/ready` remains ready during
later xDS disconnects or non-fresh periods. Operators can use the metric
surface to distinguish raw stream connectivity, current config freshness,
completed disconnect duration, and completed non-fresh duration. Sustained
freshness policy belongs in dashboards, alerts, or external controllers.

## Problem

Ztunnel readiness is currently a startup gate. Once initial xDS sync completes,
`/healthz/ready` stays healthy even if ztunnel later loses contact with istiod
long enough for workload, service, policy, or certificate state to become
stale.

Using Kubernetes readiness for that steady-state signal overloads a pod
scheduling contract. Ztunnel is not a normal Service endpoint, but readiness
still has Kubernetes meaning, so xDS freshness is better exposed as telemetry
that an external controller can interpret.

## Solution

This change separates raw transport connectivity from config freshness:

- `Connected` means the xDS gRPC stream is open, but usable config has not yet
  been ACKed on that stream.
- `Synced` means ztunnel successfully handled and ACKed a response on the
  current stream, with no watched-resource rejection outstanding. During
  startup, `Synced` is delayed until all expected watched xDS types have ACKed.
- `Disconnected` means the stream ended or failed.

The xDS client publishes the latest state through a
`tokio::sync::watch::Receiver<XdsConnectionState>` and publishes transitions
through a bounded broadcast channel. The freshness monitor starts immediately,
records completed non-`Synced` periods, and independently waits for the
existing initial sync gate before clearing readiness. It does not register a
readiness blocker after startup.

Each published state carries a producer-side transition timestamp. Completed
non-fresh duration is calculated from those transition timestamps, so queued
events are timed from when xDS changed state rather than when the monitor task
was scheduled to process them.

## Configuration

This PR does not add a ztunnel freshness threshold configuration.

The previous prototype used `XDS_UNHEALTHY_THRESHOLD`, but that name implied
readiness would become unhealthy. A later `XDS_NON_FRESH_THRESHOLD` shape was
also removed because it still placed alert policy in ztunnel. Ztunnel now
reports facts; operators set the policy on the alert/dashboard side.

For example, a Prometheus alert can use `istio_xds_config_fresh == 0` with
`for: 5m`. A dashboard can use range queries such as
`max_over_time(istio_xds_config_fresh[5m]) == 0` when it wants to show
continuous non-fresh intervals.

## Metrics

This adds:

- `istio_xds_up`
- `istio_xds_config_fresh`
- `istio_xds_disconnect_duration_seconds`
- `istio_xds_config_non_fresh_duration_seconds`
- `istio_xds_monitor_sender_dropped_total`
- `istio_xds_monitor_events_lagged_total`

These cover current xDS stream connectivity, current xDS config freshness,
completed disconnect duration, completed non-fresh duration, monitor input
sender closure, and skipped monitor events. Lagged monitor events are counted
separately from sender drops. On lag, the monitor reconciles from the latest
watch snapshot and continues.

## Behavior

Default behavior:

- Ztunnel waits for initial xDS sync before becoming ready.
- After startup, xDS disconnects do not affect `/healthz/ready`.
- `istio_xds_config_fresh` reports `0` while the current stream is
  disconnected, connected but not ACKed, or has a watched-resource rejection
  outstanding.
- `istio_xds_config_fresh` reports `1` only after a usable ACK on the current
  stream and no watched-resource rejection is outstanding.
- A completed non-fresh period is observed in
  `istio_xds_config_non_fresh_duration_seconds` when freshness is restored.
- If the monitor receiver lags, `istio_xds_monitor_events_lagged_total`
  increments and the monitor reconciles from the current xDS state snapshot.

Remote-xDS startup sender-drop handling remains fail-closed: if the startup xDS
signal sender disappears before initial sync, readiness remains blocked with
`xds monitor dead` and data-plane listeners do not start. Startup-gated
data-plane tasks stay parked until drain instead of returning immediately, so
the readiness server remains available to report the fail-closed state.
Post-startup monitor input closure records
`istio_xds_monitor_sender_dropped_total` and does not re-arm readiness.

If the startup sync signal and monitor input closure are both ready in the same
poll, the startup signal wins and clears the startup readiness gate. That avoids
turning a completed startup sync into a permanent `xds monitor dead` blocker.

## Implementation Notes

The main code paths are:

- `src/xds/client.rs`: publishes `XdsConnectionState` snapshots, tracks
  `Connected`, `Synced`, and `Disconnected`, updates `xds_config_fresh`,
  records watched-resource ACK and NACK state, waits for all expected startup
  watched types before first freshness, publishes `Synced` before clearing the
  startup readiness blocker, only collects accepted watched resources for
  response types with an outstanding rejection, attaches transition timestamps,
  observes disconnect duration after a stream has reached `Connected`, raises
  `istio_xds_up` before publishing `Connected`, and logs the same retry delay
  it will sleep.
- `src/xds/freshness_monitor.rs`: owns startup readiness waiting and completed
  non-fresh duration metrics. Broadcast lag is counted separately from sender
  drop, retained stale backlog is discarded on lag, startup first consumes an
  already-retained broadcast event before falling back to the watch snapshot,
  startup sync wins over a simultaneous monitor exit, and reconciliation happens
  through the watch snapshot.
- `src/xds/metrics.rs`: adds stream and freshness monitor metrics.
- `src/app.rs` and `src/state.rs`: wire the monitor, metrics, connection-state
  watch receiver, transition broadcast, and startup drain handling for
  startup-gated data-plane tasks.
- `src/test_helpers/app.rs`, `src/test_helpers/xds.rs`, and `tests/direct.rs`:
  add in-process xDS integration coverage for freshness and metrics behavior.
  Direct metric assertions wait on a test-only freshness observation signal and
  then scrape `/metrics` once, instead of polling the stats endpoint for metric
  propagation.

## Risk

After startup, freshness restoration is intentionally based on the first
successfully handled and ACKed `DeltaDiscoveryResponse` on the current xDS
stream, with no watched-resource rejection outstanding. Raw transport
reconnection is not enough, and NACKed config does not restore freshness.

ACK-side clearing of a prior watched-resource rejection is applied only after
the outbound ACK has been successfully queued. NACK-side recording and
freshness demotion remain fail-closed before the outbound NACK send.

This does not prove every watched xDS type has been reasserted after reconnect.
A stronger freshness contract could require per-type or per-resource freshness,
but that would need an explicit istiod/ztunnel guarantee that every
freshness-critical type sends a confirmatory response after reconnect, including
quiet or empty types. Without that guarantee, requiring all watched types to ACK
could leave freshness false indefinitely in otherwise healthy clusters.

## Validation

Commands run for this branch:

```bash
cargo fmt -- --check
cargo check --lib --tests
git diff --check upstream/master...HEAD
git diff --check
cargo test --lib app::tests::
cargo test --lib xds::freshness_monitor::
cargo test --lib xds::client::tests::
cargo test --test direct xds_freshness_metrics
```

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -436,7 +436,7 @@ async fn handle_jemalloc_pprof_heapgen(
         .expect("builder with known status code should not fail"))
 }
 
-#[cfg(not(feature = "jemalloc"))]
+#[cfg(all(not(feature = "jemalloc"), target_os = "linux"))]
 async fn handle_jemalloc_pprof_heapgen(
     _req: Request<Incoming>,
 ) -> anyhow::Result<Response<Full<Bytes>>> {
--- a/src/app.rs
+++ b/src/app.rs
@@ -36,6 +36,33 @@ pub async fn build_with_cert(
     config: Arc<config::Config>,
     cert_manager: Arc<SecretManager>,
 ) -> anyhow::Result<Bound> {
+    // Startup orchestration overview:
+    //
+    //  ┌─────────────────────────────────────────────────────────────────┐
+    //  │ 1. Bootstrap                                                    │
+    //  │    data plane pool, drain channel, readiness tasks, metrics     │
+    //  └──────────────────────────┬──────────────────────────────────────┘
+    //                             │
+    //  ┌──────────────────────────▼──────────────────────────────────────┐
+    //  │ 2. ProxyStateManager (creates xDS client)                       │
+    //  └──────────────────────────┬──────────────────────────────────────┘
+    //                             │
+    //  ┌──────────────────────────▼──────────────────────────────────────┐
+    //  │ 3. xDS monitoring task                                          │
+    //  │    Phase 1: wait for initial xDS sync → drop "state manager"    │
+    //  │    Phase 2: freshness metrics                                   │
+    //  │             non-Synced ──► record non-fresh period               │
+    //  │             Synced ACK ──► report completed duration             │
+    //  └──────────────────────────┬──────────────────────────────────────┘
+    //                             │
+    //  ┌──────────────────────────▼──────────────────────────────────────┐
+    //  │ 4. Proxy / DNS listeners (wait for xDS sync before accepting)   │
+    //  └──────────────────────────┬──────────────────────────────────────┘
+    //                             │
+    //  ┌──────────────────────────▼──────────────────────────────────────┐
+    //  │ 5. Admin + metrics servers                                      │
+    //  └─────────────────────────────────────────────────────────────────┘
+    //
     // Start the data plane worker pool.
     let (data_plane_pool, data_plane_handle) = new_data_plane_pool(config.num_worker_threads);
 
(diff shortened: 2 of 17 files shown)


howardjohn:
I am a bit concerned this is abusing the readiness probe to mean something aside from its desired intent. Presumably the goal for this is for something to understand "is XDS fresh" and we are shoe-horning this into the "is this pod ready" bit that Kubernetes already has.

Instead of doing that, why not just expose a metric? then we can expose a variety of bits of information instead of conditionally choosing what the one -- readiness -- means, and instead keep that to mean "ready".

In particular I am worried that while we do not use readiness in ztunnel as a typical Service would, it does still have meaning around pod scheduling etc.

iandvt:
> I am a bit concerned this is abusing the readiness probe to mean something aside from its desired intent. Presumably the goal for this is for something to understand "is XDS fresh" and we are shoe-horning this into the "is this pod ready" bit that Kubernetes already has.
> 
> Instead of doing that, why not just expose a metric? then we can expose a variety of bits of information instead of conditionally choosing what the one -- readiness -- means, and instead keep that to mean "ready".
> 
> In particular I am worried that while we do not use readiness in ztunnel as a typical Service would, it does still have meaning around pod scheduling etc.
Agreed. I reworked this so readiness remains a startup gate only.

The PR now exposes xDS freshness through metrics instead of re-arming `/healthz/ready` after startup. I also removed the ztunnel-side non-fresh threshold config and the derived threshold metrics. Ztunnel now reports observable state, and sustained non-fresh policy can live in alerts, dashboards, or runbooks. For example, an alert could use `istio_xds_config_fresh == 0` with a Prometheus `for: 5m` window.

The metric surface now covers raw xDS stream connectivity, current config freshness, completed disconnect duration, completed non-fresh duration, monitor sender drops, and lagged monitor events.

The source diff includes the replacement freshness monitor and removes the old readiness monitor.

PR #1912, opened 2026-05-19
Disable autobenches

fraenkel:
Do not include all files in /benches as benchmarks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ name = "ztunnel"
 version = "0.0.0"
 edition = "2024"
 rust-version = "1.90"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -3,6 +3,7 @@ name = "ztunnel"
 version = "0.0.0"
 edition = "2024"
 rust-version = "1.90"
+autobenches = false
```
ilrudie:
```suggestion
autobenches = false # benches must be explicitly declared
```

ilrudie (approved):
Thanks for figuring this out. I added a dnm for my proposed comment

PR #1917, opened 2026-05-22
bump cargo deps incl fuzz lock

PetrMc:
Replaces #1916 (cannot push to dependabot branch). Same all-dependencies bump plus refreshed fuzz/Cargo.lock from make gen.


PetrMc:
cc @ilrudie - hope you have enough permssions and `hickory` background to approve

ilrudie (approved):

PR #1918, opened 2026-05-22
Resolve DNS for headless service pods

### LINKED ISSUE #1119: Fixes around workload.hostname

mzacho:
Currently ztunnel doesn't resolve DNS for headless service pods (there's an explicit TODO in `dns/server.rs` for this limitation). This PR fixes that by resolving workload IPs for hosts of the form `<pod>.<subdomain>.<ns>.svc.<cluster-domain>` (pod FQDN) or `<pod>.<subdomain>.<ns>.svc` / `<pod>.<subdomain>.<ns>` (partially qualified pod host) in the following way.

First the set of workload candidates is computed:

1) lookup all workloads named `<pod>` in the `WorkloadStore` (an extra `by_name` index has been created for this).
2) for each workload `w`: If any service `s` in `w.services` has the hostname `<subdomain>.<ns>.svc.<cluster-domain>` and `s.vips` is empty, add `w` as a workload candidate.

Step 2) relies on the existing logic that if `s.vips` is empty and `s.hostname` ends in `svc.<cluster-domain>`, then `s` is a headless service. However there's an existing comment saying this is a bit fragile, but the best we can do given the information currently available from WDS.

Then the best matching workload candidate is selected, giving preference to
1) workloads running in the same cluster and namespace as ztunnel
2) workloads running in the same cluster as ztunnel (but maybe foreign namespace)
3) workloads running in the same namespace as ztunnel (but maybe remote cluster)

---

At the company I work for we currently need this for onboarding some workloads to ambient, since we rely on Istio for resolving DNS for headless pods that reside in a remote cluster (since forwarding to k8s doesn't work in that case).

Probably closes #1119.


Stevenjin8:
@mzacho I think this PR is welcome, but just so you know:

> since we rely on Istio for resolving DNS for headless pods that reside in a remote cluster

Officially, we only support multinetwork multicluster, so cross-cluster pod-to-pod connectivity is not officially supported. Though with some effort, I'm sure you can get it to work.

mzacho:
Ah thanks - I recall this limitation of ambient now that you mention it. We're running sidecar-mode in a same-network primary-remote multicluster installation, so for incrementally migrating onto ambient I think our only option is to run it in the same cluster / network topology.

mzacho:
But thanks for taking a look at this, let me know what you think. This would make life a bit easier for us at least, and it seems to align pretty well with [how the agent resolves headless service pod IPs](https://github.com/istio/istio/blob/master/pkg/dns/server/name_table.go#L88) at the moment.

Stevenjin8:
I'm adding this to my list

Stevenjin8:
ok, it seems like single network mc is WIP, so we want this: https://github.com/istio/istio/issues/42137

mzacho:
So.. you want the PR you mean?..

src/dns/server.rs
```diff
@@ -329,6 +346,16 @@ impl Store {
         out
     }
 
+    /// Returns the cluster-local domain suffix used to identify Kubernetes headless services,
+    /// e.g. `.svc.cluster.local` (no trailing dot).
```
Stevenjin8: This is just used to identify k8s services. And even then its not perfect. Could you update the docs?

src/dns/server.rs
```diff
@@ -234,19 +234,32 @@ impl Store {
             }
         };
 
+        let name_len = name.iter().len();
+        let svc_domain_len = self.svc_domain.iter().len();
+        // Check if name maybe headless pod hostname, i.e. either one of
+        // 1) <pod>.<subdomain>.<ns>.svc.<cluster-domain> : pod FQDN
+        // 3) <pod>.<subdomain>.<ns>.svc                  : partially qualified domain
+        // 2) <pod>.<subdomain>.<ns>                      : partially qualified domain
```
Stevenjin8: This is actually kind tricky. What if the hostname is web-1.web? Is that the web-1 pod of the web headless service in the current namespace? Or is the the web-1 service in the web namespace? If both are defined? which takes priority?

Stevenjin8: Also, I think we can just get rid of `maybe_pod_host`. Ultimately, we are going to require the fqdn to end with `.svc.cluster.local`

src/dns/server.rs
```diff
@@ -403,7 +420,35 @@ impl Store {
...
+                let pod_name = str::from_utf8(name.next().unwrap()).unwrap();
+                let service_suffix = Name::from_labels(name).unwrap().to_utf8();
+                let service_suffix = service_suffix
+                    .strip_suffix('.')
+                    .expect("the svc domain must have a trailing '.'");
+
+                let cluster_local_domain = self.kubernetes_cluster_local_domain();
+                if let Some(wl) = HeadlessServiceMatch::find_best_match(
+                    state.workloads.find_headless_pod_workloads(
+                        pod_name,
```
Stevenjin8, line 454: For example here, `find_headless_pod_workloads` will only return if service_suffix ends with `svc.cluster.local`. I think it would be much simple if we go of the `maybe_pod_host` and here we check if alias.name has 6 labels. Then, the first label must be the pod name and the rest is the svc fqdn. Then, we can check if that svc fqdn exists, etc ,etc

src/state/workload.rs
```diff
@@ -731,6 +826,8 @@ pub struct WorkloadStore {
     pub(super) by_uid: HashMap<Strng, Arc<Workload>>,
     // Identity->Set of UIDs. Only stores local nodes
     node_local_by_identity: HashMap<WorkloadIdentity, HashSet<Strng>>,
+    /// by_name maps workload names to workload UIDs.
+    by_name: HashMap<Strng, HashSet<Strng>>,
```
Stevenjin8: More often than not, there well be one pod uid per pod name and allocating a whole HashSet is fairly wasteful. Can you something similar to WorkloadByAddr

Stevenjin8 (changes requested):
Apologies this took so long. After this is merged, could you also add an integration test to istio/istio and a release not?

src/dns/server.rs
mzacho: Updated the doc.

src/dns/server.rs
mzacho: Agree, it's not obvious how this should behave. 

>  If both are defined? which takes priority?

web-1 service in web namespace would take priority in this case, since we attempt to resolve the host using the service registry first.

> Also, I think we can just get rid of `maybe_pod_host`. Ultimately, we are going to require the fqdn to end with `.svc.cluster.local`

I think you're right. fqdn ending with `.svc.cluster.local` is checked [here](https://github.com/mzacho/ztunnel/blob/headless-svc-dns/src/state/workload.rs#L334) so we should be fine.

src/dns/server.rs
mzacho: Do we even need to check if `alias.name.num_labels == 6`?.. I don't really see a a case where we may resolve incorrectly if we don't.

If we do, we need to check using something like `name_len.wrapping_sub(svc_domain_len) == 3` I think, to take into account cluster domains other than `cluster.local`.

src/state/workload.rs
mzacho: Good point. I'll fix that.

src/dns/server.rs
mzacho: Removed `maybe_pod_host`.

src/state/workload.rs
mzacho: Fixed.

mzacho:
> Apologies this took so long.

No worries, thanks for getting back to it, appreciate it.

> After this is merged, could you also add an integration test to istio/istio and a release not?

Sure thing.

src/dns/server.rs
Stevenjin8: sorry this fell of my radar, but tyeah this is fair.

src/dns/server.rs
```diff
@@ -403,7 +403,35 @@ impl Store {
...
                 }
-                // TODO(): add support for workload lookups for headless pods
+            }
+
+            if alias.name.num_labels() > 1 {
+                trace!("checking headless svc {:#?}", alias);
+
+                let mut name = alias.name.into_iter();
+                let pod_name = str::from_utf8(name.next().unwrap()).unwrap();
+                let service_suffix = Name::from_labels(name).unwrap().to_utf8();
```
Stevenjin8: we can use `alias.name.base_name` here

src/dns/server.rs
```diff
@@ -403,7 +403,35 @@ impl Store {
...
                     });
                 }
-                // TODO(): add support for workload lookups for headless pods
+            }
+
+            if alias.name.num_labels() > 1 {
+                trace!("checking headless svc {:#?}", alias);
+
+                let mut name = alias.name.into_iter();
+                let pod_name = str::from_utf8(name.next().unwrap()).unwrap();
```
Stevenjin8: Can we use .to_utf8 here?

Stevenjin8: I think we have all the tools here to avoid unwraps

src/dns/server.rs
Stevenjin8: Ok, I looked into this a bit more, something that's not great is that `find_headless_pod_workloads` for every alias, when we know that `find_headless_pod_workloads` will only return if the service is an fqdn. I think a better way of doing this is to check `if name_len.wrapping_sub(svc_domain_len) == 3 && svc is headless`. At this point, we know the svc is headless, so instead of `find_headless_pod_workloads` we can do `find_pod_workloads` where we find pods with a name that are attached to some (not necessarily headless) service.

Stevenjin8 (changes requested):
sorry this fell off my radar. Had a few comments.

src/dns/server.rs
Stevenjin8: The main concern here is that we do a bunch of unnecessary workload and service lookups.

mzacho: Hmm.. Changing the guard to `name_len.wrapping_sub(svc_domain_len) == 3` and only supporting pod FQDN lookup to do less work makes sense. It'll suit the majority of (probably all) use cases I think.

Currently we lookup the workload by name, iterate through any services for the workload (typically just one I guess), lookup the svc by name if it matches the name, and check that it's headless (i.e. no vips and svc cluster domain).

IIUC then what you propose is to lookup the service by name, check that its headless, if yes lookup the workload, iterate through any services for the workload (again, usually just one or a couple) and check that it matches the name.

In the first case we save a service lookup if there's no workload with the name, or we may do an unnecessary service lookup if the workload doesn't back a headless svc.

In the second case we save a workload lookup if there's no headless svc with the name, or we may do an unnecessary workload lookup if there's a headless service but no backing pod.

I'd suspect the two strategies have similar performance, and would thus prefer to keep it as is. But if you think otherwise I'm OK doing the other thing.

src/dns/server.rs
mzacho: cleaned up the pod fqdn deconstruction. `alias.name.iter()` has an annoying `&[u8]` associated type for its items, so some unwrapping is necessary when converting this back to utf8.

src/dns/server.rs
Stevenjin8: The case I'm a bit worried about if there are two pods with the same name and namespace across multiple clusters (with stateful sets, this is very plausible). In this case, my understanding is that this implementation would do the service lookup multiple times.

My gut says that we should look up the pod before the service as pods tend tohave a lot more churn than services, but I care about this order much less. The main point is that we should do at most one service lookup per fqdn.

mzacho: Yeah, you're right, that's an issue. Good catch. I'll fix it one of these coming days.

mzacho: Fixed.

src/dns/server.rs
```diff
@@ -417,7 +417,43 @@ impl Store {
...
+                .name
+                .num_labels()
+                .wrapping_sub(self.svc_domain.num_labels())
+                == 3
+            {
+                // name is on the form pod-0.subdomain.ns.svc.cluster.local, i.e. it may
+                // be a pod FQDN - try to lookup in workload registry.
+                trace!("checking headless svc {:#?}", alias);
+
+                let pod_name = alias.name.iter().next().expect("at least one label");
```
Stevenjin8: hey, can we try to avoid "expects" at all costs. I think here, we can log an error and `continue`

src/state/workload.rs
```diff
@@ -1104,8 +1104,22 @@ impl WorkloadStore {
...
+                    .get_by_host(&strng::new(service_suffix))
+                    .into_iter()
+                    .flat_map(|svcs| svcs.into_iter())
+                    .filter(|svc| svc.is_kubernetes_headless(cluster_local_domain))
+                    .map(|svc| svc.namespaced_hostname())
+                    .collect_vec()
+            });
+
+            wl.services.iter().any(|svc| headless_svcs.contains(svc))
+        })
```
Stevenjin8: It still seems like we are doing extra work here. I think before we enter this function, we can check whether svc is headless and then just check if the workload is part of the service right? Here, we are checking whether every service each workload is part of has a headless service. We only care about `svc` that we know ahead of time.

Stevenjin8 (changes requested):

src/state/workload.rs
mzacho: > Here, we are checking whether every service each workload is part of has a headless service.

No, we're checking whether any service each workload is part of is headless.

> I think before we enter this function, we can check whether svc is headless and then just check if the workload is part of the service right?

I'm confused - by this function, do you mean `find_headless_pod_workloads`? Or is what you're suggesting simply to drop the lazy initialization of `headless_svcs` so we don't call `self.find_by_name` if it's empty?

src/dns/server.rs
mzacho: Yeah, fair, updated.

src/dns/server.rs
mzacho: Resolving, discussion moved to https://github.com/istio/ztunnel/pull/1918#discussion_r3846835765.

src/state/workload.rs
Stevenjin8: I might be getting confused, but this is my understanding of the code:

for each workload with the name `pod_name`, we loop through its services, filter out the ones that are not headless, and then check if the filtered set contains `svc`.
We have the `headless_svcs` array to store the results of lookups so we don't have to do it again, but I'm arguing that a lot of these lookups shouldn't even be happening. Instead, we should do something like

```
if !svc.is_kubernetes_headless(services.get_by_host(svc)) {
  return []
}
self.find_by_name(pod_name).filter(move |wl| {
  wl.services.iter().any(|svc| contains(svc))
}
```

For each workload `wl` that `self.find_pod_by_name` returns, we only care if it is an endpoint of `svc`. So the order should be:

if `svc` is not a kubernetes headless service:
  return empty
find all the pods with `pod_name` and filter out the ones that aren't part of `svc`.

**This way, we don't check if some irrelevant service is headless or not**

Concretely, I think we should do the following.

First, rename `find_headless_pod_workloads` to `find_pod_workloads_by_svc`.
This function will just return all the workloads with a certain name that is part of an svc.
Then, vefore we call `find_pods_workloads_by_svc`, we check if the svc is headless

mzacho: Alright, thanks for spelling it out - so basically I think what you want is this
> Or is what you're suggesting simply to drop the lazy initialization of headless_svcs so we don't call self.find_by_name if it's empty?

I've refactored again - WDYT?

Stevenjin8 (approved):

PR #1921, opened 2026-05-27
introduce crl enforcement for new outbound connections

conradhanson:
PR adds CRL enforcement for **_new_** outbound connections.
`crl_policy_rejections` metric counter added which increments every time a connection is rejected due to a cert revocation error (applies to both inbound and outbound).
Unit tests and integration tests added as well. 

approved [design doc here](https://docs.google.com/document/d/10lhVGCIHhmrqJE-lkLieoZSJCHCWU-QujI4tK8zveXQ/edit?tab=t.0#heading=h.xw1gqgyqs5b)


Stevenjin8:
@conradhanson this is on my list, but feel free to give me a ping.

src/hyper_util.rs
```diff
@@ -42,17 +42,32 @@ pub fn tls_server<T: ServerCertProvider + Clone + 'static>(
     cert_provider: T,
     listener: TcpListener,
 ) -> impl Stream<Item = tokio_rustls::server::TlsStream<TcpStream>> {
+    tls_server_with_error_handler(cert_provider, listener, |_| {})
+}
+
+pub fn tls_server_with_error_handler<T, F>(
```
Stevenjin8, line 48: this is just helpful for testing right?

Stevenjin8 (commented):
This seems good to me, but would like @ilrudie (or another ztunnel maintainer to take a look). @conradhanson could you make a follow up PR to test this in istio/istio with integration tests?

src/hyper_util.rs
conradhanson: pretty much, in prod code we always use the error handler version so the metric is incremented when the expected error is encountered.

src/proxy/outbound.rs
```diff
@@ -273,7 +275,8 @@ impl OutboundConnection {
                 .local_workload_information
                 .fetch_certificate()
                 .await?;
-            let connector = cert.outbound_connector(wl_key.dst_id.clone())?;
+            let connector =
+                cert.outbound_connector(wl_key.dst_id.clone(), self.pi.crl_manager.clone())?;
             let tls_stream = connector.connect(upgraded).await?;
```
stevenctl: the pool.rs diff does this

cert.outbound_connection(...)
let tls_stream = ...

but that side does `inspect_err` and calls `record_crl_rejection`; seems like that behavior should be matched here or pulled out to a common spot

src/tls/crl.rs
```diff
@@ -37,10 +38,12 @@ pub enum CrlError {
...
 #[derive(Clone)]
-/// NOTE: CRL updates take effect when new ServerConfigs are created, which happens
-/// on certificate refresh (~12hrs). For immediate CRL enforcement, a custom
-/// ClientCertVerifier wrapper would be needed, but rustls doesn't provide a
-/// built-in mechanism like `with_cert_resolver` for dynamic CRL updates.
+/// NOTE: For **inbound** mesh TLS, `ServerConfig` is rebuilt when the workload certificate
+/// refreshes (~12h), so CRL changes apply on that cadence unless the server config path changes.
+///
+/// For **outbound** mesh TLS, [`crate::tls::WorkloadCertificate::outbound_connector`] calls
+/// [`CrlManager::get_crl_ders`] when building each new client connector (e.g. new HBONE pool
```
stevenctl: i'm a bit confused by the comment here: `get_crl_ders` doesn't seem to be called by that path? seems to be only called inbound and during handshake

src/hyper_util.rs
```diff
@@ -42,17 +42,32 @@ pub fn tls_server<T: ServerCertProvider + Clone + 'static>(
...
+    cert_provider: T,
+    listener: TcpListener,
+    on_handshake_error: F,
```
stevenctl, line 51: super nit: on_handshake_error is unused

stevenctl (approved):

src/tls/crl.rs
conradhanson: comment is referencing the wrong method, should be `get_crls` used by outbound connector. will change

src/hyper_util.rs
conradhanson: not sure i follow, its used on line 70

src/proxy/outbound.rs
conradhanson: i will match the logic here. 
Just so my understanding is sound, this method is `proxy_to_double_hbone` which is opening an outer HBONE connection and then inner hbone requests (no connection pooling at all)? and then in pool.rs, the `ConnSpawner new_pool_conn`method is creating a pooled tls connection that can be reused if the IDs of client and dst are the same?

PR #1931, opened 2026-05-28
bump all-dependencies group (manual replacement for #1923)

PetrMc:
manual replacement for #1923 with fuzz/Cargo.lock regenerated. closes #1923


fjglira (approved):
Lgtm

PR #1947, opened 2026-06-10
fix: avoid panic on invalid DNS_PROXY_ADDR

immanuwell:
Invalid `DNS_PROXY_ADDR` currently makes ztunnel panic during startup.

This switches that path to return a normal config error instead, and adds a regression test.

Repro:
`PROXY_CONFIG='{"proxyMetadata":{"DNS_PROXY_ADDR":"not-an-address"}}' cargo run --quiet -- proxy`

Before:
panic at `src/config.rs`

After:
`Error: invalid configuration: failed to parse DNS_PROXY_ADDR: not-an-address (invalid socket address syntax)`

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -611,8 +611,11 @@ pub fn construct_config(pc: ProxyConfig) -> Result<Config, Error> {
     // Note: since DNS proxy runs in the pod network namespace, we will recompute IPv6 enablement
     // on a pod-by-pod basis.
     let dns_proxy_addr: Address = match pc.proxy_metadata.get(DNS_PROXY_ADDR_METADATA) {
-        Some(dns_addr) => Address::new(ipv6_localhost_enabled, dns_addr)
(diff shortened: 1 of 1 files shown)


keithmattix:
Why wouldn't we want to panic on startup here? Doesn't this mean a large portion of ztunnel wouldn't work and the app developer has no idea?

immanuwell:
@keithmattix after my changes ztunnel still fails to start - the ? propagates the error out of construct_config just like a panic would

the only difference is the output:
- before: panic backtrace pointing into config.rs
- after: Error: invalid configuration: failed to parse DNS_PROXY_ADDR: not-an-address (invalid socket address syntax)

the error message is more visible and actionable for the developer, not less

keithmattix (approved):
Sorry missed the greater context of the file. This is good; thanks for the contribution

ilrudie (approved):

PR #1951, opened 2026-06-12
revert dep bump #1949 on release-1.29

PetrMc:
after this bump integ-ambient-mc TestServiceRestart fails consistently on istio/istio#60563 (3/3 runs, same signature: single request failing over to the e/w gateway at restart trigger). reverting to unblock 1.29 updates while the root cause is investigated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -425,9 +425,9 @@ dependencies = [
 
 [[package]]
 name = "chrono"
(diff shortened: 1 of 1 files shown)


PetrMc:
closing in favor of #1952 - pinning only prost to bisect the integ-ambient-mc failure

PR #1952, opened 2026-06-12
pin prost crates back to 0.14.3 on release-1.29

PetrMc:
since #1949 merged, integ-ambient-mc TestServiceRestart on istio/istio#60563 fails 3/3 with the same signature (single request failing over to the e/w gateway at restart trigger, gateway 503). pinning prost/prost-types/prost-derive back to 0.14.3 while keeping the rest of the bump to bisect which dep triggers it.


fjglira (approved):
lgtm, let's see next test executions

PR #1961, opened 2026-06-17
bump all-dependencies group and regen fuzz lock (1.29)

PetrMc:
replaces #1956, adds fuzz/Cargo.lock regen so check-clean-repo passes


fjglira (approved):
lgtm

PR #1965, opened 2026-06-26
dns: filter dns records by the requesting client ip family

PetrMc:
ztunnel DNS returns AAAA records (e.g. an auto-allocated IPv6 VIP for a ServiceEntry) to a single-stack IPv4 client, which then prefers the IPv6 address and fails with "network is unreachable". This filters records by the requesting client workload IP family, matching the sidecar/NDS FilterAddressesByIPFamily behavior.

### DIFF
--- a/src/dns/server.rs
+++ b/src/dns/server.rs
@@ -632,7 +632,12 @@ impl Resolver for Store {
         // From this point on, we are the authority for the response.
         let is_authoritative = true;
 
-        if !service_family_allowed(&service_match.server, record_type, self.ipv6_enabled) {
+        if !service_family_allowed(
+            &service_match.server,
+            &client,
+            record_type,
+            self.ipv6_enabled,
+        ) {
             access_log(
                 request,
                 Some(&client),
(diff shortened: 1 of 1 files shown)


PetrMc:
feedback request @istio/wg-networking-maintainers-ztunnel:

filtering ztunnel DNS so a single-stack client isn't handed the other family's record (auto-allocated v6 VIP to a v4-only pod, which fails with "network is unreachable"). signal is the requesting client's IP family.

two options for sourcing it:
(a) (current) derive from workload_ips at lookup. walk over 1-2 ips, effectively O(1), no extra state.
(b) precompute/store ip_family on Workload at registration. faster in theory, but Workload has no Default, so it touches 40+ Workload {...} literals across ~12 files (mostly tests).

currently (a). happy to switch to (b), or do it as a follow-up, if you'd prefer the stored field. preference?

howardjohn (changes requested):
I don't think this is right.

If you go into 99% of clusters, **the cluster** is IPv4 only. That means the pod gets a single IPv4 address. That does NOT mean they are unable to send IPv6 traffic.

By doing this ,we are basically unconditionally turning IPv6 off for a ton of clients when we have no reason to.

If a user is in the rate condition where they do not support sending IPv6 traffic, they should (1) realize its 2026 and ask wtf they are doing (2) disable IPv6 at the ztunnel level, which is already handled

(I am not 100% confident I am right here 🙂 )

PetrMc:
thanks @howardjohn for the good point.
the pod does have a v6 stack, so it's not v6-incapable. but in a v4-only cluster it has no v6 route and ambient doesn't set up v6 capture, so connecting to the auto-allocated `2001:2::x` VIP is `ENETUNREACH` (verified). and in the issue i looked at `grpc-go/otel` don't fall back to the v4 VIP, so it just wedges.

two things i'd flag before we drop it:

- this is mesh-scoped, not all AAAA. [`service_family_allowed`](https://github.com/istio/ztunnel/blob/94ffa3a16f2835738f4bf447170fed813d184018/src/dns/server.rs#L599-L602) only runs for names that resolve to a mesh service; external names are forwarded upstream untouched. so it's not turning v6 off broadly, just the synthetic VIPs.
- it mirrors what the sidecar/NDS path already does `getAllAddressesForProxy` ([service.go#L1808](https://github.com/istio/istio/blob/9f7f71384a0e/pilot/pkg/model/service.go#L1808)). ztunnel doing the same is consistency, not new policy.

that said, if you'd rather fix it at the source, the synthetic v6 VIP arguably shouldn't be allocated/advertised on a single-stack cluster at all (istiod side). happy to move it there. or if `ipv6_enabled=false` is the intended answer for these envs, having auto-allocator to stop minting an address it knows can't be reached is the goal. wdyt?

PetrMc:
so the behavior we were addressing here is not failed connections, it's a noise of failing v6 and going v4. The right env setup with "IPV6_ENABLED=false" would remove the logs noise.

PR #1971, opened 2026-07-06
fix fd count underflow

### LINKED ISSUE #1970: Integer underflow in encode_open_fds after an error occurs when reading the FD_PATH
In the event that an error occurs while reading `FD_PATH`, the block returns `0` which is defined to be a `u64`. In the very next line, this will result in an integer underflow when constructing the `ConstGauge` type..

From https://github.com/istio/ztunnel/blob/94ffa3a16f2835738f4bf447170fed813d184018/src/metrics/process.rs#L31

```rust
// Count open fds by listing /proc/self/fd
let open_fds = match std::fs::read_dir(FD_PATH) {
    Ok(entries) => entries.count() as u64,
    Err(e) => {
        error!("Failed to read {}: {}", FD_PATH, e);
        0
    }
};
// exclude the fd used to read the directory
let gauge = metrics::gauge::ConstGauge::new(open_fds - 1);
```

Stevenjin8:
This should never happen in the happy path, but just in case. resolves https://github.com/istio/ztunnel/issues/1970

### DIFF
--- a/src/metrics/process.rs
+++ b/src/metrics/process.rs
@@ -31,14 +31,14 @@ impl ProcessMetrics {
     fn encode_open_fds(&self, encoder: &mut DescriptorEncoder) -> Result<(), std::fmt::Error> {
         // Count open fds by listing /proc/self/fd
         let open_fds = match std::fs::read_dir(FD_PATH) {
-            Ok(entries) => entries.count() as u64,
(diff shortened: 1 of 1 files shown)


ilrudie (approved):
Thanks @Stevenjin8

PR #1972, opened 2026-07-07
Support weighted waypoint sets for per-connection canary routing

ymesika:
## Summary

Adds ztunnel support for a **weighted set of waypoints** on a `Service`, so a configurable share of a service's in-mesh connections can be shifted between waypoints (e.g. canarying a waypoint revision or migration) without client changes.

This is the dataplane half: istiod emits the weighted set, and ztunnel samples one waypoint **per connection**, proportional to each entry's weight. The proto mirrors `istio.workload.Service` (`weighted_waypoints = 13`).

Part of: istio/istio#60801

## Backward compatibility

When `weighted_waypoints` is empty, behavior is unchanged — the single `waypoint` field is used as before, and it stays populated with the primary so older ztunnels keep sending all connections there. An all-zero-weight set falls back to the primary.

## Testing

New unit test `test_sample_weighted_waypoint` (absent set, `[100,0]`, `[0,100]`, all-zero, and a 50/50 split). Verified e2e with the istiod change: measured split tracks the configured weight.


src/state.rs
```diff
@@ -996,9 +1014,19 @@ impl DemandProxyState {
         source_workload: &Workload,
         original_destination_address: SocketAddr,
     ) -> Result<Option<Upstream>, Error> {
-        let Some(gw_address) = &service.waypoint else {
-            // no waypoint
-            return Ok(None);
+        // When a weighted set is present, sample one waypoint per connection proportional to
+        // weight (used to shift a share of connections between waypoints during a canary).
+        // Otherwise (or if the set is not samplable, e.g. all-zero) fall back to the single
+        // `waypoint` field, preserving existing behavior.
```
ilrudie: **nit:** this basically just describes what the function does.

```suggestion
        // Choose a weighted waypoint, else fall back to the single
        // `waypoint` field, preserving existing behavior.
```

src/state.rs
```diff
@@ -518,6 +518,24 @@ impl DemandProxyState {
     }
 }
 
+/// sample_weighted_waypoint picks one waypoint from a service's weighted set, proportional to
+/// each entry's weight (used to shift a share of connections between waypoints during a canary).
+/// Returns None when there is no usable weighted set — an empty set, or one that is not samplable
+/// (e.g. all weights zero) — in which case the caller falls back to the single `waypoint` field.
+fn sample_weighted_waypoint<'a>(
+    service: &'a Service,
+    rng: &mut impl rand::Rng,
```
ilrudie, line 527: Is there some reason we need rng as a variable? It seems like we only ever pass it `rand::Rng()`.

src/state.rs
```diff
@@ -1183,6 +1211,108 @@ mod tests {
...
+            match sample_weighted_waypoint(&split, &mut rng) {
+                Some(x) if *x == a => saw_a = true,
+                Some(x) if *x == b => saw_b = true,
+                other => panic!("unexpected selection: {other:?}"),
+            }
+        }
+        assert!(
+            saw_a && saw_b,
+            "both waypoints should be selected across samples"
+        );
```
ilrudie: Should we make an assertion about the shape of a few different distributions here? It's probably a bit tricky since the sample size likely needs to be pretty large to get higher assurance that a distribution test will not flake.

ilrudie (commented):
mostly lgtm, couple of comments

src/state.rs
ymesika: Yes, It's injected so the tests can seed it deterministically (see `test_sample_weighted_waypoint`).

src/state.rs
ymesika: Good call. turned this into a seeded 75/25 distribution assertion (deterministic seed + wide band, so it checks the split shape without flaking).

ilrudie:
Seems good now, but we should probably wait on https://github.com/istio/api/pull/3735 for final merge.

ilrudie (approved):
LGTM

PR #1976, opened 2026-07-08
Safer ServiceEntry

ilrudie:
This is the ztunnel implementation of the [Safer ServiceEntry](https://docs.google.com/document/d/17CLPeo1Jzi283GBAutqhXLTa1LZxDili5EIhllmbPn0/edit?usp=sharing) proposal.

The behavior is about filtering during matching and name resolution. A client which has no visibility for a specific WDS service is expected to behave as if the service doesn't exist. There should be no "shadowing" where the existence of a service which isn't directly visible to a client influences how that client's traffic is handled (a synthetic NXDOMAIN introduced when the best resolution of a host is a "not visible" service for instance).

Telemetry at this point is simple DEBUG logging, which emits a message per-service which is not visible at the point of filtering. This may be quite chatty, in an environment where dozens of teams have a namespace-visible `github.com` service entry the result would be many unique services being filtered as "not visible". Likely suitable enough at the debug level though.

An istio/istio PR will be x-linked once filed. This should however be wire and behavior compatible with older versions of istio. When visibility isn't present in the WDS message, the service is "public" visibility preserving the previous behavior.

### DIFF
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -490,8 +490,10 @@ impl Inbound {
         })
     }
 
-    // Selects a service by hostname without the explicit knowledge of the namespace
-    // There is no explicit mapping from hostname to namespace (e.g. foo.com)
+    // A hostname has no namespace (e.g. foo.com), so resolve against the local workload's.
+    // Leak-safe even on inbound, where local_workload is the dst and not the client: the dst may
+    // only handle traffic for a service it can itself see, so long as the rule barring cross-namespace
+    // binding of NAMESPACE-visibility services holds.
     fn find_service_by_hostname(
         state: &DemandProxyState,
         local_workload: &Workload,
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -490,10 +490,8 @@ impl Inbound {
         })
     }
 
-    // A hostname has no namespace (e.g. foo.com), so resolve against the local workload's.
-    // Leak-safe even on inbound, where local_workload is the dst and not the client: the dst may
-    // only handle traffic for a service it can itself see, so long as the rule barring cross-namespace
-    // binding of NAMESPACE-visibility services holds.
+    // Select a service by hostname, preferring one in the namespace of the given workload.
+    // The given workload may be in the perspective of outbound or inbound.
     fn find_service_by_hostname(
         state: &DemandProxyState,
         local_workload: &Workload,
--- a/src/proxy/inbound.rs
+++ b/src/proxy/inbound.rs
@@ -784,7 +784,7 @@ mod tests {
         rbac::Connection,
         state::{
             self, DemandProxyState, WorkloadInfo,
-            service::{Endpoint, EndpointSet, Service},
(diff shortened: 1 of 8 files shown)


src/state/service.rs
```diff
@@ -389,7 +427,29 @@ impl ServiceStore {
         ns: Option<&Strng>,
     ) -> Option<Arc<Service>> {
         if let Some(services) = self.by_vip.get(vip) {
-            return Some(ServiceMatch::find_best_match(services.iter(), ns, None)?.clone());
+            // When a client namespace is provided, skip services not visible to it so the lookup
+            // misses and the caller falls through to passthrough (as if the VIP were not a mesh
+            // service). With no client namespace (e.g. the inbound path), don't enforce.
+            return Some(
```
stevenctl: If we miss here on visibility should we fall through to the cidr lookup?

src/state/service.rs
stevenctl, line 456: find_hostname/get_best_by_host should also have visibility filtering I think

stevenctl (commented):
Pretty much LGTM!

src/state/service.rs
ilrudie: absolutely it should, excellent catch!

src/proxy/inbound.rs
```diff
@@ -490,8 +490,10 @@ impl Inbound {
         })
     }
 
-    // Selects a service by hostname without the explicit knowledge of the namespace
-    // There is no explicit mapping from hostname to namespace (e.g. foo.com)
+    // A hostname has no namespace (e.g. foo.com), so resolve against the local workload's.
```
stevenctl: This comment is a bit confusing now. What about

```
Select a service by hostname, preferring one in the namespace of the given workload.
The given workload may be in the perspective of outbound or inbound. 
```

Also, isn't it usually the case that a workload can't serve a non-namespace local service by construction? I think the visibility rules don't really need a mention in this comment, for inbound.

src/state/service.rs
```diff
@@ -389,7 +427,30 @@ impl ServiceStore {
         ns: Option<&Strng>,
     ) -> Option<Arc<Service>> {
         if let Some(services) = self.by_vip.get(vip) {
-            return Some(ServiceMatch::find_best_match(services.iter(), ns, None)?.clone());
+            // When a client namespace is provided, skip services not visible to it. If a visible
```
stevenctl: this comment is also quite verbose for relatively narrow behavior, could have a oneline here, and a oneline above the  cidr fallback

// only return a service that is visible to the passed in workload's namespace

// fallback to CIDR lookup if there are no matching and visible services

stevenctl (commented):
lgtm, but the comments seem kind of verbose/confusing

src/proxy/inbound.rs
ilrudie: Thanks @stevenctl, I cleaned up the two comments you highlighted as well as shortening a comment which was likely overly-verbose and should if anything, be an issue in github instead of a code comment.

stevenctl (approved):

PR #1979, opened 2026-07-09
Support compression for responses for metrics

fraenkel:
Allow http requests to suggest compression encodings on responses.

### DIFF
--- a/src/hyper_util.rs
+++ b/src/hyper_util.rs
@@ -12,9 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::convert::Infallible;
 use std::net::SocketAddr;
-use std::sync::Arc;
 use std::task::{Context, Poll};
 use std::{
     future::Future,
@@ -25,13 +23,14 @@ use std::{
 use crate::drain::DrainWatcher;
 use crate::{config, proxy};
 use bytes::Bytes;
-use futures_util::TryFutureExt;
 use http_body_util::Full;
+use hyper::body::Incoming;
 use hyper::client;
 use hyper::rt::Sleep;
 use hyper::server::conn::{http1, http2};
 use hyper::{Request, Response};
 use hyper_util::client::legacy::connect::HttpConnector;
+use hyper_util::service::TowerToHyperService;
 use tokio::net::{TcpListener, TcpStream};
 use tokio_stream::Stream;
 use tracing::{Instrument, debug, info, warn};
@@ -182,19 +181,17 @@ pub fn plaintext_response(code: hyper::StatusCode, body: String) -> Response<Ful
 /// Server implements a generic HTTP server with the follow behavior:
 /// * HTTP/1.1 plaintext only
 /// * Draining
-pub struct Server<S> {
+pub struct Server {
     name: String,
     binds: Vec<TcpListener>,
     drain_rx: DrainWatcher,
-    state: S,
 }
 
-impl<S> Server<S> {
+impl Server {
     pub async fn bind(
         name: &str,
         addrs: config::Address,
         drain_rx: DrainWatcher,
-        s: S,
     ) -> anyhow::Result<Self> {
         let mut binds = vec![];
         for addr in addrs.into_iter() {
@@ -204,7 +201,6 @@ impl<S> Server<S> {
             name: name.to_string(),
             binds,
             drain_rx,
-            state: s,
         })
     }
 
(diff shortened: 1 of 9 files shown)


fraenkel:
feature parity - istiod already supports metrics compression

src/metrics/server.rs
```diff
@@ -13,21 +13,43 @@
...
 
 use crate::config::Config;
 use crate::drain::DrainWatcher;
 use crate::hyper_util;
 
+struct FmtToIoWriter<'a, W: std::io::Write> {
+    inner: &'a mut W,
+}
+
+impl<'a, W: std::io::Write> std::fmt::Write for FmtToIoWriter<'a, W> {
```
Stevenjin8:
```suggestion
impl<'_, W: std::io::Write> std::fmt::Write for FmtToIoWriter<'_, W> {
```

src/metrics/server.rs
```diff
@@ -36,49 +58,66 @@ impl Server {
...
+) -> Response<BoxBody<Bytes, Infallible>> {
+    let (writer, reader) = duplex(4096);
+
+    tokio::task::spawn_blocking(move || {
+        let mut sync_writer = SyncIoBridge::new(writer);
+        let reg = reg.lock().expect("mutex");
+        let mut fmt_target = FmtToIoWriter {
+            inner: &mut sync_writer,
+        };
+        let _ = encode(&mut fmt_target, &reg);
```
Stevenjin8: This should probably never fail, but lets do a warning if it does so its not silent.

src/metrics/server.rs
```diff
@@ -13,21 +13,43 @@
 // limitations under the License.
```
Stevenjin8, line 13: I don't think changes to this files are necessary, but they seem like a general improvement

Stevenjin8 (commented):
minor nits. Thanks

src/metrics/server.rs
fraenkel: Unfortunately that doesn't work.
```
error[E0637]: `'_` cannot be used here
  --> src/metrics/server.rs:42:6
   |
42 | impl<'_, W: std::io::Write> std::fmt::Write for FmtToIoWriter<'_, W> {
   |      ^^ `'_` is a reserved lifetime name
   |
   = help: use another lifetime specifier
```

src/metrics/server.rs
```diff
@@ -36,49 +58,68 @@ impl Server {
...
+        let mut sync_writer = SyncIoBridge::new(writer);
+        let reg = reg.lock().expect("mutex");
+        let mut fmt_target = FmtToIoWriter {
+            inner: &mut sync_writer,
+        };
+        if let Err(e) = encode(&mut fmt_target, &reg) {
+            tracing::warn!("metric registry encoding failed: {}", e)
+        }
+        let _ = sync_writer.flush();
+    });
```
Stevenjin8: I thought about this more and 4096 seems really small, especially because we are holding the reg mutex the whole time.

Stevenjin8: I would prefer that we have an unbounded pipe or just read the whole thing and then send like we did before

fraenkel: I was trying to avoid having memory grow unbounded and then need even more to do the compression in memory as well.
But this will simplify the code.

Stevenjin8: that's fair, I'm just worried if there is network backpressure and we are stuck holding the lock for a long time.

fraenkel:
@Stevenjin8 any other changes needed?

src/hyper_util.rs
```diff
@@ -239,37 +234,21 @@ impl<S> Server<S> {
...
-
-                                        // Failures would abort the whole connection; we just want to return an HTTP error
-                                        f(state, req).or_else(|err| async move {
-                                            Ok::<Response<Full<Bytes>>, Infallible>(Response::builder()
-                                                .status(hyper::StatusCode::INTERNAL_SERVER_ERROR)
-                                                .body(err.to_string().into())
-                                                .expect("builder with known status code should not fail"))
-                                        })
-                                    }),
-                                );
```
Stevenjin8, line 272: @fraenkel sorry for the delay. I think we still need to preserve this behaviour for the admin server.

fraenkel: To add this back I would need 2 different server implementations. Is that ok?

fraenkel: I could also get away with 2 different spawns.

fraenkel: @Stevenjin8 This is one approach to support both types of services.

Stevenjin8 (approved):

PR #1980, opened 2026-07-09
The Network loadbalancer scope is preferred but not required

fraenkel:
When computing the rank of endpoints, if network is the first scope, we will prefer those that match, but will allow the remaining scopes to calculate its rank.
Those that match on network will have a higher ranking than those that do not.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -435,14 +435,20 @@ impl ProxyState {
 
         let options = match svc.load_balancer {
             Some(ref lb) if lb.mode != LoadBalancerMode::Standard => {
+                let network_preferred = !lb.routing_preferences.is_empty()
+                    && lb.routing_preferences[0] == LoadBalancerScopes::Network;
                 let ranks = endpoints
                     .filter_map(|(ep, wl)| {
                         // Load balancer will define N targets we want to match
                         // Consider [network, region, zone]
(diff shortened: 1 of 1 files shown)


stevenctl:
This is a pretty big change to the semantics of `failoverPriority`, so just making I sure I understand the implications:

Previously the network went down, traffic would go across all remote endpoints regardless of priority, now they'll failover to same-region first. 

The worst in-network zone is still higher ranked than the best out-of-network zone.

Previously a network outage would have distributed traffic to all gateways, regardless of locality. Now it will choose the closest gateway(s). This could be seen as a positive, but it's theoretically possible someone ended up relying on the previously weird behavior. 

Overall I think this is good, although it's awkward that it's based on the positioning within the list; we're only saying that network gets this special "orthogonal to the rest of locality" treatment in a  magical-convention .

fraenkel:
What you describe is correct.

The ideal would have been to specify a list of failoverPriorities so you can more accurately control how a ranking should be computed once the first scope is no longer true. But that isn't possible with the current configs for both istio or ztunnel.

There is also a bit of a nuance with the Region, Zone, SubZone scopes. These are nested which doesn't necessarily apply to Network and whatever follows.

stevenctl (approved):

PR #1986, opened 2026-07-14
Crl existing conns

conradhanson:
Introduces CRL enforcement for existing connections, both inbound and outbound. 

CRL Manager now has a rw-locked index (tree structure representing **verified** cert chain) that:
- is written to when a connection is registered (opened), or deregistered (closed)
- is read after CRL file load event to navigate index and collect affected subtrees for processing

To determine that a CRL affects existing connections, a CRL's issuer is directly compared to the subject field of a CA, and upon match, revoked serials are compared to the CA's child CA serials (if present, chained intermediate scenario) and leaf cert serials.
For a CA/IA revocation, a single (representative) leaf cert is used as the end entity as well as its cert chain in the webpki verifier call, and upon successful revocation, **_bulk-drop_** all its associated connections (leaves and child CA leaves to corresponding depth).

When IA or leaf cert is determined to be revoked, index increments crl revocation metric and sends a watcher signal which triggers a tunnel's tokio select arm for the connection to be terminated immediately. 
Access log attribution has been added so that when a connection is being terminated due to cert revocation the log correctly states why the connection has been closed (without it the connection would appear to close successfully instead of forcefully). 

Webpki verifier is the sole source of truth for cert parsing, revocation, and a connection close event.
**Note**: during benchmarking it was discovered that the webpki crate only supports cert chains up to 6 certs in length, more than that and webpki fails/exits early with `MaximumPathDepthExceeded`. So the benchmarking below is representative of this limitation.

Benchmarked the 3 main methods of the index:
- `register` is webpki-bound (~100 microsecs) and connection count-independent
- `deregister` is sub-microsec
- `navigate` is "free" (nanosecs) on non-matching CRL reloads and, on a match, is bounded by an O(leaves-under-issuer x CRL-count x Log of CRL-content-size) scan plus a single confirming webpki verification
- for reference, the webpki verification for single cert is ~30-60microsecs and scales linearly with the cert chain length.

[design doc](https://docs.google.com/document/d/1LztLf9QoprjR4B3ZUSDweyYvbkvJz7DGQ9llCk8wWc0/edit?tab=t.0) - benchmark details at end


src/tls/verifier.rs
```diff
@@ -0,0 +1,158 @@
...
+        revocation,
+        None,
+    )
+    .map_err(webpki_error_to_rustls)?;
+    Ok(())
+}
+
+// The signature-verification helpers below are identical for inbound and outbound:
+// both delegate to the active crypto provider's algorithms.
+
```
Stevenjin8:
```suggestion
```

src/tls/revocation.rs
```diff
@@ -0,0 +1,1404 @@
...
+/// Leaf revocation re-checks just the leaves whose serial the CRL lists.
+/// webpki remains the sole judge of revocation.
+///
+/// register/deregister take write lock only for map mutations, and
+/// navigation snapshots the affected subtree under read lock and then
+/// webpki verification and teardown signals run outside the lock.
+#[derive(Clone)]
+pub struct RevocationIndex {
+    metrics: Arc<Metrics>,
+    inner: Arc<RwLock<IndexInner>>,
```
Stevenjin8, line 353: @conradhanson I gave this a rough look and I think I'm happy with how everything is plumbed. My only point of contention is the use of a tree rather than straight hash map where the keys is just the certifacte serials concatenated. so if i had a peer that presented a cert chain with CA, IA, leaf, the following entries would be in the hash map

* <ca serial>
* <ca serial>.<ia serial>
* <ca serial>.<ia serial>.<leaf serial>

curious if you considered something like this.

Stevenjin8: wow that markdown rendered horribly. Anyways, maybe a better idea is key by <issuer der>.<cert serial> really just anything to avoid looping over every connection.

conradhanson: so just a flat hashmap, not nested, of every cert serial presented in a connection?
We cannot key by concatenated serials, this would give us no relevant info to do a lookup on. 
CRLs have two identifying pieces of info, an issuer and a list of serials that have been revoked. 

So for flat hashmap, the keys generated for every connection would be `subject DN + child serial` for every link in the cert chain presented (webpki limit is 6 cert depth) so we can easily lookup whether a CRL entry is being used, and a value that holds a list of entries where each entry has a sender unique to a connection (so it can be terminated) as well as all info needed to do an identical webpki verify call for that connection. So a lot of potential duplication on the value side for each link in the presented cert chain. 

Tree is more representative of the hierarchical nature of cert chains and has no info duplication scenario. 
In this tree implementation, we never loop over every single connection, its always confined to the matched issuer of the CRL.
If an IA has been revoked (either by a root or a parent IA), then we do a single webpki verify call on a representative leaf cert and, on revoked, bulk drop that IAs connections.
If a leaf has been revoked, then we find the matched issuer, and iterate its leaves to determine the serial match and, on revoked, drop that connection. 
We could instead have a map keyed by serial instead for the leaf scenario so it is constant lookup, but I believe  it adds a couple allocations on the `register/deregister` path.

Stevenjin8:
@ilrudie @stevenctl  I'll try my best to review this before the branch cut, but at the very least would like someone else to look at it

src/tls/revocation.rs
```diff
@@ -0,0 +1,1494 @@
...
+/// Contents only used for index navigation and pre-checks — webpki remains the sole judge of revocation.
+fn parse_cert(der: &[u8]) -> Option<CertInfo> {
+    use x509_parser::prelude::*;
+    let (_, cert) = X509Certificate::from_der(der).ok()?;
+    Some(CertInfo {
+        subject_dn: der_sequence_content(cert.subject().as_raw())?.to_vec(),
+        issuer_dn: der_sequence_content(cert.issuer().as_raw())?.to_vec(),
+        serial: cert.raw_serial().to_vec(),
+    })
+}
```
ilrudie: Could we do `webpki::EndEntityCert::try_from(der)` here rather than `x509_parser::from_der(der)`?

This would reduce the possibility that the two different parsers diverge in some subtle way and also probably let us get rid of the der_sequence_content function below.

src/tls/revocation.rs
```diff
@@ -0,0 +1,1494 @@
...
+                "crl index: peer presented no certificates; connection not tracked"
+            );
+            return None;
+        };
+        let Some(leaf) = parse_cert(leaf_der.as_ref()) else {
+            warn!(
+                peer = conn.peer(),
+                "crl index: could not parse leaf cert; connection not tracked"
+            );
+            return None;
```
ilrudie, line 546: If we hit this, the connection isn't added to the index, so navigate will never re-check it — it quietly degrades to "enforce on new connections only" for this tunnel (a revoked cert could keep serving an existing connection until it closes). That's important to surface: a mesh operator should know they aren't getting full existing-connection CRL enforcement for some connections. Suggest a counter here — e.g. istio_crl_untracked_connections_total, labeled by reporter (source/destination) to match istio_crl_policy_rejections_total — so this blind spot is alertable rather than silent.

src/tls/certificate.rs
```diff
@@ -311,25 +317,14 @@ impl WorkloadCertificate {
...
-        let mut builder = WebPkiClientVerifier::builder_with_provider(
+        // Custom client-cert verifier that shares the webpki chain + CRL logic with the outbound verifier.
+        // CRLs are read from the manager on every handshake, so CRL updates apply to new inbound connections immediately.
+        // TODO: check if our own certificate is revoked in the CRL and log warning
+        let client_cert_verifier = crate::tls::workload::WorkloadClientVerifier::new(
             self.root_store.clone(),
-            crate::tls::lib::provider(),
+            td,
+            crl_manager,
         );
```
ilrudie: This carries a real risk of behavior change, even for users who have not engaged with the feature. Since registering with the CRL manager re-runs a revocation check we could ostensibly keep the upstream client verifier + TD verifier and rely on the registration check to enforce the behaviors we want. WDYT?

src/tls/revocation.rs
```diff
@@ -0,0 +1,1494 @@
...
+type LeafId = u64;
+
+/// x509 serial number, stored as the DER `INTEGER` *content* (value, leading `0x00` sign byte included).
+/// used in webpki's [`CertRevocationList::find_serial`] for comparison.
+/// created by x509-parser's `raw_serial()`.
+/// `Vec` to hold all bytes verbatim so serial of any length isn't truncated (false negative).
+type Serial = Vec<u8>;
+
+/// One tracked tunnel, owned by its issuing CA node's `child_leaves`
+struct Leaf {
```
ilrudie, line 271: Don't change the PR, but we probably want to follow up on this. We could treat leaf nodes more like an intermediate and have the children just be the senders needed to revoke each connection. All the logic for dropping and cleaning up the data structure exists to handle the IAs. We'd only need to store one chain per identity instead of per tunnel and that might save a decent amount of memory and CPU of a busy node.

tests/namespaced.rs
```diff
@@ -1687,6 +1687,196 @@ mod namespaced {
...
+        info!("initial connection should succeed, CRL is currently empty");
+        tx.send_and_wait(()).await?;
+
+        // revoke client cert by writing the pre-built CRL into the file watched by the server ztunnel
+        info!("revoking client cert and waiting for CRL to update");
+        crl_file.write_all(crl_pem.as_bytes())?;
+        crl_file.flush()?;
+
+        // wait for the CRL file-watcher debounce (~2s) + reload + abrupt GOAWAY teardown to propagate
+        tokio::time::sleep(Duration::from_secs(4)).await;
```
ilrudie: Can we switch these tests to follow an `eventually_true` pattern instead of a fixed wait? Just give them a generous timeout to avoid flakes and check that eventually the verify_metric_exists is ok before we move on.

src/tls/revocation.rs
```diff
@@ -0,0 +1,1494 @@
...
+        ));
+        let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
+        let addr = listener.local_addr().unwrap();
+        let accept = tokio::spawn(async move {
+            let (stream, _) = listener.accept().await.unwrap();
+            server_tls.accept(stream).await.unwrap()
+        });
+
+        let stream = TcpStream::connect(addr).await.unwrap();
+        // avoid the
```
ilrudie: **nit** delete or finish the thought?

src/tls/revocation.rs
```diff
@@ -0,0 +1,1494 @@
...
+}
+
+#[derive(Default)]
+struct NavWork {
+    leaf_probes: Vec<LeafProbe>,
+    ca_probes: Vec<CaProbe>,
+}
+
+/// Index of every existing connection, owned by [`CrlManager`].
+/// Navigation walks only connections under CRL issuing CA(s).
```
ilrudie:
```suggestion
/// Index of every tracked leaf, owned by [`CrlManager`].
/// Navigation walks only leaves under CRL-issuing CA(s).
```

ilrudie (commented):
First pass, mostly looking solid.

src/tls/revocation.rs
conradhanson: 100% agree on this, will change.

tests/namespaced.rs
conradhanson: nice catch, absolutely

src/tls/revocation.rs
conradhanson: great callout, this is definitely an edge case we need to surface to the user so that it is known. will add a metric counter for this

src/tls/certificate.rs
conradhanson: tldr; ian and i talked about this and opted to keep existing behavior to mitigate behavior change risk here.

src/tls/crl.rs
```diff
@@ -40,9 +41,13 @@ pub enum CrlError {
...
 
 #[derive(Clone)]
-/// Both **inbound** and **outbound** mesh TLS verifiers call [`CrlManager::get_crls`] on every
+/// **Outbound** mesh TLS verifiers call [`CrlManager::get_crls`] on every
 /// handshake, so CRL file updates apply to new connections immediately.
 ///
+/// **Inbound** `ServerConfig` creation calls [`CrlManager::get_crl_ders`] and
+/// is rebuilt when the workload certificate refreshes (~12h),
+/// so CRL changes apply on that cadence for new connection handshakes unless the server config path changes.
+///
```
ilrudie: Are we counting anything that makes it past handshake an existing connection? The post-registration CRL check to close the race still occurs before we drive the bi-directional copy.

conradhanson: yes. technically a crl could revoke after successful handshake and before the connection actually starts moving data.

conradhanson: going to simplify this comment so its crystal clear

src/tls/revocation.rs
```diff
@@ -338,6 +338,7 @@ impl RevocationIndex {
         // Verify again to obtain the verified path, and route the index by that instead of the peer's presented chain order.
         // Presented chains can be compromised or non-conforming that still handshake (webpki ignores unused ones)
         // and corrupt index structure causing a missed connection revocation.
+        let mut verified_chain: Option<Vec<CertificateDer<'static>>> = None;
```
ilrudie: I think the match can produce a (tracked, verified_chain) tuple which is a little clearer to read than the mutable variable where you need to go searching for assignments

conradhanson: agreed, that is definitely cleaner. changing now

src/tls/revocation.rs
```diff
@@ -0,0 +1,1657 @@
...
+/// CA node id (root or IA) — globally unique, minted by a monotonic counter (never reused).
+type NodeId = u64;
+/// Per-tunnel id, minted by a monotonic counter (never reused).
+/// Distinguishes tunnels sharing identical peer cert (tunnel multiplex limit is 100).
+/// `u64` so the counter cannot realistically wrap over a long-lived proxy's uptime.
+type LeafId = u64;
+
+/// x509 serial number, stored as the DER `INTEGER` *content* (value, leading `0x00` sign byte included).
+/// used in webpki's [`CertRevocationList::find_serial`] for comparison.
+/// created by x509-parser's `raw_serial()`.
```
ilrudie: **nit** Is this still true? I think we use the webpki types instead in all cases now?
```suggestion
```

src/tls/revocation.rs
```diff
@@ -0,0 +1,1657 @@
...
+                    // handshake already verified this chain, so a failure here is unexpected
+                    Err(e) => {
+                        warn!(
+                            peer = conn.peer(),
+                            error = %e,
+                            "crl index: re-verification failed; connection not tracked"
+                        );
+                        self.metrics.record_crl_untracked_connection(conn.reporter);
+                        (None, None)
+                    }
```
ilrudie: This might happen due to an inherent race caused by doing two different verifies. The canonical one for _most_ of the checks happens at handshake and so most failures here are reduced from **kill the connection** to **untracked by CRL manager** except for a revoke where this check does have the final say. Example (admittedly contrived, but it shows the idea): peer cert expires between accept and this check => untracked because it was valid at accept and this check is non-canonical for expiration.

ilrudie: FWIW, this is just part and parcel with the way our verify APIs presently function. We'd probably need pretty major surgery or an upstream PR to materially close the race out.

src/proxy/metrics.rs
```diff
@@ -399,6 +404,13 @@ impl Metrics {
             crl_policy_rejections.clone(),
         );
 
+        let crl_untracked_connections = Family::default();
+        registry.register(
+            "crl_untracked_connections",
+            "The total number of connections not tracked for CRL existing-connection enforcement because a peer certificate failed to parse (unstable)",
```
ilrudie: There are other reasons this may increment, such as the race discussed in previous comments causing a non-canonical verify fail (cannot track instead of kill connection). Maybe we should clean up the description to not be prescriptive about why.

conradhanson: agreed, this was formed before adding the metric elsewhere so ill make the description more generic.

src/tls/revocation.rs
conradhanson: dang it, missed that one. will remove

conradhanson: its all webpki now, no x509 parser.

src/tls/revocation.rs
conradhanson: agreed. as previously discussed, if we could move the register call to the point at which the handshake verification happened, it could simply this func and close the race gap between handshake verification and insertion into the crl manager index. But as it stands we can't get the verified chain from that handshake flow so the APIs around it would need to change in an insignificant way.

src/tls/revocation.rs
```diff
@@ -0,0 +1,1656 @@
...
+            && chain_is_revoked(
+                crl_manager,
+                chain,
+                &conn.roots,
+                conn.key_usage,
+                conn.established,
+            )
+        {
+            drop_revoked(&self.metrics, conn.reporter, &tx);
+        }
```
ilrudie: I don't think this code path has an explicit test. Is it possible to add?

src/tls/revocation.rs
```diff
@@ -0,0 +1,1656 @@
...
+    revoked_tx: Arc<watch::Sender<bool>>,
+    roots: Arc<RootCertStore>,
+    /// leaf cert serial number
+    serial: Serial,
+}
+
+/// CA node (root or intermediate)
+struct CaNode {
+    /// Child CA nodes issued by this CA, keyed by the child's serial (unique per issuer)
+    child_cas: HashMap<Serial, NodeId>,
```
ilrudie, line 289: Technically this isn't incorrect. It would require a CA bug to happen, but if a CA did issue a duplicate serial number this mapping wouldn't be unique anymore the trees underneath would merge. I don't know if we need to block the merge on it, but we should consider if this is something we want to try to protect against by keying on dn+serial or something...

src/tls/revocation.rs
```diff
@@ -0,0 +1,1656 @@
...
+    pub fn new(metrics: Arc<Metrics>) -> Self {
+        Self {
+            metrics,
+            inner: Arc::new(RwLock::new(IndexInner::default())),
+        }
+    }
+
+    /// Track a connection at handshake and return its [`RevocationHandle`].
+    /// Inserts the leaf + its verified CA path into the index.
+    pub fn register(&self, crl_manager: &CrlManager, conn: ConnRegistration) -> RevocationHandle {
```
ilrudie, line 368:
```suggestion
    /// Track a connection at handshake and return its [`RevocationHandle`].
    /// Inserts the leaf + its verified CA path into the index.
    /// For both inbound and outbound, register owns the final decision for revocation. All other
    /// decisions are owned by the handshake.
    pub fn register(&self, crl_manager: &CrlManager, conn: ConnRegistration) -> RevocationHandle {
```

src/tls/revocation.rs
conradhanson: it isn't testable as part of the register call itself since it is determined by crl reload timing.
we would need to pull that logic into its own func and just test that. 
is that acceptable?

src/tls/revocation.rs
ilrudie: maybe something like this for clarity?
```suggestion
                    // handshake already verified this chain, so a failure here is likely the result of a race
                    // between state updates and verifyForUse calls.
                    Err(e) => {
                        warn!(
                            peer = conn.peer(),
                            error = %e,
                            "crl index: re-verification failed; connection not tracked"
                        );
                        self.metrics.record_crl_untracked_connection(conn.reporter);
                        (None, None)
                    }
```

src/tls/revocation.rs
conradhanson: yea i did consider that scenario as well as the dn+serial keying because of it.
A CA issuing duplicate serials goes directly against industry standards and causes a cascade of issues.
current implementation does assume that CAs follow industry standards in general, otherwise things breakdown.

I opted for keying child CAs by serial since it simplifies the lookup aspect of tree traversal. 
If we were to key by dn+serial there is still a scenario where a key collision could occur because the dn could potentially be identical too. The only way to truly have a unique key here is to use spki+dn+serial which I think adds major complexity for a pretty edgy edge case.

conradhanson: the other major consideration for why i used serial for child cas is because of how we traverse the tree based on CRL content. A CRL only has the issuer and revoked serials to operate off of. keying by anything other than a pure serial means that we need a way of getting the additional info to form the proper key.

src/tls/revocation.rs
conradhanson: moved this to a func with the always inline directive and added unit test.

src/tls/revocation.rs
```diff
@@ -0,0 +1,1759 @@
...
+type LeafId = u64;
+
+/// raw cert serial number, stored as the DER `INTEGER` content (value, leading `0x00` sign byte included).
+/// used in webpki's [`CertRevocationList::find_serial`] for comparison.
+/// `Vec` to hold all bytes verbatim so serial of any length isn't truncated (false negative).
+type Serial = Vec<u8>;
+
+/// One tracked tunnel, owned by its issuing CA node's `child_leaves`
+struct Leaf {
+    /// cert chain (leaf -> IA(s), no root) exactly as presented at handshake for identical webpki verification
```
ilrudie: comment no longer true
```suggestion
    /// webpki-validated cert chain (leaf -> IA(s), no root), retained for webpki verification
    /// against updated CRLs during navigation
```

ilrudie (approved):

Issue #1966, opened 2026-06-30
ztunnel data plane freezes for several seconds on every ~30 min XDS reconnect at scale: the single ProxyState RwLock is held across the full workload re-insert, blocking all reads on the 2-worker runtime


es-ma-n:
### Summary

On a large ambient mesh (\~69.5k workloads), every ztunnel periodically (\~30 min, istiod keepalive-driven) reconnects its XDS stream. On reconnect, istiod re-sends the **full** workload set, and ztunnel applies it by taking the single `Arc<RwLock<ProxyState>>` **write** lock **once** and inserting the **entire batch under that one guard**. Because

1. the lock is a synchronous `std::sync::RwLock`,
2. the data path takes a **read** lock on the same `ProxyState` for every connection/upstream lookup, and
3. the runtime defaults to only **2 worker threads**,

the data-plane worker threads **block (park) on `state.read()` for the whole duration of the re-insert — we measured 1.7–4.9 s** depending on workload count and node single-core speed. During that window the data plane makes \~no forward progress: **we observe ztunnel's connection processing collapse to near-zero, then a backlog flush** (Evidence 2). Because reads are synchronous and there are only 2 worker threads, **in-flight forwarding for already-established connections is expected to stall as well** (both workers park on `state.read()`) — though our direct measurement is of connection-event processing, not per-connection byte rate. Either way, a latency-sensitive, no-retry protocol traversing that ztunnel fails on a \~30 min cadence.

This is a concrete, production-impacting instance of #203 (`Reduce Xds config race`, open since 2022). We'd like to advocate for moving the data-plane read path to a **lock-free / thread-local snapshot** design (Envoy-style TLS). **We're willing to do the work — prototype the change and submit a PR** — and would welcome maintainers' guidance on the preferred direction before we start.

### Environment

- Node OS: **RHEL 8.9 / 8.10** (x86_64); container runtime **containerd 1.6.28**; Kubernetes **v1.23.15**.
- ztunnel/Istio **1.26.0** in production. We verified the relevant code is **unchanged on `master`** (HEAD `94ffa3a1`, 2026-06) **and 1.30.2**.
- Ambient mode; ztunnel as a per-node DaemonSet.
- `numWorkerThreads = 2` (the default; confirmed live via the admin `config_dump`).
- Two clusters for comparison:
  - **Cluster A** — \~69.5k workloads, nodes 40–80 cores.
[...]

PR #1987, opened 2026-07-14
improved wds full push handling

ilrudie:
Working on #1966

## TLDR
Diff full pushes against state to reduce them to a delta before taking the write lock. It saves a lot of time under write lock when a running ztunnel reconnects to istiod.

## Goals:
- improved handling of the large "full push" which is received when ztunnel reconnects
- keep small delta pushes fast

## Non-goals:
- Handling a true super large push which isn't mostly no-op updates like we receive when re-connecting to istiod

## Also considered:
ArcSwap|mem::Swap type of implementation. This would require doing a clone of potentially large state and applying the push off lock, then swapping it out for an O(1) operation under write lock. This would be a more optimal path for true very large diffs, but the basic approach would render small delta pushes intolerably slow I think. A form of swapping undertaken when the prepared + diff-reduced batch is still quite large would be a potential enhancement to follow up with, which I have not yet explored very thoroughly.

There are some immutable map type data structures we could consider to reduce the costs of this "clone to update" operation, however they seem to be more optimal for write at a cost of more expensive read. I suspect the additional cost paid on every read outweighs the more optimal write many-fold.

## High level implementation:
- Move preparation of the batched update off lock
- If the update is large, reduce it by diffing against present state
- Apply the prepared batch under write lock

## Result
I tested two different "real world" cluster shapes based on the information graciously provided by @es-ma-n. This was simulated in a small KinD cluster augmented by `pilot-load` to handle the many mock workloads and services that make the full push bite. The times shown in the table are the P100 latency observed by fortio driving all new connections through ztunnel across 6 reconnects (full push).

| cluster shape | baseline | this PR |
| --- | --- | --- |
| clusterA (max ~2,700-endpoint service) | 597 ms | 17 ms |
| clusterB (max ~1,600-endpoint service) | 509 ms | 13 ms |

### DIFF
--- a/src/xds.rs
+++ b/src/xds.rs
@@ -386,11 +386,9 @@ impl Handler<XdsWorkload> for ProxyStateUpdater {
         &self,
         updates: Box<&mut dyn Iterator<Item = XdsUpdate<XdsWorkload>>>,
     ) -> Result<(), Vec<RejectedConfig>> {
-        // Phase 1: convert every resource off-lock. A malformed resource only
-        // rejects itself and doesn't abort the batch.
-        let mut prepared = Vec::new();
+        let mut batch = PreparedBatch::default();
         let result = handle_single_resource(updates, |res: XdsUpdate<XdsWorkload>| {
-            prepared.push(match res {
+            batch.push(match res {
                 XdsUpdate::Update(w) => {
                     PreparedUpdate::Workload(PreparedWorkload::try_from_xds(w.resource)?)
                 }
@@ -398,7 +396,15 @@ impl Handler<XdsWorkload> for ProxyStateUpdater {
             });
             Ok(())
         });
-        // Phase 2: take the write lock once and apply the converted batch.
+        // For large batches, drop no-op updates by diffing against
+        // current state under a read lock (shared; doesn't block readers).
+        let prepared = if batch.should_diff() {
+            let state = self.state.read().unwrap();
+            reduce_by_diff(&state, batch.updates)
+        } else {
+            batch.updates
+        };
+        // Take the write lock once and apply the reduced batch.
         let mut state = self.state.write().unwrap();
         self.updater.apply_prepared(&mut state, prepared);
         result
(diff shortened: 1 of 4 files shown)


src/xds.rs
```diff
@@ -313,38 +416,159 @@ impl Handler<XdsAddress> for ProxyStateUpdater {
...
+fn reduce_by_diff(state: &ProxyState, prepared: Vec<PreparedUpdate>) -> Vec<PreparedUpdate> {
+    // Nothing to diff against on the initial push (empty state) — every resource
+    // is new, so skip the comparison pass and apply the batch as-is.
+    if state.workloads.is_empty() && state.services.is_empty() {
+        return prepared;
+    }
+    prepared
+        .into_iter()
+        .filter(|update| match update {
+            PreparedUpdate::Workload(pw) => state
```
jgkawell, line 545: Comparing Workloads here means we might miss endpoint changes and accidentally drop them. Since ports are defined outside of a Workload, if the PortList changed but the Workload was otherwise the same we'd lose that port update, right?

jgkawell: Comparing endpoints might get tricky since their inclusion depends on the Service health policy though

src/xds.rs
```diff
@@ -292,19 +386,28 @@ impl Handler<XdsWorkload> for ProxyStateUpdater {
...
+                    PreparedUpdate::Workload(PreparedWorkload::try_from_xds(w.resource)?)
                 }
-            }
+                XdsUpdate::Remove(name) => PreparedUpdate::Remove(strng::new(name)),
+            });
             Ok(())
+        });
+        // For large batches, drop no-op updates by diffing against
+        // current state under a read lock (shared; doesn't block readers).
+        let prepared = if batch.should_diff() {
```
howardjohn, line 401: I solved this in AGW at the control plane fwiw: http://github.com/agentgateway/agentgateway/pull/2497. should be feasible in istio too

ilrudie:
Superseded by https://github.com/istio/ztunnel/pull/1996 and https://github.com/istio/istio/pull/60935

PR #1993, opened 2026-07-15
CPU-aware defaults for worker threads

jgkawell:
ztunnel's worker pool previously defaulted to a hardcoded 2 worker threads, regardless of how many cores the node had. On large nodes with no CPU limit this capped throughput at 2 threads. This change makes the default cpu-aware while keeping existing overrides.

### details

When neither `ZTUNNEL_WORKER_THREADS` nor proxy `concurrency` is set, derive the worker-thread count from the pod's CPU configuration instead of using a flat default of 2:

- `ZTUNNEL_RESOURCE_CPU_LIMIT` (set by the chart only when `resources.limits.cpu` is configured) is used as the worker-thread count (floored at 2, and raised to the request if the request is higher).
- `ZTUNNEL_RESOURCE_CPU_REQUEST` (set by the chart only when `resources.requests.cpu` is configured) raises the worker-thread floor so a pod that reserves N cores gets at least N worker threads.
- With neither set, fall back to a fraction of the node's logical cores (default of 25%)

These are separate from `ZTUNNEL_CPU_LIMIT`, which the downward API also fills with node-allocatable cores when no limit is set and therefore cannot distinguish an explicit limit from an unlimited pod.

Also handle `concurrency == 0` as 'use all available cores' (previously this panicked the runtime).

### scenarios

- **No config, large node:** ztunnel now scales to ~25% of node cores instead of
  being stuck at 2 worker threads
- **CPU limit configured:** worker count tracks the limit, so a pod sized for N
  CPUs actually uses N worker threads
- **CPU request configured:** worker count won't drop below the request, so a pod
  that reserves N cores always has at least N worker threads
- **Existing overrides:** `ZTUNNEL_WORKER_THREADS` as a number or `%`, proxy
  `concurrency` still work
- **Small/limited pods are protected:** a binary-only upgrade (new env var not yet injected by the chart) won't over-thread a limited pod, because num_cpus is cgroup-aware and the 25% default is computed from the limited core count.

### istio PR: https://github.com/istio/istio/pull/60899

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -456,20 +456,41 @@ fn get_cpu_count() -> Result<usize, Error> {
     }
 }
 
+/// Parse an env var holding a Kubernetes CPU quantity (e.g. "2", "1.5", "500m") into a
+/// whole number of CPUs, rounding up.
+fn parse_cpu_quantity(env: &str) -> Result<Option<usize>, Error> {
+    let Some(raw) = parse::<String>(env)? else {
+        return Ok(None);
+    };
+    let cores = match raw.strip_suffix('m') {
+        Some(millis) => millis.parse::<f64>().map(|m| m / 1000.0),
+        None => raw.parse::<f64>(),
+    }
+    .map_err(|e| Error::EnvVar(env.to_string(), raw.clone(), e.to_string()))?;
+    if !cores.is_finite() || cores <= 0.0 {
+        return Err(Error::EnvVar(
+            env.to_string(),
+            raw,
+            "must be a positive CPU quantity".to_string(),
+        ));
+    }
+    Ok(Some(cores.ceil() as usize))
+}
+
 /// Default worker-thread count when neither ZTUNNEL_WORKER_THREADS nor proxy
 /// `concurrency` is set.
 fn default_worker_threads() -> Result<usize, Error> {
-    // Maximum of user set request and default
-    let floor = match parse::<usize>(ZTUNNEL_RESOURCE_CPU_REQUEST)? {
-        Some(request) => request.max(DEFAULT_WORKER_THREADS as usize),
-        None => DEFAULT_WORKER_THREADS as usize,
-    };
-    // User set limit or fraction of available CPUs
-    let target = match parse::<usize>(ZTUNNEL_RESOURCE_CPU_LIMIT)? {
-        Some(limit) => limit,
-        None => (num_cpus::get() as f64 * DEFAULT_WORKER_THREAD_CPU_FRACTION).ceil() as usize,
-    };
-    Ok(target.max(floor))
+    let request = parse_cpu_quantity(ZTUNNEL_RESOURCE_CPU_REQUEST)?.unwrap_or(0);
+    match parse_cpu_quantity(ZTUNNEL_RESOURCE_CPU_LIMIT)? {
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -443,6 +456,22 @@ fn get_cpu_count() -> Result<usize, Error> {
...
+/// Default worker-thread count when neither ZTUNNEL_WORKER_THREADS nor proxy
+/// `concurrency` is set.
+fn default_worker_threads() -> Result<usize, Error> {
+    // Maximum of user set request and default
+    let floor = match parse::<usize>(ZTUNNEL_RESOURCE_CPU_REQUEST)? {
+        Some(request) => request.max(DEFAULT_WORKER_THREADS as usize),
+        None => DEFAULT_WORKER_THREADS as usize,
+    };
+    // User set limit or fraction of available CPUs
+    let target = match parse::<usize>(ZTUNNEL_RESOURCE_CPU_LIMIT)? {
```
stevenctl: these could be fractional or expressed as mCPU eg `500m`

src/config.rs
```diff
@@ -443,6 +456,22 @@ fn get_cpu_count() -> Result<usize, Error> {
...
+        Some(limit) => limit,
+        None => (num_cpus::get() as f64 * DEFAULT_WORKER_THREAD_CPU_FRACTION).ceil() as usize,
+    };
+    Ok(target.max(floor))
```
stevenctl: An explicit CPU limit of < 2 isn't allowed/desired?

src/config.rs
stevenctl, line 446: should this contain all of the possible ways to do the lookup:

```
   match parse::<usize>(ZTUNNEL_RESOURCE_CPU_LIMIT)? {
        Some(limit) => Ok(limit),
        None => match parse::<usize>(ZTUNNEL_CPU_LIMIT)? {
            Some(limit) => Ok(limit),
            None => Ok(num_cpus::get()),
        },
    }
```

src/config.rs
```diff
@@ -443,6 +456,22 @@ fn get_cpu_count() -> Result<usize, Error> {
...
+}
+
 /// Parse worker threads configuration, supporting both fixed numbers and percentages
 fn parse_worker_threads(default: usize) -> Result<usize, Error> {
```
stevenctl, line 501: We say concurrency = 0 means all cores, but I think this flows into Builder::worker_threads(0) at src/app.rs:271; we should clamp whatever we assign to config.num_worker_threads to be 1 at least

src/config.rs
ilrudie: Honestly, the right thing to do with limit(1) is hard to sus out I think. The rub is the main thread lurking behind the scenes.

`limit:1, worker:1` results in 2 threads competing for 1 cpu allotment. They'll stall waiting for the next window when the main thread is active already. They may also block each other since the main thread can take a write lock on state, and read lock is not presently async. If that happens the proxy moves no data until the write lock releases. Brutal.

With a busy control plane, 2 threads might win out. In a quiet cluster where the main thread isn't working hard, 1 thread probably give you the best performance.

ilrudie: I did allow 1 worker for this case. All options are probably bad though

src/config.rs
ilrudie: Good call out. Clamped in construct_config now.

stevenctl (approved):

PR #1995, opened 2026-07-17
Working on Memory Management

ilrudie:
Inspired and influence by https://github.com/agentgateway/agentgateway/pull/1696

glibc exhibits a behavior where memory which ztunnel has freed is not returned. This causes the working set to grow in a way which is often mistaken for a memory leak. From a system perspective, freed memory which isn't given back is leaky, but it's not indicative of ztunnel retaining stale references. jemalloc has better behaviors in this regard.

| arm | idle | peak* | after idle |
| --- | --- | --- | --- |
| master (glibc) | ~2 | 33–37 | **34.6–35.1 (frees ~0)** |
| patched (jemalloc no profiling)| ~3.6 | 24–27 | **8.8–9.8** |
| patched (jemalloc profiling enabled, not active) | ~3.8 | 24–27 | **9.0–10.1** |

*peak is working set size measured during fortio 100 concurrency no keep alive benchmarking session

Additionally, glibc complicates the collection of heap profiles. Today, this requires a custom build of ztunnel. You cannot profile any normal build published by the Istio project. Switching to jemalloc would allow us to have heap profiling capability compiled in and enabled on our default builds.

Performance is mostly on par. Slightly worse on c1 but better under higher churn c100 benchmarks.

| arm | c1 max | c10 max | c100 max | c100 no-keepalive |
| --- | --- | --- | --- | --- |
| master (glibc) | 9574 | 40830 | 59487 | 11085 |
| patched (jemalloc profiling enabled, not active) | 9477 | 40996 | 59741 | 11436 |

**note** This round of benchmarking was done on my consumer-grade, not server-grade, hardware.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,11 @@
 [build]
 target-dir = "out/rust"
+
+[target.'cfg(target_os = "linux")']
+rustflags = ["-C", "force-frame-pointers"]
+
 [env]
 BORING_BSSL_FIPS_PATH = { value = "vendor/boringssl-fips/linux_x86_64", force = true, relative = true }
 BORING_BSSL_FIPS_INCLUDE_PATH = { value = "vendor/boringssl-fips/include/", force = true, relative = true }
+CFLAGS = "-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"
+CXXFLAGS = "-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ rust-version = "1.90"
 autobenches = false # benches must be explicitly declared
 
 [features]
-default = ["tls-aws-lc"]
+default = ["tls-aws-lc", "jemalloc"]
 jemalloc = ["dep:tikv-jemallocator", "dep:jemalloc_pprof"]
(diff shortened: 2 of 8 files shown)


.cargo/config.toml
```diff
@@ -1,5 +1,11 @@
...
 target-dir = "out/rust"
+
+[target.'cfg(target_os = "linux")']
+rustflags = ["-C", "force-frame-pointers"]
+
 [env]
 BORING_BSSL_FIPS_PATH = { value = "vendor/boringssl-fips/linux_x86_64", force = true, relative = true }
 BORING_BSSL_FIPS_INCLUDE_PATH = { value = "vendor/boringssl-fips/include/", force = true, relative = true }
+CFLAGS = "-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"
+CXXFLAGS = "-fno-omit-frame-pointer -mno-omit-leaf-frame-pointer"
```
Stevenjin8, line 11: I'm guessing we need these to do the heap profile?

Stevenjin8 (approved):

.cargo/config.toml
ilrudie: Its aligning the compilation for any c code we depend on with our rust compilation so we shouldn't wind up with work/memory we can't attribute in our profiles just because it's not from a pure rust dependency.

PR #1996, opened 2026-07-20
reduce wds full push on ztunnel reconnect

ilrudie:
This is the ztunnel side of https://github.com/istio/istio/pull/60935

When combined they address the freeze/stall reported in #1966 caused by the full WDS push sent on a reconnect.

Istiod now sends a version hash with each resource, which ztunnel retains. When it connects, ztunnel sends these versions along with the names of its resources. Istiod can use this information reduce the full push to a delta.

## Benchmarking

Both profiles are real customer shapes at ~69k workloads: clusterA has fewer, larger
services (one 2,700-endpoint service; ΣEi²≈8.1M), clusterB many smaller ones (one
1,600-endpoint service; ΣEi²≈2.9M).

| arm | profile | reconnects | max p100 (ms) |
| --- | --- | ---: | ---: |
| baseline | clusterA | 5 | 391 |
| patched | clusterA | 5 | 12 |
| baseline | clusterB | 6 | 309 |
| patched | clusterB | 5 | 12 |

### DIFF
--- a/src/state/service.rs
+++ b/src/state/service.rs
@@ -125,6 +125,20 @@ impl EndpointSet {
         self.inner.insert(k, Arc::new(v));
     }
 
+    pub fn extend(&mut self, eps: impl IntoIterator<Item = (Strng, Endpoint)>) {
+        for (k, v) in eps {
+            self.inner.insert(k, Arc::new(v));
+        }
+    }
+
+    pub fn len(&self) -> usize {
+        self.inner.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.inner.is_empty()
+    }
+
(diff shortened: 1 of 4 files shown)


src/state/service.rs
```diff
@@ -593,63 +613,62 @@ impl ServiceStore {
...
+                changed = true;
+                svc.endpoints.extend(accepted);
+            }
 
-                // Update the service.
+            // Single reindex for the whole batch.
+            if changed {
                 self.insert_endpoint_update(svc);
             }
+        } else {
```
jgkawell, line 653: nit: unless I'm missing something, this else branch isn't being tested

stevenctl (approved):
LGTM

PR #2002, opened 2026-07-23
Add test coverage for `apply_endpoints`

jgkawell:
Adding a few tests to validate wonky endpoint insertion ordering. An add-on to #1996

### DIFF
--- a/src/state/service.rs
+++ b/src/state/service.rs
@@ -1621,11 +1621,10 @@ mod tests {
         assert_eq!(store.endpoint_reindexes(), 0);
     }
 
-    // A workload re-insertion puts the same uid in both removals and upserts in
-    // one batch. Removals are applied before upserts, so when the new endpoint
-    // is health-filtered out the stale one is still dropped.
+    // A workload going unhealthy on re-insertion (same uid in both the removal and
+    // the upsert) must still evict its previously-healthy endpoint.
     #[test]
-    fn apply_endpoints_present_removal_before_filtered_upsert() {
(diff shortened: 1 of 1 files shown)


src/state/service.rs
```diff
@@ -1544,4 +1559,98 @@ mod tests {
...
+
+        // Re-insertion of the same uid, now unhealthy: the removal drops the stale
+        // endpoint and the unhealthy upsert is filtered out, so ep1 disappears.
+        let mut upserts = HashMap::new();
+        upserts.insert("ep1".into(), endpoint("ep1", HealthStatus::Unhealthy));
+        store.apply_endpoints(&host, upserts, HashSet::from(["ep1".into()]));
+        assert_eq!(store.endpoint_reindexes(), 2);
+        let svc = store.get_by_namespaced_host(&host).unwrap();
+        assert_eq!(svc.endpoints.len(), 0);
+        assert!(!svc.endpoints.contains(&"ep1".into()));
```
ilrudie, line 1653: Because the endpoint is unhealthy and thus filtered, wouldn't the same assertions pass if we had done the removal after the upsert?

jgkawell: hmm yeah I was initially thinking ordering was important here but I think you're right. We still want this test but the wording around ordering can be removed

ilrudie: Reword on this one seems right, I think you could have a test that asserts the remove before upsert invariant too.

jgkawell: added

ilrudie (approved):
lgtm

PR #2009, opened 2026-07-27
Don't use gsutil in aws

Stevenjin8:
In AWS, we should not use gsutil.


ilrudie (approved):
thanks for adding the context!

keithmattix (approved):

PR #2013, opened 2026-07-28
Log improvements from agw

ilrudie:
Based on work in AgentGateway. This provides a modest improvement in logging performance under high load.


src/telemetry/nonblocking.rs
```diff
@@ -0,0 +1,451 @@
...
+
+/// The default maximum number of buffered log lines.
+///
+/// If [`NonBlocking`] is lossy, it will drop spans/events at capacity.
+/// If [`NonBlocking`] is _not_ lossy, backpressure will be exerted on
+/// senders, causing them to block their respective threads until there
+/// is available capacity.
+///
+/// Recommended to be a power of 2.
+pub const DEFAULT_BUFFERED_LINES_LIMIT: usize = 128_000;
```
Stevenjin8, line 38: This is not a power of two.

src/telemetry/worker.rs
```diff
@@ -0,0 +1,227 @@
...
+        buf.push(msg);
+        let mut res = Ok(());
+        while buf.can_push() {
+            let try_recv_result = self.receiver.try_recv();
+            match self.handle_try_recv(try_recv_result) {
+                Ok(Some(msg)) => {
+                    buf.push(msg);
+                }
+                Ok(None) => break,
+                Err(e) => res = Err(e),
```
Stevenjin8, line 86: If we get multiple errors, all but the last get swallowed.

Stevenjin8 (approved):
some minor nits that I don't super care about. Merge when you're happy with it.

src/telemetry.rs
```diff
@@ -576,3 +585,166 @@ pub mod testing {
...
+                c.wait().ok();
+                d
+            }
+            other => panic!("unknown sink {other}"),
+        }
+    }
+
+    #[test]
+    #[ignore = "benchmark; run explicitly with --ignored --nocapture"]
+    fn bench_access_log_end_to_end() {
```
howardjohn: nit: would either make it a normal benchmark or just remove it probably

howardjohn (commented):
LGTM but given this is (mostly) copying code I wrote will leave to someone else to approve?

src/telemetry/nonblocking.rs
ilrudie: This is verbatim from tokio (and agw). I'm not opposed to fixing per se, but there is perhaps value in leaving it matching? lmk if you'd disagree with leaving it matching the source

src/telemetry/worker.rs
ilrudie: Good point, I think it is ~harmless because both errors converge on the same behaviors :thinking:

src/telemetry/nonblocking.rs
Stevenjin8: its close enough

PR #2017, opened 2026-07-28
Improve hbone pool coordination

mmcloud:
This PR changes how ztunnel manages shared HBONE connections.

### Current
ztunnel manages shared HBONE connections using a separate asynchronous mutex for each source-destination key. This ensures that only one task creates a new connection for that key at a time. However, the mutex remains locked while the TCP, TLS, and HTTP/2 connection is being established, and waiting tasks use a notification shared across the entire pool. As a result, connection setup can hold the lock for a relatively long time, and creating a connection for one destination can unnecessarily wake tasks waiting for other destinations.

### Target state
Coordination model with a separate pool entry for each source-destination key. Each entry tracks connection creation, waiting tasks, failures, retry backoff, available connections, and cleanup. The entry mutex is held only briefly while this state is checked or updated; the slower TCP, TLS, and HTTP/2 connection setup happens after the mutex has been released. Waiting tasks are notified only when something changes for their own source-destination key.

### Graceful drain behaviour

The pool now observes proxy drain events directly. The graceful drain currently appears only on outbound connections but not on the hbone connection.

During graceful drain:

New HBONE connection creation stops.
Blocked creators and waiters are released.
Existing HBONE streams remain active until normal drain completion.

### Reusing full HTTP/2 connections

An HBONE connection can carry several HTTP/2 streams at the same time.

When a connection reaches its configured stream limit, it is temporarily removed from the available pool. This PR adds tracking that can put the connection back into the pool when the stream counter shows that capacity is available again.

This can reduce the need to create additional HBONE connections after existing streams finish.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,10 @@ harness = false
 name = "basic"
 harness = false
 
+[[bench]]
+name = "pool"
+harness = false
+
 [dependencies]
 # Enabled with 'tls-boring'
 boring-rustls-provider = { git = "https://github.com/janrueth/boring-rustls-provider", optional = true } #
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,10 @@ harness = false
 name = "pool"
 harness = false
 
+[[bench]]
+name = "pool_compare"
+harness = false
+
 [dependencies]
 # Enabled with 'tls-boring'
 boring-rustls-provider = { git = "https://github.com/janrueth/boring-rustls-provider", optional = true } #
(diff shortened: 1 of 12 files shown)


mmcloud:
This PR does not change:

HBONE protocol behaviour
Workload identity or certificate handling
The configured HTTP/2 stream limit
How traffic destinations are selected
The number of streams carried by an individual proxied connection

Benchmark #TODO

This PR is being split to make the performance measurements and behavioural changes easier to review. A benchmark-only PR will be opened first. Once that merges, this PR will either be reduced to the coordination change or replaced by smaller follow-up PRs.

mmcloud:
benchmark: https://github.com/istio/ztunnel/pull/2018

PR #2029, opened 2026-07-30
Use R2 credential region for uploads

Stevenjin8:
The release script passed the Cloudflare R2 endpoint to `aws s3 cp` but inherited the workload region (`us-west-2`). R2 requires the region stored in the credential secret (`auto`).

Read `.region` from `CF_CREDENTIALS` and scope `AWS_REGION` to the R2 upload invocation.

we do a similar thing here: https://github.com/istio/proxy/blob/98e840f31e7a7079bd4f74bc95cd3d72fe91ebe8/scripts/release-binary.sh#L183-L183

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -56,9 +56,11 @@ fi
 
 R2_DEST="${DEST/gs:\/\//s3:\/\/}"
 ENDPOINT=$(echo "${CF_CREDENTIALS}" | jq -r '.endpoint')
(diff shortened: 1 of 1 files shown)


ilrudie (approved):

Stevenjin8:
@dhawton @keithmattix @jwendell @kfaseela Could one of you take a look?

keithmattix (approved):

PR #2032, opened 2026-08-03
Parse the certificate once

fraenkel:


src/tls/certificate.rs
```diff
@@ -60,18 +61,18 @@ pub struct WorkloadCertificate {
...
 
-pub fn identity_from_connection(conn: &CommonState) -> Option<Identity> {
+pub fn certificate_from_connection(conn: &CommonState) -> X509Result<'_, X509Certificate<'_>> {
     use x509_parser::prelude::*;
     conn.peer_certificates()
         .and_then(|certs| certs.first())
-        .and_then(|cert| match X509Certificate::from_der(cert) {
-            Ok((_, a)) => Some(a),
-            Err(e) => {
-                warn!("invalid certificate: {e}");
```
stevenctl, line 70: should the warn here be preserved somewhere else?

stevenctl (approved):

stevenctl:
lgtm, hold is optional if you want to restore the warn! somehow

src/tls/certificate.rs
fraenkel: I struggle with this warn only because it lacks context.
But I will add it back.

PR #2039, opened 2026-08-04
Use R2 credential region for uploads

Stevenjin8:
manual cherry pick of https://github.com/istio/ztunnel/pull/2029 to unblock release


Stevenjin8:
lets just use the cherry pick commit that comes out of https://github.com/istio/ztunnel/pull/2029

PR #1936, opened 2026-05-30
[DONOTMERGE] WIP for SPIFFE Broker API

MikeZappa87:
Please ignore for now. The SPIFFE broker api is still not merged however the majority of the specification is ironed out, small changes still need to be made. I will rebase and update this branch. This however fully works with the k8s workload reference approach, the pid approach has not been implemented as it probably wouldn't be used per the last istio community sync (The pid code was actually removed so it would explain why some of the abstractions exist, truth to be told, they probably could be removed as well). It can be added as the design is extensible enough. Will need to move some of the code possibly out of this code base since it can be used as a larger client library in rust, everything is just in one place to make things easier. This branch includes a script to setup/teardown/verify in kind.

PR #2067, opened 2026-08-24
identity: support per-workload keys in the certificate cache

nddq:
`SecretManager` caches certificates in a map [keyed by `Identity`](https://github.com/istio/ztunnel/blob/62d826530e5ce85764e309170205b293efcbc80b/src/identity/manager.rs#L219). A workload's identity is its [ServiceAccount-based SPIFFE ID](https://github.com/istio/ztunnel/blob/62d826530e5ce85764e309170205b293efcbc80b/src/state/workload.rs#L304), so, in ztunnel today, all workloads that share a ServiceAccount share one cached certificate. This matches the Istio CA flow: the CSR names [only the SPIFFE identity](https://github.com/istio/ztunnel/blob/62d826530e5ce85764e309170205b293efcbc80b/src/identity/caclient.rs#L140), so one certificate serves every workload with that identity.

The [SPIFFE Broker API](https://github.com/spiffe/spiffe/blob/dc4e9d9b4eff8aa181a54cd330ff9f877186060e/standards/SPIFFE_Broker_API.md) is a new SPIRE agent API, released in [SPIRE `v1.15.2`](https://github.com/spiffe/spire/releases/tag/v1.15.2) under the agent's [`experimental` config block](https://github.com/spiffe/spire/blob/v1.15.3/doc/spire_agent.md#spiffe-broker-api). It lets a node agent such as ztunnel obtain certificates on behalf of the workloads it proxies. Every request names one concrete workload by a [`WorkloadReference`](https://github.com/spiffe/spiffe/blob/dc4e9d9b4eff8aa181a54cd330ff9f877186060e/standards/brokerapi.proto#L108); the agent attests that workload and returns the SVIDs its registration entitles it to. The spec [requires per-workload isolation](https://github.com/spiffe/spiffe/blob/dc4e9d9b4eff8aa181a54cd330ff9f877186060e/standards/SPIFFE_Broker_API.md#42-remote-procedure-scope): a response is valid only for the referenced workload and must not affect others. An identity-keyed cache cannot meet this. Every fetch and every background refresh must name one live pod, and one pod losing entitlement must not drop a certificate shared with its siblings. With Istio-compatible registration, pods that share a ServiceAccount keep one SPIFFE ID. ztunnel support for this API is tracked in #1936 and istio/istio#42339.

This PR adds per-workload cache keys as groundwork:

- `CacheKey` is `Identity(Identity)` or `Workload { id, uid, workload }`. Identity keys keep today's behavior. Workload keys are keyed by (identity, pod UID); the workload context rides on the key as payload, excluded from Hash/Eq and pinned by a unit test. Pod UID is unique, so (identity, uid) is a total key and the payload is a pure function of it; the map keeps the first inserted key. Same pattern as `ProxyRbacContext` in [`src/state.rs`](https://github.com/istio/ztunnel/blob/62d826530e5ce85764e309170205b293efcbc80b/src/state.rs#L154-L160).
- `CaClientTrait::fetch_certificate` takes `&CacheKey`. The Istio CA client reads only the identity part and fails closed on a workload key, so the CSR path is unchanged.
- `fetch_certificate_by_workload` and `forget_certificate_by_workload` are the per-workload entry points, named after the existing `fetch_workload_by_*` convention. Nothing calls them yet, so behavior is unchanged.
- `forget_certificate` now also removes the workload-keyed entries for the identity, and the worker drops the per-key retry backoff on forget. The [workload-removal hook](https://github.com/istio/ztunnel/blob/62d826530e5ce85764e309170205b293efcbc80b/src/cert_fetcher.rs#L112) forgets by identity only, so without this the entries for deleted pods refresh forever.
- The admin certs dump gains an optional `uid` field, so two same-identity rows stay distinguishable and `identity` stays a parseable SPIFFE URI.
- `cert_fetcher` keeps its own identity-based request enum. Prefetch stays a per-identity operation.

Derived from #1936 by @MikeZappa87. The workload context now lives on the cache key instead of a parallel request type. The broker client and its configuration follow in later PRs.

Behavior: the certificate path is unchanged for the Istio CA. Two internal changes on forget: `forget_certificate` also removes workload-keyed entries, and the worker drops a key's retry backoff when it drains a forget, so forget-then-re-add starts from a fresh backoff instead of inheriting the old one. No user-facing change.

## Testing

`cargo test --lib`: 256 pass, including 6 new tests — per-pod cache entries, rotation carries the pod context, forget removes workload keys and stops their refresh, per-pod forget, key semantics of the educe derive, and the fail-closed CA guard. `cargo clippy --lib --tests` and `cargo fmt` are clean.


src/identity/manager.rs
```diff
@@ -292,42 +346,48 @@ impl Worker {
...
                     },
-                    Some(Request::Forget(id)) => {
-                        if self.has_id(&id).await {
+                    Some(Request::Forget(key)) => {
+                        // A forget destroys the entry, so its retry backoff dies with
+                        // it. Remove before the has_key check so a forgotten key never
+                        // keeps backoff state. Cost: a stale Forget after a re-add
+                        // resets the live backoff once. The reverse placement is worse:
+                        // a re-added key would inherit a near-cap backoff.
+                        pending_backoffs_by_key.remove(&key);
```
jaellio, line 379: Could you explain this change in a bit more depth? Is this only required for the workload keys (and if so why) and if not is it a bug in the current implementation?

Stevenjin8:
@nddq I think the mechanism seem ok, to me, but I'd like to better understand where this fits into the big picture. Is this PR derieved from #1936 or enables #1936 because 1936 seems very stale.

nddq:
@Stevenjin8 thanks for taking a look! My plan is to break down #1936 into core changes that are easier to review, largely keeping the original author's design decisions where possible. The end goal is still the same as the original PR - adopting SPIFFE Broker API, making ztunnel understand Spire/SPIFFE as a CA. This PR changes the first piece in the cache so that it allows ztunnel to cache attested certificate per workload in addition to per identity to maintain backward compat.

src/identity/manager.rs
nddq: It is for both, currently `Forget` never clears the backoff entry, only a later successful fetch of the same key does, so a key that failed and then got forgotten keeps its `ExponentialBackoff` around. That's bounded now since identity keys come back. Workload keys are per pod uid and never come back, so with pod churn + a flaky CA the map just grows, which is why the fix is in this PR. By placing this before the has_key check, a forget-then-re-add still clears the old backoff instead of handing the re-added key a near-cap wait, at the cost of a stale Forget occasionally resetting a live backoff once.

src/identity/caclient.rs
```diff
@@ -212,13 +218,15 @@ pub mod mock {
...
-    use crate::identity::Identity;
+    use crate::identity::{CacheKey, Identity};
 
     use super::*;
 
     #[derive(Default)]
     struct ClientState {
-        fetches: Vec<Identity>,
+        // Records the full cache key of each fetch. Tests assert the identity
+        // and, for workload keys, the workload context.
```
Stevenjin8:
```suggestion
```

src/identity/manager.rs
```diff
@@ -145,7 +147,59 @@ impl Default for Identity {
...
+        }
+    }
+
+    /// The pod UID for workload-keyed entries, `None` for identity-keyed ones.
+    pub fn uid(&self) -> Option<&Strng> {
+        match self {
+            CacheKey::Identity(_) => None,
+            CacheKey::Workload { uid, .. } => Some(uid),
+        }
+    }
```
Stevenjin8: not a huge fan here. Can we just do the `match` on the outside of where this is getting called?

src/identity/manager.rs
```diff
@@ -292,42 +346,48 @@ impl Worker {
...
                         }
                     },
-                    Some(Request::Forget(id)) => {
-                        if self.has_id(&id).await {
+                    Some(Request::Forget(key)) => {
+                        // A forget destroys the entry, so its retry backoff dies with
+                        // it. Remove before the has_key check so a forgotten key never
+                        // keeps backoff state. Cost: a stale Forget after a re-add
+                        // resets the live backoff once. The reverse placement is worse:
+                        // a re-added key would inherit a near-cap backoff.
```
Stevenjin8:
```suggestion
                        // keeps backoff state.

**Review by Stevenjin8 — CHANGES_REQUESTED** [review]
Mostly lgtm, just some thoughts on the comments

**Comment by nddq** [comment]
@Stevenjin8 thanks for taking a look! All comments have been addressed

File: src/identity/manager.rs [hunk 5]
```diff
@@ -669,20 +728,71 @@ impl SecretManager {
...
         // We would ideally drop any pending or new requests to rotate.
-        if self.worker.certs.lock().await.remove(id).is_some() {
-            self.post(Request::Forget(id.clone())).await;
-        }
+        //
+        // Removes every entry with this identity: the identity-keyed entry and
+        // all workload-keyed entries. To remove one workload only, use
+        // forget_certificate_by_workload.
+        self.forget_matching(|key| key.identity() == id).await
+    }
```
ilrudie, line 761: This turns the istio csr path from a constant time direct remove into a full cache walk under lock. Can we keep the direct remove?

nddq: I've reverted back to direct remove for the istio csr path. Workload-keyed entries are removed via `forget_certificate_by_workload`, which now takes the same args as the fetch so it's a direct remove too.

src/identity/caclient.rs
```diff
@@ -199,8 +199,14 @@ impl CaClient {
...
-    async fn fetch_certificate(&self, id: &Identity) -> Result<tls::WorkloadCertificate, Error> {
-        self.fetch_certificate(id).await
+    // Istio CA issues per identity. Only the identity part of the key is used.
+    async fn fetch_certificate(&self, key: &CacheKey) -> Result<tls::WorkloadCertificate, Error> {
+        // Workload-keyed fetches must not reach a per-identity CA. Each pod
+        // would mint a duplicate certificate, so fail closed instead.
+        if !matches!(key, CacheKey::Identity(_)) {
+            return Err(Error::BugWorkloadKeyedFetch(key.to_string()));
+        }
+        self.fetch_certificate(key.identity()).await
```
ilrudie: **nit** doesn't need to block, imo this reads nicer though.
```suggestion
    // Istio CA issues per identity. Only the identity part of the key is used.
    async fn fetch_certificate(&self, key: &CacheKey) -> Result<tls::WorkloadCertificate, Error> {
        // Workload-keyed fetches must not reach a per-identity CA. Each pod
        // would mint a duplicate certificate, so fail closed instead.
        let id = match key {
             CacheKey::Identity(id) => id,
             CacheKey::Workload(_) => {
                return Err(Error::BugWorkloadKeyedFetch(key.to_string()))
             }
       };
self.fetch_certificate(id).await
```

(don't just accept the suggestion, the white space will be all messed up).

ilrudie (approved):
I added the hold because of my non-blocking readability suggestion. Other than that, lgtm

ilrudie:
Oh, also I noticed you are squashing commits and force-pushing. We'll squash automatically when it merges so you don't need to do that if you don't want. Appreciate the instinct to keep a tidy commit history though.

nddq:
@ilrudie force of habit 😅 I'd keep separate commits on a PR if the changes are modular and large since it helps reviewers going through the changes, otherwise for small nits, I'd usually just amend and force push

ilrudie:
> @ilrudie force of habit 😅 I'd keep separate commits on a PR if the changes are modular and large since it helps reviewers going through the changes, otherwise for small nits, I'd usually just amend and force push

@nddq, sgtm. Just wanted to make sure you knew.

PR #2086, opened 2026-09-03
fix: correct CIDR service entry mapping updates

### LINKED ISSUE #2085: Service endpoints for CIDR look ups are left stale
When using a ServiceEntry that utilizes a CIDR, e.g.
```
spec:
  addresses:
  - 0.0.0.0/0
  exportTo:
  - '*'
  hosts:
  - internal-catchall
  location: MESH_INTERNAL
  ports:
  - name: tcp
    number: 80
    protocol: TCP
    targetPort: 80
  resolution: STATIC
  workloadSelector:
    labels:
      app: agentgateway-egress
```

When the egress gateway pods are recycled, ztunnel doesn't attempt to send to new pod ips until ztunnel or istiod is also recycled. This leads to errors like:
```
2026-09-03T02:04:23.060428Z warn access connection failed src.addr=192.168.68.244:55956 dst.addr=140.82.112.21:80 direction="outbound" error="no healthy upstream: 140.82.112.21:80" 2026-09-03T02:04:23.215356Z warn access connection failed src.addr=192.168.65.176:37424 dst.addr=140.82.112.21:80 direction="outbound" error="no healthy upstream: 140.82.112.21:80"
```

zac-nixon:
fixes: https://github.com/istio/ztunnel/issues/2085

It's a small fix to remove the existing service entry, and replace it with the current service snapshot, which matches the other spots in the code which also need to do service updates


zac-nixon:
Hi @ilrudie can you take a look here! :D

src/state/service.rs
```diff
@@ -1228,6 +1233,42 @@ mod tests {
...
+        // Replace uid-a's with uid-b
+        store.apply_endpoints(
+            &name,
+            HashMap::from([("uid-b".into(), endpoint("uid-b", HealthStatus::Healthy))]),
+            HashSet::from(["uid-a".into()]),
+        );
+        let svc = store.get_best_by_vip(&target, None).unwrap();
+        assert!(
+            !svc.endpoints.contains(&"uid-a".into()),
+            "removed endpoint must be in CIDR lookup results"
```
ilrudie:
```suggestion
            "removed endpoint must not be in CIDR lookup results"
```

ilrudie (approved):
lgtm, added a hold because I think one test message is slightly wrong and we should correct before merging.

Thanks for the patch!

ilrudie:
Thanks for the patch @zac-nixon, I hope you don't mind I dropped the hold. Would like to see this merge

PR #2091, opened 2026-09-07
bump hickory crates to 0.26.2, add a regression test for non-EDNS UDP responses over 512 bytes

### LINKED ISSUE #2060: DNS proxy returns UDP responses >512 bytes to non-EDNS(0) clients without setting TC
### Bug Description

  With DNS capture enabled, ztunnel's DNS proxy returns UDP responses larger than 512 bytes to                                                                    
  clients that sent no OPT record, and does not set the TC bit.

  RFC 1035 §4.2.1 limits a UDP DNS message to 512 bytes; that ceiling is only raised when the
  requestor advertises a larger buffer via EDNS(0), and RFC 6891 §7 says absence of an OPT record
  in the request MUST be taken as the requestor not implementing EDNS(0) at all.

  Resolvers that read into a 512-byte buffer (Ruby's `Resolv`, musl libc) therefore receive a
  kernel-truncated datagram, fail to parse it, silently discard it, and hang until their retry
  budget is exhausted. Querying CoreDNS directly behaves correctly — only the captured path is affected.

  ### Evidence

  Test name is an internal CNAME to an AWS OpenSearch VPC endpoint with 42 A records. Same cluster,
  same CoreDNS, same query (`A`, `rd=1`, no OPT). Only variable is ambient enrollment:

  | Pod | Response | ANSWER | TC bit |
  |---|---|---|---|
  | ambient-captured | **821 bytes** | 43 RRs | **not set** |
  | `istio.io/dataplane-mode: none` | **509 bytes** | 24 A records (of 42) | **set** |

  From an ambient-captured pod:

  ```
  $ dig +noedns +ignore opensearch.internal.example.com A
  ;; flags: qr rd ra; QUERY: 1, ANSWER: 43, AUTHORITY: 0, ADDITIONAL: 0
  ;; MSG SIZE  rcvd: 821
  ```

  A 512-byte-buffer resolver in that same pod:

  ```
  $ ruby -rresolv -e 'p Resolv.getaddress("opensearch.internal.example.com")'
  Resolv::ResolvError: no address for opensearch.internal.example.com
  # returns after 75s — Resolv's 5+10+20+40s retry ladder, at ~0 CPU
  ```

  Mechanism, same datagram read with two buffer sizes:

  ```
  recvfrom(512)   -> 512 bytes: Resolv::DNS::DecodeError: limit exceeded
  recvfrom(65535) -> 821 bytes: decodes OK, 42 A records
  ```

  ### Impact

  A fluentd DaemonSet (`fluent-plugin-opensearch`, Ruby) went into CrashLoopBackOff: the output
  plugin blocks during worker startup retrying resolution at 75s per attempt, so fluentd never binds
  its health-check port, and the liveness probe SIGKILLs the container at ~4 minutes. Cluster-wide
  log shipping was down.

  Diagnosis was expensive because the failure presents as a DNS outage while `dig`, `getent`, `ping`
  and `nc` all succeed from the same pod — glibc advertises EDNS(0), so the oversized answer is
  legitimate for it. Only non-EDNS(0) clients break.

  Likely also affects musl/Alpine images, whose resolver has historically used a 512-byte buffer with
  no EDNS(0) and no TCP fallback. Possibly the same underlying cause as #1196.

  ### Likely root cause

  [hickory-dns/hickory-dns#3814](https://github.com/hickory-dns/hickory-dns/issues/3814) —
  *"server: UDP responses to queries without EDNS are limited to 4096 bytes, not 512"*.
  `MessageResponse::encode` falls back to `MAX_RECEIVE_BUFFER_SIZE` (4096) when the response carries
  no OPT record, which is exactly the non-EDNS case. Reported against `hickory-server` 0.26.1; fixed
  upstream but not yet in a published release (latest tag is v0.26.1).

  `release-1.30` and `master` both pin `hickory-server 0.26.1`. The 4096-byte ceiling matches what we
  observed: an 821-byte answer passed through untruncated.

  Caveat: I have not confirmed that ztunnel's proxy path routes forwarded responses through the
  server-side `MessageResponse::encode`. Component, version, symptom and ceiling all line up, but
  please treat this as *consistent with* #3814 rather than confirmed.

  ### Workaround

  `istio.io/dataplane-mode: none` on the affected pods — resolved immediately, 4/4 Ready, no restarts.

  ### Version

  - Istio 1.30.0 (istiod and ztunnel both 1.30.0), ambient mode
  - `ISTIO_META_DNS_CAPTURE: "true"` in meshConfig `proxyMetadata`
  - EKS, stock CoreDNS (forwards non-`cluster.local` to the VPC resolver)

  ### Related

  - #1457 — ztunnel DNS capture merging ADDITIONAL into ANSWER (also traced to hickory-dns)
  - istio/istio#29675 — DNS proxy inflating response size (sidecar era)

yeongjukang:
fixes: #2060

hickory-server 0.26.1 used a 4096-byte UDP payload limit for non-EDNS(0) queries instead of the RFC 1035 512-byte limit, so oversized responses reached non-EDNS(0) clients (musl libc, Ruby's `Resolv`, etc.) untruncated and got cut off by the kernel, causing lookups to hang.

Fixed upstream in hickory-dns/hickory-dns#3825, released in hickory-server 0.26.2.
Adds a regression test that fails against 0.26.1 and passes against 0.26.2.

Upstream release notes: https://github.com/hickory-dns/hickory-dns/releases#release-v0.26.2

---

Additional note: cargo audit reports no advisories related to the updated hickory-* crates or prefix-trie. It does report two pre-existing findings in crossbeam-epoch and h2, neither of which is modified by this PR.


ilrudie (approved):
lgtm, thanks for the patch

PR #2101, opened 2026-09-14
pin h2 crate at 0.4.15

### LINKED ISSUE #2100: h2 GOAWAYs on 0.14.16+
https://github.com/hyperium/h2/issues/944

A recent h2 PR can impact ambient clients/servers sending many small TCP packets over the hbone tunnel. This has been seen in the wild for certain types of databases. When the fix trips, it sends GOAWAY for the tunnel severing the connections riding on that hbone tunnel.

We probably want to roll back to h2 0.4.15, and check if https://github.com/hyperium/h2/pull/946 or something more would be needed to repair.

ilrudie:
Pin h2 crate to 0.4.15 until we can fix https://github.com/istio/ztunnel/issues/2100.


Cargo.toml
```diff
@@ -112,7 +112,7 @@ backoff = "0.4"
 pin-project-lite = "0.2"
 pingora-pool = "0.8"
 flurry = "0.5"
-h2 = "0.4"
+h2 = "0.4.15" # https://github.com/istio/ztunnel/issues/2100
```
howardjohn:
```suggestion
h2 = "=0.4.15" # https://github.com/istio/ztunnel/issues/2100
```

syntax might be off but I think your current thing just says "at least 0.4.15"

howardjohn (approved):

Cargo.toml
ilrudie: ty

PR #2113, opened 2026-09-16
Windows ambient prototype: fix cross-build and make -D warnings presubmit pass

MartinForReal:
This PR resolves the build issues on the **experimental Windows ambient** prototype (`experimental-windows-ambient`) so the branch builds (and its presubmit, which runs with `RUSTFLAGS="-D warnings"`, passes) on Windows.

## Cross-build fix
- **`Cargo.toml`**: switch the default TLS provider from `aws-lc` to the pure-Rust `ring`. `aws-lc-sys` shells out to cmake/nasm and a C toolchain, which is fragile across Windows toolchains (the `cmake` crate panics on newer Visual Studio versions and `-WX -W4` rejects C11 atomics). `ring` builds everywhere with no native toolchain. `aws-lc`/`boring`/`openssl` remain selectable via the existing `TLS_MODE`/feature knobs.
- **`src/socket.rs`**: `socket::socket_tests::test_keep_alive_options` read keepalive/user-timeout options back via `SockRef` getters that only exist on Unix, so `cargo test` did not compile on Windows. Gate the read-back test (and its imports) to Unix; the keepalive setters are still exercised on every platform.

## Make the `-D warnings` presubmit gate pass
- **`src/proxy.rs`**: drop the dead Windows `set_tcp_user_timeout` stub (no callers on Windows).
- **`src/inpod/windows/{namespace,statemanager}.rs`**: allow the test-only helpers currently consumed solely by commented-out tests.
- **`benches/basic.rs`, `benches/throughput.rs`**: on non-Linux the benchmark body is compiled out, leaving imports/helpers as dead code. Allow `dead_code`/`unused_imports` only on non-Linux so Linux stays strict while the benches cross-compile.

## Cross-platform warning cleanups (incl. #1)
- `src/metrics/process.rs`, `src/app.rs`, `src/main.rs`, `src/test_helpers.rs`, `examples/inpodserver.rs`

## Validation
On `x86_64-pc-windows-msvc`:
- `cargo build` -> produces `ztunnel.exe`
- `cargo check --all-targets` with `RUSTFLAGS="-D warnings"` -> clean
- `cargo test --lib --benches --tests --bins --no-run` with `RUSTFLAGS="-D warnings"` -> clean
- `cargo test --lib` -> **176 passed, 0 failed**


keithmattix:
Should the target be experimental-windows-ambient branch?

Cargo.toml
```diff
@@ -5,7 +5,20 @@ edition = "2024"
...
+# CI). However, aws-lc (via `aws-lc-sys`) shells out to cmake/nasm and needs a
+# C toolchain, which makes *cross-building* to Windows fragile: it can panic in
+# the `cmake` crate for newer Visual Studio versions and requires `-WX -W4`
+# clean C11 compilation. The pure-Rust `ring` provider builds on every platform
+# (Linux, macOS and Windows) with no native assembler/compiler dependency.
+#
+# This makes the Windows prototype default to a buildable provider while still
+# allowing aws-lc/boring/openssl to be selected explicitly with the matching
+# `tls-*` feature (see the Makefile `TLS_MODE` knob).
+default = ["tls-ring"]
```
keithmattix: Not that this branch is meant to reflect what master would look like in a windows world; we should not be changing Linux defaults

README.md
```diff
@@ -36,11 +36,16 @@ Rustls has support for plugging in various crypto providers to meet various need
 
 | Name                                               | How To Enable                                  |
 |----------------------------------------------------|------------------------------------------------|
-| [aws-lc](https://github.com/aws/aws-lc-rs)         | Default (or `--features tls-aws-lc`)           |
```
keithmattix, line 39: -1 to this. Aws-lc is faster; we should be able to just have a different crypto provider for windows

keithmattix (changes requested):
Don't mind most of the directives but we shouldn't change Linux defaults

MartinForReal:
Addressed the review concern about not changing Linux defaults.

I have rewritten the changes so the **Linux default TLS provider is left untouched**:
- `Cargo.toml` default remains exactly `["tls-aws-lc"]` (unchanged from the base branch); only a brief comment was added pointing Windows cross-build users to `WINDOWS.md`.
- `README.md` was fully reverted to the base-branch version (aws-lc listed as the default).
- Instead of changing the default, Windows cross-builds now opt into the pure-Rust `ring` provider explicitly via `--no-default-features --features tls-ring` (documented in `WINDOWS.md`).

This keeps the Linux build/defaults identical while still making the Windows prototype cross-build cleanly without a cmake/nasm C toolchain.

Also confirmed `experimental-windows-ambient` is the correct target base; PR is now mergeable (MERGEABLE) against it.

Validated on x86_64-pc-windows-msvc with `--no-default-features --features tls-ring`: `cargo check --all-targets` and `cargo test --lib --benches --tests --bins --no-run` pass with `RUSTFLAGS="-D warnings"`; `cargo test --lib` = 176 passed / 0 failed.

keithmattix (approved):

PR #2122, opened 2026-09-18
windows: IPv6 parity with the Linux data-plane design (source preservation + family-aware errors)

MartinForReal:
## Summary
Outbound IPv6 on the Windows ambient prototype silently degraded in two places. Both now **follow the Linux data-plane design**.

### 1. Source-IP-preserving connect no longer silently downgrades
`socket::set_freebind_and_transparent` on non-Linux returned
`Err("IP_TRANSPARENT and IP_FREEBIND are not supported on this operating system")`, and `proxy::freebind_connect` treated that as *best-effort*: it logged a warning and connected anyway **without** the source address — misattributing the connection's source identity.

On Linux, transparent mode is a hard requirement of the capture design; a transparent-mode failure surfaces as a **connect failure**, never a silent downgrade.

**Changes (mirroring the Linux domain-aware design):**
- `socket.rs`: replaced the blanket `#[cfg(not(target_os = "linux"))]` stub with a `#[cfg(target_os = "windows")]` implementation (the other-OS stub is kept). Windows has no `IP_TRANSPARENT`/`IP_FREEBIND`, and none are needed for the **in-pod** prototype: the workload's source address is *local to the process's network compartment*, so source preservation is achieved by the plain `bind(local_addr)` that `freebind_connect` performs. A bind of a genuinely non-local address (e.g. host-compartment mode) still **fails loudly at bind time** — the same observable behavior as Linux when transparent mode is unavailable.
- `proxy.rs`: `freebind_connect` now **returns** the error from the freebind/transparent setup and the source bind instead of warning and proceeding. Only the HBONE paths pass a source address (`inbound.rs`, `inbound_passthrough.rs`) and source preservation is a requirement there (identity); `outbound.rs` passes `None` and is unaffected.

### 2. Family-aware error when the IPv6 socket cannot be created
`create_socket` surfaced the bare `"socket creation failed"` `io::Error` when `new_tcp_v6()` failed, making a dual-stack misconfig (a pod compartment without IPv6) undiagnosable. The error now **names the address family**.

There is deliberately **no v6→v4 fallback** at connect time: an IPv4 socket cannot connect to an IPv6 address, and Linux has no such fallback either. (A v6→v4 fallback for dual-stack *services* belongs in destination resolution, not at connect time.)

### Also
Dropped the now-unused bare `std::io::Error` alias in `socket.rs` (fully qualified `io::Error` is used everywhere) — it was only used by Linux/other-OS paths and broke `-D warnings` on Windows.

## Validation
On `x86_64-pc-windows-msvc` with `RUSTFLAGS="-D warnings"` and `--no-default-features --features tls-ring`:
- `cargo check --all-targets` — clean
- `cargo test --lib` — 176 passed, 0 failed

## Files changed
- `src/socket.rs` — domain-aware Windows transparent/freebind mirror; `io::Error` qualification
- `src/proxy.rs` — loud source-preservation failure; family-aware IPv6 socket error


MartinForReal:
Consolidating the IPv6 support work into a single PR (#2121) as requested. This PR's changes are included there (rebased on top of the IPv6-detection/reuse_port changes), so closing this one to keep the review in one place.

